PR #185, opened 2014-12-29
OCSADV-196

sraaphorst:
Created the interface for science models and extracted the GMOS science area code from the TPE into the model.
Minor changes to ImPolygon for Java interoperability.


sraaphorst:
Lots of changes in this modification to the original PR. The GMOS (and by extension, BHROS) guide probe arm / pickoff mirror are now part of the model. This was a fairly extensive change that required a lot of analysis of the code and refactoring because the original representation was calculated in TPE screen coordinates and was dependent on some TPE settings. Furthermore, the drawing code was tangled with a bunch of affine transformations (scalings and reflections) that needed to be factored out so that the geometry could instead be calculated in arcsec and then have transformations applied after to the geometry to achieve the same shapes in the same positions for the TPE.

The most challenging part was finding the necessary technique to extract the tangled in transformations from the calculation of the probe arm angle and then find the equivalent transformations to be applied afterwards. I am quite certain (barring any calculation errors) that I have proven the correctness of this on paper through linear algebra that the result is the same, but the correctness proof doesn't seem suitable as something to be typed up here in github.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/inst/FeatureGeometry.scala
```diff
@@ -0,0 +1,81 @@
...
+  def fullGeometry: Shape = {
+    val allShapes = geometry
+    if (allShapes.length <= 1)
+      allShapes.headOption.getOrElse(new Area())
+    else {
+      val fullArea = new Area()
+      geometry.foreach(g => fullArea.add(new Area(g)))
+      fullArea
+    }
+  }
```
tpolecat: You can save a traversal by pattern-matching (`List.length` is O(N)).

```scala
def fullGeometry: Shape = 
  geometry match {
    case Nil      => new Area
    case a :: Nil => a
    case as       => as.foldLeft(new Area)((a, g) => a.add(new Area(g)))
  }
```

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/inst/FeatureGeometry.scala
```diff
@@ -0,0 +1,81 @@
...
+   * A convenience method to calculate the full geometry and then apply a transformation to it.
+   * @param transform the transformation to apply
+   * @return          the geometry for the features under the transformation in arcsec
+   */
+  def transformedFullGeometry(transform: AffineTransform): Shape = {
+    val area = new Area(fullGeometry)
+    area.transform(transform)
+    area
+  }
+}
```
tpolecat: This is fine, but just FYI you can use `<|` from scalaz to do this kind of thing in one step. I am **not** recommending that you change your code here and elsewhere, it's just a fun thing to know.

```scala
def transformedFullGeometry(t: AffineTransform): Shape =
  new Area(fullGeometry) <| (_ transform t)
```

tpolecat:
I don't see any obvious problems, but it's kind of mathy ... I would feel a little better if there were some tests for this, maybe comparing results with baselines computed from the old code? Is this feasible? This PR has been sitting here for a while so I'm wondering how we nudge it forward.

PR #186, opened 2014-12-31
OCSADV-27: Support for results filtering

cquiroz:
This is in intermediate PR, it basically adds the filtering logic on the existing catalog module to support VoTable queries and the new target model.
After this PR the work should move into the core of AGS though changes to catalog classes are to be expected


cquiroz:
I put several updates based on the comments, no new commits though

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/api/CatalogQuery.scala
```diff
@@ -1,8 +1,11 @@
 package edu.gemini.catalog.api
 
-import edu.gemini.spModel.core.{Angle, Coordinates}
+import edu.gemini.spModel.core.Coordinates
+import edu.gemini.spModel.core.Target.SiderealTarget
 
-case class CatalogQuery(base: Coordinates, coneRadius: Angle, catalog: CatalogName = sdss)
+case class CatalogQuery(base: Coordinates, radiusConstraint: RadiusConstraint, magnitudeConstraints: MagnitudeConstraints, catalog: CatalogName = ucac4) {
+  def filter: SiderealTarget => Boolean = (t) => radiusConstraint.targetsFilter(base)(t) && magnitudeConstraints.filter(t)
+}
```
swalker2m: There may be some confusion (in my head at least).  There are two distinct concepts here.  There's the question of whether other (non-gemini) catalog servers must be supported vs whether only a single catalog being hosted by a server need be supported. My understanding is that no other servers need be supported (and so no UI choice) but that we will still need a few of the catalogs.  AGS of course has its own defaults and doesn't prompt the user but a manual search still might provide catalog options.  I guess we should seek clarification here.

tpolecat:
Looks good to me. :+1:

PR #187, opened 2015-01-01
OCSADV-154: Remove support for FK4/B1950

tpolecat:
This PR removes support for the FK4/B1950 coordinate system. It includes the following changes:

- B1950 is removed as a constant (in several places) and all special handling that switches on J2000 vs B1950 is collapsed into a single code path.
- Code is added in `To2015B` to precess existing targets in B1950 forward in time to J2000 on import. Required math code is copied into the migration package to allow future deletion of the original code in `jsky`. There are associated tests.

Following this merge J2000 will be the only supported coordinate system, and a follow-on PR will remove the associated enumerations and so on, which now have only a single member.


bundle/edu.gemini.spModel.io/src/main/java/edu/gemini/spModel/io/impl/migration/to2015B/WSCon.java
```diff
@@ -0,0 +1,225 @@
+package edu.gemini.spModel.io.impl.migration.to2015B;
+
+import java.awt.geom.*;
+
+// taken from wscon.java from NASA
+public class WSCon {
```
tpolecat: FYI this code is copy/pasted from `wscon.java` and has a bunch of inlined methods to remove external dependencies. The testcase compares its output with that from Simbad and `jprecess` (which is apparently the standard implementation of this algorithm).

bundle/edu.gemini.spModel.io/src/main/scala/edu/gemini/spModel/io/impl/migration/to2015B/To2015B.scala
```diff
@@ -4,14 +4,17 @@ import java.util.UUID
```
tpolecat: I turned a bunch of literals into constants here, so those changes are interleaved with the important bits. Sorry.

bundle/edu.gemini.spModel.io/src/main/java/edu/gemini/spModel/io/impl/migration/to2015B/WSCon.java
```diff
@@ -0,0 +1,225 @@
...
+        input.y = ((d2000) * 180.0 / Math.PI);
+        pm.x = ((dr2000) * 180.0 / Math.PI);
+        pm.y = ((dd2000) * 180.0 / Math.PI);
+
+        if (diag > 0) {
+            scon = ((3.6e3) * 180.0 / Math.PI);
+            tcon = ((2.4e2) * 180.0 / Math.PI);
+            dra = tcon * (r2000 - r1950);
+            ddec = scon * (d2000 - d1950);
+            //printf("J2000-B1950: dra= %11.5f sec  ddec= %f11.5f arcsec\n",
```
cquiroz: printf? :wink:

bundle/edu.gemini.spModel.io/src/main/scala/edu/gemini/spModel/io/impl/migration/to2015B/To2015B.scala
```diff
@@ -116,6 +194,36 @@ object To2015B {
     ps
 
   }
+
+  // Proper Motion conversion to degrees/yr
```
cquiroz: On the new spModel I don't think the units for PM are defined. Should we do that?
UCAC4 gives them as miliarcs/year

tpolecat: In the new model my intent was to standardize on mas/yr but this isn't documented. I'll add a doc comment.

tpolecat: I added comments in commit dd0375ef3c2a775f2235d9d635657a147e2d8e8a. It might be sensible to create numeric types that carry units, but that's more than I want to bite off right now.

cquiroz:
Seems good to me :smile:

tpolecat:
@swalker2m would you mind giving this a glance?

bundle/edu.gemini.spModel.io/src/main/scala/edu/gemini/spModel/io/impl/migration/to2015B/To2015B.scala
```diff
@@ -21,38 +24,113 @@ object To2015B {
...
-    uselessSystemsToJ2000
+    uselessSystemsToJ2000,
+    b1950ToJ2000
   )
 
+  // Convenience op for \/
+  implicit class MoreDisjunctionOps[A, B](ab: A \/ B) {
+    def unsafeExtract(implicit ev: A <:< Throwable): B =
+      ab.fold(throw _, identity)
+  }
```
swalker2m: Cool.  Seems like that should be there.

bundle/edu.gemini.spModel.io/src/main/scala/edu/gemini/spModel/io/impl/migration/to2015B/To2015B.scala
swalker2m: I kind of lament the loss of units in our `Angle` class but at least with the new `Angle` we know we've got degrees.   I realize PM is angle / year but would it make sense to represent it with an `Angle` instead of a raw `Double`?

swalker2m:
I definitely support the goal and can't see any obvious issues with the changes themselves so :+1: from me.

bundle/edu.gemini.spModel.io/src/main/scala/edu/gemini/spModel/io/impl/migration/to2015B/To2015B.scala
tpolecat: Yep, changed these to `Angle` and parallax to `Option[Angle]` in b529845d4a2b88de551a18990cda9e4515a8b406.

swalker2m: Cool, :+1:

PR #188, opened 2015-01-05
OCSADV-154: Remove machinery for unsupported conic target types

tpolecat:
OCS currently supports seven conic target representations, but only two of them are supported in the OT; the other five just get shoved into the same fields as one of the two supported types (or in the case of old planet representations the OT goes into an inconsistent state, showing conic parameters rather than the planet selector). Anyway, this PR removes all that crap and adds conversion methods in `To2015B` to switch the systems for comets and minor planets, and actually select, you know, a real planet for old planet formats. I also deleted some related dead code.

Note that I made no attempt to reinterpret the old conic parameters in terms of the new systems. The OT has been doing this blindly for years, so I didn't try to raise the bar.

There is a ridiculous amount of complexity in the OT's target editor surrounding this stuff, but simplifying that is left for a later task.


cquiroz:
Looks good, lots of deleted code (it has to be good :smile: )

swalker2m:
Good to see all that go away.  :+1:

Usually in the semester converters I check the program "version" attribute and skip the conversion if it is newer than the semester in which the conversion is being applied.  Presumably that helps keep the full semester import time down a bit by skipping old unnecessary conversions.

````scala
val Version_2015A = Version.`match`("2015A-1")

def isPre2015A(c: Container): Boolean =
    c.getVersion.compareTo(Version_2015A) < 0
...
getContainers(doc).find(_.getKind == SpIoTags.PROGRAM).filter(isPre2015A)
````

If we're serious about making a break with the XML format though maybe it doesn't matter too much.

tpolecat:
Ok I added the suggested logic to filter out newer programs.

tpolecat:
Ok I updated the version in a9ae3c2. Does that look right?

tpolecat:
Thanks, btw. The versioning stuff there is not obvious.

PR #189, opened 2015-01-06
OCSADV-27: Implicit conversions and port of edu.gemini.ags.impl to the new target model

cquiroz:
This is another PR on the road to make AGS use the new catalog.

The changes are restricted to the package `edu.gemini.ags.impl` porting it use the new target model.
In summary these are the changes:
* Implicit conversions of the new target model classes to the old and viceversa as needed
* Enabling of tests to make sure changes don't break existing logic
* Port of classes on the `edu.gemini.ags.impl` to use the new target model classes as much as possible

There are still a few uses of the old model as long as are method signatures used in other packages.


tpolecat:
Just FYI - although it was common in the early days of Scala, there is a consensus now that implicit conversion between "real" types is generally a bad idea. The preferred approach is to use an implicit class to add syntax that allows for explicit conversion. This is the distinction between  `JavaConversions` (which is deprecated) and `JavaConverters` for example. In this case we could say

```scala
implicit class SkyCalcAngleOps(val a: skycalc.Angle) extends AnyVal {
  def toNewModel = Angle.fromDegrees(a.toDegrees.getMagnitude)
}
```
which lets you convert easily by saying `a.toNewModel` but makes the point of application explicit.

**However**, if these conversions are temporary and only used within this package (or bundle) then I'm ok leaving them here. But if we expose them to more of the codebase or think they will still be around for the 15B release then I would like to change them to use the safer pattern above.

cquiroz:
I was wondering about the same topic comparing the two Java conversions style. this methods of direct conversion hides some of the conversions so using the implicit classes could be better

The conversions are in the `ags.impl` package but I can imagine them being used for a while, there are just too many uses of the old classes to expect them to switch in just one release cycle

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/RadiusLimitCalc.scala
```diff
@@ -0,0 +1,86 @@
...
+    val p1 = l.getP1
+    val p2 = l.getP2
+    if ((p1.getY <= 0) && (0 <= p2.getY)) {
+      // If the center falls between the vertical extremes of the
+      // segment, it's the horizontal distance to the line.
+      abs(p1.getX)
+    } else {
+      min(Zero.distance(p1), Zero.distance(p2))
+    }
+  }
```
swalker2m: Seems like we might as well collapse these two `shortestDistance` into one passing in the `Point2D => Double` function to use?

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/CandidateValidator.scala
```diff
@@ -1,40 +1,41 @@
...
-          val diff = new CoordinateDiff(ctx.getBaseCoordinates, soCoords)
-          diff.getDistance.compareToAngle(min) >= 0
+          val soCoords = so.coordinates
+          val diff = Coordinates.difference(ctx.getBaseCoordinates, soCoords)
+          diff.distance >= min
         }
 
       // Only keep candidates that fall within the magnitude limits.
-      def brightnessOk = so.getMagnitude(params.band).asScalaOpt.exists(magLimits.contains)
+      def brightnessOk = so.magnitudeOn(params.band).exists(m => magLimits.contains(m))
```
swalker2m: Trivial: not sure if it is part of this PR but "in" seems to me a better preposition than "on" here.  As in `magnitudeIn(R)`.

swalker2m: Also, was something wrong with the original eta reduced version?

cquiroz: Nothing really, I'll change it back

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/GemsStrategy.scala
```diff
@@ -70,7 +76,7 @@ object GemsStrategy extends AgsStrategy {
 
   // Convert from catalog results to GeMS-specific results.
   private def toGemsCatalogSearchResults(ctx: ObsContext, futureAgsCatalogResults: Future[List[CatalogResultWithKey]]): Future[List[GemsCatalogSearchResults]] = {
-    val anglesToTry = (0 until 360 by 45).map(x => new Angle(x, Angle.Unit.DEGREES))
+    val anglesToTry = (0 until 360 by 45).map(x => Angle.fromDegrees(x))
```
swalker2m: Another trivial point but I think the η-reduction makes it read a bit better:  `(0 until 360 by 45).map(Angle.fromDegrees)`

swalker2m:
I skimmed through these changes and they look fine to me. :+1: It's great to see the old target model giving a bit of ground to the new!

cquiroz:
I converted the `implicit val`s to `implicit class` and integrated @swalker2m suggestions

PR #190, opened 2015-01-07
OCSADV-154 Remove reference frame as a concept

tpolecat:
All targets in OCS are now specified in the J2000 reference frame, so there is no need to track equinox in the new model and we can get rid of the J2000 conversion methods in the old model. That's what this PR does.

N.B. the `getTargetAsJ2000` method on `ITarget` always returned a clone, but no caller ever retained a reference or mutated the result. It is therefore safe to just operate with the existing reference rather than making a copy.


tpolecat:
Oops didn't have the new catalog bundle in the project. Standby. :-1:

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/Epoch.scala
```diff
@@ -0,0 +1,11 @@
+package edu.gemini.spModel.core
+
+/** Epoch in Julian years. */
+final case class Epoch(value: Double)
```
swalker2m: I know case classes are technically extensible but since you're really not supposed to do that I just treat them as if that weren't a possibility.  Just as `null` is technically possible but we treat it as if it doesn't exist unless working with Java.  Just wondering if you recommend `final`ing our case classes in general.

cquiroz: Isn't extending a case class actually deprecated in 2.10?

swalker2m:
Looks good.  :+1:

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/Epoch.scala
tpolecat: Yes, I do recommend `final case class`, if only for the potential benefit of inlining and monomorphic method dispatch. I don't think it's a huge deal because as you say, they are "morally" final ... but there's no harm and it may help.

@cquiroz you can extend a case class with a non-case class. I don't know why this is allowed. I'll ask around.

PR #192, opened 2015-01-07
OCSADV-154 push mags into itarget

tpolecat:
This is a small refactor that moves the `ImList<Magnitude>` member from `spTarget` to `ITarget` (which has been made into a subclass of `CoordinateSystem` for this purpose).

The diff for `ITarget` is a little confusing because it's trying to interleave deletions with unrelated pasted code from `SPTarget`. Sorry.


swalker2m:
@cquiroz I think the only reason to use `_name` would be to be consistent with some existing Java class.

swalker2m:
This looks good to me.  :+1:

cquiroz:
Look fine for me too

PR #193, opened 2015-01-08
OCSADV-154 Simplify ITarget hierarchy (part 1)

tpolecat:
This is the first of several simplifications to flatten out the `ITarget` hierarchy. Goal is to make the old stuff look more like the new stuff so the switchover is easier. I'm just removing some unnecessary classes here.


cquiroz:
It looks good to me
How do we make sure these changes to core classes don't break stuff? Do you do a full compile/run tests/run the OT before the PR is submitted?

swalker2m:
Because it is so convoluted, I don't have a good mental model of the old non-sidereal stuff so it's hard to evaluate any potential problems. This is clearly simpler though and I don't have a concrete reason think it won't be okay. :+1:

tpolecat:
Ok yeah this is very simple. Both of these interfaces had only one implementation (`NonSiderealTarget`) so it was safe to just get rid of them.

tpolecat:
And yeah I do a full build with all the tests.

PR #194, opened 2015-01-09
OCSADV-200: work-in-progress on diff-based merge

swalker2m:
This PR updates the difference calculation that will be employed by the sync/merge algorithm.  In particular, it:

* moves new code into a temporary package `edu.gemini.sp.vcs.diff` to help me separate classes involved in the old algorithm in `edu.gemini.sp.vcs` from the new.  When the new algorithm is done, the old code will be removed and the content of this package can be moved back into `edu.gemini.sp.vcs`.
* simplifies and flattens difference nodes into just two varieties: those present in the local program vs. those that are missing.  That is, the previous `InUse` turned into `Present` and the old `Missing` and `Removed` were merged into just `Missing`.
* updates the comparison to take into account whether a node is present in the remote program.  Even when a node’s `VersionVector` is the same across the local and remote program, we still want to know about it if the node is missing in one instance of the program.
* implements `scalaz.Equal` instances for `VersionVector`.

Testing feels inadequate.  I could use some pointers for how/whether to turn this into property based tests checked by ScalaCheck, presumably using a Generator that produces program edit functions.


swalker2m:
I should note that I renamed the existing class `RichRemoteNode` to `RichNode` (to match the previous renaming of `ISPRemoteNode` to `ISPNode`).  It appears as an entirely new class in the difference.  I did add a few methods there but not everything is really new.

swalker2m:
If all the commits seem confusing it's all about the 50 line `compare` method in the final version of `ProgramDiff`.

bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs/diff/ProgramDiff.scala
```diff
@@ -0,0 +1,125 @@
...
+        case _                 =>
+          val childDiffs = (List.empty[Diff]/:r.children) { (ns, child) =>
+            presentDiffs(child) ++ ns
+          }
+
+          // If there is even one descendant that differs, include this node
+          // in the results.  Otherwise, only include it if it differs.
+          if (childDiffs.isEmpty) diffNode(r).toList
+          else present(r) :: childDiffs
+      }
```
tpolecat: I think I would find this a little easier to follow if it were a single expression, although as you noted your fold might be more efficient in the end.

```scala
case _ =>
  // If there is even one descendant that differs, include this node
  // in the results.  Otherwise, only include it if it differs.
  r.children.flatMap(presentDiffs) match {
    case Nil => diffNode(r).toList
    case xs  => present(r) :: xs
  }
```

swalker2m: Doesn't appear to be consistently slower in the ridiculous 10,000 observation diff timing test.  I like it.

tpolecat:
This all looks good to me. :+1: 

As for property-based testing, you can state some invariants. For instance:

- Basic stuff; if a given node differs then it must appear in the diff list, etc.
- If `a` appears in the diff list and `b` is an ancestor of `a`, then `b` must also appear in the diff list.
- Any pair of nodes that share the same ancestor observation must both appear in the diff list, or both not appear in the diff list.

So you can maybe start with a simple constant program and apply random changes to a pair of them and ensure that the invariants are always met. I'm not sure how helpful this will be in the end but it might help detect regressions in the end, and might be useful for handling more complex invariants that have to deal with program status and so on.

PR #195, opened 2015-01-09
OCSADV-27: Another step to port AGS

cquiroz:
This PR contains many small commits to port the internals and API of AGS to use the new spModel classes

Most of the changes are quite mechanical and sometimes they are reverted on the next commit when other parts of the code is ported, however, each commit compiles and passes the tests.

This is close to how far I can get without ripping the old catalog out and replace it with the new one


bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsUtils4Java.scala
```diff
@@ -0,0 +1,19 @@
...
+import edu.gemini.shared.util.immutable.ScalaConverters._
+
+/**
+ * Utility methods for Java classes to access scala classes/methods
+ */
+object GemsUtils4Java {
+  // Returns true if the target magnitude is within the given limits
+  def checkMagLimit(target: SPTarget, magLimits: MagnitudeConstraints): Boolean =
+    // TODO This method has too many conversions, should be simplified
+    target.getMagnitude(magLimits.band.toOldModel).asScalaOpt.map(m => magLimits.contains(m.toNewModel)).getOrElse(true)
```
cquiroz: According to @swalker2m the option should default to false (`getOrElse(false)`) but the original code uses true

bundle/edu.gemini.ags/src/main/java/edu/gemini/ags/gems/GemsGuideStarSearchOptions.java
```diff
@@ -281,22 +281,22 @@ public GemsCatalogSearchCriterion searchCriterion(ObsContext obsContext,
...
+        Option<Angle> searchPA = (posAngles.size() == 1) ? new Some<>(posAngles.iterator().next()) : None.<Angle>instance();
         CatalogSearchCriterion criterion = new CatalogSearchCriterion(name, magLimits, radiusLimits, searchOffset, searchPA);
         GemsCatalogSearchKey key = new GemsCatalogSearchKey(gType, gGroup);
         return new GemsCatalogSearchCriterion(key, criterion);
     }
 
     public Set<String> getCatalogs() {
-        Set<String> catalogs = new HashSet<String>(2);
+        Set<String> catalogs = new HashSet<>(2);
         catalogs.add(nirCatalog);
```
swalker2m: Super Trivial: this part is some one else's code but I think an initial capacity of 2 with a default load factor of 0.75 result in growing the backing hash table as soon as the second catalog is added to the set, which is probably not intended.  We might as well star with an initial capacity of 4.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsMagnitude.scala
```diff
@@ -31,10 +28,8 @@ object AgsMagnitude {
    * Gets the widest possible range limits incorporating the given conditions
    * and speeds.
    */
-  def rangeLimits(mc: MagnitudeCalc, c1: (Conditions, GuideSpeed), c2: (Conditions, GuideSpeed)): MagnitudeConstraints = {
-    val ml = mc(c1._1, c1._2).union(mc(c2._1, c2._2)).getValue
-    MagnitudeConstraints(ml.getBand.toNewModel, FaintnessConstraint(ml.getFaintnessLimit.getBrightness), ml.getSaturationLimit.asScalaOpt.map(s => SaturationConstraint(s.getBrightness)))
-  }
+  def rangeLimits(mc: MagnitudeCalc, c1: (Conditions, GuideSpeed), c2: (Conditions, GuideSpeed)): Option[MagnitudeConstraints] =
+    mc(c1._1, c1._2).union(mc(c2._1, c2._2))
```
swalker2m: I see what is going on here but I'm a little hesitant about bubbling up the `Option`.  Let's analyze this a bit and see if we can find a way to guarantee that the union produces a valid `MagnitudeConstraints`.  The band shouldn't change here for example.  Or maybe this doesn't have major ramifications on the clients?

swalker2m: A bit of thinking out loud:  in general, `MagnitudeConstraint.union`  has got to produce an `Option` since you don't know whether they contain the same bands.  Here it *seems* like a safe assumption that a given `MagnitudeCalc` isn't going to produce `MagnitudeConstraints` with different bands given different `Conditions` and `GuideSpeed`.  In fact we make that assumption in the `band` method below.

In general, I think the `union` method could be made to work even if there is no saturation limit or if one of the two constraints being compared doesn't define saturation limits.  In those cases, the union could just be the one with the faintest faintness limit.  If both define a saturation limit, then it takes the max as it does now.  Basically think of no saturation limit as a limit of negative infinity.

So does this mean we should make this method private, call `union` and assume the value is defined?  That doesn't make me feel good either. :-/  I think this is a sign that the API is not perfect.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/conf/MagLimitsId.scala
```diff
@@ -0,0 +1,10 @@
+package edu.gemini.ags.conf
+
+import edu.gemini.spModel.gemini.obscomp.SPSiteQuality.{SkyBackground, ImageQuality}
+import edu.gemini.spModel.guide.GuideSpeed
+
+sealed case class MagLimitsId(name: String)
```
swalker2m: A sealed case class?  Since case classes shouldn't be extended anywhere (whether in the same file or not) maybe you meant `final`?

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsMagnitude.scala
cquiroz: bubbling up the Option does work though, the method is called in a couple of places where you can handle the option, e.g.
`GemsStrategy.queryConstraints`

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/conf/MagLimitsId.scala
cquiroz: Right, I just moved it from conf package object

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/mascot/util/MatrixUtil.scala
```diff
@@ -236,14 +236,14 @@ object MatrixUtil {
        }
     }
     sb.append(")\n")
-    sb.toString;
+    sb.toString
```
swalker2m: I guess the original author hadn't discovered the `mkString` method.  As in `a.map(sFormat).mkString("Array(\n", ",\n", ")\n")`

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsUtils4Java.scala
tpolecat: Can we rename `checkMagLimit` to something that better indicates what the return value means?

cquiroz: certainly, I'd also expect this to migrate and perhaps to be a method on `SiderealTarget`, like `target.isInLimits(constraints)`

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/api/MagnitudeConstraints.scala
```diff
@@ -83,7 +83,7 @@ case class MagnitudeConstraints(band: MagnitudeBand, faintnessConstraint: Faintn
       val faintness = faintnessConstraint.max(that.faintnessConstraint)
 
       // Calculate the max out of the saturation limit if both are defined
-      val saturation = (saturationConstraint |@| that.saturationConstraint)(_ max _)
+      val saturation = (saturationConstraint |@| that.saturationConstraint)(_ min _)
```
swalker2m: we should update the comment to match i suppose

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/SingleProbeStrategy.scala
```diff
@@ -45,7 +43,7 @@ case class SingleProbeStrategy(key: AgsStrategyKey, params: SingleProbeStrategyP
     // throw away the guide probe (which we know anyway), and obtain just the
     // list of guide stars
     candidates(ctx, mt).map { lst =>
-      lst.headOption.fold(List.empty[SiderealTarget]) { case (_, so) => so.map(_.toNewModel) }
+      lst.headOption.fold(List.empty[SiderealTarget]) { case (_, so) => so }
```
tpolecat: FYI in cases like this where you have a monoid you can just say `.orZero`

```scala
lst.headOption.map(_._2).orZero
~lst.headOption.map(_._2) // same thing, but I prefer .orZero
```

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsMagnitude.scala
swalker2m: The `Option` bubbling up is unfortunate since it shouldn't be possible to end up with a `None` there if we could fix the AGS API.  That isn't the point of this task though so I'm willing to let it slide.

cquiroz: `union` only returns `None` for mismatched bands, it will handle saturation limits correctly

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/mascot/util/MatrixUtil.scala
cquiroz: I just realized most of the formatting code isn't in use, so it could be removed altogether

cquiroz:
I added an extra commit with some of the suggested changes and a note to re-evaluate the return type of `AgsMagnitudes.rangeLimits`

swalker2m:
Looks good. :+1:

PR #196, opened 2015-01-13
REL-2918: Change default target name resolver to SIMBAD@ESO

cquiroz:
Small changes to remove non working name server catalogues


tpolecat:
:+1:

PR #197, opened 2015-01-13
REL-2193: format user target numbers correctly

swalker2m:
The so-called "user" blind offset targets are supposed to be indexed from 1 when formatted for display.


cquiroz:
I'm not sure that +1 looks suspicious :smile: 
:+1:

PR #198, opened 2015-01-14
OCSADV-154 Simplify ITarget hierarchy (part 2)

tpolecat:
This is a large PR, sorry. There was no easy to way to break this up. This PR removes the `CoordinateSystem` superclass and essentially all of its functionality. It provided subclasses with the ability to have a number of "system options" that could be mutated at runtime, but in practice this operation **never** worked correctly and would usually throw an exception. This greatly complicated a lot of things including the target editor. So this mess is gone now. There was also undocumented interaction with catalog and TCC code that depended on particular string values.


cquiroz:
This seems good, nice to have much more deleted code than added :+1:

tpolecat:
Hmm yeah the target editor is being terrible. Need to fix a few things.

tpolecat:
Ok at least one test is failing. Will pick it up tomorrow. :-1:

tpolecat:
Ok I was lying about the test failing. Merging, whee!

PR #199, opened 2015-01-14
OCSADV-207: Regression tests for ITC

fnussber:
Main purpose of this PR is to provide means for substantial regression testing of the ITC app in order to be able to go ahead with some (much needed) aggressive refactoring and simplification of the code in preparation of disentangling the presentation code from the business logic. (The functions doing the bulk of the work in creating the output are each several 100 lines long with many nested if and for statements.)

This PR contains a suite of utilities that allows to create an "ITC baseline" based on a set of defined input values. For creation (updating) and exhaustive testing of the baseline some tests in BaselineTest.scala are provided, which by default are ignored. For each set of input values a hash value for the input values and the expected output value (currently a string, i.e. the html response that is sent to the client) is stored in a text file. A Scala test spec implements tests of a random subset of all defined input values against that baseline and fails if the output of the test run does not correspond with the one on record. This allows for quick testing, as opposed to run an exhaustive test from BaselineTest that checks *all* defined inputs against their baselined outputs (which currently takes about an hour).


tpolecat:
Curious, why are the `Test` classes in `src/main` rather than `src/test`?

fnussber:
These classes have been there before and are not used by the new regression tests, but I had to keep them in sync with the changes I made to the parameter objects. (Also: in the old ant build there was no differentiation between main/test). I am planning to remove those, but I need to have the ok from science first...

cquiroz:
It is absolutely great that you'd tackle this task, it is critical to have a good tests suite especially as nobody seems to understand how ITC works
Why did the `BaselineXXXSpec` become plain `BaselineXXX`? Are all of them called from `BaselineAllSpec`?

fnussber:
First I had a ScalaCheck test for every instrument separately, but then I decided to move the relevant bits to the BaselineAllSpec, so this is the only remaining test executed in automated testing now. All other Baseline files just define the set of input values to be used for the different instruments, so I thought they shouldn't be called *Spec anymore. Nice thing is that ScalaCheck executes tests in a file in parallel by default, too.

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/nici/Test.java
```diff
@@ -73,8 +74,12 @@ public static void main(String[] argv) throws IOException {
...
             
             NiciParameters np =
-                    new NiciParameters("R",
-                    NiciParameters.NDA);
+                    new NiciParameters(
+                        "H",                    // channel1
+                        "ndfilt_clear",         // channel2
+                        "80",                   // pupil
+                        "",                     // instrument mode, unused?
+                        "h5050");               // dichroic position
```
swalker2m: Is the idea that stuff like this will be based on the program model (e.g., `edu.gemini.spModel.gemini.nici.NICIParams`) when you're done refactoring?  So you'd pass in `DichroicWheel.H5050_BEAMSPLITTER` there instead of the `String` "h5050"?

bundle/edu.gemini.itc/src/test/scala/edu/gemini/itc/baseline/util/Baseline.scala
```diff
@@ -0,0 +1,78 @@
...
+    w.flush()
+    Output(o.toString)
+  }
+
+  def checkAgainstBaseline(b: Baseline): Boolean =
+    baseline.get(b.in).map(_ == b.out) match {
+      case Some(true)  => true
+      case Some(false) => false
+      case None        => throw new Exception("Unknown input, try recreating baseline!")
+    }
```
swalker2m: Maybe `baseline.get(b.in).fold(throw new Exception("...")) { _ == b.out }` ?

swalker2m:
Looks like an heroic effort to me.  :+1:

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/nici/Test.java
fnussber: Yes absolutely. I would like to replace as many of those strings as possible by the types from the OCS core and instruments. I might have to make some compromises depending on available time etc. But as long as nobody stops me I am planning to clean this up as much as possible. This is also the main reason why I was pushing for integrating the ITC into the OCS source tree so that I can use that code.

PR #200, opened 2015-01-15
OCSADV-200-C

swalker2m:
This installment of OCSADV-200 introduces ScalaCheck Science Program generators and program edit generators and a collection of `ProgramDiff` test properties.  I've also added ScalaTest as a universal testing option.  As noted in `ProgramDiffPropertyTest`, the testing is a bit unorthodox because of the mutable science program and the time that it takes to generate all the programs and edited versions.

The `ProgramGen` test class is conceivably useful outside of Science Program merge/sync.  Is it worth making generally available?  How would that be done?

(Credit to Rob for [pointing the way](https://gist.github.com/tpolecat/0ba1a178a3881a29a0c6#file-x-scala-L24) with the generated function idea, by the way.)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -45,7 +45,8 @@ libraryDependencies in ThisBuild ++= Seq(
   "junit"           % "junit"           % "4.11"   % "test",
   "com.novocode"    % "junit-interface" % "0.9"    % "test",
   "org.scalacheck" %% "scalacheck"      % "1.10.1" % "test",
-  "org.specs2"     %% "specs2"          % "1.12.3" % "test"
+  "org.specs2"     %% "specs2"          % "1.12.3" % "test",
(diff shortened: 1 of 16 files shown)


tpolecat:
This looks great. :+1: 

Can we put `ProgramIdGen` with the tests for `spModel.core` so it can be reused elsewhere? To use it you would need to change the dependency in `OcsBundle.scala` to look like this:

```scala
bundle_edu_gemini_spModel_core % "test->test;compile->compile"
```

Could also do the same thing by putting `ProgramGen` in `spMode.core`, looks like.

swalker2m:
Moved `ProgramIdGen` to `edu.gemini.spModel.core` and `ProgramGen` to `edu.gemini.pot.sp`.

PR #201, opened 2015-01-15
OCSADV-27: Port of gems parts of AGS to the new model

cquiroz:
This is largish PR containing multiple patches to convert the classes in AGS related to Gems (package `edu.gemini.ags.gems`)

It is mostly mechanical conversion of classes and a few bug fixes discovered during testing
There is some important effort to talk between the Java classes in the `edu.gemini.ags.gems` and the scala classes on `edu.gemini.spModel.core`. `GemsUtils4Java` fills that gap but ideally we should deprecate most of it at some moment

Some simple value classes on the Java side were converted to Scala


bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsUtils4Java.scala
```diff
@@ -1,17 +1,91 @@
...
+  // TODO Star should be replaced by SiderealTarget
+  def starToSiderealTarget(star: Star): SiderealTarget = {
+    val ra = RightAscension.fromAngle(Angle.fromDegrees(star.ra))
+    val dec = Declination.fromAngle(Angle.fromDegrees(star.dec)).getOrElse(Declination.zero)
+    val coords = Coordinates(ra, dec)
+
+    val magnitudes = List((star.bmag, MagnitudeBand.B), (star.vmag, MagnitudeBand.V), (star.rmag, MagnitudeBand.R), (star.jmag, MagnitudeBand.J), (star.hmag, MagnitudeBand.H), (star.kmag, MagnitudeBand.K))
+
+    val invalid = MascotConf.invalidMag
+    val mags = magnitudes.filter(_._1 != invalid).map(v => new Magnitude(v._1, v._2))
```
swalker2m: When I see `x.filter(...).map` I think `collect`.  It doesn't get much shorter but debatably easier to read without the `._1`s.

````scala
magnitudes.collect { case (mag, band) if mag != invalid => new Magnitude(mag, band) }
````

swalker2m:
It's great that you're doing all this.  Requires the patience of saints I imagine.  I didn't study every line in detail but it looks good to me.  :+1:

PR #202, opened 2015-01-15
REL-2197: AGS feedback for VisitorInstrument

sraaphorst:
Added a `VisitorRule` to force P2 AGS checks for visitor instruments.

Also removed the `UnknownError` I previously added and returned when the `Site` couldn't be determined since we don't want an obscure message of `Unknown error.` being displayed for AGS / P2 / QPT when this happens.

Made `AgsAnalysis` methods into `Option[AgsAnalysis]` methods to compensate for the removal to allow for no analysis when the analysis does not make sense / cannot be done.

### DIFF
--- a/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsAnalysis.scala
+++ b/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsAnalysis.scala
@@ -40,10 +40,6 @@ sealed trait AgsAnalysis {
 }
 
 object AgsAnalysis {
-  case object UnknownError extends AgsAnalysis {
-    override def message(withProbe: Boolean): String  = "Unknown error."
-  }
-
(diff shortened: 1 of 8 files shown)


cquiroz:
I have modified the `AgsAnalysis` class as part of my porting work. When this PR is merged on develop there will be conflicts. I'm not sure what would be the best way to handle this
Any ideas?

tpolecat:
@cquiroz you'll need to rebase and resolve the conflict before merging your PR.

sraaphorst:
@cquiroz If you predict that the conflicts will be extensive, perhaps we can discuss the best way to handle this? Let me know.

cquiroz:
I don't know how extensive but some of the signatures have changed to use `SideralTarget` instead of `SPtarget`. Since this PR is needed for the patch release we should likely go ahead and fix the conflicts when we merge into develop

swalker2m:
Looks fine to me.  The pending conflict is unfortunate but I don't see a way around it. :+1:

PR #203, opened 2015-01-16
REL-1301

sraaphorst:
This PR was originally proposed for develop, but it falls under the OCS patch release, so I have closed the previous PR and opened a new one as appropriate.

The issue being addressed here is in the QPT and involves the instrument filters. The filter choices were not preserved when a reset was done on the plan.

The original issue is that a refresh was trigger an entire (and completely unnecessary) rebuild of the Swing tree that holds the filter choices, which resulted in resetting all of the filter choices to their original default values. I originally prevented the rebuilding of the tree once it had already been built for the plan, but it was reported that there were still problems with filter choices being reset.

An investigation revealed that the issue was some ugly hackery in `RefreshAction`, where in order to force-trigger certain updates, two property change events were fired: one with `null` and then one with the original model. This caused the code in `InstViewAdvisor` to think that the model had changed and thus the tree needed to be rebuilt. In order to avoid this, I now check in the property change listener whether or not the new value is `null`, and if it is, I ignore it and wait for the second property change event, which is what we actually care about.


cquiroz:
Looks good

tpolecat:
There's a lot of unfortunate subtlety in the way events fly around in QPT, so good job tracking this down. 

:+1:

PR #204, opened 2015-01-16
OCSADV-27: Improvements to mascot's classes

cquiroz:
I decided to punish myself and improve the mascot classes a bit, in particular:

* the `Star` class contains a lot of redundant target information, now it includes a `SiderealTarget`
* Remove the use of band names as strings and use `MagnitudeBand` instead
* Remove the use of `null` values and use `Option` instead

There is plenty of opportunities to do more improvement but I'll get as far as here give this is not really part of OCSADV-27


tpolecat:
You are awesome.  :+1:

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/mascot/MagLimits.scala
```diff
@@ -1,26 +1,27 @@
...
-  def filter(star: Star): Boolean = {
-    star.bmag <= bmag &&
-      star.vmag <= vmag &&
-      star.rmag <= rmag &&
-      star.jmag <= jmag &&
-      star.hmag <= hmag &&
-      star.kmag <= kmag
-  }
+  def filter(star: Star): Boolean =
+    filters.map(f => star.target.magnitudeIn(f.get.band) <= f).map(Tags.Conjunction).suml
```
swalker2m: Had to puzzle that out!  Assuming a `filters` list without the `Option` wrappers, is this the same as

````scala
filters.forall { m => star.target.magnitudeIn(m.band).exists(_.value <= m.value) }
````

PR #205, opened 2015-01-16
OCSADV-154 Target editor cleanup

tpolecat:
This PR cleans up some of the terrible mess surrounding nonsidereal target editing in the OT. It is still terrible of course but it's substantially easier to follow now. The main culprit was `NonSiderealTargetSupport`, which had a bunch of enums encoded as arrays indexed by constants, and an amazing `String[][][]` for storing information about which fields to show and how to label them for different types of target. So this is all greatly simplified now.

The diff will be hard to follow, sorry. You might just glance at the before and after versions of `NonSiderealTargetSupport`.

### DIFF
--- a/bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
+++ b/bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
@@ -1178,21 +1178,21 @@ public final class EdCompTargetList extends OtItemEditor<ISPObsComponent, Target
                     switch (reply.getObjectType()) {
                         case COMET: {
                             final ConicTarget target = newOrExistingTarget(ITarget.Tag.JPL_MINOR_BODY);
-                            _nonSiderealTargetSup.showNonSiderealTarget(target, ITarget.Tag.JPL_MINOR_BODY);
+                            _nonSiderealTargetSup.showNonSiderealTarget(target);
                             _w.orbitalElementFormat.setValue(_nonSiderealTargetSup.getNonSiderealSystems()[NonSiderealTargetSupport.JPL_COMET]);
                             _curPos.setTarget(target);
                             }
                             break;
                         case MINOR_BODY: {
                             final ConicTarget target = newOrExistingTarget(ITarget.Tag.MPC_MINOR_PLANET);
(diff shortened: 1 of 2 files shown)


bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
```diff
@@ -1079,20 +1029,12 @@ protected HorizonsReply doInBackgroundWithSubject() {
...
-                    service.setObjectType(HorizonsQuery.ObjectType.COMET);
-                } else if (system == _nonSiderealTargetSup.getNonSiderealSystem(ITarget.Tag.MPC_MINOR_PLANET)) {
-                    service.setObjectType(HorizonsQuery.ObjectType.MINOR_BODY);
-                } else if (system == _nonSiderealTargetSup.getNonSiderealSystem(ITarget.Tag.NAMED)) {
-                    service.setObjectType(HorizonsQuery.ObjectType.MAJOR_BODY);
+                ITarget.Tag tag = (ITarget.Tag) _w.orbitalElementFormat.getSelectedItem();
+                switch (tag) {
+                    case JPL_MINOR_BODY:   service.setObjectType(HorizonsQuery.ObjectType.COMET);
+                    case MPC_MINOR_PLANET: service.setObjectType(HorizonsQuery.ObjectType.MINOR_BODY);
+                    case NAMED:            service.setObjectType(HorizonsQuery.ObjectType.MAJOR_BODY);
```
fnussber: With Java switch statements on enums I like having a default case that throws an exception.

swalker2m:
Much better. :+1:

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
tpolecat: In this case the old behavior was to fall through and do nothing. Although it does look like I need `break`s in there doesn't it? ;-)

fnussber: Ahm, ja. That might help ;)

PR #206, opened 2015-01-16
REL-2162

sraaphorst:
We now suppress warnings about missing ODGW and Canopus guide stars in P2 checks and the QPT since these messages are already taken care of by other warnings.


swalker2m:
We all hate special cases but, well, it's a special case.  :+1:  Not sure how to easily generalize that.  There is overlap in general with P2 checks and QPT warnings.  "Someday" we could look into that I guess.

PR #207, opened 2015-01-20
Updated configuration for CfP PW and DT for 2015A

cquiroz:
Small update to p1-monitor configuration, should be integrated to develop too


swalker2m:
:+1:

PR #211, opened 2015-01-21
Hotfix: Revert url for Poor Weather and Directors Time backends to non SV

cquiroz:
Before I forget, I noticed the backend urls are incorrect for Poor Weather and Directors Time.
This patch fixes that


swalker2m:
:+1:

PR #212, opened 2015-01-22
REL-2179: Added missing to/from xml conversion for GSAO filters.

fnussber:
The code here needs an overhaul, it is easy to forget to keep the xml translations in sync when code is added. A type class based approach and some additional testing is on the todo list for later releases. For now, this solves the issue.


swalker2m:
:+1:

PR #213, opened 2015-01-22
OCSADV-154 Collapse coord/non-coord updates for WatchablePos

tpolecat:
OCS has the notion of a `WatchablePos` which is either an `SPTarget` or an offset. When these objects change state they notify listeners by saying "the coordinates changed" or "something else changed". This distinction makes it difficult to implement a generic property change when the [immutable] target is replaced, and it turns out that nobody ever really cares what kind of update it is. So this change collapses this distinction into a single kind of update event.


swalker2m:
Looks good to me. :+1:

PR #214, opened 2015-01-23
REL-2177: Custom QV charts are lost when application is quit.

fnussber:
Another missing filter xml transformation and fixed node names in the xml. Again (as stated with REL-2179): This code is brittle and needs some rework. As part of this bugfix release I am going to check in the smallest set of changes that fixes the problems at hand.


bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/filter/core/FilterXMLFormatter.scala
```diff
@@ -28,9 +28,9 @@ object FilterXMLFormatter {
     <qvTool>
       <filters>{filters.map(format)}</filters>
       <axes>{axes.map(format)}</axes>
-      <charts>{histograms.map(_.toXml)}</charts>
+      <histograms>{histograms.map(_.toXml)}</histograms>
       <tables>{tables.map(_.toXml)}</tables>
-      <visCharts>{barCharts.map(_.toXml)}</visCharts>
```
swalker2m: I'm quite confident that you've fixed the problem, and it looks like more of the same so I'm also sure it's fine.  Still I feel a bit sheepish giving the thumbs up as if I knew what is going on.  Just for my own edification a word or two about what the difference between, for example, labeling this "barCharts" vs "visCharts" actually does would be great.

fnussber: The renaming brings the node names in sync with the parser that tries to find those nodes in the xml.

swalker2m:
:+1:

fnussber:
Abandoning this one because of the mess I made with the last two commits which don't belong here. The relevant changes have been merged in a separate (new and clean) PR.

PR #215, opened 2015-01-23
REL-2214: Fixes in how values in tables are exported (formatting).

fnussber:
Some improvements in how values are formatted when tables are exported from QV.

### DIFF
--- a/bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/table/ObservationTableModel.scala
+++ b/bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/table/ObservationTableModel.scala
@@ -10,6 +10,7 @@ import edu.gemini.skycalc.TimeUtils
 import edu.gemini.spModel.`type`.{DisplayableSpType, LoggableSpType}
 import edu.gemini.spModel.core.{Affiliate, Angle}
 import edu.gemini.spModel.gemini.obscomp.SPSiteQuality.{ElevationConstraintType, TimingWindow}
+import jsky.coords.DMS
 
 import scala.collection.JavaConverters._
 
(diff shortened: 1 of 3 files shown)


bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/table/ObservationTableModel.scala
```diff
@@ -368,20 +369,34 @@ object ObservationTableModel {
   // ==== Wrapper case classes for some values which allow us to define custom renderers and sorting. ===
   case class RaValue(ra: Double) extends Comparable[RaValue] {
     def compareTo(other: RaValue): Int = ra - other.ra match {
-      case d if d < 0 => -1
-      case d if d == 0 => 0
-      case d if d > 0 => 1
```
swalker2m: `math.signum(ra - other.ra).toInt`

Somewhere the `ra` `Double`s are normalized, or we know we're working with normalized values?  Not not for the patch release of course but we can switch this to `edu.gemini.spModel.core.RightAscension` in the future and get normalization, formatting, ordering and more.

fnussber: The values are taken from the mini model which in turn gets them from the OT targets. I was *assuming* that those values are normalized and I have never seen this not being the case, but, it is not enforced in QV.

Do we currently have the case in the OT that we have non-normalized targets? However this would (I think) also cause problems in the QPT.

I am all for using the new target model, this is definitely going to simplify a few things. Unfortunately some of the stuff I would need right now (formatters) is not yet available on the bugfix release branch.

swalker2m:
:+1:

PR #216, opened 2015-01-24
REL-2178: QV is using wrong calendar

fnussber:
The production version of the QV was using the test calendar instead of the production one. Since in the north all relevant events for the current semester have now been added to the test calendar to facilitate planning we have decided to rename the test calendars into the production ones and vice versa and change the config properties accordingly.

I have also moved the url/calendar name properties form the ocs credentials project to here (see http://source.gemini.edu:8060/browse/Software/ocs-credentials/trunk/project/OcsCredentials.scala?r1=60915&r2=61421 ) since these are not really secret. Furthermore I shortened the property name a bit. In theory it would be sufficient to store the ID, the URL can be deduced from it, I wasn't sure in the beginning if this additional flexibility was going to be needed, but for now I leave it as it currently is.

I did a local dry run with ocsDist Linux64 and that seemed to yield the expected configuration. But please verify that the odbproduction config is the right place for this; i.e. these configuration entries will replace the ones from the common config in the production environment.


swalker2m:
This looks fine to me.  It implies redeploying the ODB which I'd hoped to avoid but at least we don't need an export/import cycle. :+1:

fnussber:
Yes, sorry, this configuration is server side unfortunately.

PR #217, opened 2015-01-26
REL-2175: NPE bugfix

swalker2m:
This is a fix for a `NullPointerException` that sometimes occurs in the TPE.  There exists the concept of a history list of image and catalog query results.  In the past this history list would be initialized upon creation of the `HTMLViewer` class in the constructor by loading serialized data from a file.  At some point this was turned off but the history list was unfortunately left altogether uninitialized.

This change simply initializes the history list to an empty list upon creation of the class.  There are many opportunities for improvement in this class, starting with parameterizing the `LinkedList` correctly.  I'm bypassing those opportunities though since I want to touch this code as little as possible for the patch release.


tpolecat:
Ugh. :+1:

PR #218, opened 2015-01-26
OCSADV-154  tidy up SPTarget

tpolecat:
Many methods on `SPTarget` delegate to the contained `ITarget`, which is unnecessary in some cases and nonsensical in others (since many operations are defined only for specific target types). This is the first set of changes and is all "easy" refactorings: inlining, renaming, and in a couple cases pushing methods down. 

It you're reviewing this commit-by-commit note that I reverted and re-did `push getSkycalcCoordinates down to ITarget`.


swalker2m:
I looked through it all and it seems fine to me.  :+1:

PR #220, opened 2015-01-26
Add command to open dash from sbt prompt

tpolecat:
I find myself looking stuff up a lot in Dash and it's kind of clicky. So this lets you do it from the sbt prompt.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -58,10 +58,19 @@ publishArtifact in (ThisBuild, packageSrc) := false
 // No poms
 publishMavenStyle in ThisBuild := false
 
-// Pull stuff up in Dash, if you have it: > dash List
-commands +=
-  Command.single("dash") { (state, topic) =>
-    import scala.sys.process._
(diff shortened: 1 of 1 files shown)


swalker2m:
Two thumbs up.  :+1: :+1: How did you get the scalaz docset, btw?  Generating it from mkscaladocset looks annoying.

tpolecat:
You can generate a docset from anything on Maven Central; it downloads and converts the doc jar. So go to **Preferences > Downloads > Scala Docsets**, search for `scalaz` and then select `org.scalaz:scalaz_2.11 7.1.0` and it will  generate the docset. Takes a minute or two. If you don't see this option you may need to update Dash.

swalker2m:
Excellent, got it.

swalker2m:
Note there's also an Idea Dash plugin which lets you highlight something and do a Dash search.  I had to reassign the keyboard shortcut but otherwise it seems to work.

PR #221, opened 2015-01-27
OCSADV-154  ITarget Cleanup

tpolecat:
This PR removes a bunch of unnecessary methods from `ITarget` and renames `X/Y` and `C1/C2` as `Ra/Dec`. This was almost entirely mechanical refactoring.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/SPTargetPio.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/SPTargetPio.java
@@ -88,8 +88,8 @@ public class SPTargetPio {
             final HmsDegTarget t = (HmsDegTarget) target;
             Pio.addParam(factory, paramSet, _SYSTEM, t.getTag().tccName);
             paramSet.addParam(t.getEpoch().getParam(factory, _EPOCH));
-            Pio.addParam(factory, paramSet, _C1, t.c1ToString());
-            Pio.addParam(factory, paramSet, _C2, t.c2ToString());
+            Pio.addParam(factory, paramSet, _C1, t.getRaHMS());
+            Pio.addParam(factory, paramSet, _C2, t.getDecDMS());
             paramSet.addParam(t.getPM1().getParam(factory, _PM1));
             paramSet.addParam(t.getPM2().getParam(factory, _PM2));
             paramSet.addParam(t.getParallax().getParam(factory, _PARALLAX));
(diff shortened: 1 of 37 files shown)


bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/SPTargetPio.java
```diff
@@ -89,9 +88,8 @@ public static ParamSet getParamSet(final SPTarget spt, final PioFactory factory)
             final HmsDegTarget t = (HmsDegTarget) target;
             Pio.addParam(factory, paramSet, _SYSTEM, t.getTag().tccName);
             paramSet.addParam(t.getEpoch().getParam(factory, _EPOCH));
-            Pio.addParam(factory, paramSet, _BRIGHTNESS, t.getBrightness());
-            Pio.addParam(factory, paramSet, _C1, t.c1ToString());
-            Pio.addParam(factory, paramSet, _C2, t.c2ToString());
+            Pio.addParam(factory, paramSet, _C1, t.getRa().toString());
+            Pio.addParam(factory, paramSet, _C2, t.getDec().toString());
```
swalker2m: I suppose we would migrate the name of the PIO param as well here but since this is going away anyway there's no real need?

tpolecat: Yeah that's what I was thinking. Changing the PIO names would require conversion code for old programs and it doesn't seem worth the trouble; the eventual conversion code will be building completely new types.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/obsComp/TargetObsComp.java
```diff
@@ -90,14 +91,18 @@ public String getTitle() {
...
+                ITarget t = tp.getTarget();
+                String name = t.getName();
+                if (name == null || name.trim().length() == 0) {
+                    name = "<Untitled>";
+                }
+                switch (t.getTag()) {
+                    case JPL_MINOR_BODY:   return "Comet: " + name;
+                    case MPC_MINOR_PLANET: return "Minor Planet: " + name;
+                    case NAMED:            return "Solar System Object: " + name;
+                    case SIDEREAL:         return "Sidereal: " + name;
```
swalker2m: I like this change. I suppose we could get some flak because it is different but it makes sense to me.

tpolecat: The `position` method is vestigial and made no sense at all (see the old impl on `ConicTarget` below). I talked to Andy and we decided to try this and see what people say. If we had s distinct icon for each target type we could get rid of the "type" prefix. I opened an issue for this.

swalker2m:
Looks good. :+1:

PR #223, opened 2015-01-28
OCSADV-200-D: (preliminary) preliminary merge algorithm

swalker2m:
This PR is not ready to be merged.  :-1:  I'm posting it early in order to get feedback on the `PreliminaryMerge` algorithm and the tree of `MergeNode`s that it creates.  You can focus on the last commit if you have time to review this now.  I will continue cleaning it up and testing regardless.

The basic idea is to pair up the `Diff` nodes retrieved from the remote version of a program with their local `ISPNode` science program counterparts and come up with merged `Diff` nodes based upon a comparison of the vector clocks.  A merged `Diff.Present` will choose the data object and children from the local program and/or the remote diff according to the vector clock comparison.  A merged `Diff.Missing` is simply an update to the deleted node's vector clock

We then traverse the merged `Diff.Present`s from the root down, following the children in a DF traversal and building a `MergeNode` tree.  There are two types of `MergeNode`: `ModifiedNode` and `UnmodifiedNode`.  `ModifiedNode`s are created for every surviving (i.e., reachable from the root) merged `Diff.Present`.  When a `ModifiedNode` has a child for which there was no `Diff.Present`, the child is represented as an `UnmodifiedNode` leaf which simply refers to the matching `ISPNode`.  I realize this probably makes little or no sense as described.  Perhaps a picture would help ...

The `MergeNode` tree will ultimately be manipulated to handle special cases like observation renumbering.  This PR will cover just up to the initial creation of the `MergeNode` tree with no additional corrections.  In other words, a preliminary merge (see `PreliminaryMerge`).

### DIFF
--- a/bundle/edu.gemini.pot/src/main/scala/edu/gemini/pot/sp/version/VersionComparison.scala
+++ b/bundle/edu.gemini.pot/src/main/scala/edu/gemini/pot/sp/version/VersionComparison.scala
@@ -1,0 +1,46 @@
+package edu.gemini.pot.sp.version
+
+/** A comparison between two NodeVersions or VersionMaps.
+  *
+  * This class simply provides a more readable enum of the version comparison
+  * outcomes.
+  */
+sealed trait VersionComparison
+
+import scalaz._
+import Scalaz._
+
+object VersionComparison {
+  case object Same        extends VersionComparison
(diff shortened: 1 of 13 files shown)


tpolecat:
I will continue looking at this. It's kind of involved, as you noted in your message above.

PR #224, opened 2015-01-29
OCSADV-209: Simplifying parsing of dat files and caching of results.

fnussber:
The original ITC code loaded and parsed the data resource files every time anew when they were needed. This task was mainly about making sure that *.dat resource files are loaded and parsed only once and the resulting data is cached. In the wake of doing that I also cleaned some of the code and unified some copy/pasted things I came across (of which there are many..).

ITC was using a home grown TextFileReader which was replaced with the standard java.util.Scanner for parsing. Using Scala parse/combinators turned out to be very inefficient for some of the larger *.dat files (~230000 lines of pairs of doubles) and is maybe a bit overkill for such simple file structures anyways, so I choose a more "manual" and maybe less elegant but much more efficient approach based on the Scanner class; see very last incarnation of DatFile.scala for details.

Currently some of the caching is done outside of the DatFile object. I am planning to move some/most/all of those caches into the DatFile object and/or replace some of them (is it really necessary to have config files for everything or could we hard code some of those values, e.g. for instrument configs? ==> I will discuss that with Andy at some point). But I think I will do that in separate tasks in order to not inflate this task anymore, because it is enormous as it is already anyways.

The good news: For my current test set with ~5000 different configurations covering all instruments execution time went down from 75 to 20 minutes.

Regarding the memory consumption of the caching: There is a total of 700 dat files left (hopefully I will be able to eliminate some more) with a total of ~58M lines. Assuming that most of the dat files are of the form of two doubles per line (wavelength + transmission values) this results in an upper bound of approximately ~928MB of memory in case all files were cached. Currently the caching is implemented using Scalaz's Memo.immutableHashMapMemo. I will keep an eye on how memory consumption looks like in reality and then maybe change that to weak hash maps or use floats for the dat files etc. as needed. On the other hand, as always my opinion in such matters is, that memory is cheap and should be plentiful..

As an additional note: I am still playing around with parser combinators just because the resulting code is quite a bit nicer. I am down from 4* slower to 2* slower, which is still annoying when executing single tests but over all - assuming that files in production are only read once/occasionally - probably doesn't matter too much. So a future version of DatFile might go back to parser combinators.


swalker2m:
I can't go through all the changes but it's clearly getting better from what I've seen.  You're doing the lord's work, so to speak. (And thanks for taking my name out of the headers!) :+1:

fnussber:
My hope is, that over time the PRs will become more targeted, smaller and easier to review, right now the problem is that touching something almost always means touching a lot of other stuff, so it's all getting very confusing. Thanks to my tests I am nevertheless reasonably confident that I am not totally crippling the whole application without noticing it. You may not quote me on that one.

PR #225, opened 2015-01-29
OCSADV-154  remove SPTarget mutators

tpolecat:
This PR removes the remaining mutators on `SPTarget` with the exception of those associated with proper motion. These will be pushed down into `HmsDegTarget` in a distinct PR. The implementations have been inlined and in some cases include an explicit call to tell the `SPTarget` to notify listeners. This is terrible but it will go away soon enough; all target mods will soon be done via `setTarget`.

### DIFF
--- a/bundle/edu.gemini.oodb.too.url/src/main/java/edu/gemini/spdb/rapidtoo/impl/TooUpdateServiceImpl.java
+++ b/bundle/edu.gemini.oodb.too.url/src/main/java/edu/gemini/spdb/rapidtoo/impl/TooUpdateServiceImpl.java
@@ -238,11 +238,10 @@ public final class TooUpdateServiceImpl implements TooUpdateService {
         SPTarget base = targetEnv.getBase();
 
         base.getTarget().setName(tooTarget.getName());
-        base.notifyOfGenericUpdate();
         base.getTarget().getRa().setAs(tooTarget.getRa(), CoordinateParam.Units.DEGREES);
         base.getTarget().getDec().setAs(tooTarget.getDec(), CoordinateParam.Units.DEGREES);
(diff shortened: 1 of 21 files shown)


bundle/edu.gemini.oodb.too.url/src/main/java/edu/gemini/spdb/rapidtoo/impl/TooUpdateServiceImpl.java
```diff
@@ -236,9 +237,12 @@ private void _setTarget(ISPObservation obs, TooUpdate update)  {
...
 
-        base.setName(tooTarget.getName());
-        base.setRaDecDegrees(tooTarget.getRa(), tooTarget.getDec());
-        base.setMagnitudes(tooTarget.getMagnitudes());
+        base.getTarget().setName(tooTarget.getName());
+        base.notifyOfGenericUpdate();
+        base.getTarget().getRa().setAs(tooTarget.getRa(), CoordinateParam.Units.DEGREES);
+        base.getTarget().getDec().setAs(tooTarget.getDec(), CoordinateParam.Units.DEGREES);
+        base.notifyOfGenericUpdate();
+        base.getTarget().setMagnitudes(tooTarget.getMagnitudes());
```
swalker2m: I suppose it is irrelevant given where this is going, but why not just a single `base.notifyOfGenericUpdate()` at the end?

tpolecat: Yeah it's a consequence of inlining. I will go through and clean those up.

swalker2m:
:+1:

PR #226, opened 2015-01-31
OCSADV-215: Use enums and types from spModel in TeleParameters

fnussber:
As a long term goal all (or at least most) stringly typed values in ITC should be replaced with the corresponding values from spModel.

Apart from general type-safety this will also make it easier the keep ITC in sync with changes in the model (e.g. new instrument capabilities like filters etc) and integrate it into the OT and PIT. Furthermore I want to use the enums for a more automated approach in the translation from and to string values for the HTML pages using the enum class and value names. This should also simplify a potential future rewrite of the web front end since it will be easier to create option drop downs etc from the enums in the spModel.

This is a first step exercising these changes with the TeleParameters object. This PR also includes a rework of the Exception handling, mainly a change from checked exceptions to unchecked ones and the removal of catching and re-throwing exceptions when parsing input parameter objects. Missing parameter exceptions (which result from errors in the html pages) etc are now allowed to bubble all the way up to the servlet before being caught. All user relevant configuration problem exceptions (e.g invalid instrument configurations) have been changed into IllegalArgumentExceptions and are presented to the user as before.

Another bunch of obsolete data files has been identified and eliminated. And occasionally some more code has been cleaned up here and there.

Additional remark: I did some manual testing and found an error that went undetected by the current tests. In addition to fixing the error I also improved the test infrastructure to also take the resulting spectra files into account when calculating / comparing the baselines. Hopefully this will catch any future unintentional changes.

Another add-on: I fixed a problem which caused sometimes the same chart to be displayed twice instead of the two different ones for spectroscopy. This was caused by a race condition in the servlet (filename was "shared" between different threads executing doGet()). This includes also a first refactoring/simplification step of the charts in the recipes.


fnussber:
Anyone wants to comment on this PR? I know it's crazy big, mostly because it just contains a lot of unrelated/unplanned but somehow necessary refactorings I stumbled upon. I am not expecting anyone to go through all of that, but if you're interested in some commits that are exemplary for the work I am doing right now check out:

TeleParameter object goes typed:
https://github.com/fnussber/ocs/commit/bd9d07dea7dabfc0366de5ad663d0f278f482748
Clean up exception handling:
https://github.com/fnussber/ocs/commit/f5f9b85a04ca4fe8f863e6674a0d8fe63b37534b
https://github.com/fnussber/ocs/commit/a7077abd103d8e38e23486874f106eafa35b6952
Bug fix (race condition) and code simplification:
https://github.com/fnussber/ocs/commit/337652b4fec2d3e60e6a2d82b0ab068f4eeaa6e2

cquiroz:
Thanks for the pointers, I checked the commits and certainly I'm happy to transform strings to types and better exception handling

PR #229, opened 2015-02-02
REL-2200: Expose Graces

cquiroz:
Graces is an option on 2015B, this PR makes it available to end users


tpolecat:
I really admire the way you have tests for all these things. It's great. :+1:

PR #230, opened 2015-02-03
REL-1841: Update GRACES PIT options

cquiroz:
This PR contains two updates to the Graces model:

* Rename of fiber modes
* Addition of read mode

The PR takes care of the xml, model and decision tree changes plus tests

### DIFF
--- a/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/dtree/inst/Graces.scala
+++ b/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/dtree/inst/Graces.scala
@@ -2,24 +2,38 @@ package edu.gemini.model.p1.dtree.inst
 
 import edu.gemini.model.p1.dtree._
 import edu.gemini.model.p1.immutable._
-import scala.Some
 
 object Graces {
   def apply() = new FiberNode
 
-  class FiberNode extends SingleSelectNode[Unit, GracesFiberMode, GracesBlueprint](()) {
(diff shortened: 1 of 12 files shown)


cquiroz:
Comments?

arturog8m:
Everything seems consistent at least to my eyes! :+1: Others can comment on the scala code in more detail, but looks fairly straightforward.

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/dtree/inst/Graces.scala
```diff
@@ -22,4 +21,19 @@ object Graces {
     }
   }
 
+  class ReadNode(fm: GracesFiberMode)  extends SingleSelectNode[GracesFiberMode, GracesReadMode, GracesBlueprint](fm) {
```
swalker2m: Typo here, I presume.  Should be `ReadMode`.

swalker2m:
Other than the typo, everything looks fine to me to.  :+1:

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/dtree/inst/Graces.scala
cquiroz: Actually it refers to the Node on the decision tree, like the previous one is called `FiberNode`
Perhaps a better name is `ReadModeNode`

PR #232, opened 2015-02-04
REL-2032: Do not submit to partners with zero requested time

cquiroz:
Updated a problem robot to forbid submissions when any partner's request is zero

Th relevant commit is cc574d4, the rest is code improvements and formatting

### DIFF
--- a/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
+++ b/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
@@ -468,7 +468,7 @@ object TimeProblems {
         } yield for {
             ps <- sub
             if ps.request.time.value <= 0.0
-          } yield new Problem(Severity.Error, s"Please specify a time request for partner ${Partners.name.getOrElse(ps.partner, "")}", SCHEDULING_SECTION, s.showPartnersView())
+          } yield new Problem(Severity.Error, s"Please specify a time request for ${Partners.name.getOrElse(ps.partner, "")} or remove partner", SCHEDULING_SECTION, s.showPartnersView())
       probs.right.getOrElse(Nil)
     case _                            => Nil
   }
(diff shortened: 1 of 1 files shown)


bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
```diff
@@ -459,10 +459,20 @@ object TimeProblems {
...
-}
 
-import ProblemRobot._
-import TimeProblems._
+  // REL-2032 Check that none of the requested times per partner are zero
+  def partnerZeroTimeRequest(p: Proposal, s:ShellAdvisor): List[ProblemRobot.Problem] = p.proposalClass match {
+    case g:GeminiNormalProposalClass =>
+      val probs = for {
+          sub <- g.subs.swap.right
+        } yield for {
```
sraaphorst: Can someone explain to me why you would do a:
`for { ... } yield for { ... } yield ...`?
I'm not sure what this does. Is it equivalent to a nested loop in this case so the result is an iterable of iterable of Problem?

sraaphorst:
I'm not familiar with the logic, but the cleanup looks great. :+1:

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
cquiroz: The first one gets you out of the `Either`, the second iterates over sub which is a `List[NgoSubmission]`

tpolecat: Might be easier to read without the `for` syntax. It's just 

```scala
g.subs.swap.right.map { sub =>
  sub.filter(_.request.time.value <= 0.0).map(ps => new Problem(...)))
}
```

swalker2m:
:+1:

PR #233, opened 2015-02-05
REL-2177B: Custom QV charts are lost when application is quit (2)

fnussber:
When loading bar charts from disk the lookup for the "dynamic" axes for Observations and Programs failed because they are currently not part of the default set of axes. (This could/should probably be changed, but when I tried to do that I run into some side-effects which I did not want to deal with in a bugfix release.)

In order to fix this for now, the two axes are manually added to the axes lookup table used to resolve axes by names when a bar chart is loaded from disk.


cquiroz:
Looks good :+1:

PR #234, opened 2015-02-05
Added NSHighResolutionCapable=true in Info.plist for OSX

fnussber:
Adding NSHighResolutionCapable in Info.plist turns on high resolution support for OSX which results in properly rendered fonts instead of fuzzy fluffiness on retina displays. Icons are still scaled up and would need to be replaced with icons with higher resolution. However having crisp texts already goes a long way. I tested this with a locally edited version, but we should probably do a proper test with a properly built OT version before releasing this.

@swalker2m @cquiroz : Is this going to take care of OT, QPT and PIT?

### DIFF
--- a/project/src/main/scala/edu/gemini/osgi/tools/app/MacDistHandler.scala
+++ b/project/src/main/scala/edu/gemini/osgi/tools/app/MacDistHandler.scala
@@ -124,8 +124,8 @@ case class InfoPlist(meta: ApplicationMeta, version:String, fwJar: File, mainCla
         </array>
         <key>WorkingDirectory</key>
         <string>$APP_ROOT/Contents/Resources</string>
-	<key>NSHighResolutionCapable</key>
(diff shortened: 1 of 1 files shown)


cquiroz:
Awesome, truly awesome. It probably was quite a detective work, thanks for taking the time to research it.
We should incorporate this change to the PIT and QPT :+1:

sraaphorst:
Glad someone was able to get to the bottom of this mystery! :+1:

tpolecat:
This is great. @fnussber can you do a sanity check on PIT and QPT?

fnussber:
I tested this with a locally built OT, QPT and PIT and it solves the issue for all three apps. Hopefully there are no anticipated side effects, but I think it should not influence the rendering other than render fonts and lines etc properly. I opened some dialogs including the TPE and it all seemed to look ok.

PR #235, opened 2015-02-05
REL-1391: Increase proposal size limit and give helpful error when too large

cquiroz:
Simple PR containing a new `ProblemRobot` for the PIT


swalker2m:
Looks good. :+1:

PR #236, opened 2015-02-05
REL-1298: Show GN/GS observation time sums on PDF summary cover page

cquiroz:
This request requires changes to the xsl used to generate pdfs out of p1 templates.
Tests are included that check the intermediate FOP representation


arturog8m:
I see the changes to the NOAO and the default templates. Don't we have an Australian template as well?

cquiroz:
No, there are a few customizations for Australia on the default template

arturog8m:
Everything looks good to go :+1:

PR #237, opened 2015-02-06
REL-2226: Add CFHT as a new exchange request type

cquiroz:
This PR contains updates to the model and UI to include CFHT as an Exchange Partner for Classical Proposals


bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/editor/BlueprintEditor.scala
```diff
@@ -31,6 +31,7 @@ object BlueprintEditor {
       case e:ExchangeProposalClass => e.partner match {
         case ExchangePartner.KECK   => Site.Keck
         case ExchangePartner.SUBARU => Site.Subaru
+        case ExchangePartner.KECK   => Site.CFHT // Shouldn't happen
```
fnussber: This looks fishy, KECK probably shouldn't be twice here(?).

cquiroz: Good catch, this is not logically reachable code but it should be correct in any case

swalker2m:
Looks good. :+1:

PR #238, opened 2015-02-06
REL-2231: Display full PIT version in compatibility error message

cquiroz:
Improve parsing of the backend response to give a better error message to the user


swalker2m:
:+1:

PR #242, opened 2015-02-07
REL-2177C: Additional bug fixes for dealing with dynamic axes.

fnussber:
Another bug fix for QV. Hopefully it can still make it into the bug fix release. It has been tested by Kristin with a test version I created locally for her.

This is a set of fixes that deal with problems with dynamic or "on-the-fly" axes in QV. (An axis in QV is a sequence of filters that is used to define a sequence of bins for an axis of a histogram or table. The same concept btw. is used in QV for color coding in graphs, i.e. splitting the data into several bins which each are differentiated by a different color.)

First the changes in this PR address a problem that is caused by the fact that dynamic axes are actually stubs that don't contain any filters because those are created on-the-fly based on the currently available data (e.g. the available observations if separate bins for all available observations are needed). This caused a problem with creation of the axis editors which rely on at least one filter being available. However since dynamic axes are not meant to be user editable anyway this problem was addressed by differentiating between editable and non-editable axes and not displaying an edit button nor creating an axis editor for dynamic axes.

The second problem addressed with this PR is that QV has a set of "default" axes which are meant to be a collection of basic axes, that can not be edited. Those axes (and other default elements) are not meant to be imported/exported in order to save space and also to make sure they are always present/valid. Unfortunately older versions of QV incorrectly exported the dynamic axes "Observations" and "Programs" which after importing will cause a second axis with the same name to appear in QV. This is not very grave but annoying. For now these elements are filtered out on import as a temporary solution; this will go away/be replaced with a more permanent solution.

A more permanent solution for the second issue will hopefully be implement as part of the next release. Making the dynamic axes default axes (which they actually are) will also allow to simplify the code in a few places. See TODOs in code.

### DIFF
--- a/bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/QvStore.scala
+++ b/bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/QvStore.scala
@@ -171,13 +171,17 @@ object QvStore extends Publisher {
     JOptionPane.showMessageDialog(null, msg, "QV Defaults Error", JOptionPane.ERROR_MESSAGE)
   }
 
+  // TODO: On import we should make sure that default elements are not overridden.
   def loadFromFile(file: File) = {
     for {
       xml <- Try(XML.load(file.getAbsolutePath))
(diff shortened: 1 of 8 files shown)


cquiroz:
Seems good and it is nice that you mark the places where further work is needed.

PR #243, opened 2015-02-10
REL-2030: Invalid character causes proposal to be rejected

cquiroz:
Quick PR, sometimes people edit their proposals by hand and break the XML, this PR takes care of filtering out non valid xml files

### DIFF
--- a/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/ProposalIo.scala
+++ b/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/ProposalIo.scala
@@ -50,15 +50,24 @@ object ProposalIo {
   def writeToXml(p: Proposal): Node =
     XML.loadString(writeToString(p)) // there is probably a more direct route ...
 
-  //private def tryLatin1(f: File)
-
   private def toProposal(r: ConversionResult): ProposalConversion = ProposalConversion(r.transformed, r.from, r.changes, toProposal(unmarshaller.unmarshal(new StringReader(r.root.toString))))
 
-  def readAndConvert(s: String):Validation[Either[Exception, NonEmptyList[String]], ProposalConversion] = try {
-      UpConverter.upConvert(XML.loadString(s)).fold(Right(_).fail, toProposal(_).success)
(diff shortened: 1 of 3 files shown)


bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/ProposalIo.scala
```diff
@@ -50,15 +50,26 @@ object ProposalIo {
...
-  //private def tryLatin1(f: File)
-
   private def toProposal(r: ConversionResult): ProposalConversion = ProposalConversion(r.transformed, r.from, r.changes, toProposal(unmarshaller.unmarshal(new StringReader(r.root.toString))))
 
-  def readAndConvert(s: String):Validation[Either[Exception, NonEmptyList[String]], ProposalConversion] = try {
-      UpConverter.upConvert(XML.loadString(s)).fold(Right(_).fail, toProposal(_).success)
+  // Not all UFT-8 characters are valid in XML, strip them if needed
+  // Reference http://www.w3.org/TR/2000/REC-xml-20001006#NT-Char
+  private def stripNonValidXMLCharacters(in: String) = in.filter {
+    case c if c == 0x9                      => true
```
sraaphorst: I'm not sure why these single cases have the form:
`case c if c == 0x9 => true`
Wouldn't it be simpler to do:
`case 0x9 => true`?

(Or even: `case c if c in List(0x9, 0xA, 0xD) => true` and just combine them?)

tpolecat:
```scala
case 0x9 | 0xA | 0xD => true
```

sraaphorst: Yup. That makes more sense. I always forget about that syntax.

cquiroz: Thanks for the suggestions, I updated the code acoordingly

swalker2m:
:+1:

PR #244, opened 2015-02-10
OCSADV-154  push proper motion into HmsDegTarget

tpolecat:
This PR pushes proper motion parameters from `ITarget` down into `HmsDegTarget`, which is the only subtype where they make sense. Consequences are as follows:

- Had to rewrite part of `GeneralRule.WFS_RULE` so I added a test for it.
- Discovered a bug in `clone()` for all target types that caused the RA and Dec to be shared with the previous instance. This was probably introduced by me in an earlier refactoring. Fixed and added tests for cloning.
- Several OT editors had to be updated to deal with the possibility that their target might not have proper motion. Mostly trivial changes here.


cquiroz:
Seems very good :+1:

bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/general/GeneralRule.java
```diff
@@ -241,8 +252,30 @@ public IP2Problems check(ObservationElements elements)  {
...
+
+                case SIDEREAL:
+                    final HmsDegTarget hms1 = (HmsDegTarget) t1;
+                    final HmsDegTarget hms2 = (HmsDegTarget) t2;
+                    return hasSameCoordinates(hms1, hms2) &&
+                           hasSameProperMotion(hms1, hms2) &&
+                           hasSameTrackingDetails(hms1, hms2);
+
+                default:
+                    throw new Error("Unpossible target tag: " + t1.getTag());
```
swalker2m: "Unpossible", a perfectly cromulent word.

PR #245, opened 2015-02-11
REL-2246: hack to work around bug in hdiutil

swalker2m:
This is a hack to work around an error encountered while creating a Mac `.dmg` for the QPT.  

    release/ocs-2015A.1.1.2 * 19$ hdiutil create -srcfolder /Users/swalker/dev/ocs/app/qpt/target/qpt/2015A-test.1.1.2/MacOS/mac-test/Gemini\ QPT\ 2015A-test.1.1.2.app -volname qpt_2015A-test.1.1.2_macos /Users/swalker/dev/ocs/app/qpt/target/qpt/2015A-test.1.1.2/MacOS/mac-test/qpt_2015A-test.1.1.2_macos.dmg
    .....................................................................................................................................................................................
    hdiutil: create failed - error -5341

Googling that one lead me to this [sketchy “solution”](http://stackoverflow.com/questions/18621467/error-creating-disk-image-using-hdutil) which is basically to add a .Trash folder to the root directory.  This change is not end-user visible and the .Trash folder doesn't even appear to end up in the installed application directory.

### DIFF
--- a/project/src/main/scala/edu/gemini/osgi/tools/app/MacDistHandler.scala
+++ b/project/src/main/scala/edu/gemini/osgi/tools/app/MacDistHandler.scala
@@ -19,6 +19,7 @@ case class MacDistHandler(jre: Option[String], jreName: String) extends DistHand
     // Output dirs
     val name = meta.osxVisibleName(version)
     val appDir = mkdir(outDir,  s"$name.app")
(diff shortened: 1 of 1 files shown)


cquiroz:
Do you think that is due to a change in e.g. Yosemite?
Anyway it should be harmless, so let's merge it :smile:

swalker2m:
This seems to have been an issue for a long time now and we were just never bitten.

PR #246, opened 2015-02-11
Prevent RPM binary stripping

cquiroz:
QPT RPM failed on execution. It was noted that the jar files are different on the rpm than on the plain distribution.

This effect was traced to the binary stripping step for rpm
http://livecipher.blogspot.com/2012/06/disable-binary-stripping-in-rpmbuild.html

This PR changes .rpmmacros to prevent binary stripping


swalker2m:
Beautiful. :+1:

PR #247, opened 2015-02-11
OCSADV-200-D: preliminary merge algorithm

swalker2m:
This is the latest in a series of PRs to address updating the merge / sync algorithm to work with program differences.  This installment is focused on the "preliminary merge" phase, which refers to the idea that the remote differences are combined with the local science program considering only the vector clocks.  This ignores permissions and can even produce invalid observations so a series of corrections must be applied to the preliminary merge, but that will be the focus of future PRs.

I wrote a bit of documentation on [Program Synchronization](https://github.com/gemini-hlsw/ocs/wiki/Science-Program-Synchronization-Notes) that may help orient a review.  A few additional notes:

* This is unfortunately rather large.  The last commit is the important one and you could focus on `PreliminaryMerge` there to speed things up if you like.
* `MergeRecord` is not used and shouldn't have been included.  I didn't want to mess with taking it out and plan on evolving it in the future.  Please ignore.
* I changed `ProgramDiff` to work with a `scalaz.Tree[MergeNode]` because I realized there was a lot of overlap between the difference information and the preliminary merge result.
* The lord's anger burned against me and I was lost wandering in the wilderness for a while so I ended up squashing the last meandering commits into one mega commit.  Sorry about that.
* I believe this is the worst, most convoluted part of the algorithm.  The good news is that the remainder will focus on the output of this step, the `scalaz.Tree[MergeNode]` and the preliminary merge can be revisited without impacting subsequent steps to some degree.


cquiroz:
Thanks for the documentation, this will be very useful. 
A couple of questions:
* After you do a sync, wouldn't the two merged `NodeVersions` produce one with the highest edit counter + 1?
* Is there some locking on the remote in case two simultaneous edits are attempted?

swalker2m:
> After you do a sync, wouldn't the two merged `NodeVersions` produce one with the highest edit counter + 1?

You can't glean any information from comparing the edit counter for program instance `a` vs the counter for program instance `b`.  There was an idea at one point to make an edit bump the counter from `n` to `max(all edit counters) + 1` instead of just `n + 1` but I could never figure out how to make it work.  Might be worth revisiting because you could use that information to know when you have pending changes without having to have the remote `NodeVersions` to compare with.

So the sync produces a new `NodeVersions` with the highest count for all known program instances to indicate that it has taken into account any corresponding changes.  You definitely don't want to bump any counters unless you have something new to consider.  Otherwise, once something is touched it would always appear out of date on a sync.  You'd send the changes back to the remote ODB and it's counters would be bumped, etc.

I know all of this is terribly confusing and realize it might be a sign that it isn't very well done.

swalker2m:
> Is there some locking on the remote in case two simultaneous edits are attempted?

In order to commit something to the remote database, your `VersionMap` as a whole has to be strictly newer than the remote program.  If it isn't, someone has slipped in changes after you grabbed the diffs but before you could send your changes back.  In that case, your "push" is rejected and you have to sync again.  The OT automates that process because it can happen more often than you'd think as datasets are being added to the program during observing.

cquiroz:
It is difficult to comment on the details, it seems well thought and I know it has been extensively tested

PR #248, opened 2015-02-12
Export edu.gemini.catalog.votable package

cquiroz:
SPDB is broken due to a missing package on the catalog bundle, this PR fixes that without bringing other non-completed AGS work


swalker2m:
:+1:

PR #249, opened 2015-02-17
OCSADV-200-E: observation renumbering

swalker2m:
This is the first of a series of corrections to be applied to the preliminary merge.  A `MergeCorrection` is just a function `MergePlan => MergePlan` that potentially modifies the provided plan in some way to correct for a flaw produced by the preliminary merge.  In this particular instance, the issue is that observations created locally can be given the same observation numbers as those created independently in the remote program.  Any such observations must be renumbered to be greater than the maximum remote observation number.  Note that no `ISPObservation` instances themselves are actually changed in this process.  Updates to the `ISPNodes` will be handled when the `MergePlan` is actually applied in a later step.

### DIFF
--- a/bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs/diff/ObsNumberCorrection.scala
+++ b/bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs/diff/ObsNumberCorrection.scala
@@ -1,0 +1,84 @@
+package edu.gemini.sp.vcs.diff
+
+import edu.gemini.pot.sp.SPNodeKey
+import edu.gemini.sp.vcs.diff.NodeDetail.Obs
+import edu.gemini.sp.vcs.diff.ProgramLocation.{Remote, Local}
+import edu.gemini.spModel.rich.pot.sp._
+
+import scalaz._
+import Scalaz._
+
+/** Corrects observation numbers such that all local-only observations have
+  * numbers greater than the maximum remote observation number.
+  *
+  * There are a few implicit assumptions at work here.  First, once an
+  * observation has been created its key is always remembered in the
+  * `VersionMap` (that's true of any node actually).  Second, as new
+  * observations are created they are given the next sequential integer.
+  * That is, observation numbers aren't reused when deleted.  Finally
+  * observation numbers are never modified once assigned (other than by
+  * the merge plan application itself).
+  *
+  * This means that there cannot exist an observation known to both local and
+  * remote program versions with an observation number greater than a
+  * local-only observation.
+  */
+class ObsNumberCorrection(isKnown: (ProgramLocation, SPNodeKey) => Boolean) extends MergeCorrection {
+  def apply(mp: MergePlan): MergePlan = {
+    val t = (mp.update.loc/:renumberedObs(mp)) { case (loc, (key, num)) =>
+      loc.find(_.tree.rootLabel.key === key).fold(loc) { obsLoc =>
+        val mergeNode = obsLoc.tree.rootLabel match {
+          case Modified(k,nv,dob,Obs(_)) => Modified(k, nv, dob, Obs(num))
+          case lab                       => lab
+        }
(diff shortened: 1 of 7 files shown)


bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs/diff/ObsNumberCorrection.scala
```diff
@@ -0,0 +1,84 @@
...
+  def apply(mp: MergePlan): MergePlan = {
+    val t = (mp.update.loc/:renumberedObs(mp)) { case (loc, (key, num)) =>
+      loc.find(_.tree.rootLabel.key === key).fold(loc) { obsLoc =>
+        val mergeNode = obsLoc.tree.rootLabel match {
+          case Modified(k,nv,dob,Obs(_)) => Modified(k, nv, dob, Obs(num))
+          case lab                       => lab
+        }
+        obsLoc.setLabel(mergeNode).root
+      }
+    }.toTree
```
swalker2m: I'm not pleased with this bit.  Here we take the `Tree[MergeNode]` and for each observation to be renumbered, we find it in the tree and update the observation number.  Instead of repeatedly finding nodes in the tree from the root, it seems like we should be able to walk the tree applying corrections but every attempt I made was awkward.

cquiroz:
What happens to already executed observations? Wouldn't a number change affect the already taken images?

swalker2m:
In general you're right, there is nothing that prevents an already executed observation from being renumbered.  In practice that cannot happen since an observation must be in the ODB in order to be executed.  Once in the ODB an observation won't be renumbered unless we initiate a sync from the ODB to a remote database instead of the other way around.  If we assume that observations can be executed anywhere then we also have a problem.  I'm not sure what to do about that though.  I had thought that all the corrections would work for any `MergePlan` but I suppose I could make `MergeCorrection` be a function `MergePlan => Unmergeable \/ MergePlan` or something and then sequence the series of corrections.

cquiroz:
Cannot comment on the details of the code, but it seems good and with plenty of tests :+1:

cquiroz:
How is execution guaranteed to be done only from the DB? Because the TCC and Seqexec will only talk to the ODB via WDBA?

swalker2m:
Right, we cannot support execution in multiple databases with this codebase.  That is, though you could configure everything to work with a different database (and we do for testing of course), we have no mechanism for working with multiple production ODBs at a given site.

bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs/diff/ObsNumberCorrection.scala
```diff
@@ -0,0 +1,84 @@
...
+    val or = mp.update.foldObservations(ObsRenum(None, List.empty)) { (mod, i, or) =>
+      val k = mod.key
+      (isKnown(Local, k), isKnown(Remote, k)) match {
+        case (false, true) => or.addRemote(i)
+        case (true, false) => or.addLocal(k, i)
+        case _             => or
+      }
+    }
+
+    or.maxRemote.fold(Set.empty[(SPNodeKey, Int)]) { max =>
```
tpolecat: You can use `foldMap` here since you're folding into a monoid and starting with its zero.

tpolecat:
As we discussed on iChat I think I would just go for expressiveness initially and use `map` and `foldRight` on your tree rather than trying to optimize the traversals. But either way this looks good to me. :+1:

cquiroz:
Could you, in principle, mark a local observation as executed?

swalker2m:
Yes but that shouldn't be an issue.  The issue would be when there are datasets or events in the `ObsExecLog`.  That's when it should count as "executed".

swalker2m:
Updated to handle the unexpected case of locally-executed observations plus simplification of the renumbering to eliminate the use of the zipper in favor of a `map` over the `Tree[MergeNode]`.

PR #250, opened 2015-02-17
REL-1841: Changes to Graces fiber modes' names

cquiroz:
This PR contains renaming of Graces fiber modes and updates to test accordingly


swalker2m:
:+1: I suppose a new Graces class is coming to the SP Model soon?

cquiroz:
Ehh, we probably need to create a task for that along the templates...

PR #253, opened 2015-02-17
REL-2032: Do not submit to partners with zero requested time

cquiroz:
On this PR the UI has been changed so when a PI selects a 0 time for a partner it is the same as removing the partner
The other change will open the partner time editor if the zero partner time problem is selected


swalker2m:
Looks reasonable to me. :+1:

PR #254, opened 2015-02-17
REL-2231: Display full PIT version in compatibility error message

cquiroz:
Some small changes to the way the error message is presented to the user if the backends are incompatible with the PIT version


bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/view/submit/SubmitStatus.scala
```diff
@@ -41,12 +41,16 @@ object SubmitStatus {
 
   def submissionClosed(destination: Option[SubmitDestination]) = s"The ${destinationName(destination)} proposal server is currently closed. Please see the Gemini web pages for proposal submission dates"
 
-  private val ErrorRegex = """.* (\d\d\d\d.) ([\d\.]*).*""".r // Extract semester and version from the backend error message
+  private val VersionMismatchRegex = """.* (\d\d\d\d)([A|B]) \d{4}(\.[\d\.]*).*""".r // Extract semester and version from the backend error message
+
+  def parseServiceError405(destination: Option[SubmitDestination], message: String) = {
+    val VersionMismatchRegex(year, sem, version) = message
```
tpolecat: If this doesn't match you get a `MatchError` ... do you need to handle this case?

cquiroz: Good point, this assume the backend is well behaved, which is not a 100% solid assumption

cquiroz:
Fixed to consider @tpolecat comment

tpolecat:
:+1:

PR #255, opened 2015-02-18
REL-2248: Limited partners and joint proposals

cquiroz:
This PR fulfills a request to mark certain partners as un-elegible for joint proposals.
No changes to the model are required, just a new Problem Robot

### DIFF
--- a/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
+++ b/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
@@ -496,6 +496,24 @@ case class TimeProblems(p: Proposal, s: ShellAdvisor) {
     case _ => None
   }
   lazy val b3ReqOrZero = b3Req.map(_.time).getOrElse(TimeAmount.empty)
+  lazy val jointNotAllowed = {
+    def checkForNotAllowedJointProposals(subs: Option[List[NgoSubmission]]):List[Option[Problem]] = {
+      val r = for {
+          subs <- subs
+          if subs.size > 1
+        } yield for {
+            p <- subs.filter(s => Partners.jointProposalNotAllowed.contains(s.partner))
+          } yield new Problem(Severity.Error, s"${~Partners.name.get(p.partner)} cannot be part of a joint proposal, please update the time request.", SCHEDULING_SECTION,
+              s.inPartnersView(_.editSubmissionTime(p)))
+      r.sequence
+    }
+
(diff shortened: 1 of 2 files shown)


bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
```diff
@@ -496,6 +496,24 @@ case class TimeProblems(p: Proposal, s: ShellAdvisor) {
...
+              s.inPartnersView(_.editSubmissionTime(p)))
+      r.sequence
+    }
+
+    p.proposalClass match {
+      case p: GeminiNormalProposalClass => checkForNotAllowedJointProposals(p.subs.left.toOption)
+      case e: ExchangeProposalClass     => checkForNotAllowedJointProposals(e.subs.some)
+      case x                            => println("abc " + x);Nil
+    }
+  }
```
swalker2m: * I'm not that familiar with how the robot works but why does the check return a `List[Option[Problem]]` instead of just a `List[Problem]`.  I think just `.flatten`ing the return value would get rid of the options.

* It may be just a question of preferences but I always find the nested `for` expressions confusing.  Since we're filtering and mapping, would `collect` be easier to follow?  Maybe something like:

````scala
val r = subs.collect {
  case Some(subList) if subList.size > 1 =>
    subList.collect {
      case sub if Partners.jointProposalNotAllowed.contains(sub.partner) =>  
        new Problem(...)
    }
}
````

* `println("abc " + x);` looks like a bit of debugging left over?

swalker2m:
Beyond the trivial code questions, it seems unfortunate that we leave this to problem checking instead of prohibiting it in the UI.  Ideally if you added time for Brazil or whatever KR would be disabled and vice versa.  I realize that's a much more difficult thing to achieve though, especially at the last minute.

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
cquiroz: Hi, right a List[Problem] will do and certainly debugging will be removed
Funny, I find clearer the nested `for` (no more than 2 levels)  than a combination of operations, should we have some guidelines about that?

cquiroz:
Hi, could we bring this idea into discussion with science, I don't think it is impossible on this time frame and there is always the fallback to `Problem`

cquiroz:
I added the suggested code changes

swalker2m:
:+1:

PR #257, opened 2015-02-18
OCSADV-216: Split HTTP request parsing from parameter objects

fnussber:
Large PR which focuses primarily on splitting the HTTP request parsing from the main parameter objects (```ObservationDetailsParameters```, ```TeleParameters```, ```ObservingConditionParameters```, ```PlottingDetailsParameters```, ```SourceDefinitionParameters```, ```AltairParameters```) and introduces additional typing (objects/enums) to reduce the dependency on string typed values.

While this was meant to be a stroll in the park it turned out to be much more involving and included a lot of clean-up tasks. I also ran into some "special cases", which kept me in some places from doing the straight forward thing I had in mind and forced me to put some workarounds into place which need to be cleaned up further at a later stage (e.g. some ugly hacks in ```MichelleRecipe``` and ```TRecsRecipe``` that originally changed some values in ```ObservationDetailsParameters``` on the fly and needed to be replaced by some more localized hack that allowed to turn ```ObservationDetailsParameters``` into an immutable object, see 
https://github.com/fnussber/ocs/commit/daf32149060447996896cc4a0105c57ff6b4abb9.).

And on and on it goes. I spare you the details. It's all pretty much brain dead work and a lot of time goes into trying to make sure not to change the actual calculations in order to keep everything working as is.

Cleanups include:
* Removal of unused help pages / documents
* Removal of unused altairNiri servlet
* Removal of support for NICI (this was never released, code is in unknown/useless state)
* Removal of some obsolete/unused parameters

If you want to get an idea of what I am doing here, check out the transition of ```ObservingConditionParameters``` into an object that uses the conditions from the spModel: 
https://github.com/fnussber/ocs/commit/118f735e7e5c39febd9dcde826dadaf8ad5b2214

All in all I am not very proud of what I am doing here, many of the changes need more work. But I have to strike a balance between touching some stuff and making incremental improvements over trying to be too thorough and get into a state where I have too many changes at once.

Again, I hope future PRs will become more focused; I do think that things are getting more and more streamlined, and if nothing else at least I am building up my mental model of the ITC code which also helps with future refactorings.


swalker2m:
10 lines of code = 10 issues
2000 lines of code = :+1: 

Joking aside, I can't review all of this but it is clearly going in the right direction.  Also, it's great to see so much unused junk going away.

cquiroz:
@shane well put, I skimmed a few commits and it is so nice to see 200+K lines of deleted code. I'd say go ahead with your changes

PR #258, opened 2015-02-19
OCSADV-27: AGS on OT UI using VoTable catalogues

cquiroz:
This is a large PR where a big portion of AGS is ported to use the new VoTable catalogues. Before this PR the output of AGS should have remain identical to the previous one but after AGS will produce different results (They should be close though). Summary of changes:

* Ported `SingleProbeStrategy` to use VoTable on the OT AGS
* Reenabled Gems AGS related tests to ensure no regression is broken
* Rewrite of `GemsCatalog` in scala on the class `GemsVoTableCatalog`. This took quite a bit of work and the commits were squashed into commit 9fc2621
* Proper error handling

We should likely not merge this into develop until it has been validated by science but we can review the code.

Future steps:
* Convert `GemsStrategy` to use VoTable on the AGS Servlet
* Unify the API used by the OT for AGS (Right now the OT uses different APIs and UI for Gems and other instruments)
* Add query caching
* Restore `StatusLogger` functionality
* `rm -f GemsCatalog.java`


bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/SingleProbeStrategy.scala
```diff
@@ -94,8 +95,10 @@ case class SingleProbeStrategy(key: AgsStrategyKey, params: SingleProbeStrategyP
       case (angle, st) => new CandidateValidator(params, mt, List(st)).exists(ctx.withPositionAngle(angle.toOldModel))
     }
 
-  override def queryConstraints(ctx: ObsContext, mt: MagnitudeTable): List[QueryConstraint] =
-    params.queryConstraints(ctx, mt).toList
+  override def queryConstraints(ctx: ObsContext, mt: MagnitudeTable): List[QueryConstraint] = Nil
```
swalker2m: I take it we don't need this method anymore?  If so, we should probably remove it from the `AgsStrategy` API altogether.

swalker2m:
@sraaphorst I think you should take a close look at this PR.  It seems possible to generate conflicts with what you're doing.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/SingleProbeStrategy.scala
cquiroz: Right, I replaced `queryConstraints` by `catalogQueries` but didn't review the API
Will do on the next PR

cquiroz:
For @sraaphorst: There are no API changes to `AgsStrategy` but if you use `GemsCatalog` directly it has been deprecated and `GemsVoTableCatalog` is there instead. In the next PR I pretend that all access to AGS is via `AgsStrategy` though

cquiroz:
Rebased onto develop after merging release changes

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsVoTableCatalog.scala
```diff
@@ -0,0 +1,164 @@
...
+
+    val resultSequence = inst match {
+      case GemsInstrument.flamingos2 => searchCatalog(basePosition, criterions, statusLogger)
+      case i                         => searchOptimized(basePosition, criterions, i, statusLogger)
+    }
+
+    // sort on criteria order
+    resultSequence.map(_.toList.sortWith({
+      case (x, y) =>
+        criterions.indexOf(x.criterion) < criterions.indexOf(y.criterion)
```
sraaphorst: Is it just me, or is doing a call to `indexOf` for each comparison really inefficient? I'm not sure that it matters based on the sizes of `criterions`, but perhaps these indices should be precomputed as this looks like it'll make the sort `2 x^2 log(x)` complexity as opposed to `x log(x)` if they are precomputed.

cquiroz: criterions is normally of size 2

sraaphorst:
I can't see this generating any conflicts with my PR and I think that it should be fine. I've looked at this in detail now, and I'm going to give it a :+1: .

PR #259, opened 2015-02-19
REL-2226: Add CFHT as a new exchange request type

cquiroz:
This PR lets both classical and queue proposal to get requests from CFHT

### DIFF
--- a/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/view/partner/PartnerView.scala
+++ b/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/view/partner/PartnerView.scala
@@ -887,16 +887,6 @@ class PartnerView extends BorderPanel with BoundView[Proposal] {view =>
         // Enabled?
         enabled = canEdit
 
-        // Hide CFHT for queue proposals
-        m.map {
-          case _:QueueProposalClass     =>
-            new DefaultComboBoxModel(PartnerType.values.filterNot(_ == ExchangeCFHT).toArray)
-          case _:ClassicalProposalClass =>
-            new DefaultComboBoxModel(PartnerType.values.toArray)
-          case _                        =>
-            this.peer.getModel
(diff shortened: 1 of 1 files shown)


bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/view/partner/PartnerView.scala
```diff
@@ -889,9 +889,7 @@ class PartnerView extends BorderPanel with BoundView[Proposal] {view =>
 
         // Hide CFHT for queue proposals
```
swalker2m: I suppose this comment is out of date then?

cquiroz:
Actually it could be simplified even further, I updated the commit to reflect this

swalker2m:
:+1:

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/view/partner/PartnerView.scala
```diff
@@ -887,16 +887,6 @@ class PartnerView extends BorderPanel with BoundView[Proposal] {view =>
...
 
-        // Hide CFHT for queue proposals
-        m.map {
-          case _:QueueProposalClass     =>
-            new DefaultComboBoxModel(PartnerType.values.filterNot(_ == ExchangeCFHT).toArray)
-          case _:ClassicalProposalClass =>
-            new DefaultComboBoxModel(PartnerType.values.toArray)
-          case _                        =>
-            this.peer.getModel
-        }.map(this.peer.setModel)
```
tpolecat: Use `foreach` if you're going to side-effect; this ensures that it happens now and exactly once. `map` doesn't guarantee either.

cquiroz: good hin

PR #260, opened 2015-02-19
REL-2252: Double prompt when attempting to close the PIT

cquiroz:
Quick PR to fix a small bug closing the PIT, It will ask to confirm twice if you say no the first time


tpolecat:
It's remarkably hard to get "quit" to do the right thing.  :+1:

cquiroz:
It turns out this is OSX-only but the fix doesn't affect Linux nor Windows

PR #262, opened 2015-02-20
REL-2253: TimeRequest dialog box Ok button needs to be pressed twice

cquiroz:
One of those bugs...

### DIFF
--- a/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/util/NumberField.scala
+++ b/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/util/NumberField.scala
@@ -18,9 +18,8 @@ object NumberField {
 }
 
 class NumberField(d: Option[Double]) extends FormattedTextField(NumberField.df) with SelectOnFocus {
-  d.orElse(Some(0))foreach { d =>
-    import NumberField.df
-    text = df.format(d)
+  d.orElse(Some(0)).foreach { d =>
+    text = NumberField.df.format(d)
     commitEdit()
   }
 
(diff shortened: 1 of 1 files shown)


bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/util/NumberField.scala
```diff
@@ -39,20 +38,22 @@ class NumberField(d: Option[Double]) extends FormattedTextField(NumberField.df)
...
 
-      valid = try {
-        val pp = new ParsePosition(0)
-        valid(NumberField.df.parse(text, pp).doubleValue) && (pp.getIndex == text.length())
-      } catch {
-        case _:Exception => false
+      // REL-2253 For some reason on OSX you get sometimes an extra value change event with an empty text.
+      // This creates a null parsed value and the validation fails
+      // This is not very noticeable as there is a second event with the valid text
+      // However, it means that sometimes you need to press Ok twice
```
swalker2m: What happens now if you delete all the text, which presumably legitimately produces an event with empty text?

cquiroz:
Good question, It does in fact break. Given the time constraints I'll close the PR and leave the bug open

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/util/NumberField.scala
```diff
@@ -39,20 +38,22 @@ class NumberField(d: Option[Double]) extends FormattedTextField(NumberField.df)
...
+        background = if (valid) white else pink
       }
-
-      background = if (valid) white else pink
-
   }
 
-
-
   focusLostBehavior = FormattedTextField.FocusLostBehavior.Commit // ?
```
arturog8m: Why we have a question mark in this line, as a comment?

PR #263, opened 2015-02-20
OCSADV-200-F: `MergePlan` application

swalker2m:
In this installment, I switched the focus from merge corrections to applying the `MergePlan`.  This edits the program according to the merge results.  This step is a bit premature because the `MergePlan` is still lacking corrections required to prevent trying to make illegal programs (for example observations with two sequence nodes, etc.).  I thought it would be useful to get a vertical slice through the problem completed first though and then come back to merge corrections.  In the tests, I just ignore illegal merge plans for now.

As usual for me, there is a bit of meandering in the commits.  The important part to focus upon is the update to `MergePlan.merge` itself.  That is where the changes are applied.  As usual with mutation, this is awkward and I don't know how to do it correctly.  Should I be stuffing this into `IO` or something?

Otherwise there is a bit of ugliness down in `pot.sp`.  I replaced `MemAbstractContainer.TypedChildren.getChild` with `getOnlyChild` which does the same thing as before but throws an exception if it is not in fact the only child of the given type.  I think this is what we want because it represents a logic error.  Also I just went ahead and exposed `setObservationNumber` on `ISPObservation`.  Once all of this is over, I'll go back and remove the other truly awful method for renumbering observations which is in `ISPProgram.renumberObservationsToMatch`.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/rich/pot/sp/RichProgram.scala
+++ b/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/rich/pot/sp/RichProgram.scala
@@ -1,34 +1,49 @@
 package edu.gemini.spModel.rich.pot.sp
 
+import edu.gemini.pot.sp._
 import edu.gemini.spModel.gemini.obscomp.SPProgram
 import scala.collection.JavaConverters._
-import edu.gemini.pot.sp.{ISPObservation, ISPProgram}
 
 class RichProgram(prog:ISPProgram) {
   def spProgram:Option[SPProgram] = prog.dataObject.map(_.asInstanceOf[SPProgram])
 
-  def update(f:SPProgram => Unit) {
-    spProgram foreach {dataObj =>
+  def update(f:SPProgram => Unit): Unit =
+    spProgram.foreach { dataObj =>
       f(dataObj)
       prog.dataObject = dataObj
     }
-  }
 
   def allObservations:List[ISPObservation] = prog.getAllObservations.asScala.toList
 
   def obsByLibraryId(lid:String):Either[String, ISPObservation] =
-    allObservations.find(_.libraryId.map(_ == lid).getOrElse(false)).toRight("Observation with library id '%s' was not found.".format(lid))
+    allObservations.find(_.libraryId.exists(_ == lid)).toRight("Observation with library id '%s' was not found.".format(lid))
 
   def obsByLibraryIds(lids:Seq[String]):Either[String, List[ISPObservation]] = {
     val empty:Either[String, List[ISPObservation]] = Right(Nil)
-    (empty /: lids) {
-      (eos, lid) =>
-        for {
-          os <- eos.right
-          o <- obsByLibraryId(lid).right
-        } yield (o :: os)
+    (empty/:lids) { (eos, lid) =>
+      for {
+        os <- eos.right
+        o <- obsByLibraryId(lid).right
+      } yield o :: os
     }
   }
 
+  def copy(fact: ISPFactory): ISPProgram = {
+    val sp = fact.createProgram(prog.getNodeKey, prog.getProgramID)
 
+    def init(src: ISPNode, dest: ISPNode): Unit = {
+      dest.dataObject = src.dataObject
+      dest.children   = src.children.map(copy)
+    }
+
+    def copy(src: ISPNode): ISPNode = {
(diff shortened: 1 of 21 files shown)


bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/rich/pot/sp/RichProgram.scala
```diff
@@ -1,34 +1,30 @@
...
       f(dataObj)
       prog.dataObject = dataObj
     }
-  }
 
   def allObservations:List[ISPObservation] = prog.getAllObservations.asScala.toList
 
   def obsByLibraryId(lid:String):Either[String, ISPObservation] =
-    allObservations.find(_.libraryId.map(_ == lid).getOrElse(false)).toRight("Observation with library id '%s' was not found.".format(lid))
+    allObservations.find(_.libraryId.exists(_ == lid)).toRight("Observation with library id '%s' was not found.".format(lid))
```
cquiroz: I guess you can use string interpolation here

bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs/diff/MergePlan.scala
```diff
@@ -1,14 +1,112 @@
...
   * merge.
   */
 case class MergePlan(update: Tree[MergeNode], delete: Set[Missing]) {
 
+  /** "Encode" for serialization. The issue is that `scalaz.Tree` is not
+    * `Serializable` but we need to send `MergePlan`s over `trpc`.
+    */
+  def encode: MergePlan.Transport = {
+    def encodeTree(t: Tree[MergeNode]): MergePlan.TreeTransport =
+      MergePlan.TreeTransport(t.rootLabel, t.subForest.toList.map(encodeTree))
```
cquiroz: It seems `subForest` is a `Stream[Tree[A]]`, would it be more performant not to materialize the stream and try computing on the stream as much as possible?

bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs/diff/MergePlan.scala
```diff
@@ -1,14 +1,112 @@
...
+        case Modified(k, _, dob, _) =>
+          NodeFactory.mkNode(f, p, dob.getType, Some(k)).toRightDisjunction {
+            Unexpected("Could not create science program node of type: " + dob.getType)
+          }
+        case Unmodified(k)          =>
+          Unexpected(s"Unmodified node with key $k not found in program ${p.getProgramID}.").left
+      }
+
+    // Edit the ISPNode, applying the changes in the MergeNode if any.
+    def edit(t: Tree[(MergeNode, ISPNode)]): Unit = {
```
cquiroz: Is this recursive but with return type `Unit`?

cquiroz:
Seems good as much as I can follow. I'll trust the tests cover the functionality :+1:

bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs/diff/MergePlan.scala
swalker2m: The fact that `scalaz.Tree` is based on `Stream` is sort of unfortunate for this use case.  I don't think there is anywhere I can take advantage of it.  In particular here I need to build something that can be serialized and sent over `trpc`.  In our version of Scala at least `Stream` is not serializable.

bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs/diff/MergePlan.scala
swalker2m: Yes.  It's walking the tree mutating the `ISPNode` according to the matching `MergeNode`.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/rich/pot/sp/RichProgram.scala
swalker2m: You're right, I should have updated that along the way.  I touched this file to add a method only to later repent and take it back out.  I did a few obvious improvements along the way but I probably should have just rolled back all the changes to keep it from distracting from the rest.

PR #264, opened 2015-02-20
REL-2030: Prepare PIT 2015B release

cquiroz:
Version changes for the PIT release


bundle/edu.gemini.model.p1.submit/src/main/scala/edu/gemini/model/p1/submit/package.scala
```diff
@@ -36,11 +36,7 @@ package object submit {
     // REL-1257 It was requested to support simultaneous SV and non SV backends
     val specialProposalsTestUrls = for {
       st      <- SpecialProposalType.values.toSeq
-    } yield if (st == SpecialProposalType.DIRECTORS_TIME || st == SpecialProposalType.POOR_WEATHER) {
-        Special(st) -> url(s"${st.name}_sv")
-      }  else {
-        Special(st) -> url(st.name)
-      }
```
swalker2m: Is this what the comment above referring to REL-1257 is about?  If so I guess we can remove that.

cquiroz: Right, documentation is as always stale, fixed now

swalker2m:
:+1:

PR #267, opened 2015-02-20
OCSADV-196: Vignetting for GMOS

sraaphorst:
Sorry all... this is going to be horrid to slog through, as it represents quite a bit of work all thrown in at once.
This should handle the guide probe vignetting for GMOS, but I still need to write test / specification cases.
Any feedback, though, until then would be most appreciated!


cquiroz:
Before reviewing we should consider that we have 3 ongoing big code changes touching AGS pending:

* The merge from release to develop
* PR#258
* This PR

There are already conflicts on the first merge. I suggest we do the merge and then rebase both PRs before continuing

tpolecat:
@cquiroz this seems sensible to me. I also recommend making a backup before rebasing; it's easy to screw it up when you have conflicts.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/gmos/GmosOiwfsProbeArm.scala
```diff
@@ -0,0 +1,142 @@
...
+import edu.gemini.skycalc.Offset
+import edu.gemini.spModel.core.Coordinates
+import edu.gemini.spModel.inst.{ArmAdjustment, ProbeArmGeometry}
+import edu.gemini.spModel.obs.context.ObsContext
+import edu.gemini.spModel.telescope.IssPort
+
+object GmosOiwfsProbeArm extends ProbeArmGeometry {
+  import edu.gemini.spModel.inst.FeatureGeometry.transformPoint
+
+  val instance = this
```
swalker2m: Is this for easier / cleaner access from Java?

sraaphorst: Yes. There may be an easier way to do it, but I was struggling to find it.

tpolecat: In the worst case you can write a Java interface that provides the API you want to see from Java, then implement it with a Scala class that your Java code can instantiate and use. Depends on your situation; may not be worth it.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/SingleProbeStrategy.scala
```diff
@@ -120,4 +125,34 @@ object SingleProbeStrategy {
...
+   * @param lst              the list of possible candidates
+   * @param mt               magnitude lookup table
+   * @param ctx              context information
+   * @param probe            the guide probe
+   * @param toSiderealTarget converts candidates to SiderealTargets
+   * @tparam A               the type of the candidates
+   * @return                 Some(guideStar) if a candidate exists that can be used, None otherwise
+   */
+  def brightestByQualityAndVignetting[A](lst: List[A], mt: MagnitudeTable, ctx: ObsContext, probe: ValidatableGuideProbe with VignettingGuideProbe)(toSiderealTarget: A => SiderealTarget): Option[A] = {
+    if (lst.isEmpty) None
```
tpolecat: Do you need this case? If `lst` is empty then `candidates` will be empty and `reduceOption` will give you `None`.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/gmos/GmosOiwfsProbeArm.scala
```diff
@@ -0,0 +1,142 @@
...
+          case _                    => None
+        }
+      }
+      wfsOffset.map { w =>
+        val angle = armAngle(w, posAngle, guideStarPt, offsetPt)
+        ArmAdjustment(angle, guideStarPt)
+      }
+    }
+  }
+
```
tpolecat: In cases like this you're usually better off lifting the nullable values into `Option` and proceeding from there with your entire computation in `Option`. Something like this, perhaps

```scala
override def armAdjustment(ctx0: ObsContext, guideStarCoords: Coordinates, offset0: Offset): Option[ArmAdjustment] =
  for {
    ctx    <- Option(ctx0)
    offset <- Option(offset0)
    w      <- ctx.getInstrument match {
                case gmosn: InstGmosNorth => Some(gmosn.getFPUnit.getWFSOffset)
                case gmoss: InstGmosSouth => Some(gmoss.getFPUnit.getWFSOffset)
                case _                    => None
              }
  } yield {
    val flip = if (ctx.getIssPort == IssPort.SIDE_LOOKING) -1.0 else 1.0
    val posAngle = ctx.getPositionAngle.toRadians.getMagnitude
    val offsetPt = new Point2D.Double(-offset.p.toArcsecs.getMagnitude, -offset.q.toArcsecs.getMagnitude * flip)
    val guideStarPt = {
      val baseCoords = ctx.getBaseCoordinates.toNewModel
      val o = Coordinates.difference(baseCoords, guideStarCoords).offset
      val (x,y) = (o.p.toArcsecs, o.q.toArcsecs)
      new Point2D.Double(-x, -y * flip).toCanonicalForm
    }
    val angle = armAngle(w, posAngle, guideStarPt, offsetPt)
    ArmAdjustment(angle, guideStarPt)
  }
```

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/inst/ProbeArmGeometry.scala
```diff
@@ -0,0 +1,109 @@
...
+   * @param offset the offset for which to calculate the adjustment
+   * @return       the probe arm adjustments for this data
+   */
+  def armAdjustment(ctx: ObsContext, offset: Offset): Option[ArmAdjustment] =
+    (for {
+      c            <- Option(ctx)
+      guideTargets <- ctx.getTargets.getPrimaryGuideProbeTargets(guideProbeInstance).asScalaOpt
+      guideStar    <- guideTargets.getPrimary.asScalaOpt
+    } yield armAdjustment(ctx, guideStar, offset)).flatten
+
```
tpolecat:
```scala
for {
 c            <- Option(ctx)
 guideTargets <- ctx.getTargets.getPrimaryGuideProbeTargets(guideProbeInstance).asScalaOpt
 guideStar    <- guideTargets.getPrimary.asScalaOpt
 adj          <- armAdjustment(ctx, guideStar, offset)
} yield adj
```

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/gmos/GmosScienceAreaGeometry.scala
```diff
@@ -0,0 +1,226 @@
...
+      inst.getFPUnitMode match {
+        case GmosCommonType.FPUnitMode.BUILTIN if inst.isImaging       => List(fov(ImagingFOVSize, ImagingFOVInnerSize))
+        case GmosCommonType.FPUnitMode.BUILTIN if inst.isSpectroscopic => List(longSlitFOV(width))
+        case GmosCommonType.FPUnitMode.BUILTIN if inst.isIFU           => List(ifuFOV(inst.getFPUnit, isSouth))
+        case GmosCommonType.FPUnitMode.BUILTIN if inst.isNS            => List(nsFOV(width))
+        case GmosCommonType.FPUnitMode.CUSTOM_MASK                     => List(fov(MOSFOVSize, MOSFOVInnerSize))
+        case _                                                         => Nil
+      }
+    }
+  }
```
tpolecat: Here you could eliminate the `if` by saying

```scala
Option(inst).toList.flatMap { inst => 
  lazy val width = ...
  ...
}
```



```

**Comment by tpolecat** [comment]
Hmm GitHub is eating my comments for some reason. Just a few stylistic things as described. I don't understand what this code is doing but it seems fine to me. You mentioned that testing might be difficult ... what's the strategy for determining whether it's correct or not?

File: bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/SingleProbeStrategy.scala (hunk 2 above)
Comment (sraaphorst) [inline bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/SingleProbeStrategy.scala]: Good point. The `if`...`else` seems entirely unnecessary.

File: bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/gmos/GmosOiwfsProbeArm.scala (hunk 3 above)
Comment (sraaphorst) [inline bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/gmos/GmosOiwfsProbeArm.scala]: Nice! That's much cleaner. I like it. Thanks!

File: bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/inst/ProbeArmGeometry.scala (hunk 4 above)
Comment (sraaphorst) [inline bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/inst/ProbeArmGeometry.scala]: I feel a bit dim not having seen that :+1:

File: bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/gmos/GmosOiwfsProbeArm.scala (hunk 1 above)
Comment (swalker2m) [inline bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/gmos/GmosOiwfsProbeArm.scala]: A comment `// for Java access` or something might help.  It didn't occur to me at first why we'd do this but it makes sense.

Comment (sraaphorst) [inline bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/gmos/GmosOiwfsProbeArm.scala]: Sure. I think this is easiest, so I'll go ahead and add the comment.

File: bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/inst/ProbeArmGeometry.scala [hunk 6]
```diff
@@ -0,0 +1,109 @@
...
+      new Point2D.Double(p.getX.toCanonicalValue, p.getY.toCanonicalValue)
+  }
+}
+
+/**
+ * A representation of the adjustment made to the default list of shapes when using a specified guide star.
+ * @param angle           the angle which will be used by the probe arm
+ * @param guideStarCoords the coordinates (in arcsec) where the probe arm will be placed
+ */
+case class ArmAdjustment(angle: Double, guideStarCoords: Point2D)
```
swalker2m: Would it be a good idea to represent this with `edu.gemini.spModel.core.Angle` instead of a raw `Double`?  A `Double` value can be anything but an `Angle.toDegrees` will always give you a value [0, 360).

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/inst/FeatureGeometry.scala
```diff
@@ -0,0 +1,161 @@
...
+   *
+   *
+   */
+  def findObsContextOffset(ctx: ObsContext, ox: Double, oy: Double, bx: Double, by: Double, pixelsPerArcsec: Double): Option[Offset] = {
+    import scala.math.abs
+
+    // Convert from screen coordinates to (p,q) coordinates in arcseconds
+    val pq = List(bx - ox, by - oy).map(_ / pixelsPerArcsec)
+
+    // If the offset is (0,0), which is not included in the ObsContext's science positions, create a new Offset.
```
swalker2m: If there are no offset positions then I believe the "science positions" should be a singleton list containing (0,0).  It isn't clear to me why the first branch of the `if` is necessary.

sraaphorst: I'm pretty sure that I've encountered cases where `ObsContext.getSciencePositions` returns an empty `Set<Offset>`, which causes problems, so this is supposed to work around that. This may have been when there was no `Offset` iterator in the sequence, so I guess I need to ask: will there always be an `Offset` iterator in the sequence?

sraaphorst: BTW, I'm investigating now to find out when this happens. Are there cases where we wouldn't want (0,0) to be an valid offset? I'm not clear on that, but I imagine that it could be the case.

swalker2m: Definitely you cannot rely on having an offset iterator in the sequence.  The intention is that there should always be at least one "science position" according to my reading of the code [here](https://github.com/gemini-hlsw/ocs/blob/develop/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/offset/OffsetUtil.java#L123-L133) and [here](https://github.com/gemini-hlsw/ocs/blob/develop/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/obs/context/ObsContext.java#L122-L129). This code is not great though and it wouldn't surprise me if you are right.  We should look into this if you can confirm that you're getting an `ObsContext` without science positions.  It'll probably bite us elsewhere otherwise.

swalker2m:
This is tricky stuff by nature but the code seems clear and nicely commented to explain what is going on. I think once the tests we discussed are ready and the minor stylistic things @tpolecat pointed out are done, we're good to go with this.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/SingleProbeStrategy.scala
```diff
@@ -75,7 +76,11 @@ case class SingleProbeStrategy(key: AgsStrategyKey, params: SingleProbeStrategyP
         case FIXED_180 | PARALLACTIC_ANGLE => selectBounded(List(ctx, ctx180(ctx)), mt, candidates)
         case UNBOUNDED                     => selectUnbounded(ctx, mt, candidates)
       }
-      brightest(results, params.band)(_._2).map {
+      println(s"===resultsSize=${results.size}")
```
cquiroz: Some leftover printout?

bundle/edu.gemini.pot/src/main/scala/edu/gemini/pot/ModelConverters.scala
```diff
@@ -0,0 +1,141 @@
...
+
+  implicit class NewMagnitude2Old(val m: Magnitude) extends AnyVal {
+    def toOldModel: skyobject.Magnitude = new skyobject.Magnitude(m.band.toOldModel, m.value)
+  }
+
+  implicit class HmsDegCoords2Coordinates(val c: skyobject.coords.HmsDegCoordinates) extends AnyVal {
+    def toNewModel: Coordinates = Coordinates(RightAscension.fromAngle(c.getRa.toNewModel), Declination.fromAngle(c.getDec.toNewModel).getOrElse(Declination.zero))
+  }
+
+  implicit class SiderealTarget2SkyObject(val st:SiderealTarget) extends AnyVal {
```
cquiroz: On the 3 target converts we are missing translating the proper motion. There is code on AGS to do that but it must be ported to this class once it is merged into the main trunk

cquiroz:
As other have mentioned I can't follow the calculation but this seems good with @tpolecat and @swalker2m suggestions
I'm worried that the AGS work may collide with these changes. I'd like this to be merged sooner rather than later but I can't tell if you are ready to do so.

PR #269, opened 2015-02-23
Fixed broken tests when changing ServiceError signature

cquiroz:
A broken test related to phase-1 went undetected, this PR fixes that


tpolecat:
:+1:

PR #270, opened 2015-02-23
OCSADV-218: Simplifications in fixtures, parallel execution.

fnussber:
Some cleanup and improvements in the ITC tests, in particular parallel execution of the baseline creation and exhaustive comparison tests ```BaselineTest.create()``` and ```BaselineTest.checkAll()```.


cquiroz:
I skimmed through the code and it seems good. You can see sort of repeated patterns on the different fixtures, is there some amount of code that could be de-duplicated?

fnussber:
There's probably some room for improvement. However the fixtures are in subtle ways specific to each instrument: TReCS has special rules for its observation conditions, the wavelength of the source, the instrument filters and graters must match etc. etc. To a certain degree the fixture will always end up being a (long) enumeration of instrument specific configurations that are valid. I started to share the sources between the instruments. I might come up with more simplifications like that. One other thing is that I hope to be able to simplify the instrument configurations; some/all of them have (a lot) of unused junk parameters that can go away, which in turn will simplify the fixtures.

PR #275, opened 2015-02-26
OCSADV-227: Integrate VoTable based Gems queries into GemsStrategy

cquiroz:
This PR makes `GemsStrategy` use VoTable queries for its internal calculation
A bug was found on the search method that prevented `GemsStrategy` to work, apparently this has not been noticed as the OT calls the `GemsVoTableCatalog` class directly for Gems AGS


bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsVoTableCatalog.scala
```diff
@@ -136,11 +136,14 @@ object GemsVoTableCatalog {
 
   // Sets the min/max magnitude limits in the given query arguments
   protected [gems] def optimizeMagnitudeLimits(criterions: List[GemsCatalogSearchCriterion]): List[MagnitudeConstraints] = {
+    val magConstraints = for {
+        criteria <- criterions
+      } yield criteria.criterion.magConstraints
```
swalker2m: I'm partial to explicitly using `map` and `flatMap` unless you start to chain together too many in one expression.  So I would write that:

`val magConstraints = criterions.map(_.criterion.magConstraints)`

but it is debatable which is clearer.  In the steps that follow though, `magConstraints` get flattened twice.  Could we do that once here?  Now I might break out `for`:

````scala
val magConstraints = for {
  criteria <- criterions
  mc       <- criteria.criterion.magConstraints
} yield mc
````

File: bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsVoTableCatalog.scala [hunk 2]
```diff
@@ -69,7 +69,7 @@ object GemsVoTableCatalog {
   private def searchCatalog(basePosition: Coordinates, criterions: List[GemsCatalogSearchCriterion], statusLogger: StatusLogger): Future[List[GemsCatalogSearchResults]] = {
     val queryArgs = for {
       c <- criterions
-      q = CatalogQuery(basePosition, c.criterion.radiusLimits, c.criterion.magLimits.some)
+      q = CatalogQuery(0, basePosition, c.criterion.radiusLimits, c.criterion.magConstraints)
```
swalker2m: The id, `0` here, is new.  What is it for?

swalker2m:
Seems good to me.  :+1: Also, thanks for slogging through the GeMS stuff.  Not a fun task but it looks like it is getting better.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsVoTableCatalog.scala
cquiroz: Right, as we discussed I'll use the convention of using map for simple cases like this where for with 1-statement may be overkill

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsVoTableCatalog.scala
cquiroz: Well, the id is only useful for gems, it is ignored for non-gems, perhaps `Option[Int]` is the right approach here

tpolecat: This seems kind of smelly ... does the presence or absence of the zero determine whether it's a gems search or not? Or is there other information that does this? Would it make sense to have two constructors? `CatalogQuery.forGems(...)` etc.?

cquiroz: In non-gems queries there is one query per AGS search, while on gems there are at least two (one per guide probe). The result of those queries is indexed by the id and combined.
Having an factory like that makes sense to me

cquiroz:
I added the suggestions to the code

PR #276, opened 2015-02-27
OCSADV-230: Add catalog query caching

cquiroz:
Small PR to add caching of queries using scalaz Memo

### DIFF
--- a/bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableClient.scala
+++ b/bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableClient.scala
@@ -2,6 +2,7 @@ package edu.gemini.catalog.votable
 
 import java.net.UnknownHostException
 import java.util.concurrent.atomic.AtomicInteger
+import java.util.logging.Logger
 
 import edu.gemini.catalog.api.CatalogQuery
 import edu.gemini.spModel.core.Angle
@@ -16,6 +17,8 @@ import scalaz._
 import Scalaz._
 
 trait VoTableClient {
+  val Log = Logger.getLogger(getClass.getName)
+
   private val timeout = 30 * 1000 // Max time to wait
 
   private def format(a: Angle)= f"${a.toDegrees}%4.03f"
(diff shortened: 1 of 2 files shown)


bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableClient.scala
```diff
@@ -39,21 +42,31 @@ trait VoTableClient {
     p.future
   }
 
-  protected def doQuery(query: CatalogQuery, url: String): Future[QueryResult] = future {
-    val method = new GetMethod(s"$url/cgi-bin/conesearch.py")
-    val qs = queryParams(query)
-    method.setQueryString(qs)
+  // weak hash map should clean itself if we are using too much memory
+  val memo:Memo[(CatalogQuery, String), QueryResult] = Memo.weakHashMapMemo
```
swalker2m: This is cool but I wonder about the `WeakHashMap` option in this case.  The idea with a `WeakHashMap` in general is to avoid the presence of the key in the map from keeping the key (and therefore the map entry) from being removed.  In this case there's nothing to keep the `(CatalogQuery, String)` pair from being removed at any time no?  That is nobody holds a reference to it.  I'm not sure whether the gc necessarily waits for memory pressure to clean that up.

tpolecat: Yeah I just opened an issue about this. https://github.com/scalaz/scalaz/issues/910

Suggest switching to `arrayMemo` with some reasonable size.

cquiroz: `arrayMemo` is `Memo[Int, V]` so the key being an integer could be a problem. I guess I could manufacture a synthetic key out of the hash of `(CatalogQuery, String)`

swalker2m: Since two different `(CatalogQuery, String)` pairs could hash to the same `Int` you wouldn't know whether the value really corresponds.  Seems like they should provide an LRU cache implementation, which is easily obtainable via `LinkedHashMap`.

tpolecat: Yeah you're right, I misunderstood `ArrayMemo`. We could implement a `LinkedHashMap`-backed `Memo`.

cquiroz:
After some discussions with @swalker2m we realized this approach is too naive regardless of the choice of `Memo`. We should be able to reuse queries with different magnitude constraints and even reuse if the radius constraint is already contained. Closing for now

PR #277, opened 2015-03-02
OCSADV-200: client/server + shell commands

swalker2m:
This installment contains an initial implementation for VCS client/server interactions and shell commands.   I have updated the [documentation](https://github.com/gemini-hlsw/ocs/wiki/Science-Program-Synchronization-Notes#4-application-clientserver) to describe the main type aliases and classes and how they relate.  This is the key picture:

__Note:__ as this reimplementation grows, the amount of overlap with the old implementation increases.  When this is all over there will be a mega cleanup step to remove the existing contents of `edu.gemini.sp.vcs` and the entire `edu.gemini.sp.vcs.tui` bundle.


bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs/diff/osgi/Commands.scala
```diff
@@ -0,0 +1,160 @@
...
+
+    val handlers = List(vcsCmdHandler, hostHandler)
+
+    def vcs2(cmd: String, args: Array[String]): String =
+      try {
+        exec(cmd, args, handlers) | "Usage: vcs2 host|sync"
+      } catch {
+        case e: Exception =>
+          e.printStackTrace()
+          "???"
```
cquiroz: ???

bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs/diff/VcsServer.scala
```diff
@@ -0,0 +1,169 @@
...
+      try {
+        vcsLog.selectByProgram(id, offset, length).right
+      } catch {
+        case ex: Exception => VcsException(ex).left
+      }
+  }
+
+  private def locking[A](id: SPProgramID, user: Set[Principal], lock: SPNodeKey => Unit, unlock: SPNodeKey => Unit)(body: ISPProgram => VcsAction[A]): VcsAction[A] =
+    accessControlled(id, user) {
+      lookup(id) >>= { p =>
```
cquiroz: What does >>= do?

cquiroz:
Very impressive :smile:

bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs/diff/VcsServer.scala
tpolecat: It's just an alias for `flatMap`.

swalker2m: `>>=` comes from Haskell and is basically `flatMap` but it reads better.  Also goes well with `>>` which is a `flatMap` in which the argument is ignored: `m.flatMap { _ => ... }`.

bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs/diff/osgi/Commands.scala
swalker2m: Thanks, I need to clean that up a bit and add the other commands to the usage line.  Also, the "vcs2" command will be renamed to "vcs" once I get rid of the old version.

bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs/diff/MergePlan.scala
```diff
@@ -41,10 +51,27 @@ case class MergePlan(update: Tree[MergeNode], delete: Set[Missing]) {
...
+
   /** Accepts a program and edits it according to this merge plan. */
-  def merge(f: ISPFactory, p: ISPProgram): TryVcs[Unit] = {
+  def merge(f: ISPFactory, p: ISPProgram): VcsAction[Unit] = {
     // Tries to create an ISPNode from the information in the MergeNode.
-    def create(mn: MergeNode): TryVcs[ISPNode] =
+    def create(mn: MergeNode): VcsFailure \/ ISPNode =
       mn match {
         case Modified(k, _, dob, _) =>
           NodeFactory.mkNode(f, p, dob.getType, Some(k)).toRightDisjunction {
```
tpolecat: FYI you can use `\/>` instead of `toRightDisjunction` if you want to be cool.

bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs/diff/MergePlan.scala
```diff
@@ -75,24 +102,23 @@ case class MergePlan(update: Tree[MergeNode], delete: Set[Missing]) {
...
     // ISPNodes as necessary.
-    val mergeTree = update.map { mn =>
-      nodeMap.get(mn.key).fold(create(mn)) { _.right }.map {n => (mn, n) }
-    }.sequenceU
+    val mergeTree: VcsAction[Tree[(MergeNode, ISPNode)]] = Task.delay {
+      val nodeMap = p.nodeMap
 
+      update.map { mn =>
+        nodeMap.get(mn.key).fold(create(mn)) { _.right }.map {n => (mn, n) }
+      }.sequenceU
```
tpolecat: Two other ways to be cool:
- `fa.map(a => (x, a))` can be written `fa.strengthL(x)`
- `fa.map(f).sequenceU` can be written `fa.traverseU(f)`

bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs/diff/Vcs.scala
```diff
@@ -0,0 +1,159 @@
...
+import scala.collection.JavaConverters._
+import scalaz._
+import Scalaz._
+
+import Vcs._
+
+/** Vcs provides the public API for vcs commands such as push, pull and sync. */
+class Vcs(kc: KeyChain, server: VcsServer) {
+
+  private def user: Set[Principal] = kc.subject.getPrincipals.asScala.toSet
```
tpolecat: I would lift this into `VcsAction` because it's a side-effecting action.

```scala
private val user: VcsAction[Set[Principal]] = Task.delay(...).liftVcs
```

bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs/diff/VcsServer.scala
```diff
@@ -0,0 +1,171 @@
...
+  def write[A](id:       SPProgramID,
+               user:     Set[Principal],
+               evaluate: ISPProgram => VcsAction[A],
+               filter:   A => Boolean,
+               update:   (ISPFactory, ISPProgram, A) => VcsAction[Unit]): VcsAction[A] =
+    managed(id, user, writeLock, writeUnlock) { prog =>
+      evaluate(prog) >>= { a =>
+        if (filter(a)) {
+          val cp = odb.getFactory.copyWithSameKeys(prog)
+          update(odb.getFactory, cp, a) >> putProg(cp).map(_ => a).liftVcs
```
tpolecat: FYI, `fa.map(_ => a)` can be written `fa.as(a)` or `fa >| a`.

bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs/diff/VcsServer.scala
```diff
@@ -0,0 +1,171 @@
...
+    * same key or id with an existing program in the database and the user has
+    * appropriate keys.
+    */
+  def add(p: ISPProgram, user: Set[Principal]): VcsAction[Unit] = {
+    def failIfExists(id: SPProgramID, key: SPNodeKey): VcsAction[Unit] = {
+      val lookupFailure = Option(odb.lookupProgramByID(id)).map(_ => IdAlreadyExists(id)).orElse {
+        Option(odb.lookupProgram(key)).map(_ => KeyAlreadyExists(id, key))
+      }
+
+      (lookupFailure <\/ (())).liftVcs
```
tpolecat: Don't you have a `.fail` constructor?

tpolecat:
This looks great so far. Controlled effects make it much easier to follow.

bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs/diff/VcsServer.scala
swalker2m: So `lookupFailure` is `Option[VcsFailure]`.  Success is `None` and failure is `Some(VcsFailure)`.  I could do a `fold` I suppose:

`lookupFailure.fold(VcsAction(()))(VcsAction.fail)`

There's probably a better way?

tpolecat: Ah, right. That seems fine. If you need to do it more than once you might add a `VcsAction.void` or `.unit` constant for the no-op action.

swalker2m:
Thanks for all the excellent suggestions.  I think I've incorporated them all now.

bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs/diff/ProgramLocation.scala
```diff
@@ -4,9 +4,39 @@ import scalaz._
 
 sealed trait ProgramLocation
```
tpolecat: If you were to add `def fold[A](loc: => A, rem: => A): A = this match ...` you could remove the matches in the `def +(...)` methods below. May or may not be worthwhile.

tpolecat:
This is looking really clean.

swalker2m:
Added basic test cases for `VcsServer` and `Vcs`.  I had to [address a grim problem](https://github.com/swalker2m/ocs/commit/6fcd13c0ad907bae07f34d980f946daef2986559) with program copying.  When you checkout or add a program, it needs a distinct lifespan id so that edits are recorded correctly in the `VersionMap`.  So I added a method to the program factory for that.  Maybe I'll think of something better but otherwise I think it is time to close this and go back to merge plan corrections.

PR #278, opened 2015-03-03
OCSADV-226: Make GMOS parameter immutable and typed.

fnussber:
Some more cleanups and refactorings. Main focus was this time on GmosParameters object.


bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gmos/Gmos.java
```diff
@@ -18,79 +14,136 @@
...
+                        if (detectorCcdIndex == 0) _instruments = new Gmos[]{this};
+                        break;
+                    // E2V for GS: gmos_n_cdd_red.dat              => EEV legacy
+                    case GS:
+                        _detector = new Detector(getDirectory() + "/", getPrefix(), "ccd_red", "EEV legacy array");
+                        _detector.setDetectorPixels(DETECTOR_PIXELS);
+                        if (detectorCcdIndex == 0) _instruments = new Gmos[]{this};
+                        break;
+                    default:
+                        throw new IllegalArgumentException();
```
tpolecat: Hm, `IllegalArgumentException` is intended to indicate an error made by the caller. I usually throw `Error` in cases like this, with an indication that the condition should be impossible; if you have exhausted all the members of the `enum` you're in pretty bad trouble.

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gmos/Gmos.java
```diff
@@ -18,79 +14,136 @@
...
+                if (detectorCcdIndex == 0)
+                    _instruments = createCcdArray();
+                break;
+            default:
+                throw new IllegalArgumentException();
+        }
+
+        if (detectorCcdIndex == 0) {
+            _dtv = new DetectorsTransmissionVisitor(gp.getSpectralBinning(),
+                    getDirectory() + "/" + getPrefix() + "ccdpix_red" + Instrument.getSuffix());
```
tpolecat: Eventually these will be a `java.io.File`s rather than strings, I hope? Constructing paths with strings is error-prone (and technically you should use `File.separator` rather than `"/"` so it works on Windows).

tpolecat:
I don't understand what any of this code does, but at least mechanically your updates are clearly improving code quality and safety, so I'm all for it. :+1:

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gmos/Gmos.java
fnussber: Agreed, Error is the better choice here.

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gmos/Gmos.java
fnussber: Baby steps, baby steps ;) There's a lot of stuff like that, I hope I have time to sweep through all of that, too, at some point.

PR #279, opened 2015-03-04
OCSADV-231: Allow for Catalog offline tests

cquiroz:
This PR changes VoTableClient to have a pluggable backend. The backend by default goes to the remote catalogues but it can be switched to use files for testing.
This speeds up tests and increases robustness been assured to get the same results every time


bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/GemsStrategy.scala
```diff
@@ -27,7 +27,10 @@ import edu.gemini.spModel.core.{Angle, MagnitudeBand}
 import scalaz._
 import Scalaz._
 
-object GemsStrategy extends AgsStrategy {
+trait GemsStrategy extends AgsStrategy {
+  // By default use the remote backend but it can be overriden in tests
+  private [impl] implicit val backend:VoTableBackend = RemoteBackend
```
swalker2m: I like this idea that you can provide a different for testing.  Do you need the `implicit val` here given the one in the `VoTableBackend` companion object?

cquiroz: It is there so that I don't need to modify the `AgsStrategy` API signature passing the backend. This is used locally on GemsStrategy and it is overridable. I guess I could make it non-implicit by passing it directly to the `search` method

tpolecat: Yeah, I prefer explicit arguments or something monadic like `Reader` for passing dependencies, and only use implicit arguments for typeclass instances.

swalker2m: @tpolecat I agree but it might be okay as a backdoor for testing that doesn't clutter "normal" usage?  The alternative is to pass in `RemoteBackend` explicitly everywhere `GemsVoTableCatalog.search` is called I suppose.

cquiroz:
Replaced the implicit on `GemsStrategy` by a regular val that can be overridden in tests

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/GemsStrategy.scala
tpolecat: As awkward as it might seem, this is what I would prefer.

swalker2m: I guess this use of implicit params does seem a bit magical.  If it is too awkward to add the backend option everywhere `GemsVoTableCatalog.search` is called, it seems like you could accomplish the same thing in other ways.  Instead of the `GemsStrategy` being fitted with an implicit value for the backend, `GemsVoTableCatalog` itself could be parameterized by the backend and then a different instance used for testing.  I don't know if that helps.

cquiroz: `GemsStrategy` should not be called directly only via `AgsStrategy` and the val is private. Note that on the last commit the `val backend` is not implicit

cquiroz:
Replaced implicit usage by parameters

cquiroz:
Yep, a `def` could be useful for tests too

swalker2m:
This seems okay to me. :+1:

PR #280, opened 2015-03-05
OCSADV-230: Add catalog query caching

cquiroz:
This is a full rewrite of the previous attempt to do query caching. A traditional key-based cache (like scalaz's `Memo`) will be suboptimal in this case as we can reuse query results for nearby targets, e.g. if we had searched at the pole for a radius 1 and then we try the pole with radius 0.5, we can reuse the previous query doing client-side filtering.

The selected approach for the cache is a LRU cache based on a linear sequence where elements are moved to the front if they are used. To find items on the cache the client needs to provide a function that traverses the existing cache finding compatible queries. This limits performance to O(n) at best but given the relatively small size of the cache it is much faster than redoing the remote query

After some benchmarking `Vector` was selected as the structure for the cache. Some measurements are included on the tests cases

The signature of the API loosely matches scalaz's `Memo` but with  a very different implementation

### DIFF
--- a/bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableClient.scala
+++ b/bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableClient.scala
@@ -9,6 +9,7 @@ import edu.gemini.spModel.core.Angle
 import org.apache.commons.httpclient.{NameValuePair, HttpClient}
 import org.apache.commons.httpclient.methods.GetMethod
 
+import scala.annotation.tailrec
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.{Promise, Future, future}
 import scala.util.{Failure, Success}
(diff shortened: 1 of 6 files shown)


bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableClient.scala
```diff
@@ -20,17 +21,95 @@ trait VoTableBackend {
...
+    // Find if a search is already in the cache
+    // Note that this assumes all catalogues give the same result for a given query
+    def contains(a: QueryCache.CacheContainer[SearchKey, QueryResult], k: SearchKey):Option[(Int, QueryResult)] = {
+      @tailrec
+      def go(a: QueryCache.CacheContainer[SearchKey, QueryResult], pos: Int):Option[(Int, QueryResult)] = {
+        a match {
+          case x if x.isEmpty                                                    => None
+          case x +: _ if x.k.query == k.query || x.k.query.isSuperSetOf(k.query) => Some((pos, x.v))
+          case x +: xs                                                           => go(a.tail, pos + 1)
+        }
```
tpolecat: Destructuring a `Vector` in this way is actually quite expensive because it must construct a new instance for the tail. You might do better indexing into the vector directly:

```scala
def go(pos: Int): Option[(Int, QueryResult)] = {
  a.lift(pos) match {
    case Some(x) if x.k.query == k.query || x.k.query.isSuperSetOf(k.query) => Some((pos, x.v))
    case _ => go(pos + 1)
  }
```

cquiroz: Nice tip, I expected List-like performance on unapply. I didn't go with `zipWithIndex` thinking it would make a new `Vector` copy.

swalker2m:
Looks good. :+1:

PR #282, opened 2015-03-10
OCSADV-243: Split GMOS recipe into presentation and logic.

fnussber:
Main goal of this PR was to separate the calculation logic from the code that creates the HTML output in the GMOS recipe in order to allow to execute calculations and work with their results independently from creating HTML output.

Included in this PR is a major refactoring that pulls a lot of similar/identical code from all the recipes that deals with calculating the light path from the source to the instrument, including stuff like applying redshift etc. This is mostly independent from the actual instrument and can be done centrally. As part of this ```Gems``` and ```Altair``` have changed to implement a common interface ```AOSystem``` which allows to handle them the same way.

However, there are some subtle differences in the order and how exactly the different light path elements are applied. For details see all the weird if statements and TODO comments in ```SEDFactory.calculate()```. I hope that these exceptions can be transformed into some rules like "do this for all near-IR instruments" or "do that if there is an AO component". There are also some TODOs denoting actual errors, which need to be fixed but those changes will impact the regression tests and will have to be done in sync with science.

The current ```GmosResult``` objects will evolve further in the near future to provide simple access to the relevant figures (peak pixel flux, SN ratio..) and will probably also turn into Scala case classes once there is a TRCP service for the calculations.


bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gmos/GmosRecipe.java
```diff
@@ -72,6 +66,18 @@ public GmosRecipe(SourceDefinitionParameters sdParameters,
...
+
+
+        // TODO: this validation should be done centrally somewhere (several instruments do similar stuff here)
+        if (_sdParameters.getDistributionType().equals(SourceDefinitionParameters.Distribution.ELINE)) {
+            if (_sdParameters.getELineWidth() < (3E5 / (_sdParameters.getELineWavelength() * 1000))) {
+                throw new RuntimeException(
+                        "Please use a model line width > 1 nm (or "
+                                + (3E5 / (_sdParameters.getELineWavelength() * 1000))
+                                + " km/s) to avoid undersampling of the line profile when convolved with the transmission response");
+            }
```
swalker2m: I appreciate the predicament you’re in with this codebase.  Using exceptions for this kind of thing was probably not the optimal design decision.  These look like input validation and carefully formatted messages for the user of the application. Turning them into unchecked exceptions must mean somewhere at a high level everything is wrapped with a try catch clause and whatever the exception message might be is displayed to the end user?  

If this were new Scala code we would use scalaz’s disjunction or probably `Validation` for this kind of thing.  If it were new code even in Java I’d say we should probably wrap all the return types in something like that.  As is, I’m not sure what to suggest.  So I guess this is just a note to lament the situation we are in with the itc.

swalker2m:
I skimmed some of these changes.  The exceptions for input validation makes me nervous but I don't know what to do about that.  There's a lot more red (deletion) than green (addition) though, I can see where you're going with the separation of the logic, and there are lots of other little improvements so :+1:

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gmos/GmosRecipe.java
fnussber: This is a valid point. So far ITC just has a "catch-all" clause somewhere high up in the call stack that prints the exception message to the html page sent back to the user. I agree there's probably a need for something more sophisticated.

In particular, I've been discussing with Andy the idea to turn errors and suggestions like these into warnings and errors in the OT (using same mechanisms as the P2 checks). If we want to do that, obviously the error handling will have to change.

I haven't made up my mind yet how to attack this, so far I've just been moving around those validations with the rest of the code, so they're still randomly distributed throughout the code base. However, since I am starting now to work on the actual service interface that will be used to access the calculations in the future I'll have to address this sooner than later. I am planning to have the service layer entirely in Scala and only leave the calculation code in Java, that gives us some more possibilities here to use ```Validation```or whatever is most useful here. To be continued...

fnussber:
Regarding the amount of code I am deleting I am actually surprised that that thing is still working. However most of the code that is going away is just copy-pasted, so I guess no big surprise here. No worries about the exception handling, so far it's not any better or worse than it used to be ;) The checked exceptions were not really of any use because they were just done in a way that in the end almost every single function was marked as "throws Exception" which is no real advantage over having a big ugly "catch-all" on the highest level, which in fact is precisely how ITC did it.

PR #283, opened 2015-03-11
OCSADV-244: AGS results should include proper motion if available

cquiroz:
AGS wasn't considering the information about `ProperMotion` coming from the catalogue.
The model for `ProperMotion` has been changed to use `AngularVelocity` instead of `Angle` to represent a real velocity that can be negative or positive


bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/AngularVelocity.scala
```diff
@@ -0,0 +1,82 @@
...
+  val Zero = fromMilliArcSecondsPerYear(0.0)
+
+  /**
+   * Construct a `DeclinationAngularVelocity` from the given value in mas/y, which will be normalized to [-1296000000, 1296000000].
+   * @group Constructors
+   */
+  def fromMilliArcSecondsPerYear(mas: Double): DeclinationAngularVelocity = new DeclinationAngularVelocity {
+    override def toMilliArcSecondsPerYear: Double = mas % AngularVelocity.MilliArcSecsInADegree
+  }
+}
```
tpolecat: I think I would prefer to see something like `case class DeclinationAngularVelocity(toAngularVelocity: AngularVelocity) extends AnyVal` rather than doing this with subtyping. This would allow us to define consistent and safe `Order/Ordering` and `Monoid` instances for `AngularVelocity` as well as both safe wrappers. What do you think?

tpolecat:
My only concern is the use of subtyping rather than composition as noted above, but otherwise this looks great. I love all the tests.

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/AngularVelocity.scala
cquiroz: Does it make sense to have a `Monoid` if we are wrapping the values. The range is -360 to 360 degrees to make it compatible with the old world?

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/AngularVelocity.scala
```diff
@@ -0,0 +1,82 @@
+package edu.gemini.spModel.core
+
+import scalaz._, Scalaz._
+
+/** An angular velocity in [mas/y] in the range [-1296000000, 1296000000] equivalent to [-360, 360] degrees. */
+sealed trait AngularVelocity extends java.io.Serializable {
```
swalker2m: Since it is a velocity not an angle, I don't understand why there is a range.

cquiroz: The reason maybe not be very good. Without the limit it won't be fully compatible with the old world angle, I hit that on the tests but it could be simpler to limit the range on the tests

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/AngularVelocity.scala
tpolecat: As @swalker2m noted it probably doesn't make sense to wrap, but even in that case the monoid does make sense. Here we actually have [at least] an Abelian group, which is a much stronger statement than monoid, but scalaz doesn't give us this abstraction so we just stop with `Monoid`. This requires us to set aside the inconvenient fact that addition of floating-point values isn't actually associative, but for our purposes this is ok.

cquiroz:
Thanks for the suggestions, I've remove the range limitation and use composition instead of inheritance for the Velocities. A couple of typeclasses were added too

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/AngularVelocity.scala
```diff
@@ -0,0 +1,63 @@
...
+  implicit val ordering: scala.Ordering[AngularVelocity] =
+    scala.Ordering.by(_.masPerYear)
+
+  /** @group Typeclass Instances */
+  implicit val equal: Equal[AngularVelocity] =
+    Equal.equalA[AngularVelocity]
+  
+  /** @group Typeclass Instances */
+  implicit val monoid: Monoid[AngularVelocity] =
+    Monoid.instance((a, b) => AngularVelocity(a.masPerYear + b.masPerYear), Zero)
```
swalker2m: These look good.  Do we not also want the same typeclass instances for `RightAscensionAngularVelocity` and `DeclinationAngularVelocity`?  For our ground types like this we might also add `Show` instances, I guess?  It's kind of nice for logging/debugging.

cquiroz: Right, we should likely add them.
Just for my edification, what would be the benefit of `Show` over plain old `_.toString`

swalker2m: Actually I don't see a huge advantage over `toString` but you do get slightly more convenient syntax, e.g., `angularVel.println`.  Like with any type class in Scala you can also bring different instances into scope for different situations.  Probably not worth it though, I retract the suggestion. :-)

cquiroz: I can imagine you could make a generic function to print and change the format by providing a different instance of `Show`, e.g. one that could restrict output to 80 chars

cquiroz:
Added typeclasses to AngularVelocity classes

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/AngularVelocity.scala
tpolecat: Any use of methods defined on `Any` (including `toString`) can be problematic when reasoning about types because it violates a principle called [**parametricity**](http://ttic.uchicago.edu/~dreyer/course/papers/wadler.pdf). So `Show` gives us a way of constraining a type parameter to only those types that explicitly provide a string representation. In practice I don't think it's very useful; there are no constraints on what `Show` can do so it doesn't really mean anything.

The issue of local/alternative instances is kind of thorny because it violates *another* principle called **coherence**, which requires that typeclass instances be unique. In general this is important, but using a local instance in a way that is not observable (like sorting a list using a non-default `Order`) is probably ok. 

But in any case I don't see much reason to define `Show` instances until we hit a point where we need them, because it may never happen.

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/AngularVelocity.scala
```diff
@@ -0,0 +1,94 @@
...
+  implicit val order: Order[AngularVelocity] =
+    Order.orderBy(_.masPerYear)
+
+  /** @group Typeclass Instances */
+  implicit val ordering: scala.Ordering[AngularVelocity] =
+    scala.Ordering.by(_.masPerYear)
+
+  /** @group Typeclass Instances */
+  implicit val equal: Equal[AngularVelocity] =
+    Equal.equalA[AngularVelocity]
```
tpolecat: `Order` extends `Equal` so you don't need this instance.

```scala
scala> case class Foo(n: Int)
defined class Foo

scala> implicit val OrderFoo: Order[Foo] = Order.orderBy(_.n)
OrderFoo: scalaz.Order[Foo] = scalaz.Order$$anon$7@1024f5d6

scala> Equal[Foo]
res0: scalaz.Equal[Foo] = scalaz.Order$$anon$7@1024f5d6
```

tpolecat:
Looks great. I would get rid of those unnecessary instances but otherwise :+1:

cquiroz:
Good tip, removed `Equal` and will merge

PR #285, opened 2015-03-13
OCSADV-248,OCSADV-250: Adding ITC panels to OT

fnussber:
This PR addresses two steps for the ITC integration. 

It adds functionality in ```UniqueConfig``` that deals with extracting the "unique configurations" that are relevant for the ITC calculations, i.e. it groups all observes in the sequence according to their configurations (e.g. dispersers, filters etc). For each of those groups we will later have to trigger a call to the ITC service. (Note: I just noticed that there is already another ```UniqueConfig``` implementation in the pot bundle which seems to do some similar stuff. However since the UniqueConfig here is short and very specific to ITC I think I will leave it as is and not try to use the ```UniqueConfig``` from the pot bundle.)

It also adds UI elements to the Base Sequence Component node which display those unique configurations and - at a later stage - will display the results of the ITC service calls. There are separate panels for imaging and spectroscopy because those results are fundamentally different (e.g. spectroscopy results will also contain charts while imaging results don't). Unfortunately at least GMOS will also differ from the other instruments by displaying results for each of its CCDs separately, which results in a different table layout. Note that some instruments only support imaging or spectroscopy while others support both; depending on the instrument the panels are displayed or not.

Currently the ITC result panels are displayed for all instruments that support ITC calculations so we can check the unique configuration extraction and can play around with it. At a later stage we might only show the ones for which the ITC services have already been successfully connected to the OT.

As an example some unique spectroscopy configurations extracted from a sequence of 40 observes with iterators and offsets:

image

### DIFF
--- a/bundle/jsky.app.ot/src/main/java/jsky/app/ot/editor/seq/EdIteratorFolder.java
+++ b/bundle/jsky.app.ot/src/main/java/jsky/app/ot/editor/seq/EdIteratorFolder.java
@@ -96,10 +96,10 @@ public final class EdIteratorFolder
 
         // show/hide ITC result panels depending on instrument
         final ISPObsComponent instrument = getContextInstrument();
+        _w.tabbedPane.remove(itcImagingPanel.peer());
+        _w.tabbedPane.remove(itcSpectroscopyPanel.peer());
         if (instrument != null) {
             final SPComponentType type = instrument.getType();
-            _w.tabbedPane.remove(itcImagingPanel.peer());
-            _w.tabbedPane.remove(itcSpectroscopyPanel.peer());
             if (itcImagingPanel.visibleFor(type))      _w.tabbedPane.add("ITC Imaging",      itcImagingPanel.peer());
             if (itcSpectroscopyPanel.visibleFor(type)) _w.tabbedPane.add("ITC Spectroscopy", itcSpectroscopyPanel.peer());
         }
--- a/bundle/jsky.app.ot/src/main/java/jsky/app/ot/editor/seq/EdIteratorFolder.java
+++ b/bundle/jsky.app.ot/src/main/java/jsky/app/ot/editor/seq/EdIteratorFolder.java
@@ -6,10 +6,7 @@
 //
 package jsky.app.ot.editor.seq;
 
-import edu.gemini.pot.sp.ISPObsExecLog;
-import edu.gemini.pot.sp.ISPObservation;
-import edu.gemini.pot.sp.ISPSeqComponent;
-import edu.gemini.pot.sp.SPUtil;
+import edu.gemini.pot.sp.*;
 import edu.gemini.spModel.seqcomp.SeqBase;
 import jsky.app.ot.editor.OtItemEditor;
 import jsky.util.gui.TextBoxWidget;
@@ -18,18 +15,22 @@ import jsky.util.gui.TextBoxWidgetWatcher;
 import javax.swing.*;
 import javax.swing.event.ChangeEvent;
 import javax.swing.event.ChangeListener;
+import java.awt.*;
 import java.beans.PropertyChangeEvent;
 import java.beans.PropertyChangeListener;
 
 
 /**
- * This is the editor for the "iterator folder" or "Sequence" item.
- * It displays two tabs:
+ * This is the editor for the "iterator folder" or "Sequence" item. It displays the following tabs:
  * <p>
  *  Sequence: a table into which the iteration sequence of contained
  * iterators is written, and
  * <p>
  * Timeline: which shows the sequence as a timeline.
+ * <p>
+ * ITC Imaging: All unique imaging configurations and their ITC calculation results (if applicable).
+ * <p>
+ * ITC Spectroscopy: All unique spectroscopy configurations and their ITC calculation results (if applicable).
  */
 public final class EdIteratorFolder
         extends OtItemEditor<ISPSeqComponent, SeqBase>
@@ -37,6 +38,10 @@ public final class EdIteratorFolder
 
     /** the GUI layout panel */
     private final IterFolderForm _w;
+
+    /** ITC result panels */
+    private final ItcImagingPanel itcImagingPanel;
+    private final ItcSpectroscopyPanel itcSpectroscopyPanel;
 
     /** The data for the note */
     private final SequenceTab _sequenceTab;
@@ -60,6 +65,10 @@ public final class EdIteratorFolder
         _origSequenceTab = new OrigSequenceTab(_w, this);
 
         _w.tabbedPane.addChangeListener(this);
+
+        itcImagingPanel      = ItcPanel$.MODULE$.forImaging(this);
+        itcSpectroscopyPanel = ItcPanel$.MODULE$.forSpectroscopy(this);
+
     }
 
     /** Return the window containing the editor */
@@ -80,11 +89,21 @@ public final class EdIteratorFolder
     /** Set the data object corresponding to this editor. */
     @Override public void init() {
         final ISPObservation obs = getContextObservation();
-        // No context observation if this sequence component is in a conflict
-        // folder!
+        // No context observation if this sequence component is in a conflict folder!
         if (obs != null) {
             obs.addCompositeChangeListener(obslogUpdater);
         }
+
+        // show/hide ITC result panels depending on instrument
+        final ISPObsComponent instrument = getContextInstrument();
+        if (instrument != null) {
+            final SPComponentType type = instrument.getType();
+            _w.tabbedPane.remove(itcImagingPanel.peer());
+            _w.tabbedPane.remove(itcSpectroscopyPanel.peer());
+            if (itcImagingPanel.visibleFor(type))      _w.tabbedPane.add("ITC Imaging",      itcImagingPanel.peer());
+            if (itcSpectroscopyPanel.visibleFor(type)) _w.tabbedPane.add("ITC Spectroscopy", itcSpectroscopyPanel.peer());
+        }
+
 
         _origSequenceTab.init(this);
 
@@ -122,6 +141,11 @@ public final class EdIteratorFolder
             case 1:
                 _origSequenceTab.update();
                 break;
+            default:
+                // itc panels can be at different indices depending on instrument
+                final Component c = _w.tabbedPane.getComponentAt(index);
+                if (c == itcImagingPanel.peer())      itcImagingPanel.update();
+                if (c == itcSpectroscopyPanel.peer()) itcSpectroscopyPanel.update();
         }
     }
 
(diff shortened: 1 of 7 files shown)


bundle/jsky.app.ot/src/main/java/jsky/app/ot/editor/seq/EdIteratorFolder.java
```diff
@@ -80,12 +89,22 @@ public JPanel getWindow() {
...
 
+        // show/hide ITC result panels depending on instrument
+        final ISPObsComponent instrument = getContextInstrument();
+        if (instrument != null) {
+            final SPComponentType type = instrument.getType();
+            _w.tabbedPane.remove(itcImagingPanel.peer());
+            _w.tabbedPane.remove(itcSpectroscopyPanel.peer());
+            if (itcImagingPanel.visibleFor(type))      _w.tabbedPane.add("ITC Imaging",      itcImagingPanel.peer());
+            if (itcSpectroscopyPanel.visibleFor(type)) _w.tabbedPane.add("ITC Spectroscopy", itcSpectroscopyPanel.peer());
+        }
```
swalker2m: I definitely think this is a good place for this information and the screen-shot looks very nice.  Here in this line, if the instrument is `null` should it not remove the imaging and spectroscopy panels?  Otherwise will they be left to display whatever they were set to from the previous (if any) observation that was edited?

Also, I think it's cool to hide/show them depending on instrument and ITC availability.  I think a given observation will usually be all-imaging or all-spectroscopy though.  Maybe Andy could confirm that.  If you're just doing one or the other, perhaps the irrelevant panel shouldn't show either.  I guess you don't know that until you look at the sequence though.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/UniqueConfig.scala
```diff
@@ -0,0 +1,112 @@
...
+  def imagingConfigs(seq: ConfigSequence): Seq[UniqueConfig] =
+    uniqueConfigs(seq, c => isScience(c) && isImaging(c))
+
+  /** Gets all unique spectroscopy configurations from the given sequence. */
+  def spectroscopyConfigs(seq: ConfigSequence): Seq[UniqueConfig] =
+    uniqueConfigs(seq, c => isScience(c) && isSpectroscopy(c))
+
+  // Checks if a config is science or not; only science observations are relevant for ITC
+  private def isScience(c: Config): Boolean =
+    Option(c.getItemValue(OBS_TYPE_KEY)).fold(false)(!_.equals("CAL"))
```
swalker2m: Observation type is sadly still a `String` but I see there are other options like "DARK".  Is it true that anything besides "CAL" is science?

A trivial thing about which nobody seems to agree with me:  I find it clearer to shorten `Option` folds to either `exists` or `forall` depending on the none case.  For example:

````scala
Option(c.getItemValue(OBS_TYPE_KEY)).exists(!_.equals("CAL"))
````
If you too disagree, well that's just fine.

File: bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/UniqueConfig.scala [hunk 3]
```diff
@@ -0,0 +1,112 @@
...
+import edu.gemini.spModel.gemini.niri.{InstNIRI, Niri}
+import edu.gemini.spModel.gemini.trecs.{InstTReCS, TReCSParams}
+import jsky.app.ot.editor.seq.Keys._
+
+
+/**
+ * Unique configurations represent sets of observes in a sequence which are done with the same
+ * instrument configuration and exposure times.
+ */
+case class UniqueConfig(count: Int, labels: String, config: Config) {
```
swalker2m: I think it's fine to not try and glom additional ITC functionality into the existing `UniqueConfig` class.  We should probably rename this one though to avoid confusion.  Maybe `ItcUniqueConfig` or something.

The `labels` refer to data labels?  We have a type for that, `DatasetLabel` that you might be able to use instead?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/UniqueConfig.scala
```diff
@@ -0,0 +1,112 @@
...
+
+  /** Single exposure time is either the given exposure time or the time on source divided by the number of images. */
+  def singleExposureTime: Double = singleTime.getOrElse(totalTime.getOrElse(0.0) / count)
```
swalker2m: A cool scalaz way to do this would be:

````scala
singleTime | (~totalTime / count)
````
or maybe

````scala
singleTime | (totalTime.orZero / count)
````
if you don't like the tilde for `orZero`.

File: bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/UniqueConfig.scala [hunk 5]
```diff
@@ -0,0 +1,112 @@
...
+
+  /** Total exposure time is either total time on source (TReCS/Michelle) or the single exposure time times the number of images. */
+  def totalExposureTime: Double = totalTime.getOrElse(singleTime.getOrElse(0.0) * count)
+
+  /** TReCS and Michelle have an optional total time on source. */
+  private val totalTime = Option(config.getItemValue(INST_TIME_ON_SRC_KEY)).map(_.toString.toDouble)
+
+  /** Instruments other than TReCS and Michelle have a defined exposure time per image. */
+  private val singleTime = Option(config.getItemValue(INST_EXP_TIME_KEY)).map(_.toString.toDouble)
```
swalker2m: What's the type of the exposure time in the sequence?  I can't remember.  Just wondering about the `_.toString.toDouble`.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/UniqueConfig.scala
```diff
@@ -0,0 +1,112 @@
...
+
+  // Checks if a config is science or not; only science observations are relevant for ITC
+  private def isScience(c: Config): Boolean =
+    Option(c.getItemValue(OBS_TYPE_KEY)).fold(false)(!_.equals("CAL"))
+
+  // Decides if a configuration is for spectroscopy or not.
+  private def isSpectroscopy(c: Config): Boolean = !isImaging(c)
+
+  // Decides if a configuration is for imaging or spectroscopy based on the presence of a disperser element.
+  private def isImaging(c: Config): Boolean = c.getItemValue(INST_INSTRUMENT_KEY) match {
```
swalker2m: Since the rules are instrument specific, we might add a static method on the various instrument data objects that takes a config and evaluates it?  It would only be a marginal improvement but it does keep the instrument-specific logic in the corresponding class itself.  You'd still have to switch on the instrument key to pick the static method to call of course.

For GMOS, for example, we seem to define this differently at least in the P2 checks where we ask if the disperser `isMirror` and the FPU `isImaging`.  It would be helpful if they agreed on the definition or is "imaging for the purposes of ITC" different in some way?

bundle/jsky.app.ot/src/main/java/jsky/app/ot/editor/seq/EdIteratorFolder.java
fnussber: True, with ```null``` it should remove them, however I think that "should" never happen, but better safe than sorry.

In almost all cases it is either imaging or spectroscopy. In very few cases it can be mixed. It would be great to only show the panel that's actually needed (non-emtpy sequence), I'll check if there is a simple way to do that.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/UniqueConfig.scala
fnussber: You're right, it's a double. I replaced ```_.toString.toDouble``` with ```_.asInstanceOf[Double]```. You could argue that's potentially more dangerous(?). I guess it's "pick your poison", at least like that the double doesn't get converted to a string and back.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/UniqueConfig.scala
fnussber: ```UniqueConfig``` has been renamed to ```ItcUniqueConfig```.

Unfortunately I get the label as a ```String``` (e.g."001"), not as a ```DatasetLabel``` object. Converting the number to a dataset label to get the index seems overkill because I also need the observation id to do that.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/UniqueConfig.scala
fnussber: Ahm. Well, not sure about the ```~```. However I do like the ```.orZero```. I am in general not a big fan of stuff that you either know, or don't have a chance to figure out what it means. Maybe I am just not one of the cool kids, but I think there's a reason why we usually give readable names to functions and not just a series of operators...

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/UniqueConfig.scala
fnussber: I changed it to ```_.equals("science")``` which is maybe not yet the right thing either (I have manual flats showing up now). I think I'll need to fine-tune this with Andy anyways.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/UniqueConfig.scala
fnussber: That's a good point, some of this could probably go into the instruments, in particular if it overlaps with already existing logic. I need to look into why imaging vs. spectroscopy would be defined differently for P2 checks and ITC, that seems fishy. If ok, I would like to put this on the todo-list and come back to it in a later step (I won't have time to fix this today and I need this code in order to get the demo working)?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/UniqueConfig.scala
swalker2m: Tilde is probably not great as there are a few different usages for that.  Keep in mind that `orZero` doesn't necessarily always mean the number 0.  It does here but it is the monoid zero in general so for example an `Option[String]` `orZero` would be the empty string.  As for looking for symbols in general, the OS X Dash documentation tool is quite good for that.

cquiroz: Do we have a Gemini license for Dash?

PR #286, opened 2015-03-14
OCSADV-251: New bundles itc.web and itc.shared

fnussber:
In preparation of breaking up the itc calculations for good this PR adds two new bundles: 

itc.web is the new home for the ITC web application (servlets, html stuff etc) and more and more stuff from the ITC core module that is web app specific will  move here in the future.

itc.shared is going to house the ITC service interface including the TOs needed to pass data around between ITC service consumers and the ITC services themselves. This is what clients (e.g. OT) will have to import in the future to access ITC services.

The existing itc bundle will remain the home for the calculation logic and the data tables that need to reside on the server.

Ideally the itc.web bundle can be tossed entirely sometime in the future and be replaced with some 21st century web technology. Just saying.


swalker2m:
If I follow this is mostly just moving stuff into separate bundles which all makes sense to me.  :+1:

PR #287, opened 2015-03-17
OCSADV-200-J: Staff-only field corrections

swalker2m:
This installment introduces staff-only field corrections.  Some program nodes contain data that may only be edited by users with a staff key.  In particular, the root node itself, observations, and program notes (i.e., "pink" notes).  Examples of staff-only fields are the rollover status of the program or the QA state of an observation.  The merge correction algorithm has been updated to reset staff-only field edits to the appropriate values if a non-staff user should edit them, say via importing edited XML or an OT-bug.

To make this possible, a `copy` syntax function was added to `ISPDataObject` instances.  This permits duplicating a random data object `x` with `x.copy`.  The copy function is covered in the [first commit](https://github.com/gemini-hlsw/ocs/commit/3ccd7ba93ceb1507aaf27a2a21ca042403d3198d).  We use it in the [second commit](https://github.com/gemini-hlsw/ocs/commit/57574e8e068748023cbdf1bcf8e811eff63cf536) to edit any `MergePlan` that contains edits to staff-only fields.  The existing `ISPStaffOnlyFieldProtected` data object interface is used in this process.

### DIFF
--- a/bundle/edu.gemini.pot/src/test/java/edu/gemini/pot/spdb/test/TriggerCase.java
+++ b/bundle/edu.gemini.pot/src/test/java/edu/gemini/pot/spdb/test/TriggerCase.java
@@ -9,7 +9,6 @@ import edu.gemini.pot.spdb.IDBTriggerCondition;
 import edu.gemini.spModel.data.ISPDataObject;
 import edu.gemini.spModel.pio.ParamSet;
 import edu.gemini.spModel.pio.PioFactory;
-import org.junit.After;
 import org.junit.Before;
 import org.junit.Test;
 
@@ -35,10 +34,25 @@ public class TriggerCase extends SpdbBaseTestCase {
     private static final SPComponentType TRIGGER_TYPE = SPComponentType.INSTRUMENT_FLAMINGOS2; // SPComponentType.getInstance("trigger", "trigger", "trigger");
 
     public static final class ProgramDataObject implements ISPDataObject {
-        private List _triggerList = new ArrayList();
+        private List<String> _triggerList = new ArrayList<>();
+
+        public Object clone() {
+            try {
+                return super.clone();
+            } catch (CloneNotSupportedException e) {
+                throw new InternalError();
+            }
+        }
+
+        @SuppressWarnings("unchecked")
+        @Override
+        public <A extends ISPDataObject> A clone(A a) {
+            return (A) a.clone();
+        }
+
 
         public String[] getTriggerMessages() {
-            return (String[]) _triggerList.toArray(new String[0]);
+            return _triggerList.toArray(new String[_triggerList.size()]);
         }
 
         public void addTriggerMessage(String message) {
@@ -92,6 +106,19 @@ public class TriggerCase extends SpdbBaseTestCase {
 
     public static final class TriggerDataObject implements ISPDataObject, Serializable {
         private String _triggerMessage;
+
+        public Object clone() {
+            try {
+                return super.clone();
+            } catch (CloneNotSupportedException e) {
+                throw new InternalError();
+            }
+        }
+
+        @Override @SuppressWarnings("unchecked")
+        public <A extends ISPDataObject> A clone(A a) {
+            return (A) a.clone();
+        }
 
         public String getTriggerMessage() {
             return _triggerMessage;
@@ -207,7 +234,6 @@ public class TriggerCase extends SpdbBaseTestCase {
     private ISPProgram _prog;
     private ISPObsComponent _triggerComp;
     private ISPObsComponent _nonTriggerComp;
-    private List _leaseList = new ArrayList();
 
     @Before
     public void setUp() throws Exception {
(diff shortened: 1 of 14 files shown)


bundle/edu.gemini.pot/src/test/java/edu/gemini/pot/spdb/test/TriggerCase.java
```diff
@@ -35,10 +34,25 @@
     private static final SPComponentType TRIGGER_TYPE = SPComponentType.INSTRUMENT_FLAMINGOS2; // SPComponentType.getInstance("trigger", "trigger", "trigger");
 
     public static final class ProgramDataObject implements ISPDataObject {
-        private List _triggerList = new ArrayList();
+        private List<String> _triggerList = new ArrayList<>();
```
cquiroz: Nice that we remove the raw types :smile:

bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs/diff/StaffOnlyFieldCorrection.scala
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  def apply(mp: MergePlan, hasPermission: PermissionCheck): VcsAction[MergePlan] = {
+    // Regular staff permission is determined by program contact information. To
+    // prevent changing the contact and gaining staff permission we have to check
+    // that contacts aren't edited unless you have super-staff permission.
+    val contactsMatch = contact(mp.update.rootLabel) == remote(key).flatMap(contact)
+
+    canSetStaffFields(pid, contactsMatch, hasPermission).map { canSet =>
+      if (canSet) mp else resetFields(mp)
```
cquiroz: I think this is the key line right? Does it run on Gemini's spdb? I guess if it were run on the client a clever attacker could edit the bytecodes to by-pass the check.
I guess if somebody does that it deserves to get the permission anyway :smile:

bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs/diff/StaffOnlyFieldCorrection.scala
```diff
@@ -0,0 +1,108 @@
...
+    }
+  }
+}
+
+object StaffOnlyFieldCorrection {
+  type StaffProtected = ISPDataObject with ISPStaffOnlyFieldProtected
+
+  private def contact(mn: MergeNode): Option[String] = {
+    def contact(sp: SPProgram) = Option(sp.getContactPerson).map(_.trim.toLowerCase)
```
cquiroz: `getContactPerson` can return `null`? Is it from the old Java world?

bundle/edu.gemini.sp.vcs/src/test/scala/edu/gemini/sp/vcs/diff/TestEnv.scala
```diff
@@ -23,18 +25,34 @@ case class TestPeer(odb: IDBDatabaseService, server: VcsServer, service: Princip
...
 
   def shutdown(): Unit = odb.getDBAdmin.shutdown()
 
+  def spDataObject: SPProgram = prog.getDataObject.asInstanceOf[SPProgram]
+
   // Lookup the test program
   def prog: ISPProgram = odb.lookupProgram(TestEnv.Key)
 
+  def set(update: SPProgram => Unit): Unit =
+    spDataObject <| update |> prog.setDataObject
```
cquiroz: spaceshippy...

cquiroz:
Looks good, I put a few comments inline but nothing major

bundle/edu.gemini.sp.vcs/src/test/scala/edu/gemini/sp/vcs/diff/TestEnv.scala
fnussber: We will soon get to the point where you can feed any ASCII art into the Scala compiler and out comes .. a program. ;)

bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs/diff/StaffOnlyFieldCorrection.scala
swalker2m: I didn't expect anybody would notice before I got a chance to do something about this.  :-)  I originally wrote a [version with server side validation](https://gist.github.com/swalker2m/b5605e80412c5a63edf0) but then I realized that really I should just redo all the merge corrections when the differences are stored.  I was hoping that instead I might come up with a way to validate that the merge corrections were applied to save the time required to redo them.  I probably will end up just redoing the merge corrections on the server side and checking whether anything changes.  If so, it will fail with a `NeedsUpdate` failure.  If not, the merge plan can be applied. I'll leave that for a subsequent installment though.

bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs/diff/StaffOnlyFieldCorrection.scala
swalker2m: Yes, it can return `null`.  It's part of the ancient `SPProgram` data object for the top-level node. It is initialized to the empty string but can be assigned `null`.  We pretty much have to assume that anything coming from the old Java model will sometimes be `null`.

bundle/edu.gemini.sp.vcs/src/test/scala/edu/gemini/sp/vcs/diff/TestEnv.scala
swalker2m: Some concepts just don't map to a word or two very well.  For example I think `<|` is actually clearer than `unsafeTap` once you are initiated.  According to the doc:

> Mostly for use with dodgy libraries that give you values that need additional initialization or mutation before they're valid to use.

I think the science program model qualifies.

cquiroz: I think is right that after a while you get used to them and are clear to read.
Perhaps we should have a list of the most used ones like mappend `|+|`, tie fighter `|@|`, unsafeTap `<|`, etc

PR #288, opened 2015-03-17
OCSADV-196: TPE science area drawing bugfix

sraaphorst:
TPE would previously draw the science area in the wrong place if the base was not (0,0). This has been fixed.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/scala/edu/gemini/pot/ModelConverters.scala
+++ b/bundle/edu.gemini.pot/src/main/scala/edu/gemini/pot/ModelConverters.scala
@@ -33,10 +33,8 @@ object ModelConverters {
   private val maxArcsecs = 60 * maxArcmins
   implicit class AngleNormalizer(val angle: Angle) extends AnyVal {
     private def normalize(value: Double, maxValue: Double): Double = {
-      val half = maxValue / 2.0
-      val pos = value % maxValue
-      val neg = pos - maxValue
-      if (math.abs(pos) < math.abs(neg)) pos else neg
+      val neg = value - maxValue
+      if (value < math.abs(neg)) value else neg
     }
 
     def toNormalizedRadians:    Double = normalize(angle.toRadians, maxRadians)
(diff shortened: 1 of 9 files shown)


bundle/edu.gemini.pot/src/main/scala/edu/gemini/pot/ModelConverters.scala
```diff
@@ -26,6 +26,25 @@ object ModelConverters {
...
+  private val maxDegrees = 360
+  private val maxArcmins = 60 * maxDegrees
+  private val maxArcsecs = 60 * maxArcmins
+  implicit class AngleNormalizer(val angle: Angle) extends AnyVal {
+    private def normalize(value: Double, maxValue: Double): Double = {
+      val half = maxValue / 2.0
+      val pos = value % maxValue
+      val neg = pos - maxValue
+      if (math.abs(pos) < math.abs(neg)) pos else neg
+    }
```
swalker2m: `half` isn't used for anything and `pos` is always the same as `value` since they are being passed in from an `Angle.toDegrees` etc.?

We clearly need something like this put I'm not sure we want to add it as syntax or call it "normalized".  The [0,360) representation is arguably the normalized form.  Still, let's go with it temporarily until this sprint is over and we can plan for it.  At least this way we have a compile checked reminder of everywhere we have a problem.

Did you fix the offsetting issue with this?

sraaphorst: I've been hesitant to fix the offsetting issue with this, because it seems to happen in a few places (e.g. in multiple places throughout the `PatrolField` code), and I'm not sure where all of them are of if this is the solution that we want to ultimately use, so I've held off. I think for now, I'll tell Andy and Bryan to stick to positive offsets.

This was needed for the TPE for now, though, because if either coordinate of the base position ever becomes negative, the science area ends up being drawn in the wrong place because of the positive angle restriction.

I'll go ahead and make the changes as you recommended and then merge unless someone speaks up now and says otherwise :-).

PR #291, opened 2015-03-24
OCSADV-200-K: increment version when correcting observation numbers

swalker2m:
This is a small update to observation number corrections.  It simply increments node version information when adjusting the observation number.  This is mostly to be consistent with other corrections.  The idea is that any adjustment to the merge plan should be reflected as a local edit.  We compare the version information of the final merge plan to the original program to know whether any updates need to be made locally before sending changes to the database.

### DIFF
--- a/bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs/diff/ObsNumberCorrection.scala
+++ b/bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs/diff/ObsNumberCorrection.scala
@@ -1,6 +1,7 @@
 package edu.gemini.sp.vcs.diff
 
 import edu.gemini.pot.sp.SPNodeKey
+import edu.gemini.pot.sp.version.LifespanId
 import edu.gemini.sp.vcs.diff.MergeCorrection._
 import edu.gemini.sp.vcs.diff.NodeDetail.Obs
 import edu.gemini.sp.vcs.diff.ProgramLocation.{Remote, Local}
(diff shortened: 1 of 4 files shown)


bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs/diff/ObsNumberCorrection.scala
```diff
@@ -25,15 +26,16 @@ import Scalaz._
   * remote program versions with an observation number greater than a
   * local-only observation.
   */
-class ObsNumberCorrection(isKnown: (ProgramLocation, SPNodeKey) => Boolean) extends CorrectionFunction {
+class ObsNumberCorrection(lifespanId: LifespanId, isKnown: (ProgramLocation, SPNodeKey) => Boolean) extends CorrectionFunction {
   def apply(mp: MergePlan): TryCorrect[MergePlan] =
     renumberedObs(mp).map { obsMap =>
-      if (obsMap.isEmpty)  // usually empty so we might as well check and save a traversal in that case
-        mp
+      if (obsMap.isEmpty) mp // usually empty so we might as well check and save a traversal in that case
```
cquiroz: Wouldn't a traversal of an empty map be essentially `O(1)`?

bundle/edu.gemini.sp.vcs/src/test/scala/edu/gemini/sp/vcs/diff/MergeCorrectionSpec.scala
```diff
@@ -21,6 +21,8 @@ import Scalaz._
 class MergeCorrectionSpec extends Specification {
   import NodeDetail._
 
+  val lifespanId: LifespanId = LifespanId.random
```
cquiroz: What's the is the `LifespanId` of a `MergeNode` in general?
I was wondering about the use of random here but this is just for the test, right

cquiroz:
Looks good, plenty of tests :+1:

bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs/diff/ObsNumberCorrection.scala
swalker2m: The traversal refers to mapping over the whole `Tree[MergeNode]` contained in the plan.  If there are no renumbered observations, and usually there won't be, it skips the map.

bundle/edu.gemini.sp.vcs/src/test/scala/edu/gemini/sp/vcs/diff/MergeCorrectionSpec.scala
swalker2m: There is lifespan id per instance of a program.  So while a program exists in a database it uses the same lifespan id to mark edits made to that instance.  Here we're just checking that the version information is incremented so any lifespan id will do.

PR #293, opened 2015-03-25
OCSADV-253: End-to-end implementation of ITC service for GMOS imaging

fnussber:
This PR provides a basic but working end-to-end implementation of the ITC service for GMOS (imaging only) without any bells and whistles. Improvements like client-side caching of calculation results etc. are planned for upcoming PRs.

Unfortunately this PR is a bit biggish because I had to rename/refactor a few things to get the TRCP service to work which rippled through to a lot of the files.

Main relevant changes are the inclusion of the ITC bundles into the spdb application and additions to the ```ItcTable``` and ```ItcTableModel``` classes to perform service calls and handle the results. Future refinements will probably see some of the ITC services related tasks (like creation of input parameter objects) being extracted to some utility classes.

Here it would also be nice to know if there is a way we can already anticipate the usage of ITC in PIT? However since PIT uses its own model I am afraid that the translation of instrument configurations into ITC parameter objects would have to be entirely PIT specific(?).

I also have two questions:

* Do we want to set the Serial Version UIDs for service objects manually?
* Is it correct to use the peer obtained by ```ObservingPeer``` for the service calls?

### DIFF
--- a/bundle/edu.gemini.itc.shared/src/main/Java/edu/gemini/itc/shared/ObservationDetails.java
+++ b/bundle/edu.gemini.itc.shared/src/main/Java/edu/gemini/itc/shared/ObservationDetails.java
@@ -1,9 +1,13 @@
-package edu.gemini.itc.service;
+package edu.gemini.itc.shared;
+
+import edu.gemini.itc.shared.*;
+
+import java.io.Serializable;
 
 /**
  * Container for observation detail parameters.
  */
-public final class ObservationDetails {
+public final class ObservationDetails implements Serializable {
 
     private final CalculationMethod calculationMethod;
     private final AnalysisMethod analysisMethod;
--- a/bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/ItcService.scala
+++ b/bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/ItcService.scala
@@ -1,5 +1,11 @@
-package edu.gemini.itc.service
+package edu.gemini.itc.shared
 
+import edu.gemini.spModel.core.Peer
+import edu.gemini.util.security.auth.keychain.KeyChain
+import edu.gemini.util.trpc.client.TrpcClient
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
 import scalaz.{Failure, Success, Validation}
 
 /**
@@ -16,7 +22,7 @@ final case class ItcSpectroscopyResult(/*TODO*/) extends ItcCalcResult
  * return one result (either because they don't have more than one CCD or because all different CCDs are assumed to
  * have the same characteristics).
  */
-sealed trait ItcResult {
+sealed trait ItcResult extends Serializable {
   /** Returns the results for all CCDs. */
   def ccds: Array[ItcCalcResult]
 
@@ -26,7 +32,7 @@ sealed trait ItcResult {
 
 object ItcResult {
 
-  import ItcService._
+  import edu.gemini.itc.shared.ItcService._
 
   /** Creates an ITC result in case of an error. */
   def forException(e: Throwable): Result = Failure(List(e.getMessage))
@@ -43,7 +49,7 @@ object ItcResult {
  */
 trait ItcService {
 
-  import ItcService._
+  import edu.gemini.itc.shared.ItcService._
 
   def calculate(source: SourceDefinition, obs: ObservationDetails, cond: ObservingConditions, tele: TelescopeDetails, ins: InstrumentDetails): Result
 
@@ -52,5 +58,11 @@ trait ItcService {
 object ItcService {
 
   type Result = Validation[List[String], ItcResult]
+
+  /** Performs an ITC call on the given host. */
+  def calculate(kc: KeyChain, peer: Peer, source: SourceDefinition, obs: ObservationDetails, cond: ObservingConditions, tele: TelescopeDetails, ins: InstrumentDetails): Future[Result] =
+    TrpcClient(peer).withKeyChain(kc) future { r =>
+      r[ItcService].calculate(source, obs, cond, tele, ins)
+    }
 
 }
--- a/bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/ItcService.scala
+++ b/bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/ItcService.scala
@@ -60,8 +60,8 @@ object ItcService {
   type Result = Validation[List[String], ItcResult]
 
   /** Performs an ITC call on the given host. */
-  def calculate(kc: KeyChain, peer: Peer, source: SourceDefinition, obs: ObservationDetails, cond: ObservingConditions, tele: TelescopeDetails, ins: InstrumentDetails): Future[Result] =
-    TrpcClient(peer).withKeyChain(kc) future { r =>
+  def calculate(peer: Peer, source: SourceDefinition, obs: ObservationDetails, cond: ObservingConditions, tele: TelescopeDetails, ins: InstrumentDetails): Future[Result] =
+    TrpcClient(peer).withoutKeys future { r =>
       r[ItcService].calculate(source, obs, cond, tele, ins)
     }
 
--- a/bundle/edu.gemini.itc/src/main/scala/edu/gemini/itc/service/ItcServiceImpl.scala
+++ b/bundle/edu.gemini.itc/src/main/scala/edu/gemini/itc/service/ItcServiceImpl.scala
@@ -23,7 +23,6 @@ class ItcServiceImpl extends ItcService {
   } catch {
     // TODO: for now in most cases where a validation problem should be reported to the user the ITC code throws an exception instead
     case e: Throwable =>
-      e.printStackTrace()
       ItcResult.forException(e)
   }
 
--- a/bundle/edu.gemini.itc/src/main/scala/edu/gemini/itc/service/ItcServiceImpl.scala
+++ b/bundle/edu.gemini.itc/src/main/scala/edu/gemini/itc/service/ItcServiceImpl.scala
@@ -2,6 +2,7 @@ package edu.gemini.itc.service
 
 import edu.gemini.itc.gmos.GmosRecipe
 import edu.gemini.itc.operation.ImagingS2NMethodACalculation
+import edu.gemini.itc.shared._
 
 /**
  * The ITC service implementation.
@@ -21,7 +22,9 @@ class ItcServiceImpl extends ItcService {
     }
   } catch {
     // TODO: for now in most cases where a validation problem should be reported to the user the ITC code throws an exception instead
-    case e: Throwable => ItcResult.forException(e)
+    case e: Throwable =>
+      e.printStackTrace()
+      ItcResult.forException(e)
   }
 
   private def calculateGmos(source: SourceDefinition, obs: ObservationDetails, cond: ObservingConditions, tele: TelescopeDetails, ins: GmosParameters): Result = {
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcTable.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcTable.scala
@@ -6,6 +6,7 @@ import javax.swing.table.AbstractTableModel
 import edu.gemini.itc.shared.TelescopeDetails.Wfs
 import edu.gemini.itc.shared._
 import edu.gemini.pot.sp.SPComponentType
+import edu.gemini.pot.sp.SPComponentType._
 import edu.gemini.spModel.config.ConfigBridge
 import edu.gemini.spModel.config.map.ConfigValMapInstances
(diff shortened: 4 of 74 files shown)


cquiroz:
PIT has its own limited model of instruments. Just enough to make proposals, it is even possible that there is not enough information on the model to do ITC calculations, depending on how finely detailed parameters you need.
Eventually there will be a new common model but that's probably way down the line

swalker2m:
Just a note on the future caching.  The `edu.gemini.spModel.obs.SPObsCache` is where we cache time consuming calculations that are tied to an observation.  It is automatically cleaned when something is edited so it seems like a natural fit.

bundle/edu.gemini.itc.shared/src/main/Java/edu/gemini/itc/shared/ObservationDetails.java
```diff
@@ -1,9 +1,13 @@
...
+
+import edu.gemini.itc.shared.*;
+
+import java.io.Serializable;
 
 /**
  * Container for observation detail parameters.
  */
-public final class ObservationDetails {
+public final class ObservationDetails implements Serializable {
```
cquiroz: Do they need to be serializables as they are now part of the Obs tree?

swalker2m: If it is passed over trpc then yeah it has to be `Serializable`.

bundle/edu.gemini.itc/src/main/scala/edu/gemini/itc/service/ItcServiceImpl.scala
```diff
@@ -21,7 +22,9 @@ class ItcServiceImpl extends ItcService {
     }
   } catch {
     // TODO: for now in most cases where a validation problem should be reported to the user the ITC code throws an exception instead
-    case e: Throwable => ItcResult.forException(e)
+    case e: Throwable =>
+      e.printStackTrace()
```
cquiroz: `printStackTrace`?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcTable.scala
```diff
@@ -55,29 +67,78 @@ trait ItcTable extends Table {
...
+    //val wfs  = ??? TODO
+    val src  = new SourceDefinition(PointSource(20.0, BrightnessUnit.MAG), LibraryStar("A0V"), WavebandDefinition.R, 0.0)
+    val obs  = new ObservationDetails(ImagingSN(c.count, c.singleExposureTime, 1.0), AutoAperture(5.0))
+    val tele = new TelescopeDetails(TelescopeDetails.Coating.SILVER, port, Wfs.OIWFS)
+
+    val qual = owner.getContextSiteQuality
+    val cond = new ObservingConditions(qual.getImageQuality, qual.getCloudCover, qual.getWaterVapor, qual.getSkyBackground, 1.5)
+
+    // get the instrument configuration
+    val filter    = c.config.getItemValue(new ItemKey("instrument:filter")).asInstanceOf[GmosCommonType.Filter]
```
cquiroz: I wish there would be a safer way to do this, it seems possible this could throw NPE or CCE I think it is more of a model's problem rather than ITC though

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcTableModel.scala
```diff
@@ -26,65 +32,114 @@ object ItcTableModel {
...
-  def getColumnCount: Int = headers.size + keys.size + results.size
+  override def getColumnCount: Int = headers.size + keys.size + results.size
 
-  def getKeyAt(col: Int): Option[ItemKey] = col match {
-    case c if c >= headers.size && c < headers.size + keys.size => Some(key(col))
-    case _                                                      => None
+  override def getValueAt(row: Int, col: Int): Object = col match {
+    case c if c <  headers.size             => header(col).value(uniqueSteps(row), res(row))
+    case c if c >= headers.size + keys.size => result(col).value(uniqueSteps(row), res(row))
+    case c                                  => uniqueSteps(row).config.getItemValue(key(col))
```
cquiroz: case _

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcTableModel.scala
```diff
@@ -26,65 +32,114 @@ object ItcTableModel {
...
   private def result(col: Int) = results(col - headers.size - keys.size)
 
- }
+  // Gets the result from the service result future (if present)
+  protected def imagingCalcResult(f: Future[ItcService.Result]): Option[ItcResult] =
+    for {
+      futureResult  <- f.value                // unwrap future
+      serviceResult <- futureResult.toOption  // unwrap try
+      calcResult    <- serviceResult.toOption // unwrap validation
+    } yield calcResult
```
cquiroz: Does this work because all of them become `Option` right? I wish scala would let you do a for over different types of monads

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcTableModel.scala
```diff
@@ -26,65 +32,114 @@ object ItcTableModel {
...
+
+  // TODO: display errors/validation messages in an appropriate way in the UI, for now also print them to console
+  protected def messages(f: Future[ItcService.Result]): String =
+    f.value.fold("Calculating...") {
+      case Failure(t) => t <| (_.printStackTrace()) |> (_.getMessage)  // "Look mummy, there's a spaceship up in the sky!"
```
cquiroz: :smile:

cquiroz:
Looks good to me, many package changes and some code updates
I'm not looking forward to another `sbt reload` :+1:

bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/ItcService.scala
```diff
@@ -53,4 +59,10 @@ object ItcService {
 
   type Result = Validation[List[String], ItcResult]
 
+  /** Performs an ITC call on the given host. */
+  def calculate(kc: KeyChain, peer: Peer, source: SourceDefinition, obs: ObservationDetails, cond: ObservingConditions, tele: TelescopeDetails, ins: InstrumentDetails): Future[Result] =
+    TrpcClient(peer).withKeyChain(kc) future { r =>
+      r[ItcService].calculate(source, obs, cond, tele, ins)
+    }
```
swalker2m: Perhaps we don't need the keychain?  Do we do some kind of permission checking on the server side?  It seems like this is all public information that you can get via the web service w/o an account or anything anyway?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcTableModel.scala
swalker2m: In general arbitrary monads don't compose.  For example if you have `m1: => Option[A]` and `m2: => Failure \/ B` what would it mean to write:

````scala
for {
  a <- m1
  b <- m2
} yield fun(a,b)
````
What's the result type?  What if `m1` produces `None`, etc.

This looks like it might be a case for monad transformers though.  If we have a task that contains a disjunction `\/` it could be modeled as `EitherT[Task, ItcFailure, A]`.

**Comment by swalker2m** [comment]
> Do we want to set the Serial Version UIDs for service objects manually?

If you do that then you're saying that you'll manage any future incompatibilities yourself.  As long as we're using trpc, I think it is better to just let them be incompatible when there are differences and let it break.  The OT and ODB pretty much have to be released together anyway and we check the versions of each already to see whether they are compatible.

**Comment by swalker2m** [comment]
> Is it correct to use the peer obtained by ObservingPeer for the service calls?

The observing peer won't be defined for PIs. As Carlos was suggesting, perhaps it should call the service at both sites and use the first answer it gets back.

File: bundle/edu.gemini.itc.shared/src/main/Java/edu/gemini/itc/shared/ObservationDetails.java (hunk 1 above)
Comment (fnussber) [inline bundle/edu.gemini.itc.shared/src/main/Java/edu/gemini/itc/shared/ObservationDetails.java]: Yes, this are the TOs passed around using TRPC, so they all have to be serializable.

File: bundle/edu.gemini.itc/src/main/scala/edu/gemini/itc/service/ItcServiceImpl.scala (hunk 2 above)
Comment (fnussber) [inline bundle/edu.gemini.itc/src/main/scala/edu/gemini/itc/service/ItcServiceImpl.scala]: Me being lazy, sorry. But this will go away soon (once the error messages are displayed in a more useful way in the client).

File: bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcTable.scala (hunk 3 above)
Comment (fnussber) [inline bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcTable.scala]: Agreed, this ain't pretty. I get ```Object```s back, and - depending on the instrument - I know their types, there's not much else I can do, I think.

File: bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/ItcService.scala (hunk 7 above)
Comment (fnussber) [inline bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/ItcService.scala]: That's a good point. Indeed the ITC service is public, so I actually don't need any permissions here.

**Comment by fnussber** [comment]
Would it make sense to call the service at the site of the ODB you're connected to? But of course this one could be the one that's further away from you. Calling it at both sites seems a bit of an overkill, but I guess it would make sense.

**Comment by cquiroz** [comment]
Actually we could do DNS magic for this, a global DNS entry like itc.gemini.edu and internally it would send you to the closest one. It would make the code simpler but I don't know if our DNS can do that

**Comment by tpolecat** [comment]
The OT could have a pinger that figures out which peer is closest to you, I guess. But for now I think hitting both and using the first answer makes sense. If it ends up being a performance issue we can revisit later.

File: bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcTable.scala (hunk 3 above)
Comment (tpolecat) [inline bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcTable.scala]: If you're using this API a lot it might make sense to add some bedazzlement to force a checked cast, so you get an early failure rather than a mistyped value getting passed around.

```scala
implicit class ConfigOps(c: Config) {

  def get[A](key: ItemKey(implicit ev: ClassTag[A]): ClassCastException \/ A =
    \/.fromTryCatch(ev.runtimeClass.cast(a).asInstanceOf[A]).leftMap {
      case cce: ClassCastException => cce
      case ex => throw ex
    }

  def unsafeGet[A: ClassTag](key: ItemKey(implicit ev: ClassTag[A]): A =
    get[A](key).fold(throw _, identity)

}

// and then the cast is hidden
c.config.unsafeGet[GmosCommonType.Filter](new ItemKey("instrument:filter"))
```

fnussber: What is the advantage of using ```Class.cast()``` over ```asInstanceOf[]```? Both seem to check the types and throw a CCE during runtime immediately(?).

tpolecat: Sorry, I worded that text very poorly. My point was just that **if** you abstract out the `get` operation then you end up casting to a type variable, which is unsound, and you can use a `ClassTag` to get around this problem.

Here is an example of what I mean:

```scala
scala> def hmm[A](a: Any) = \/.fromTryCatch(a.asInstanceOf[A])
hmm: [A](a: Any)scalaz.\/[Throwable,A]

scala> hmm[Int]("hello")
res15: scalaz.\/[Throwable,Int] = \/-(hello) // <-- WAT .. this is mistyped

scala> res15.toOption
res16: Option[Int] = Some(hello) // <-- still mistyped

scala> res16.map(_ + 1) // finally an operation that catches us
java.lang.ClassCastException: java.lang.String cannot be cast to java.lang.Integer
  at scala.runtime.BoxesRunTime.unboxToInt(BoxesRunTime.java:105)
  at $anonfun$1.apply(<console>:18)
  at scala.Option.map(Option.scala:146)
  ... 43 elided
```

As opposed to

```scala
scala> def hmm[A](a: Any)(implicit ev: ClassTag[A]) = \/.fromTryCatch(ev.runtimeClass.cast(a).asInstanceOf[A])
hmm: [A](a: Any)(implicit ev: scala.reflect.ClassTag[A])scalaz.\/[Throwable,A]

scala> hmm[Int]("hello")
res19: scalaz.\/[Throwable,Int] = -\/(java.lang.ClassCastException: Cannot cast java.lang.String to int)
```

fnussber: Ok, got you. Thx for clarifying.

fnussber:
I implemented some of the review findings. I will deal with using the appropriate "closest" server for the service calls in a later PR. I also added the web module to the spdb server so we can start using it and see if there are any problems with it being part of the spdb. If nobody has additional comments I am going to merge this tonight or tomorrow morning.

PR #294, opened 2015-03-25
OCSADV-245: Support Sloan's bands (Was AGS not working with NIRI+NGS nor NIRI+LGS)

cquiroz:
Out of a reported error on handling NIRI + NGS and NIRI + LGS it was detected that the processing of catalog bands was incomplete
This PR improves handling but as a side effect, AGS internal processing requires modifications. Those will be carried on OCSADV-255
A test `SingleProbeStrategySpec` reproduces the original reported error but it hasn't been fixed yet
Gems is also broken and will be fixed on OCSADV-272

### DIFF
--- a/bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableParser.scala
+++ b/bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableParser.scala
@@ -52,7 +52,7 @@ trait VoTableParser {
 
   import scala.xml.Node
 
-  private val magRegex = """(?i)em.(opt|IR).(\w)""".r
+  private val magRegex = """(?i)em.(opt|IR)(\.\w)?""".r
   private val REQUIRED = List(VoTableParser.UCD_OBJID, VoTableParser.UCD_RA, VoTableParser.UCD_DEC)
 
   protected def parseFieldDescriptor(xml: Node): Option[FieldDescriptor] = xml match {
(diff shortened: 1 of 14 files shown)


bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableParser.scala
```diff
@@ -140,24 +178,25 @@ trait VoTableParser {
       k.sequenceU
     }
 
-    def combineWithErrors(m: List[Magnitude], e: Map[MagnitudeBand, Double]) = m.map(i => i.copy(error = e.get(i.band)))
+    def combineWithErrorsAndFilter(m: List[Magnitude], e: Map[MagnitudeBand, Double]): List[Magnitude] =
+      m.map(i => i.copy(error = e.get(i.band))).filterNot(magnitudesFilter.nonValidMagnitude)
```
swalker2m: Perhaps reversing this to `.filter(magnitudesFilter.validMagnitude)` would be a bit better?

cquiroz: I'd think it is faster `nonValidMagintude` as it only needs to check 1 invalid options while a `validMagnitude` would need to check a bunch. Basically `all - 1` (amag)

swalker2m: Could it be implemented as `!amag`?

swalker2m:
I don't know much about VoTable but it all looks reasonable.  :+1:

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableParser.scala
cquiroz: The fiter actually works over field ids, it would need to be
`v.id =/= "amag" && v.id =/= "e_amag"`
Right, I could change that

cquiroz:
Added the filtering reversal and will merge

PR #295, opened 2015-03-27
OCSADV-291: Turn off Smart GCal for GPI

swalker2m:
Smart GCal doesn't work for GPI and in fact throws an exception when you try to add a smart flat or arc to a GPI sequence.  This is problematic for test cases that generate random programs and sequences.  Until Smart GCal for GPI is specified by the science staff, we should just turn it off.

I tried GPI in the OT after these changes and don't see any problem, but I'd like @fnussber to review this in case there is anything I'm overlooking.

There is one substantive commit that takes away the `CalibrationKeyProvider` interface.  The second commit is a bit of cleanup: removes the `static` modifier from enums since they are static anyway, and takes away the `private` modifier from enum constructors since they are already private.


cquiroz:
Looks good, I've been removing the `static` and `private` modifiers of enums too :smile:

fnussber:
From the top of my head I don't think there should be any side-effects by this change other than turning of smart gcal for GPI.

PR #296, opened 2015-03-27
OCSADV-200: bugfix for ancient test cases

swalker2m:
The change to make all data objects cloneable broke some old test cases.  The behavior of these old test cases relied upon having a reference to a test data object returned when requested from the program node instead of being copied.  This update fixes those test cases by making them work like the real data objects.  They are copied when requested and have to be replaced after modifying them.


cquiroz:
:+1: 
Did you run all ocs tests and olny these failed?

swalker2m:
Yes, with this fix (and a fresh start of sbt in order to have enough memory) all the test cases run.

PR #297, opened 2015-03-30
OCSADV-255: AGS changes for UCAC4

cquiroz:
This is a bit convoluted PR. AGS has always been based on using a single band to select a guide star and for validation. OCSADV-255 changes this assumption to use one of a set of bands depending on what's available.

As you'd expect this change has quite a few effects down the line touching many pieces of AGS. It also affects `VoTableCatalog`, which needs to support filtering by raw magnitudes rather than by a single particular band.

The effects were particularly felt on the `analyze` part, which wasn't tested and thus not verifiable. Several tests to the basic AGS analyze function are added. Most of the lines added are test xml files

Another side effect is that direct calls to `AgsAnalysis` have been banned and all calls from the OT must go now through `AgsStrategy`. I declare this is a *Good Thing &trade;*


cquiroz:
No takers?

tpolecat:
Sorry, I promise I'll look at this tomorrow :-)

sraaphorst:
I will as well.

swalker2m:
The probe limits table itself includes the magnitude band (singular) to which the limit table applies.  See `Guide Limits - OT Config.csv` which is ultimately just downloaded from a Google Docs spreadsheet page as a CSV file.  For example the header, if you will, for the altair table is

`Altair NGS,R,17.0,,`

where R refers to R magnitude band.  It seems like we need to support multiple bands there?

cquiroz:
With the new code, the R read from the table is understood to represent the set r',R,UC as spelled on the ticket
The limits are the same for each band

cquiroz:
After a discussion with @swalker2m I'll just merge this PR

PR #298, opened 2015-04-06
OCSADV-200-L: Observation Permission Corrections

swalker2m:
This installment introduces observation-status-based permission checking.  A user's ability edit an observation will depend upon the keys she holds and the status of the observation.  Since programs may be simultaneously edited by multiple users, upon sync a user may find that changes made to a particular observation are no longer valid because another user has advanced the observation status in the database.  For an extreme example, consider the case of a PI editing an observation's guide star in the OT. Meanwhile a staff user marks the observation ready and commits the change.  Now any update to the observation made by the PI should be considered invalid upon sync.

Observation edits are broken into three categories: creation of new observations, deletion of existing observations, and updates of existing observations.  When an observation is created with a status that the user should not have permission to assign, it is simply reset to the lowest level "Phase II" status.  When an observation with an advanced status is deleted by a user without permission to do so, it is resurrected as it exists in the remote database.  The most common case though will be edits to an observation whose status advances in the database.  In this case the observation as it exists in the remote database is restored and a new copy of the local version of the observation is added to a conflict folder.

This is a comparatively large update with four parts:

* __`ProgramDiff`__  - `ProgramDiff` was basically a function that returned a `MergePlan` describing the differences between the remote and local program versions.  Now it is a case class with the starting point `MergePlan` as before, a map from observation key to `ObservationStatus`, and a maximum observation number.  The `ObservationStatus` turned out to be impossible to compute given only the difference information so we just compute it remotely on actual `ISPObservation`s and return it along with the difference.  The maximum observation number is needed when an inappropriately edited observation is duplicated since it must be given a new unique observation number.

* __`ObsEdit`__ -  contains the logic for calculating and categorizing the observation edits that were performed by a user so that they may be subsequently validated or rejected.

* __`ObsEditValidator`__ - contains the rules for what constitutes a legal observation edit

* __`ObsPermissionCorrection`__ - uses `ObsEdit` and `ObsEditValidator` to find inappropriately edited observations and correct them as indicated above.


bundle/edu.gemini.pot/src/test/scala/edu/gemini/pot/sp/ProgramGen.scala
```diff
@@ -220,7 +235,7 @@ object ProgramGen {
 
   val genEdit: Gen[ProgEdit] =
     oneOf(
-      genEditDataObject,
+      frequency(7 -> genEditDataObject, 3 -> genEditObsPhase2Status),
```
cquiroz: Is this like generate 70% get data object and 30% edit phase 2 status?

bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs/diff/MergeNode.scala
```diff
@@ -113,11 +120,105 @@ object MergeNode {
...
+      def isConflictFolder(t: Tree[MergeNode]): Boolean =
+        t.rootLabel match {
+          case Modified(_,_, _: ConflictFolder, _) => true
+          case Unmodified(k)                       => nodeMap.get(k).exists { n =>
+            n.getDataObject.getType == SPComponentType.CONFLICT_FOLDER
+          }
+          case _                                   => false
+        }
+
+      //noinspection MutatorLikeMethodIsParameterless
```
cquiroz: Is this IDEA fodder?

bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs/diff/ObsEditValidator.scala
```diff
@@ -0,0 +1,121 @@
...
+import scalaz._
+import Scalaz._
+
+// NOTE: extracted from (and will replace):
+// edu.gemini.util.security.policy.MergeValidator
+
+/**
+ * Evaluates whether a local observation edit should be considered legal.
+ */
+class ObsEditValidator(privs: UserRolePrivileges, too: TooType) {
```
cquiroz: How do you tell the user an edit is illegal? Do you get a prompt, icon, something else?

cquiroz:
I can't really parse the details of the merge algorithms but I'll trust they are fine

bundle/edu.gemini.pot/src/test/scala/edu/gemini/pot/sp/ProgramGen.scala
swalker2m: Yes, exactly.  I think in reality it's nowhere near 30% phase 2 status changes but since it has such a big impact on permissions I wanted to make it prevalent.

bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs/diff/MergeNode.scala
swalker2m: Yeah, it keeps Idea from warning.  It may have a mutator like name but it doesn't mutate anything.  It produces a new value.

bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs/diff/ObsEditValidator.scala
swalker2m: This will be done as it is currently done in the OT, via "conflict" notes.  They appear in the program tree icons and when you click on a node with a note there is a brief description. It's much like P2 warnings but you can also cycle through the conflicts and acknowledge them to make them go away and allow you to merge.

PR #299, opened 2015-04-07
OCSADV-265: Split of presentation and business logic

fnussber:
This PR contains a loooong series of steps to split the presentation part which creates the HTML output from the actual calculations and move everything web related to the web module.

In the original code the actual calculation logic was intertwined with the "printing" logic that writes HTML code to an output stream (HttpServletResponse). All this printing code has been extracted and moved. In addition to this many print methods that were defined on several objects and some helper objects for formatting were also moved to the web module.

This is sort of a major milestone because it is basically the conclusion of a several months long process of refactorings with the goal of simplifying and straightening out the code to the point it was possible to extract all the printing stuff.

Note: I had a quick look at this with Carlos. The problem is that I think it doesn't make sense for anyone to spend too much time reviewing this. Most of the code is original ITC code that simply has been moved around and restructured. The way the HTML output is created is not great but I am at the point were I think spending any time on trying to improve that is basically a waste. If we want to spend any time on the web part of ITC we should simply toss all of that and redo it.


bundle/edu.gemini.itc.web/build.sbt
```diff
@@ -27,6 +27,7 @@ OsgiKeys.exportPackage := Seq()
 
 OsgiKeys.importPackage := Seq(
   "!com.google.inject.cglib.asm.util.*",
+  "!com.keypoint.*",
```
cquiroz: Why do we need to explicitly not import these packages? It seems they are not used in the code anyway

bundle/edu.gemini.itc.web/src/main/java/edu/gemini/itc/web/html/AcqCamPrinter.java
```diff
@@ -0,0 +1,74 @@
...
+import edu.gemini.itc.acqcam.AcquisitionCamera;
+import edu.gemini.itc.shared.ImagingResult;
+import edu.gemini.itc.shared.Parameters;
+
+import java.io.PrintWriter;
+
+/**
+ * Helper class for printing Acquisition Camera calculation results to an output stream.
+ */
+public final class AcqCamPrinter extends PrinterBase {
```
cquiroz: Any reason for not making these classes is scala :smile:

fnussber: Yeah, I was thinking about that. However these "Printer" classes are just containers for extracted code that was somewhere mixed into the calculation code. In the end it was easier to just copy the existing Java code into Java classes instead of going through the hoops to change it into Scala. So the short answer is: I am just lazy.

swalker2m:
I took your advice and didn't go through all the changes.  I completely support the goal though.  Let's keep going. :+1:

PR #300, opened 2015-04-07
OCSADV-200: fix old vcs service registration

swalker2m:
This is a quick fix for a bug with the old VCS service, which wasn't being started in some cases because the `edu.gemini.sp.vcs.tui` bundle could not find a required service, whose service class, `edu.gemini.vcs.log.VcsLog`, shares the same package root of a package exported by `edu.gemini.vcs`.

Ultimately the `edu.gemini.vcs` bundle, which houses the old VCS code, will be retired and the who-provides-what questions will become clearer.


cquiroz:
:+1:

tpolecat:
Ugh. :+1:

PR #302, opened 2015-04-07
REL-781: Update Altair w/ field lens warning.

fnussber:
From JIRA: Requirement: Change the current Phase 2 check 

WARNING if Altair Guide Star Type == Natural Guide Star, "The Altair field lens is recommended for most observations." 

to 

WARNING if Altair Guide Star Type == Natural Guide Star AND Base coordinate != Altair AOWFS coordinate , "The Altair field lens is recommended for off-axis targets."

### DIFF
--- a/bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/altair/AltairRule.java
+++ b/bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/altair/AltairRule.java
@@ -15,6 +15,7 @@ import edu.gemini.shared.skyobject.Magnitude;
 import edu.gemini.shared.util.immutable.Option;
 import edu.gemini.shared.util.immutable.Trio;
 import edu.gemini.shared.util.immutable.Tuple3;
+import edu.gemini.skycalc.Coordinates;
 import edu.gemini.spModel.gemini.altair.AltairAowfsGuider;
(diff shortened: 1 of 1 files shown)


bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/altair/AltairRule.java
```diff
@@ -15,6 +15,7 @@
 import edu.gemini.shared.util.immutable.Option;
 import edu.gemini.shared.util.immutable.Trio;
 import edu.gemini.shared.util.immutable.Tuple3;
+import edu.gemini.skycalc.Coordinates;
```
cquiroz: @swalker2m Shouldn't we try to use the new target model for new code, I guess in this case it is ok given the size of the patch but we are just adding more places to port code later on

swalker2m: @cquiroz I think I would have done the same thing as Florian here.  The new target model isn't hooked up yet to the science program model so it would mean converting to/from.  It's probably just as well to do all of that in one go.  In your catalog work, using the new target model makes a lot of sense because it is a major new effort and you're basically replacing the now redundant but less capable `SkyObject`.  Here I think I would probably just wait.

swalker2m:
:+1:

PR #303, opened 2015-04-08
REL-2239: Phase2 checks not applied to GMOS-S + Hamamatsu

fnussber:
Another one of a set of release tasks for the June release. This one is basically about repeating a set of P2 check rules for GMOS-N Hamamatsu CCDs for GMOS-S.

I will verify tomorrow with science if the gain is indeed needed (currently ```AmpGain.LOW``` everywhere) and if not clean that out, too.

While the ```GmosRule``` class would certainly benefit from some additional refactoring (splitting into separate files) since it is rather large, I decided to only do a few quick fixes for now to shorten stuff a bit and get rid of some warnings.

### DIFF
--- a/bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/gmos/GmosRule.java
+++ b/bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/gmos/GmosRule.java
@@ -17,8 +17,19 @@ import edu.gemini.spModel.config2.Config;
 import edu.gemini.spModel.config2.ConfigSequence;
 import edu.gemini.spModel.config2.ItemKey;
 import edu.gemini.spModel.data.YesNoType;
-import edu.gemini.spModel.gemini.gmos.*;
+import edu.gemini.spModel.gemini.gmos.GmosCommonType.*;
+import edu.gemini.spModel.gemini.gmos.GmosNorthType.DisperserNorth;
+import edu.gemini.spModel.gemini.gmos.GmosNorthType.FPUnitNorth;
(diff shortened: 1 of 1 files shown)


bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/gmos/GmosRule.java
```diff
@@ -291,7 +299,7 @@ public Problem check(Config config, int step, ObservationElements elems, Object
 //                        detectorManufacturer = SequenceRule.getInstrumentItem(config, InstGmosNorth.DETECTOR_MANUFACTURER_PROP);
 //                    }
 //                    // REL-231 - don't warn for GMOSN
-//                    return (amp != GmosCommonType.AmpCount.THREE) && (detectorManufacturer == GmosCommonType.DetectorManufacturer.E2V);
+//                    return (amp != AmpCount.THREE) && (detectorManufacturer == DetectorManufacturer.E2V);
```
cquiroz: Couldn't be better to just remove the commented code?

swalker2m: I'm guessing that was updated via find/replace.  The comment aside, it's nice to see the clutter of the `GmosCommonType` prefix removed.

swalker2m:
All the changes I can see look fine and splitting up the limits table seems fine to me. :+1:

fnussber:
I tossed the commented out code and removed the aforementioned AmpGain.LOW parameter. It's easy enough to add it again if this is ever necessary.

PR #304, opened 2015-04-08
OCSADV-273: negative angles

swalker2m:
This is a type-class-based solution  to the "negative" angle issue based on a suggestion from Rob.  The idea is to share all the conversion and construction methods that currently reside in our `Angle` class so that new angle-like classes can take advantage.  For example, this PR includes a simple offset in P value class that wraps an `Angle` and provides an implementation for the `IsoAngle` type class.  Using it along with the provided `AngleSyntax` support, you could write code like:

````
  import AngleSyntax._
  val p0 = 4.arcsecs[OffsetP]  // construct an OffsetP as 4 arcsecs
  val p1 = 5.arcsecs[OffsetP]
  (p0 - p1).println      // formats the resulting OffsetP assuming you want arcsecs
  // -1.000 arcsecs

  // Or if you want, construct an offset of 1 arcmin
  -1.arcmins[OffsetP].println
  // -60.000 arcsecs
````

or going the other way

````
  val d0 = p0.arcsecs    // convert an OffsetP into the Double 4.0
  val d1 = p1.arcsecs
  println(d0 - d1)       // normal floating point math
  // -0.9999999998626663   

  // Or if you want, convert a OffsetP to degrees
  println(p0.degrees)
  // 0.0011111111111290484
````

Apologies for all the trivial updates to `Angle` itself.  Those were introduced by Idea and just strip off trailing spaces.  The only substantive change is the new `signedDegrees` function at the bottom to translate [0..360) into [-180 .. 180).

`Angle` itself could be refactored to take advantage of the same construction and conversion type classes and the RA/Dec specific formatting moved out into the `RightAscension` and `Declination` classes.

### CODE REVIEW

**Comment by fnussber** [comment]
I think I would name the offset ```POffset``` or some such thing instead of just ```P```, when I see a single capital letter I immediately think "type parameter". Other than that this looks good to me.

File: bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/AngleSyntax.scala [hunk 1]
```diff
@@ -0,0 +1,39 @@
+package edu.gemini.spModel.core
+
+/**
+ * Syntax for working with angle-like types.
+ */
+object AngleSyntax {
+  implicit class ToAngleOps[A](value: A) {
+    def angle(implicit ta: ToDegrees[A]): Angle    = ta.toAngle(value)
+    def arcsecs(implicit ta: ToDegrees[A]): Double = ta.toArcsecs(value)
```
tpolecat: In cases like this (pure syntax) I think it's ethical to capture the TC instance in the ctor:

```scala
  implicit class ToAngleOps[A: ToDegrees](value: A) {
    def angle: Angle = ta.toAngle(value)
    ...
```

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/AngleSyntax.scala
```diff
@@ -0,0 +1,39 @@
...
+
+    def -(that: A)(implicit ia: IsoAngle[A]): A =
+      (value.degrees - that.degrees).degrees
+
+    def *(factor: Double)(implicit ia: IsoAngle[A]): A =
+      (value.degrees * factor).degrees
+
+    def flip(implicit ia: IsoAngle[A]): A =
+      value + 180.degrees[A]
+
```
tpolecat: As irritating as it may seem, I prefer putting these operations `plus`, `minus`, etc., on the typeclass itself, and use syntax only for delegation ... and then don't use syntax for the implementation (see below).

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/P.scala
```diff
@@ -0,0 +1,29 @@
...
+
+  import AngleSyntax._
+
+  implicit val ShowP: Show[P] =
+    Show.shows(p => f"${p.arcsecs}%4.03f arcsecs")
+
+  implicit val MonoidP: Monoid[P] =
+    new Monoid[P] {
+      val zero = 0.arcsecs[P]
+      def append(a: P, b: => P): P = a + b
```
tpolecat: I would prefer to not a data type's syntax for its own implementation. So:

```scala
def append(a: P, b: => P): P = IsoAngleP.add(a, b)
```

tpolecat:
This looks pretty good so far. I recommend a few small changes based on scalaz coding guidelines, which are kind of pedantic but tend to prevent suffering in the long run.

cquiroz:
I think we should merge this to get it into the test release, what do you think?

swalker2m:
All the existing test cases pass.  I'm not sure how to add new ones that aren't redundant.  `OffsetP` and `OffsetQ` have basically identical type class implementations which is sort of a shame.  I'm not sure how to fix that though w/o giving up something else.  But yeah, I guess I can merge it so it gets included in the manual testing.

PR #305, opened 2015-04-08
REL-2281: Updated icons for OT 2015B

cquiroz:
Just new versions of all the icons. It was tested on OSX


swalker2m:
Usually there is accompanying text that explains what we're looking at in the splash screen.  Is that already there or coming?

cquiroz:
Right, it is on the ticket, I think it goes in welcome.html?

cquiroz:
Text was added

swalker2m:
:+1:

PR #306, opened 2015-04-08
OCSADV-290: Changes to build files for Java 8.

sraaphorst:
We're officially moving to Java 8, and here are the required changes in order to do so.

Note that you will need to install a 1.8 JDK, and version 40 (or above).
When I tried compiling with JDK 1.8 version 25, the compiler crashed and died with an internal exception.


cquiroz:
Looks good :+1:

fnussber:
I can confirm that the problems we see in the ITC tests stem from the same issue that caused the other tests to fail (different rounding behavior in ```NumberFormat```). @sraaphorst : I think the easiest way to mitigate this is to ```@Ignore``` the ```BaselineTestSpec``` tests as part of this PR and then merge the PR. Once this is done I can update (i.e. recreate) the ITC baselines and then un-```@Ignore``` the test cases again.

tpolecat:
This [SO discussion](http://stackoverflow.com/questions/24426438/numberformat-rounding-issue-with-java-8-only) includes a workaround via `BigDecimal`.

swalker2m:
Okay there is still some rounding weirdness. For example,

````java
_doBothTest("20:10:3.555", "20:10:03.55", DMSFormat.DEFAULT_PRECISION);
````

but there was weirdness before, just *different* cases.  To really fix it we'd need to go back into the old code and, I guess, use `BigDecimal` everywhere.  So I hate moving on knowing this issue exists but maybe it's the thing to do.  What does everyone think?

arturog8m:
Since it seems the problem is well understood, I suggest we put a REL task to resolve this for the December release so we can start testing on other aspects that are more pressing now. If we find weird behavior during the testing stage we regroup of course.

cquiroz:
Should we merge this one in to get it into the test build?

PR #309, opened 2015-04-09
OCSADV-255: More AGS test and removal of duplicate code

cquiroz:
This PR builds on OCSADV-255 adding tests for GMOS on AGS and it replaces the ags model conversion classes with the ones on `pot`


cquiroz:
The PR has been rebased and updated to fix conflicts with https://github.com/gemini-hlsw/ocs/pull/304

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsVoTableCatalog.scala
```diff
@@ -130,10 +131,27 @@ case class GemsVoTableCatalog(backend: VoTableBackend = RemoteBackend) {
...
+      val maxLimit = radiusConstraint.maxLimit
+      val correctedMax = (c.offset |@| c.posAngle) { (o, _) =>
+          // If an offset and pos angle were defined, normally an adjusted base position
+          // would be used, however since we are merging queries here, use the original
+          // base position and adjust the radius limits
+          maxLimit + o.distance
+        } | maxLimit
+      (max(correctedMax.toDegrees, prev._1), min(radiusConstraint.minLimit.toDegrees, prev._2))
+    }
+    RadiusConstraint.between(Angle.fromDegrees(result._1), Angle.fromDegrees(result._2))
```
swalker2m: If the `criterList` is empty (which may not be possible?) this will be a strange constraint.  I'm not sure what `Double.MinValue` or `Double.MaxValue` would be when normalized to [0, 360) but probably nothing sensible.  Maybe we should return an empty constraint in that case?

cquiroz: This is more or less straight from the old code, constraints should not be empty, but nonetheless I agree it is weird.
I think the correct would be to return in that case a `RadiusConstraint.between(0, 0)` meaning we are not searching for anything, I could also check the impact of changing that to `Option[RadiusConstraint]`

swalker2m:
Other than the one possible issue with `optimizedRadiusConstraint`, this all seems fine to me. :+1:   Just returning `between(0,0)` seems okay to me if you get an empty list.

cquiroz:
I converted the method to return an option, it should be cleaner than the empty radius constraints and there is no great impact on the rest of the code

swalker2m:
:+1:

PR #312, opened 2015-04-13
REL-2281: Updated icons for OT 2015B

cquiroz:
Use the non-labeled image for the splash screen instead of the labelled one


swalker2m:
As I recall the image height is smaller than the height of the text area so there is space around the top and bottom of the image.  If so, could the image be embiggened or else the height of the text area ensmallened to match?  It just looked a bit off to see space around the top and bottom.

swalker2m:
I see the size changed too.  :+1:

PR #313, opened 2015-04-13
OCSADV-255: Missing package export to be used by AGS

cquiroz:
This missing packages break SPDB at runtime


fnussber:
:+1:  Go for it :)

PR #316, opened 2015-04-14
OCSADV-272: GEMS AGS with UCAC4

cquiroz:
This PR brings the same changes from OCSADV-255 to Gems AGS. It lets selecting of any of the r-family  of bands if present on a target. This flexibility has to be propagated all the way to mascot's algorithms.

There are also several cosmetic code fixes, refactorings and renames, plus several tests with their respective catalog responses.

There are a few things missing to complete this task:
* Transfer the logic to the new `GemsStrategy`, that is part of the AGS API
* `CatalogQuery` is more complex than it should be, containing logic for query and filters and should be refactored

The branch has been rebased with develop fixing conflicts with the changes to the `Angle` and `Offset` class


bundle/edu.gemini.ags/src/main/java/edu/gemini/ags/gems/GemsGuideStarSearchOptions.java
```diff
@@ -112,19 +112,12 @@ public GemsTipTiltMode getGemsTipTiltMode() {
...
-
-    private String opticalCatalog = DEFAULT_CATALOG;
-    private String nirCatalog = DEFAULT_CATALOG;
-    private GemsInstrument instrument;
-    private GemsTipTiltMode tipTiltMode;
-    private Set<Angle> posAngles = new HashSet<>();
-
-    public GemsGuideStarSearchOptions(final String opticalCatalog, final String nirCatalog, final GemsInstrument instrument,
+    public GemsGuideStarSearchOptions(final GemsInstrument instrument,
                                       final GemsTipTiltMode tipTiltMode, final Set<Angle> posAngles) {
```
swalker2m: The old `String` catalog identifiers won't be missed but are we sure that there's no longer a desire to allow selection of the catalogs?

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsVoTableCatalog.scala
```diff
@@ -88,23 +91,36 @@ case class GemsVoTableCatalog(backend: VoTableBackend = RemoteBackend) {
...
     val radiusLimitsList = getRadiusLimits(inst, criterions)
     val magLimitsList = optimizeMagnitudeLimits(criterions)
 
     val queries = for {
       radiusLimits <- radiusLimitsList
-      magLimits <- magLimitsList
-      queryArgs = CatalogQuery.catalogQuery(basePosition, radiusLimits, magLimits.some)
+      magLimits    <- magLimitsList
+      mr            = MagnitudeRange(magLimits.faintnessConstraint, magLimits.saturationConstraint)
+      queryArgs     = (CatalogQuery.catalogQueryWithoutBand(basePosition, radiusLimits, mr.some), magLimits)
```
swalker2m: This isn't "wrong" and I've done it many times before too so I'm not suggesting that you change anything here.  I'm wondering though in cases like this when the midstream definitions are not used until the `yield` if they should just be moved to the yield block.  I'm thinking the purpose of defining them midstream is to use them in subsequent filters and mappings, which isn't the case here.  Opinions?

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableClient.scala
```diff
@@ -160,7 +160,12 @@ case object RemoteBackend extends CachedBackend {
 
     try {
       client.executeMethod(method)
-      VoTableParser.parse(e.url, method.getResponseBodyAsStream).fold(p => QueryResult(e.query, CatalogQueryResult(TargetsTable.Zero, List(p))), y => QueryResult(e.query, CatalogQueryResult(y).filter(e.query)))
+      VoTableParser.parse(e.url, method.getResponseBodyAsStream).fold(p => QueryResult(e.query, CatalogQueryResult(TargetsTable.Zero, List(p))), y => {
+        // Print selected stars for debugging
+        // TODO Remove
+        CatalogQueryResult(y).targets.rows.foreach(println)
```
swalker2m: Do you still want this for now?

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/obscomp/SPSiteQuality.java
```diff
@@ -197,8 +199,18 @@ public SPSiteQuality clone() {
 
         PERCENT_20("20%/Darkest", 20, 21.37),
         PERCENT_50("50%/Dark", 50, 20.78),
-        PERCENT_80("80%/Grey", 80, 19.61) { public Magnitude adjust(Magnitude mag) { return adjustIf(R, mag, -0.3); } },
-        ANY("Any/Bright", 100, 0)         { public Magnitude adjust(Magnitude mag) { return adjustIf(R, mag, -0.5); }
+        PERCENT_80("80%/Grey", 80, 19.61) {
+            private double adjustment = -0.3;
+            @Override
+            public Magnitude adjust(Magnitude mag) {
+                return adjustIf(r, adjustIf(R, adjustIf(UC, mag, adjustment), adjustment), adjustment);
```
swalker2m: I suppose we could factor out this multi-band `adjustIf` since we need it five times.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/gems/StrehlFeature.java
```diff
@@ -337,7 +332,7 @@ private int getContourPlotSize() {
             if (t == GuideProbe.Type.OIWFS || t == GuideProbe.Type.AOWFS) {
                 if (t == GuideProbe.Type.OIWFS) {
                     oiwfsCount++;
-                } else if (t == GuideProbe.Type.AOWFS) {
+                } else {
                     aowfsCount++;
                 }
             }
```
swalker2m:
````java
switch (t) {
  case OIWFS: ++oiwfsCount; break;
  case AOWFS: ++aowfsCount; break;
}
````

Or something.  No reason to check twice I guess.

bundle/edu.gemini.ags/src/main/java/edu/gemini/ags/gems/GemsGuideStarSearchOptions.java
cquiroz: I guess that is a valid point, right now for GEMS there is only UCAC4 support, but there maybe for PPMXL? Perhaps Andy can comment

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsVoTableCatalog.scala
cquiroz: I don't have a strong opinion either way but I see you point, I guess it is nicer to have one liner `yield`s than blocks

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableClient.scala
cquiroz: Yes, Andy wants to see the output of parsing the VoTable xml which isn't really human-parsable

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/obscomp/SPSiteQuality.java
cquiroz: Right, that makes sense, This is one of the uglies in this PR. I'm not happy with this but I'd rather convert `SPSiteQuality` to scala before improving
Factoring out is a good option though

bundle/edu.gemini.ags/src/main/java/edu/gemini/ags/gems/GemsGuideStarSearchOptions.java
andrewwstephens: Yes, we would like to support PPMXL (which goes fainter) for December.  In the not-to-distant future we will want to support new catalogs from Pann_STARRS and LSST.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/gems/StrehlFeature.java
cquiroz: :+1:

bundle/edu.gemini.ags/src/main/java/edu/gemini/ags/gems/GemsGuideStarSearchOptions.java
cquiroz: If it is for December I'd keep it as it is now, as simple as possible, and then improve it when needed

swalker2m: I'm fine with that, as long as you're sure it will be easy to remember where and how to replace catalog options for December.

swalker2m:
I think this looks good.  :+1:

cquiroz:
Added a commit with some small updates from the comments

PR #317, opened 2015-04-14
OCSADV-265: Preparation of instrument parameter objects for OT integration

fnussber:
This PR deals with replacing the strings in a variety of parameter objects with the typed values from the spModel in the OT. I was also able to identify some more obsolete "dat" files and got rid of them.

For now I am leaving out Michelle and T-Recs since they are not really relevant (anymore) and would need some additional massaging to bring them in sync with the OT.

For all other instruments that support imaging (GMOSN/S, F2, NIRI, GSAOI and AcqCam) I should now (finally) be able to hook them up to the OT.


bundle/edu.gemini.itc.web/src/main/scala/edu/gemini/itc/web/ITCRequest.scala
```diff
@@ -41,7 +46,10 @@ sealed abstract class ITCRequest {
   def intParameter(name: String): Int = parameter(name).toInt
 
   /** Gets the named value as a double. */
-  def doubleParameter(name: String): Double = parameter(name).toDouble
+  def doubleParameter(name: String): Double = parameter(name).trim() match {
+    case ""   => 0.0
+    case s    => s.toDouble
+  }
```
swalker2m: As I understand it, the idea is to clean up validation at some point to handle invalid input like "foo" here.  Why should `intParameter` not be given the same treatment as `doubleParameter` though?  Would it make sense to pass in the default value or is `0.0` always what we want?

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/flamingos2/Flamingos2Parameters.java
```diff
@@ -1,63 +1,51 @@
...
+    public Flamingos2Parameters(final Filter filter, final Disperser grism, final FPUnit fpMask, final ReadMode readNoise) {
+        this.filter = filter;
+        this.grism = grism;
+        this.fpMask = fpMask;
+        this.readMode = readNoise;
     }
 
-    public String getColorFilter() {
-        return _colorFilter;
+    public Filter getFilter() {
```
cquiroz: Nice to get away from raw Strings

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/niri/Niri.java
```diff
@@ -208,86 +173,101 @@ public String getDirectory() {
         return ITCConstants.LIB + "/" + INSTR_DIR;
     }
 
-    // TODO: This mimics the result of the original convoluted code. (And is probably wrong?)
-    // TODO: Verify with science and use unified getObservingStart() method from the base class.
+    // TODO: This mimics the result of the original convoluted code.
```
cquiroz: Is now probably correct?

bundle/edu.gemini.itc.web/src/main/scala/edu/gemini/itc/web/ITCRequest.scala
fnussber: Yes, I need to revisit the validation of the handling of parameters here, but I think that's going to be a bigger chunk of work. Probably the int values should be handled the same, currently this code is sort of geared towards the needs of the HTML code which does define defaults for all ints but sends an empty string for the wavelength (double). I'll add the same handling for ints though. And validation is soon to come.. I hope.

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/flamingos2/Flamingos2Parameters.java
fnussber: This is the main focus of the current work. This will make everything a ton easier.

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/niri/Niri.java
fnussber: I spare you the details. I was able to simplify that method, however it still overrides some behavior from the base class that is slightly different.. so no unfortunately, I still need to check if I can replace this instrument specific piece of code with the behavior from the base class... agh. Blech.

bundle/edu.gemini.itc.web/src/main/scala/edu/gemini/itc/web/ITCRequest.scala
fnussber: As a side note: all of that stuff is only used when accessing ITC through the web pages. Since my hope is that at some point in a not so far future all of this is replaced with something based on a decent modern web framework that has support for validation etc you could argue that it doesn't make sense to spend too much time on cleaning these things up. In the end it's not worse than what they had for the last 10 years... My main goal here was only to simplify things in a way that it is less verbose (e.g. use enum class names for the HTML parameter names and use the enum value names for the HTML option values etc).

cquiroz:
I trust this is correct, after some "montecarlo" based skimming of the code it looks fine

swalker2m:
Agreed.  As usual much improved.  :+1:

fnussber:
I refined the error handling for invalid numerics a tiny bit. I also took the opportunity to do a clean sweep of how the servlet parameters are handled.

PR #318, opened 2015-04-14
OCSADV-200-O: template numbering correction

swalker2m:
This merge installment introduces the last expected `MergePlan` correction.  It makes use of the existing `VersionTokenUtil` class to renumber template groups that are split both locally and remotely.  For example, if template group "1" is split remotely and committed to the database, we end up with groups "1" and "1.1".  If the same happens locally we have the same template groups, "1" and "1.1" but the two new "1.1" template groups are different objects with different keys. In this case the local "1.1" is renumbered to "1.2".

This correction builds on `VersionTokenUtil` to manipulate the `MergePlan` in order to guarantee that all template groups are uniquely numbered.  See `TemplateNumberingCorrection`.   It also fixes a bug in which template observations were not being considered in the observation numbering correction.

### DIFF
--- a/bundle/edu.gemini.pot/src/test/scala/edu/gemini/pot/sp/ProgramGen.scala
+++ b/bundle/edu.gemini.pot/src/test/scala/edu/gemini/pot/sp/ProgramGen.scala
@@ -99,7 +99,7 @@ object ProgramGen {
       val tf  = ftf(f, p)
       val tgs = ftg.sequenceU.apply(f, p).zipWithIndex.map { case (tg, i) =>
         val dob = tg.getDataObject.asInstanceOf[TemplateGroup]
-        dob.setVersionToken(VersionToken.apply(Array(i), 1))
+        dob.setVersionToken(VersionToken.apply(Array(i+1), 1))
         tg.setDataObject(dob)
         tg
       }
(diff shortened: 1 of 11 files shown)


bundle/edu.gemini.pot/src/test/scala/edu/gemini/pot/sp/ProgramGen.scala
```diff
@@ -136,6 +166,17 @@ object ProgramGen {
       n.title = decorativeTitle(n, title)
     }
 
+  val genEditSplitTemplateGroup: Gen[ProgEdit] =
+    for {
+      tgf <- maybePickTemplateGroup
+    } yield { (f: ISPFactory, p: ISPProgram) =>
```
cquiroz: I didn't know you could do this, will that split a tuple `tgf`?

cquiroz:
I can't follow it very well but it seems good, Having the tests is probably key here :+1:

bundle/edu.gemini.pot/src/test/scala/edu/gemini/pot/sp/ProgramGen.scala
swalker2m: `tgf` stands for "template group function" here.  It's the function that when applied with a program will select a random `ISPTemplateGroup` from the program, if there is at least one.  The `for` expression in turn just yields a function that when provided with a program factory and a program, will split a random template group in the program, if there is one.

Unfortunately everything is awkward like this because of the mutable program model.  The generator doesn't actually mutate anything, it just generates functions that when handed a program do the actual mutation.

PR #320, opened 2015-04-15
OCSADV-297: All guide star quality icons change when switching stars

cquiroz:
This bug originates from the lack of a capability to to analyze a single guide star on `AgsStrategy` interface. This new method has been added on this PR, plus tests and refactorings in a few places

### DIFF
--- a/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/GemsStrategy.scala
+++ b/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/GemsStrategy.scala
@@ -69,12 +69,12 @@ trait GemsStrategy extends AgsStrategy {
     }
 
     VoTableClient.catalogs(adjustedConstraints, backend).flatMap {
-      case result if result.exists(_.result.containsError) => Future.failed(CatalogException(result.map(_.result.problems).flatten))
+      case result if result.exists(_.result.containsError) => Future.failed(CatalogException(result.flatMap(_.result.problems)))
       case result                                          => Future.successful {
-        result.map { r =>
+        result.flatMap { r =>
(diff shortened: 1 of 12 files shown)


bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/GemsStrategy.scala
```diff
@@ -105,6 +105,9 @@ trait GemsStrategy extends AgsStrategy {
     cans ++ odgw
   }
 
+  // TODO Implement for GEMS
+  override protected [ags] def analyze(ctx: ObsContext, mt: MagnitudeTable, guideProbe: ValidatableGuideProbe, guideStar: SiderealTarget): Option[AgsAnalysis] = None
```
swalker2m: This may be the reason why this wasn't implemented before.  Will it be possible to validate one probe in isolation when they depend upon each other as for Canopus?

cquiroz: Good question, it should be possible as it was done before, the `analyze` method in `AgsAnalysis` used to be called for GEMS and non-GEMS irrespective but I think it will default in GEMS to do an analyze of all the probes and return the same value for each guide star.
This should be part of the unfinished OCSADV-272 task

swalker2m:
If the analysis takes into account the context, as it must, it will know the guide stars assigned to the various probes.  In particular, Canopus 3 which controls the range of 1 and 2 as I recall.

From what I can tell, it looks good.  Use of Java 8 lambdas cleans up that old Java code too.  :+1:

cquiroz:
No going back to 1.7 :smile:

PR #321, opened 2015-04-16
OCSADV-299: fixed case of directory name.

jluhrs:


arturog8m:
after careful review, I think this is good to go :+1:

cquiroz:
Seems fine :smile:

PR #322, opened 2015-04-16
OCSADV-266: Integration of ITC imaging calculations except for GeMS

fnussber:
This PR deals with translating the instrument configurations that are extracted from the sequence model into simple transfer objects (Scala case classes for the most part) that can be used in order to talk to the ITC service.

I also hooked up the service calls for all instruments that support imaging (except for Michelle and TRecs, which may or may not be added later on depending on time and priorities). Results are displayed in the corresponding ITC result tables, which need some improvements but all in all this allows already to get a first impression of how things will work.

One thing that is missing is getting the correct parameters for GeMS, this will follow in a future PR.

There have also been quite a few refactorings and code cleanups. The most relevant point for review I guess is the ```ConfigExtractor``` class.


bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/ConfigExtractor.scala
```diff
@@ -0,0 +1,214 @@
...
+      None.right
+    }
+  }
+
+  private def extractObservingWavelength(c: Config): String \/ Double =
+    // Note: observing wavelength will only be available if instrument is configured for spectroscopy
+    if (c.containsItem(ObsWavelengthKey)) extractDoubleFromString(c, ObsWavelengthKey) else 0.0.right
+
+  // Extract a value of the given type from the configuration
+  private def extract[A](c: Config, key: ItemKey)(implicit clazz: ClassTag[A]): String \/ A =
```
cquiroz: Nice, @swalker2m Should we promote these methods to be usable from anywhere on the OCS?

bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/ConfigExtractor.scala
```diff
@@ -0,0 +1,214 @@
...
+
+  // Calculate distance between two coordinates in arc seconds
+  private def distance(t0: SPTarget, t1: SPTarget) = {
+    val c0 = toCoordinates(t0)
+    val c1 = toCoordinates(t1)
+    Coordinates.difference(c0, c1).distance.toArcsecs
+  }
+
+  // Turn SPTarget into new target model coordinates, note that DMS.getValue() guarantees value to be in [-90..90]
+  private def toCoordinates(t: SPTarget) = {
```
cquiroz: A very similar method is present on `ModelConverters`, perhaps we could put it there

bundle/edu.gemini.itc/src/main/scala/edu/gemini/itc/service/ItcServiceImpl.scala
```diff
@@ -1,48 +1,64 @@
...
+  // === Imaging
+
+  private def calculateImaging(source: SourceDefinition, obs: ObservationDetails, cond: ObservingConditions, tele: TelescopeDetails, ins: InstrumentDetails): Result =
+    ins match {
+      case i: AcquisitionCamParameters  => imagingResult(new AcqCamRecipe(source, obs, cond, tele, i))
+      case i: Flamingos2Parameters      => imagingResult(new Flamingos2Recipe(source, obs, cond, i, tele))
+      case i: GmosParameters            => imagingArrayResult(new GmosRecipe(source, obs, cond, i, tele))
+      case i: GsaoiParameters           => imagingResult(new GsaoiRecipe(source, obs, cond, i, tele))
+      case i: NiriParameters            => imagingResult(new NiriRecipe(source, obs, cond, i, tele))
+      case _                            => throw new NotImplementedError
```
cquiroz: I think you need the default case even though `InstrumentDetails` is a sealed trait because you are handling fewer cases, right?
Would it be better that instead of throwing the error we handle the case via `Option` or `\/`?

cquiroz:
Seems fine to me, I wonder if the `ConfigExtractor` should be made into a common utility class for as long as we deal with the old model from scala

bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/ConfigExtractor.scala
fnussber: If those methods are considered generally useful I can extract them and add them to the core and spruce them up a bit.

bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/ConfigExtractor.scala
fnussber: Ok, I changed that. Thanks for pointing this out.

bundle/edu.gemini.itc/src/main/scala/edu/gemini/itc/service/ItcServiceImpl.scala
fnussber: I changed some of the error handling here to makes use of ```Validation``` however I kept the lower level error message as an exception. The point here is that the error/validation/problem handling in the original Java code uses exceptions everywhere which are caught on the highest level (i.e. servlet and now the service implementation). Changing this might be worthwhile for some parts but probably not for everything (just because it's probably not worth the time to rewrite all of that in Scala). I haven't put any effort into this yet because it doesn't seem to be a hight priority, however I might change some of the input parameter checks at some point in the future.

PR #323, opened 2015-04-16
OCSADV-296: Bug fix: only look at last part in data label for index

fnussber:
Small bug fix.

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcUniqueConfig.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcUniqueConfig.scala
@@ -92,11 +92,17 @@ object ItcUniqueConfig {
   // Creates a text label based on the spans of steps covered by the steps.
   // E.g. ((1,2,3),(10,11),(15)) is turned into "001-003, 010-011, 015"
   private def labels(cs: Seq[Config]): String = {
-    val ls = cs.map(_.getItemValue(DATALABEL_KEY).asInstanceOf[String].toInt).sorted
+    val ls = cs.map(labelIndex).sorted
     findSpans(ls).map {
       case i :: Nil => i.toString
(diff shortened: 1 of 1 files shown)


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcUniqueConfig.scala
```diff
@@ -92,13 +92,19 @@ object ItcUniqueConfig {
...
     findSpans(ls).map {
       case i :: Nil => i.toString
       case i :: is => f"$i%03d-${is.last}%03d"
     }.mkString(", ")
   }
 
+  // Only the number at the very end is relevant
+  private def labelIndex(c: Config): Int = {
+    val label = c.getItemValue(DATALABEL_KEY).asInstanceOf[String]
+    label.split('-').last.toInt
```
cquiroz: I think this may NPE if there is no "-" on the string

fnussber: This will return an empty list if there is no ```'-'```.

cquiroz: I think `last` on an empty list will throw an exception

fnussber: Even for an empty string ```split()``` does not return an empty array but an array with an empty string:

```scala
scala> "".split('-')
res0: Array[String] = Array("")

scala> "123".split('-')
res1: Array[String] = Array(123)

scala> "123-123".split('-')
res2: Array[String] = Array(123, 123)
```

cquiroz: my bad :+1:

PR #324, opened 2015-04-17
OCSADV-200-P: merge conflict notes

swalker2m:
This installment adds merge conflict notes as appropriate.  Conflict notes are issues that must be brought to the attention of the user.  The already exist in the program model and are already handled by the OT.  The user is forced to acknowledge all conflicts before they can send changes to the server.

This code updates the `Modified` `MergeNode` with a field for `Conflicts` from the existing program model and augments the merge algorithm to add the appropriate notes.  It also updates `VcsServer` to reject attempts to store differences that contain conflict notes.

### DIFF
--- a/bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs/diff/PreliminaryMerge.scala
+++ b/bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs/diff/PreliminaryMerge.scala
@@ -161,15 +161,15 @@ object PreliminaryMerge {
     def addDataObjectConflicts(in: Tree[MergeNode]): TryVcs[Tree[MergeNode]] = {
       val commonKeys = mc.remote.diffMap.keySet & mc.local.nodeMap.keySet
 
-      def dobConflicts(k: SPNodeKey): Boolean = {
-        lazy val local  = mc.local.nodeMap(k).getDataObject
-        lazy val remote = mc.remote.diffMap(k).rootLabel match {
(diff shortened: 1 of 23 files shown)


bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs/diff/PreliminaryMerge.scala
```diff
@@ -156,6 +158,110 @@ object PreliminaryMerge {
         }
       }
 
-    go(Both(mc.local.prog, mc.remote.plan.update))
+    def addDataObjectConflicts(in: Tree[MergeNode]): TryVcs[Tree[MergeNode]] = {
+      val commonKeys = mc.remote.diffMap.keySet & mc.local.nodeMap.keySet
+
+      def dobConflicts(k: SPNodeKey): Boolean = {
+        lazy val local  = mc.local.nodeMap(k).getDataObject
```
cquiroz: What's the benefit of being lazy here? I always think that `lazy` is a rare situation to be in

cquiroz:
I can't really follow the details but with so many tests we cannot go wrong :smile: 
Nice to see Java8 lambdas creeping up :+1:

bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs/diff/PreliminaryMerge.scala
swalker2m: I think you're right, I just rearranged it to produce the same result w/o `lazy`.

PR #325, opened 2015-04-18
OCSADV-256: Fixed bug with regards to vignetting calculation

sraaphorst:
Science area was translated to base position + science offset, but guide stars were only translated to their offset from the base position, thus resulting in vignetting of 0 when base position was not near (0,0).

This breaks the previous test cases for AGS, probably because the brightest guide star is no longer picked for GMOS and F2 since guide stars are binned by image quality and then picked by min vignetting, with ties broken by brightness. I don't have time to fix this, though, before leaving for vacation, and I wanted to make sure that these changes were available for the next build, hence the PR.

I will have limited e-mail access as of Sunday evening should anything come up.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/guide/ScalaGuideProbeUtil.scala
+++ b/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/guide/ScalaGuideProbeUtil.scala
@@ -30,15 +30,27 @@ object ScalaGuideProbeUtil {
           area.add(new Area(s))
           area
       }
+
+      // Base position coordinates.
+      val baseX = ctx.getBaseCoordinates.getRa.toArcsecs.getMagnitude
+      val baseY = ctx.getBaseCoordinates.getDec.toArcsecs.getMagnitude
+
       val vignettingSum = ctx.getSciencePositions.toList.foldLeft(0.0){
         case (currentSum,offset) =>
           // Find the probe arm adjustment, which consists of the arm angle and guide star location in arcsec.
           // If an adjustment exists, calculate the vignetting for this offset.
           val vignetting = probeArmGeometry.armAdjustment(ctx, coordinates, offset).map { armAdjustment =>
             // Adjust the science area for the current offset.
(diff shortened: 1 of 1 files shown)


cquiroz:
Could it help if we fixed the test cases or is it better to ignore the tests?

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/guide/ScalaGuideProbeUtil.scala
```diff
@@ -30,15 +30,27 @@ object ScalaGuideProbeUtil {
...
-            val trans = AffineTransform.getTranslateInstance(x, y)
+            val offX = -(offset.p.toNewModel.toNormalizedArcseconds)
+            val offY = -(offset.q.toNewModel.toNormalizedArcseconds * flip)
+
+            // Instead of performing the intersection of the science area at its real position of (base pos + sci offset)
+            // and the probe arm at its real position of (base pos + guide star offset), we instead perform the
+            // intersection by factoring out the base position from both and translating the science area to sci offset
+            // and the probe arm position at guide star offset. This should always be equivalent.
+            // Since FeatureGeometry.transformScienceAreaForContext moves the science area to the base position, we
+            // move it back to (0,0) + offset in order to factor out the base position.
```
swalker2m: Why does `FeatureGeometry` translate the science area to the absolute base position?  I think all the math should be done as offsets from (0,0) and the absolute base position doesn't enter into the picture.  It is surprising to me at least to see these giant translations.

swalker2m: It seems like we might be mixing distances represented as the angular separation of offset positions, detector / probe arm definitions, etc. with the absolute spherical coordinates of the base position, guide stars, etc.

swalker2m: Trying these changes out, with offset positions I'm still seeing 0 vignetting for selections that clearly vignette.

PR #326, opened 2015-04-19
OCSADV-293: Support Manual GS Search for AGS

cquiroz:
This PR contains a quick solution to integrate VoTable searches into the existing Catalog Navigator. It would be very labor intensive to integrate `VoTableClient` directly into the SkyCat catalog code, and that code will be removed in the future. Hence, a short term solution was devised where a servlet running on the ODB would act as a proxy calling VoTable and exposing an SkyCat compatible output.

Given that this is a short time solution, `Votable2SkyCatalogServlet` and `Ucac4SkyObjectFactory` are born @Deprecated and there are no tests.

This was also a good chance to clean a bit some of the warnings about rawtype usage and other bits on the catalog legacy code

Note that to simplify development, the catalog proxy is running on VM on SBF, and called from the OT. It needs to be updated before the release to point to the real ODB.


bundle/edu.gemini.catalog/src/main/java/edu/gemini/catalog/skycat/table/CatalogValueExtractor.java
```diff
@@ -82,47 +80,17 @@ public int hashCode() {
         Option<T> extract(CatalogRow row, int index) throws CatalogException;
     }
 
-    private static final Extractor<Angle> RA_EXTRACTOR = new Extractor<Angle>() {
-        @Override
-        public Option<Angle> extract(CatalogRow row, int index) throws CatalogException {
-            return row.getRa(index);
-        }
-    };
+    private static final Extractor<Angle> RA_EXTRACTOR = CatalogRow::getRa;
```
swalker2m: I suppose I'm going to have to actually read what is new in Java 8!

bundle/edu.gemini.catalog/src/main/java/jsky/catalog/AbstractCatalogDirectory.java
```diff
@@ -582,7 +549,7 @@ private TreeModelEvent _getTreeModelEvent(Catalog cat) {
         int n = l.size();
         Catalog[] ar = new Catalog[n];
         for (int i = 0; i < n; i++)
-            ar[n - i - 1] = (Catalog) l.get(i);
+            ar[n - i - 1] = l.get(i);
```
swalker2m: As long as we're cleaning things up, we could make `l` an `ArrayList` and then something like this:

```java
Collections.reverse(l);
return l.toArray(new Catalog[l.size()]);
```

bundle/edu.gemini.catalog/src/main/java/jsky/catalog/skycat/SkycatOutputAdapterFactoryRegistrar.java
```diff
@@ -36,38 +36,27 @@
...
+            } catch (Exception ex) {
+                LOG.log(Level.WARNING, "Problem extracting object from " + className, ex);
             }
+            return None.instance();
+
         };
 
     private static final SkycatCatalogObjectRegistrar<SkycatOutputAdapter.Factory> reg =
-            new SkycatCatalogObjectRegistrar<SkycatOutputAdapter.Factory>(OUTPUT_ADAPTER, INST);
+            new SkycatCatalogObjectRegistrar<>(OUTPUT_ADAPTER, INST);
```
swalker2m: `SkycatOutputAdapterFactoryRegistrar`, viva OO!

swalker2m:
Given the circumstances this all seems reasonable.  I find it a bit hard to follow `Votable2SkyCatalogServlet` with (what _I think_ are) disjunctions nested in `Option`s, etc. outside of an IDE and w/o explicit type labels. I think that's the nature of the business though.  :+1:

bundle/edu.gemini.catalog/src/main/java/jsky/catalog/AbstractCatalogDirectory.java
cquiroz: The cleanup was mostly about getting rid of compiler warnings, making the code proper may be too big an enterprise if we are going to delete the skycat code.

cquiroz:
Perhaps I could document the servlet a bit more. It all boils down to extract parameters from the request (Some optional/some not) and parsing them into doubles or magnitude bands when necessary

PR #327, opened 2015-04-19
OCSADV-300: shell seqexec client

swalker2m:
This PR creates a skeleton seqexec command line client and a shared server bundle.  It is meant to be used as an example and starting point for building the seqexec, and accordingly adds a new "seqexec" application to the build.

The client bundle offers Felix shell extensions with simple commands to view observation sequences.  These are clearly primitive but may serve as an example:

* Set the database from which sequences may be fetched:
```
g! seq host localhost:8443
Default seq host set to localhost 8443
```

* View the database from which sequences will be fetched:
```
g! seq host
Default seq host set to localhost 8443
```

* Get the number of datasets that will be produced by an observation:
```
g! seq show GS-2014B-Q-2-355 count
GS-2014B-Q-2-355 sequence has 24 steps.
```
* Show all static (unchanging) settings for an observation sequence:
```
g! seq show GS-2014B-Q-2-355 static
GS-2014B-Q-2-355 Static Values
calibration:basecalDay             -> true
calibration:basecalNight           -> false
calibration:coadds                 -> 1
calibration:diffuser               -> IR
calibration:exposureTime           -> 60.0
calibration:filter                 -> ND4-5
calibration:lamp                   -> [IR grey body - high]
instrument:decker                  -> IMAGING
instrument:disperser               -> NONE
instrument:exposureTime            -> 85.0
instrument:filter                  -> J
instrument:fpu                     -> FPU_NONE
instrument:instrument              -> Flamingos2
...
```
* Show all static (unchanging) settings for a particular subsystem of an observation sequence:
```
g! seq show GS-2014B-Q-2-355 static instrument
GS-2014B-Q-2-355 Static Values (instrument only)
instrument:decker                  -> IMAGING
instrument:disperser               -> NONE
instrument:exposureTime            -> 85.0
instrument:filter                  -> J
instrument:fpu                     -> FPU_NONE
instrument:instrument              -> Flamingos2
instrument:issPort                 -> Side-looking
instrument:lyotWheel               -> OPEN
instrument:mosPreimaging           -> No
instrument:observingWavelength     -> 1.25
instrument:posAngle                -> 0.0
instrument:readMode                -> BRIGHT_OBJECT_SPEC
instrument:useElectronicOffsetting -> false
instrument:version                 -> 2009A-1
```
* Show dynamic (iterated) values for a particular step of an observation:
```
g! seq show GS-2014B-Q-2-355 dynamic 10
GS-2014B-Q-2-355 Dynamic Values (Step 10)
calibration:shutter  -> Closed
observe:dataLabel    -> GS-2014B-Q-2-355-010
observe:exposureTime -> 60.0
observe:object       -> GCALflat
observe:observeType  -> FLAT
```
* Show dynamic (iterated) values for a particular step and subsystem:
```
g! seq show GS-2014B-Q-2-355 dynamic 10 calibration
GS-2014B-Q-2-355 Dynamic Values (Step 10, calibration only)
calibration:shutter -> Closed
```

### DIFF
--- a/app/all-bundles/build.sbt
+++ b/app/all-bundles/build.sbt
@@ -35,6 +35,8 @@ ocsAppManifest := {
     BundleSpec((sbt.Keys.name in bundle_edu_gemini_qpt_server).value, Version.parse((sbt.Keys.version in bundle_edu_gemini_qpt_server).value)),
     BundleSpec((sbt.Keys.name in bundle_edu_gemini_qpt_shared).value, Version.parse((sbt.Keys.version in bundle_edu_gemini_qpt_shared).value)),
     BundleSpec((sbt.Keys.name in bundle_edu_gemini_qv_plugin).value, Version.parse((sbt.Keys.version in bundle_edu_gemini_qv_plugin).value)),
+    BundleSpec((sbt.Keys.name in bundle_edu_gemini_seqexec_client).value, Version.parse((sbt.Keys.version in bundle_edu_gemini_seqexec_client).value)),
+    BundleSpec((sbt.Keys.name in bundle_edu_gemini_seqexec_shared).value, Version.parse((sbt.Keys.version in bundle_edu_gemini_seqexec_shared).value)),
     BundleSpec((sbt.Keys.name in bundle_edu_gemini_services_client).value, Version.parse((sbt.Keys.version in bundle_edu_gemini_services_client).value)),
     BundleSpec((sbt.Keys.name in bundle_edu_gemini_services_server).value, Version.parse((sbt.Keys.version in bundle_edu_gemini_services_server).value)),
     BundleSpec((sbt.Keys.name in bundle_edu_gemini_shared_ca).value, Version.parse((sbt.Keys.version in bundle_edu_gemini_shared_ca).value)),
--- a/app/seqexec/build.sbt
+++ b/app/seqexec/build.sbt
@@ -1,0 +1,118 @@
+
+import OcsKeys._
+import edu.gemini.osgi.tools.Version
+import edu.gemini.osgi.tools.app.{ Configuration => AppConfig, _ }
+import edu.gemini.osgi.tools.app.Configuration.Distribution.{ Test => TestDistro, _ }
+import OcsCredentials.Qpt._
+
+ocsAppSettings
+
+// Application project for Queue Planning Tool
+ocsAppManifest := {
+  val v = ocsVersion.value.toBundleVersion
+  Application(
+    id = "seqexec",
+    name = "SeqExec",
+    version = ocsVersion.value.toString,
+    configs = List(
+      common(v),
+        development(v),
+        with_test_dbs(v),
+          mac_test(v),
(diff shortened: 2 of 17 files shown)


cquiroz:
This is interesting, some general questions:

Is this really another application, it looks more like an spdb bundle, or will this be the basis for the future seqexec?

Will it use the same version as the OCS?

I'm a bit worried about adding more apps to the build, reloads are already taking 30m+ and though infrequent it can add especially if you need to switch elements on the build

swalker2m:
> Is this really another application, it looks more like an spdb bundle, or will this be the basis for the future seqexec?

The idea was to build upon these bundles to create the seqexec. I guess actually there are two applications though, the seqexec server and the seqexec client.

> Will it use the same version as the OCS?

Seems like it should, yes.  It's closely tied to the model just as the ODB, OT, and QPT.

> I'm a bit worried about adding more apps to the build, reloads are already taking 30m+ and though infrequent it can add especially if you need to switch elements on the build.

I share your concern but it's what we've got until the build is broken up.  I would guess the issue is more likely the number of bundles than the number of apps but I don't know.

cquiroz:
In that sense the better candidate to be another project is PIT which has its own release cycle
I also think is the number of bundles and their interdependencies, maybe is time to think about merging some bundles

tpolecat:
@cquiroz what? 30 minutes to reload? Surely you mean 30 seconds. If it's getting slow, do `reload plugins` then `clean` then `reload return` and then exit sbt. Next time you start up will be faster.

cquiroz:
Yes, I've seen 30 minutes if the structure of the build changes, so it happens not that often. It can take that much when e.g. you add a dependency or when we moved to Java 1.8

Regular reloads are much faster

On abr 19, 2015, at 23:00, "Rob Norris" <notifications@github.com<mailto:notifications@github.com>> wrote:


@cquiroz<https://github.com/cquiroz> what? 30 minutes to reload? Surely you mean 30 seconds. If it's getting slow, do reload plugins then clean then reload return and then exit sbt. Next time you start up will be faster.

-

swalker2m:
@cquiroz @tpolecat  I just switched to a different branch w/o all the updates in this PR and it took about 28 minutes.  Maybe there is some obvious issue that would fall out of profiling.

jluhrs:
Thanks for the example. If I understand correctly, the "service" in the name of SeqExecService comes from the point of view of the ODB, where SeqExecService exposes a service to retrieve sequences.
I had already created a seqexec-server app and bundle, and I was planning on calling the UI application  seqexec-client.

swalker2m:
@jluhrs That sounds fine.  If you want you can just incorporate the "shared" bundle into the work you're doing and rename or rewrite `SeqExecService` as necessary.  I realized after the fact that we really need two applications and that "client" probably wasn't a good name for the one with the command extensions.

The commands do work though and it serves as an example of part of the `ConfigSequence` API.

swalker2m:
Dropping in favor of OCSADV-275.

PR #328, opened 2015-04-21
OCSADV-256: restore TPE version of GMOS and F2 OI and science area

swalker2m:
There is an issue with the new vignetting calculations and the new GMOS and F2 science area and probe arm models.  This PR restores the old OT/TPE definitions of GMOS/F2 features and disables the vignetting calculation while leaving all the code for the new work in place but unused.  This accomplishes a few goals:

* Allows us to put together a test release for the other OT work for 2015B
* Provides a fallback for the 2015B release if we cannot sort the problems out in time
* Gives us a quick reference to the old implementation to help work on the issue with the new model

Note that the GMOS/F2 classes `Flamingos2_OIWFS_Feature`, `Flamingos2_SciAreaFeature`, `GMOS_OIWFS_Feature`, and `GMOS_SciAreaFeature` are all simply restored from the repository and are intended to be temporary.  No effort was placed in modernizing them since that has already been done in the new model.

### DIFF
--- a/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/SingleProbeStrategy.scala
+++ b/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/SingleProbeStrategy.scala
@@ -94,7 +94,8 @@ case class SingleProbeStrategy(key: AgsStrategyKey, params: SingleProbeStrategyP
           }
           val bestPerCtx = for {
             (c, soList) <- results
-            rating      <- brightestByQualityAndVignetting(soList, mt, c, v, params)
+//            rating      <- brightestByQualityAndVignetting(soList, mt, c, v, params)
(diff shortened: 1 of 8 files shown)


cquiroz:
Do you need to also update the `match` on `SingleProbeStrategy` on the `select` function?

cquiroz:
Seems a good rollback :+1:

swalker2m:
> Do you need to also update the match on SingleProbeStrategy on the select function?

I just removed the vignetting factor but I think you're right.  It's probably better to just skip the whole vignetting block.  In fact I just did so and found that breaks a test case.  Upon investigation the test case was expecting the wrong result. So there is something a bit off with the vignetting block other than the vignetting calculation.

krumpi:
Looks good, thanks for the effort tracking this one

cquiroz:
Looks good, thanks for the effort tracking this one

PR #329, opened 2015-04-21
OCSADV-200-Q: Rename edu.gemini.sp.vcs.diff to edu.gemini.sp.vcs2

swalker2m:
This is a trivial PR that just moves and renames the package `edu.gemini.sp.vcs.diff` to `edu.gemini.sp.vcs2`.  That gives it a distinct package root and makes it easier to export the services and commands in the contained `edu.gemini.sp.vcs2.osgi` package.  This is required to start the next step, OT integration with the new VCS service.


tpolecat:
:+1:

PR #331, opened 2015-04-21
Google Services Ignore

swalker2m:
This is a tiny PR that adds a bundle property to the spdb app.  It's called `edu.gemini.services.server.start` and if not present it defaults to `true`.  You can define it and set it to `false` though and the spdb won't try to do the google services login.  I added that property to my own spdb configuration since I've never worked with anything that needs it.

In theory relinking the credentials files will make the login work, and it does for a time.  Eventually though it stops working and you have to relink.  Perhaps it is because the resources directory is missing, I'm not sure.  The end result though is that usually while trying to use the spdb command line during development it is spinning and writing out google services login exceptions.  Also, I'm often at home with a saturated 3G network and logging in to anything is problematic.  This update just allows anyone to turn the google services login and `TelescopeScheduleService` registration off.

I noticed a potential race condition in the `Activator`, by the way.  Since `start` kicks off a thread that retries the login, it seems like it could happen that you've stopped the bundle before the registration is complete.  I'm not going to touch that though since the goal is just to make it possible to turn off the login attempts.

### DIFF
--- a/app/spdb/build.sbt
+++ b/app/spdb/build.sbt
@@ -235,6 +235,7 @@ def swalker(version: Version) = AppConfig(
     "edu.gemini.dataman.xfer.tempDir.base"       -> "/Users/swalker/dataman/base/.xfer",
     "edu.gemini.dataman.xfer.tempDir.gsa"        -> "/Users/swalker/dataman/gsa/.xfer",
     "edu.gemini.datasetfile.workDir"             -> "/Users/swalker/dataman/work",
+    "edu.gemini.services.server.start"           -> "false",
     "edu.gemini.smartgcal.host"                  -> "localhost",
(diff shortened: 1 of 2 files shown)


cquiroz:
Nice, I also get those exceptions
I guess you don't need to explicitly set it to `true` on the spdb, right?

swalker2m:
Correct, if you do nothing it will try to login and register.  I'll leave the test and production configurations that way.

tpolecat:
In principle you should be able to stop that bundle to achieve the same outcome, but right now this doesn't work, as reported in [OCSINF-436](http://swgserv01.cl.gemini.edu:8080/browse/OCSINF-436). I think your solution is good for now but we need a more systematic approach eventually. :+1:

PR #332, opened 2015-04-22
OCSADV-303:Ignore bogus magnitude values on UCAC4

cquiroz:
The UCAC4 catalog sometimes includes magnitudes that should be ignored, this PR updates those rules to discard magnitudes with value 20 OR with error >= 0.9
Before, magnitudes were discarded if both conditions were true

### DIFF
--- a/bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableParser.scala
+++ b/bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableParser.scala
@@ -62,8 +62,8 @@ case object UCAC4Filter extends MagnitudesFilter {
 
   // UCAC4 ignores A-mags
   override def ignoredMagnitudeField(v: FieldId) = v.id === "amag" || v.id === "e_amag"
-  // Magnitudes with value 20 and error over 0.9 are invalid
-  override def validMagnitude(m: Magnitude) = m.value =/= ucac4BadMagnitude || m.error.map(math.abs) < ucac4BadMagnitudeError
+  // Magnitudes with value 20 or error over or equal to 0.9 are invalid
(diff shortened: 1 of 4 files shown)


bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableParser.scala
```diff
@@ -63,7 +63,7 @@ case object UCAC4Filter extends MagnitudesFilter {
   // UCAC4 ignores A-mags
   override def ignoredMagnitudeField(v: FieldId) = v.id === "amag" || v.id === "e_amag"
   // Magnitudes with value 20 and error over 0.9 are invalid
-  override def validMagnitude(m: Magnitude) = m.value =/= ucac4BadMagnitude || m.error.map(math.abs) < ucac4BadMagnitudeError
+  override def validMagnitude(m: Magnitude) = m.value =/= ucac4BadMagnitude && m.error.map(math.abs) <= ucac4BadMagnitudeError
```
tpolecat: It makes me so happy to see `===` and `=/=`.

tpolecat:
:+1:

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableParser.scala
```diff
@@ -63,7 +63,7 @@ case object UCAC4Filter extends MagnitudesFilter {
   // UCAC4 ignores A-mags
   override def ignoredMagnitudeField(v: FieldId) = v.id === "amag" || v.id === "e_amag"
   // Magnitudes with value 20 and error over 0.9 are invalid
```
cquiroz: Just noted that, as always, documentation is outdated, it should be 'or' rather than 'and'

tpolecat: Or just delete the comment. The code is quite clear.

PR #333, opened 2015-04-22
OCSADV-295: UI improvements

fnussber:
This PR incorporates a collection of suggestions and improvements from Andy after he played around with the ITC integration for a bit. It's mostly improvements in the ITC tables (tooltips, alignment etc).

Furthermore, the source magnitude and band are now extracted from the science target depending on the instrument configuration, i.e. the magnitude of the band is taken that corresponds most closely to the observing wavelength of the instrument configuration. Unifying the bands from spModel and ITC is planned as a future task, currently there's a mapping between the two. The source mag and band that's been selected for each instrument configuration is now displayed in the results table.

Converting the peak pixel flux to ADU has been left out for now, because currently there is no general knowledge of the gain for each instrument. This may or may not be added in a future PR.

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcTable.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcTable.scala
@@ -32,17 +32,17 @@ object ItcTable {
   abstract sealed class Renderer[A](alignment: Alignment.Value, f: A => (Icon, String)) extends LabelRenderer[A](f)  {
     override def componentFor(table : Table, isSelected : Boolean, hasFocus : Boolean, a : A, row : Int, column : Int) : Component = {
       val c = super.componentFor(table, isSelected, hasFocus, a, row, column)
-      val l = c.asInstanceOf[Label]
       val model = table.model.asInstanceOf[ItcTableModel]
       // Cell renderer based on the sequence cell renderer used for other sequence tables. This gives us coherent
       // formatting and color coding throughout the different tables in the sequence node.
       val bg = model.getKeyAt(column).map(SequenceCellRenderer.lookupColor)
       val tt = model.tooltip(column)
       // set horizontal alignment, bg color and tooltip as needed
-      l <|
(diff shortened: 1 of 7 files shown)


cquiroz:
Would you upload some screenshots?

fnussber:
An example screenshot for F2:
image

cquiroz:
Looks nice, perhaps the alignment on the last columns is a bit tight, could there be same space between the S/N Total column and the Messages column?

fnussber:
Yes, I need to prettify that a bit more. The last column is a bit special because I still don't know exactly what to do about that. Some of those messages might turn into Phase2 warnings/errors. Some could maybe show up at the bottom of the panel (like error messages for AGS).

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcTable.scala
```diff
@@ -128,6 +179,62 @@ trait ItcTable extends Table {
...
+    }
+
+    for {
+      wl  <- ConfigExtractor.extractObservingWavelength(c)
+      mag <- mags(wl*1000) // convert wavelength from micrometer [um] to nanometer [nm]
+    } yield {
+      val b = mag.getBrightness
+      mag.getBand match {
+        // TODO: unify band definitions from spModel core and itc shared so that we don't need this translation anymore
+        case Magnitude.Band.u  => (b, WavebandDefinition.U)
```
cquiroz: You could consider going straight to the new model band classes

cquiroz:
Looks good to me :+1:

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcTable.scala
fnussber: Oh. We actually have 3 band definitions now: new and old from core model and one from ITC. Ok, I'll clean that up when I replace the ITC bands with the new core model ones... watch out for OCSADV-305.

PR #334, opened 2015-04-23
OCSADV-294: Support Manual GS Search for GEMS and UCAC4

cquiroz:
This PR updates the manual guide star search feature to work with UCAC4.

Note that this is not the final PR as we may need to support non UCAC4 catalogs too, but we can use to test the basic functionality. Currently, queries to any other catalog will go to UCAC4.

The usual code cleanups and conversions of lambdas, raw types, etc.. are included

Here are a couple of screenshots of the Dialog box

screenshot 2015-04-23 08 17 27

screenshot 2015-04-23 09 31 02


fnussber:
Looks good to me. In particular the cleanups in the Java code are great. It totally pays that we moved to Java 8 if for nothing else than the lambdas!

PR #336, opened 2015-04-25
OCSADV-307B: UI control elements for conditions, source profile and source distribution

fnussber:
This PR adds user controls to change the conditions on-the-fly and define the source profile and distribution to be used for the ITC calculations.

The source profile and distribution UI elements will eventually find a new home on the TargetEditor 2.0 once it is ready. The user controls for the conditions are the result of some user feedback that was asking for a quick way to compare the values calculated for the program conditions with the values that result from other conditions (e.g. current conditions during the night or for what-if analysis). The conditions editor will stay on the ITC panel, values that differ from the program conditions are highlighted in red. Those values are not meant to be persistent.

The edit fields for the source (label, value, unit) might be interesting in other places as well? I don't think we have something like that. Values that are invalid (i.e. not numbers) are highlighted in red. If there is any interest, we can maybe extract that code and move it to a more central place. 

image

This PR will give Andy (and other users) the possibility to start checking the main functionality. There are still some details missing (e.g. Gems integration) but we're slowly getting there.

Note that the UI still won't win any beauty awards, but since stuff will keep moving around in the next PRs I am not yet willing to spend time to do fine adjustments.


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcPanel.scala
```diff
@@ -21,13 +30,313 @@ object ItcPanel {
...
+      private def color() = if (inSync()) Color.BLACK else Color.RED
+
+      private def tt() = if (inSync()) "" else s"Program condition is ${renderFunc(programValue)}"
+
+      private def inSync() = programValue == selection.item
+
+    }
+
+    private val t  = new Label("Conditions:")
+    private val sb = new ConditionCB[SkyBackground]  (SkyBackground.values,                       _.displayValue())
```
cquiroz: Where is the `_` coming from? Does `ConditonCB` take a function param?

cquiroz:
Looks good :+1:

arturog8m:
Looks good - I think we need to improve the usability aspects of the UI though, but as you say this is to get the conversation moving. I don't like the combo boxes for the conditions - I believe they should be radio buttons. Actually, if the options in the other controls aren't too many, we should  probably show them all with radio buttons so it's easier and save users a lot of clicking. It might look a bit crowded, so I guess we need to mock it up.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcPanel.scala
fnussber: Yes, a render function which is then used to create a ```Renderer```.

fnussber:
Ok, I see your point with the combo boxes. For the conditions I was going for combo boxes to make it similar to the Conditions node and also I was trying to use a layout that uses little space vertically (remember that the stuff to the right is going away), so that most of the room is available for the table.

For the combo boxes on the right (source type and source distribution) I was thinking along similar lines, however having seen the new target editor I am now thinking about using a tabbed panel for the different aspects of a source. This would give me more room for those UI elements. 

Other than that these combo boxes are in sync with how the target type is selected in the target editor and then depending on that selection different detail information can be entered in the other panels.

So as a compromise here's what I am going to do: I am going to merge this today so people can play around with it. And then we can finalize this when I integrate it into the target editor.

PR #337, opened 2015-04-27
REL-2208: GPI Halfwave plate configuration is 30s but should be 5s

cquiroz:
Small PR that updates the overheads calculation for GPI when there is iteration on the the Halfwave plate component
Added two unit test to check a few scenarios and their calculations


fnussber:
:+1:

PR #338, opened 2015-04-27
OCSADV-294: Support search on Gemini's ppmxl catalog

cquiroz:
On this PR the option to do manual search using the PPMXL backend of the Gemini's catalog is enabled, additionally the table is updated to show only the relevant bands in each case
An extra is the switch to search on Gemini's new urls that will be open to the internet soon

### DIFF
--- a/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsVoTableCatalog.scala
+++ b/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsVoTableCatalog.scala
@@ -26,7 +26,7 @@ import jsky.util.gui.StatusLogger
  * The catalog search will provide the inputs to the analysis phase, which actually assigns guide stars to guiders.
  * See OT-26
  */
-case class GemsVoTableCatalog(backend: VoTableBackend = RemoteBackend) {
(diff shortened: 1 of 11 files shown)


bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsVoTableCatalog.scala
```diff
@@ -26,7 +26,7 @@ import jsky.util.gui.StatusLogger
  * The catalog search will provide the inputs to the analysis phase, which actually assigns guide stars to guiders.
  * See OT-26
  */
-case class GemsVoTableCatalog(backend: VoTableBackend = RemoteBackend) {
+case class GemsVoTableCatalog(backend: VoTableBackend = RemoteBackend, catalog: CatalogName = ucac4) {
```
tpolecat: What is the reasoning behind selecting a default catalog (and back end) in the constructor? I generally prefer a companion method with an irritating name like `.forRemoteUcac4` for this kind of thing.

cquiroz: There is only one case where the catalog may be different (Thought there may be more in the future). It seems to me a good use for defaults

PR #340, opened 2015-04-28
REL-2012: Configuration from GPI/Seqexec must set Optimize and No FPM

cquiroz:
As part of REL-2012 it is necessary to add two parameters (aoOptimize, NoFPMPinhole) to let the Seqexec configure the instrument properly
This PR contains code adding the parameters and UI updates to GPI's Engineering Tab


cquiroz:
After some discussion, we reverted the logic from a **No Align FPM/Pinhole/Bias** with default true to a positive **Align FPM/Pinhole/Bias** flag with default false

Below is a screenshot of the new UI
screenshot 2015-04-29 11 56 09

swalker2m:
:+1:

PR #341, opened 2015-04-29
OCSADV-304: vignetting updates

swalker2m:
This PR reworks the vignetting model, plotting, and calculation to fix issues with non-zero base coordinates and offsets.  In particular:

* _At the Model Level_ - changes probe arm calculations to work with new model `Coordinates` and `Offsets` rather than `Point2D`.  All calculations are now done with positions relative to the base rather
than absolute coordinates.  Moves the GMOS detector gaps to the model itself.

* _In the OT TPE_ - introduces a `SciAreaPlotFeature` and a `OiwfsPlotFeature` that work with the model and just convert the model elements to screen coordinates for plotting.  Now handles the correction for image north.  There is overlap with existing `SciAreaFeatureBase` and `OIWFS_FeatureBase` that should be addressed in the future by moving the plotting code in their subclasses to model elements just as we've done for GMOS.

* _Vignetting Calculation_ - changed to approximate the area of the vignetted region and compare it with the overall science area.

Note that the F2 model and plotting are not being used yet and that the integration with AGS itself has been commented out.  I wanted to put this out a bit early for review though since I'll need to merge it into the test release.


bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/flamingos2/Flamingos2.java
```diff
@@ -1411,6 +1411,6 @@ public void restoreScienceDetails(final SPInstObsComp oldData) {
 
     @Override
     public ScienceAreaGeometry getVignettableScienceArea() {
-        return new F2ScienceAreaGeometry(this);
+        return F2ScienceAreaGeometry$.MODULE$;
```
cquiroz: We have added in a few places an `instance` val to the object to avoid the `$MODULE$` call in java

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/flamingos2/F2ScienceAreaGeometry.scala
```diff
@@ -1,57 +1,52 @@
...
-            (fpu.getSlitWidth * inst.getLyotWheel.getPixelScale, LongSlitFOVHeight * plateScale)
+            (fpu.getSlitWidth * f2.getLyotWheel.getPixelScale, LongSlitFOVHeight * plateScale)
         }
       case _ => (0.0, 0.0)
-    }}.getOrElse((0.0, 0.0))
-  }
+    }
+
+  def javaScienceAreaDimensions(f2: Flamingos2): Array[Double] =
+    scienceAreaDimensions(f2) |> { case (w, h) => Array(w, h) }
```
cquiroz: Why did you use an `unsafeTap` here could you have done `Array(scienceAreaDimensions(f2)._1, scienceAreaDimensions(f2)._2)

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/guide/VignettingCalculator.scala
```diff
@@ -0,0 +1,64 @@
...
+          case os  => os.map { case (off, sciShape) =>
+            val probeShape = probeArm.geometry(ctx, guideStar, off) | new Area()
+            val vigShape   = new Area(sciShape) <| (_.intersect(new Area(probeShape)))
+            whole.map { area => approximateArea(vigShape) / area } | 0.0
+          }.max
+        }
+
+      override def min[A](candidates: Iterable[A])(f: A => Coordinates): Option[A] =
+        if (candidates.isEmpty) None
+        else Some(candidates.minBy(c => calc(f(c))))
```
cquiroz: I think scalaz has a method to do `candidates.minimze` that would give you a `None` for empty collections

arturog8m:
Looks good. I think it's clear the separation between the model to perform the calculation to determine the vignetting ratio and the display business. That helps to understand how offsets are taking into account as they are on the same space and we aren't dealing with moving from screen coordinates back and forth. Assuming the calculations are correct then this should work!. I guess some manual testing will be useful, so I'd say ship it and have it tested! :+1:

cquiroz:
You probably deserve a medal for this :+1:

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/flamingos2/F2ScienceAreaGeometry.scala
tpolecat: That's actually not `unsafeTap`, it's the Thrush combinator. `a |> f` is the same as `f(a)` but is sometimes more convenient. In this case you could replace `|>` with `match` for the same effect with less machinery.

Also @cquiroz your suggestion computes the dimensions twice :wink:

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/guide/VignettingCalculator.scala
tpolecat: Why are we using `Iterable` here? Let's try to use `Foldable` in these cases, and yes you get a `.minimumBy` operation that yields an `Option`.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/gmos/GmosOiwfsProbeArm.scala
```diff
@@ -35,82 +37,64 @@ class GmosOiwfsProbeArm[D  <: Enum[D]  with GmosCommonType.Disperser,
...
+      val t   = (if (flip) T.flipQ else T).rotate(posAngle)
+      t + gsOffset - adj
+    }.toPoint
 
-      // The final adjusted offset as modified by the offset adjustment required by the IFU / WFS.
-      val offsetAdj = {
-        val ifuOffset = transformPoint(new Point2D.Double(wfsOffset, 0.0), posAngleRot)
-        new Point2D.Double(offset.getX - ifuOffset.getX, offset.getY - ifuOffset.getY)
-      }
+    val r     = math.sqrt(p.getX * p.getX + p.getY * p.getY)
```
tpolecat: Just FYI, `math.hypot(p.getX, p.getY)` is potentially faster and more accurate.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/gmos/GmosScienceAreaGeometry.scala
```diff
@@ -1,86 +1,96 @@
...
+  val ImagingGapWidth         =   3.00
+
+  val MosFovSize              = ImagingFovSize - 2 * 8.05
+  val MOSFovInnerSize         = 139.38 * 2
+
+  val LongSlitFovHeight       = 108.00
+  val LongSlitFovBridgeHeight =   3.20
+
+  override def unadjustedGeometry(ctx: ObsContext): Option[Shape] = {
+    def gmosGeo[P <: FPUnit](mode: FPUnitMode, fpu: P, isSouth: Boolean): Option[Shape] =
```
tpolecat: The type parameter isn't doing anything useful here; you can just say `fpu: FPUnit`.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/inst/FeatureGeometry.scala
```diff
@@ -1,143 +1,107 @@
...
-   * Given a shape, transform it as needed for display on the screen.
-   * @param shape           the shape to transform
-   * @param pixelsPerArcsec the pixel density per arcsec on the current display
-   * @param flipRA          a scaling in the y-axis: should be either 1.0 or -1.0, and not sure if this is ever -1.0
-   * @param screenPos       the final position on the screen where the adjusted guide probe arm should be placed
-   * @return                the transformed shape
-   */
-  def transformProbeArmForScreen(shape: Shape, pixelsPerArcsec: Double, flipRA: Double, screenPos: Point2D): Shape =
-    transformProbeArmForScreen(List(shape), pixelsPerArcsec, flipRA, screenPos).head
+    def go(area: Double, cur: Option[AreaCalc]): Double =
```
tpolecat: Can you add `@tailrec` here just for futureproofing? It's easy for someone to come in and change a pattern-match to a fold and break tail recursion inadvertently.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/inst/OiwfsPlotFeature.scala
```diff
@@ -0,0 +1,195 @@
...
+  import FeatureGeometry._
+
+  def getFigures(tpeCtx: TpeContext): List[Figure] = {
+    def go(obsCtx: ObsContext, patField: PatrolField): List[Figure] = {
+      // Extract info from the TPE context, converted to new model.
+      val posAngle = Angle.fromDegrees(tpeCtx.instrument.posAngleOrZero)
+      val selPos   = tpeCtx.offsets.selectedPos
+      val offset   = selPos.map { opb =>
+        Offset(opb.getXaxis.arcsecs[OffsetP], opb.getYaxis.arcsecs[OffsetQ])
+      } | Offset.zero
```
tpolecat: You can just say `.foldMap(...)` and remove the `| Offset.zero` if you define `Monoid[Offset]`, which is probably worthwhile at some point.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/inst/OiwfsPlotFeature.scala
```diff
@@ -0,0 +1,195 @@
...
+          gpt <- tpeCtx.targets.envOrDefault.getPrimaryGuideProbeTargets(probe).asScalaOpt
+          gs  <- gpt.getPrimary.asScalaOpt
+        } yield gs.toNewModel.coordinates // TODO: when?
+
+        if (showProbeArm)
+          guideStar.flatMap { gs =>
+            probeArm.geometry(obsCtx, gs, offset).map {
+              new Figure(_, ProbeArmColor, Blocked, ProbeArmStroke)
+            }
+          }
```
tpolecat: This might be more readable as a `for` comprehension.

tpolecat:
I can't believe how fast you did this. Looks great. :+1:

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/guide/VignettingCalculator.scala
swalker2m: I ended up changing it altogether.  I wanted it to stop when it finds a zero vignetting case so now it uses explicit recursion.  By the way, `minimumBy` doesn't seem to be available in our version of scalaz.

swalker2m:
Merging for initial test release.  Needs more automated testing and some cleanup in `SingleProbeStrategy`.

PR #344, opened 2015-04-30
OCSADV-275: Basic sequence executor

jluhrs:
First ultra very mucho simple implementation of a sequence execution.
Includes Shane's code to get the sequences from the ODB.


bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Executor.scala
```diff
@@ -0,0 +1,46 @@
...
+import edu.gemini.spModel.obscomp.InstConstants.INSTRUMENT_NAME_PROP
+import edu.gemini.spModel.seqcomp.SeqConfigNames.INSTRUMENT_KEY
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.{Future, Promise}
+
+/**
+ * Created by jluhrs on 4/28/15.
+ */
+sealed class Executor(sequenceConfig: ConfigSequence) {
```
tpolecat: You probably mean `final class` here.

bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Executor.scala
```diff
@@ -0,0 +1,46 @@
...
+
+  def run = runSequence(sequenceConfig.getAllSteps.toList)
```
tpolecat: Please put return types on **all** methods.

cquiroz:
Does it include the code from https://github.com/gemini-hlsw/ocs/pull/327?

bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Executor.scala
```diff
@@ -0,0 +1,46 @@
...
+
+  private def runSequence(config: List[Config]): Future[Int] = {
+    def chainStep(config: List[Config], prevFuture: Future[Int]): Future[Int] = config match {
+      case List() => prevFuture
+      case x :: ys => chainStep(ys, prevFuture flatMap (i => runStep(x, i)))
+    }
+    config match {
+      case List() => Promise.failed(new Exception("Empty sequence configuration")).future
```
cquiroz: I think this could be `Future.failed(...exception...)`

bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Executor.scala
```diff
@@ -0,0 +1,46 @@
...
+import scala.concurrent.{Future, Promise}
+
+/**
+ * Created by jluhrs on 4/28/15.
+ */
+sealed class Executor(sequenceConfig: ConfigSequence) {
+
+  def run = runSequence(sequenceConfig.getAllSteps.toList)
+
+  private def runSequence(config: List[Config]): Future[Int] = {
```
cquiroz: What is the Int in the `Future[Int]`? Number of steps executed?

bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Executor.scala
```diff
@@ -0,0 +1,46 @@
...
+    def chainStep(config: List[Config], prevFuture: Future[Int]): Future[Int] = config match {
+      case List() => prevFuture
```
cquiroz: You could use `case Nil` but I don't if there is one recommended way or another

bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Executor.scala
```diff
@@ -0,0 +1,46 @@
...
+      case x :: ys => chainStep(ys, prevFuture flatMap (i => runStep(x, i)))
+    }
+    config match {
+      case List() => Promise.failed(new Exception("Empty sequence configuration")).future
+      case x :: ys => chainStep(ys, runStep(x, 0))
+    }
+  }
```
tpolecat: Just a few stylistic things:
- Use `Nil` when matching on an empty list.
- Use `Future.failed(ex)` rather than `Promise` if you want to construct a failed future.
-  This method can be written as a fold, something like this:

```scala
private def runSequence(config: List[Config]): Future[Int] =
  config match {
    case Nil => Future.failed(new Exception("Empty sequence configuration")) // but is this really an error?
    case _   => config.foldLeft(Future(0))((f, s) => f.flatMap(runStep(s, _)))
  }
```

bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Instrument.scala
```diff
@@ -0,0 +1,13 @@
...
+
+import edu.gemini.spModel.config2.Config
+
+import scala.concurrent.Future
+
+/**
+ * Created by jluhrs on 4/22/15.
+ */
+trait Instrument extends System {
+  def observe(config: Config): Future[Unit]
```
cquiroz: I wonder about the `Future[Unit]`, could it make sense for example to return the data label, or another object with a summary of the observation execution, containing like name, execution time, etc?

bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GMOS_S.scala
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Created by jluhrs on 4/27/15.
+ */
+object GMOS_S extends Instrument {
+
+  override val name: String = INSTRUMENT_NAME_PROP
+
+  val Log = Logger.getLogger(getClass.getName)
+
+  override def configure(config: Config): Future[Unit] = Future {
```
cquiroz: Perhaps using `future` or `Future` consistently?

tpolecat:
Looks great for your first code in Scala. I suggest sitting down with @swalker2m and going through the code for stylistic things and opportunities to simplify.

cquiroz:
It seems very good and congratulations on your first usage of `Future`. I think we should consider if the `Instrument` and `System` traits  should return `Future[Unit]` or something more meaningful, I'm pretty sure we want to know a lot about the execution of a sequence once it ends
And we should have some tests :smile:

tpolecat:
I hate to bring it up, but we really should be using `scalaz.concurrent.Task` rather than `Future`. But it's ok for now.

jluhrs:
Carlos: Yes, it include the code from #327, well, most of it. I based the application seqexec-server on Shane's seqexec-client (BTW, the server-client terms are getting a bit confusing here. Shane's app is a client from the POV of the ODB, while my app is the server from the POV of the Seqexec architecture)

bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Executor.scala
jluhrs: Yes, although it is a placeholder for the step execution result. The details of such artifact will reveal themselves in future sprints.

bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Executor.scala
jluhrs: But, but... master Odersky uses List() in is course...

bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Instrument.scala
jluhrs: Yes, I think it will make sense, but the time to dwell on those details has yet to come.

bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/GMOS_S.scala
jluhrs: Done

bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Executor.scala
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Created by jluhrs on 4/28/15.
+ */
+final class Executor(sequenceConfig: ConfigSequence) {
+
+  def run: Future[Int] = runSequence(sequenceConfig.getAllSteps.toList)
+
+  private def runSequence(config: List[Config]): Future[Int] = {
+    config match {
+      case Nil => Future.failed(new Exception("Empty sequence configuration"))
```
swalker2m: I think an empty sequence might not be an error case?  If it is empty, it just "runs" immediately and can't fail.

At any rate, since we're just getting started on something new, I think it makes sense to switch to `Task` as Rob mentioned and use an application-defined failure type instead of exceptions.  Something like the `TryVcs` and `VcsAction` in the `vcs2` package object:

````scala
type TrySeq[A] = SeqFailure \/ A
type SeqAction[+A] = EitherT[Task, TrySeq, A]
````

A `Future` launches right away when it is created but this would allow us to decide when and how to run the effects later.

File: bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Executor.scala [hunk 10]
```diff
@@ -0,0 +1,42 @@
...
+      case GMOS_S.name => Some(GMOS_S)
+      case _ => None
+    }
+
+    instrument map (a => {
+      val systems: List[System] = List(TCS, a)
+
+      Future.sequence(systems map (x => x.configure(stepConfig))) flatMap
+        (x => a.observe(stepConfig)) map (x => stepIdx + 1)
+    }) getOrElse Future.failed(new Exception("Unrecognized instrument"))
```
swalker2m: I think this would be easier to see as a `for` expression, if I'm reading it correctly.

````scala
for {
  _ <- Future.sequence(systems.map(_.configure(stepConfig)))
  _ <- a.observe(stepConfig)
} yield stepIdx + 1
````

File: bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/ItemEntryUtil.scala [hunk 11]
```diff
@@ -0,0 +1,32 @@
...
+
+    def seqValue(o: Object): String = o match {
+      case s: SequenceableSpType => s.sequenceValue()
+      case d: DisplayableSpType => d.displayValue()
+      case l: LoggableSpType => l.logValue()
+      case _ => o.toString
+    }
+
+    val pad = width(ks)
+    (ks.sortWith((u, v) => u.getKey.compareTo(v.getKey) < 0) map (p => {
```
swalker2m: A minor style thing we've adopted for multi-line blocks, I think because it is most commonly done this way, is

````scala
foo.map { f =>
  ...
}
````

instead of

````scala
foo map (f => {
  ...
})
````

File: bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/ItemEntryUtil.scala [hunk 12]
```diff
@@ -0,0 +1,32 @@
...
+      case Array() => 0
+      case _ => ks.maxBy(_.getKey.toString.length).toString.length
+    }
+
+    def seqValue(o: Object): String = o match {
+      case s: SequenceableSpType => s.sequenceValue()
+      case d: DisplayableSpType => d.displayValue()
+      case l: LoggableSpType => l.logValue()
+      case _ => o.toString
+    }
```
swalker2m: I remember having written this but nevertheless I'll comment.  :-) It's easier to read in general if the arrows `=>` line up:

````scala
o match {
  case s: SequenceableSpType => s.sequenceValue()
  case d: DisplayableSpType  => d.displayValue()
  case l: LoggableSpType     => l.logValue()
  case _                     => o.toString
}
````

I know Idea likes to crap on your careful alignment of code so it's a constant struggle.

**Comment by swalker2m** [comment]
Since this stuff isn't for June, we should create a release branch for stuff that is before this is checked into develop.

File: bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/ItemEntryUtil.scala (hunk 12 above)
Comment (jluhrs) [inline bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/ItemEntryUtil.scala]: Yes, that was the result of applying Idea format tool.

File: bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Executor.scala (hunk 9 above)
Comment (jluhrs) [inline bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Executor.scala]: But, shouldn't the user be informed that he/she is trying to run an empty sequence?

Comment (tpolecat) [inline bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Executor.scala]: I agree with @swalker2m.  I think one of the following is true:
- There is no logical problem with executing an empty sequence; it is simply a no-op, or
- An empty sequence makes no sense, in which case `type Sequence = NonEmptyList[Config]` and your program will not compile unless you can prove your sequence is non-empty.

**Comment by jluhrs** [comment]
Replaced the use of Scala Futures for Scalaz Tasks, and created classes for action responses (they are placeholders for now).
I tried to use a specific class for application errors using VcsFailure as an example, but I couldn't understand how to make it work with Task, specifically how to make gatherUnordered to stop on the first SequenceFailure when configuring all systems in parallel.

File: bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Executor.scala [hunk 13]
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * Created by jluhrs on 4/28/15.
+ */
+final class Executor(sequenceConfig: ConfigSequence) {
+
+  def run: \/[Throwable, List[StepResult]] = runSequence(sequenceConfig.getAllSteps.toList).attemptRun
+
+  private def runSequence(config: List[Config]): Task[List[StepResult]] = {
+    (config map (runStep(_))).sequenceU
```
swalker2m: Just a minor style thing.  I believe this can be

```scala
private def runSequence(config: List[Config]): Task[List[StepResult]] =
  config.map(runStep).sequenceU
````

since `runStep` is a function that takes a single `Config` argument.  I understand the style `a map foo` (as opposed to `a.map(foo)`) is not encouraged except maybe for some DSLs.

File: bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Executor.scala [hunk 14]
```diff
@@ -0,0 +1,49 @@
...
+
+  private def runStep(stepConfig: Config): Task[StepResult] = {
+    val instName = stepConfig.getItemValue(new ItemKey(INSTRUMENT_KEY, INSTRUMENT_NAME_PROP))
+    val instrument = instName match {
+      case GMOS_S.name => Some(GMOS_S)
+      case _           => None
+    }
+
+    instrument map (a => {
+      val systems: List[System] = List(TCS, a)
```
swalker2m: Here let's go with

````scala
instrument.map { a =>
...
}
```
It's a bit less noisy than the alternative:
````scala
instrument.map (a => {
...
})
```

bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Executor.scala
```diff
@@ -0,0 +1,49 @@
...
+      for {
+        r <- Task.gatherUnordered( systems map (x => x.configure(stepConfig)) , true)
```
swalker2m: `Task.gatherUnordered(systems.map(_.configure(stepConfig)), true)`

bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Executor.scala
tpolecat:
```scala
config.traverseU(runStep)
```

tpolecat:
> … specifically how to make gatherUnordered to stop on the first SequenceFailure when configuring all systems in parallel

If you were to do this how would you stop the remaining tasks? Is this even possible in EPICS?

bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Executor.scala
```diff
@@ -0,0 +1,49 @@
...
+        s <- a.observe(stepConfig)
+      } yield StepResult(r, s)
+    }) getOrElse Task.fail(UnrecognizedInstrument(instName.toString))
```
swalker2m: I think this should work fine but the issue is that every failure case has be an exception. That may be okay for this application I don't know.  I found it better to be able to enumerate the possible application failure cases (say under the trait `SeqFailure`) and add one failure case that wraps connection exceptions, etc. to turn them into an application failure case (say `case class SeqException(t: Throwable) extends SeqFailure`).

You get the fail-fast behavior with `EitherT` as in

````scala
type SeqAction[+A] = EitherT[Task, SeqFailure, A]
````

which is a combination `Task` and disjunction `\/`.  You `run` the action to get the `Task[SeqFailure \/ A]` which can then be executed with `attemptRun` or whatever to produce a result of type `\/[Throwable, \/[SeqFailure, A]]` which can be merged into a `SeqFailure \/ A` with something like:

````scala
result.fold(t => SeqException(t), identity)
````

to end up with just `SeqFailure \/ A`.

Maybe it is overkill for your application though.  Do you anticipate application level failure cases?

tpolecat: Agree, let's get rid of exceptions. With `EitherT` you can use `.sequence` to go from `List[EitherT[SeqFailure,A]]` to `Task[EitherT[SeqFailure, List[A]]]` which I believe is what you want.

jluhrs:
In theory, our EPICS systems can abort a command. This is done writing STOP to the main APPLY record. But Seqexec does not use it, and I am not certain that STOP actually works in all the systems.

The current Seqexec reacts immediately to an error while configuring all systems, and notifies the user. The configuration of the other systems continues, but when they complete the sequence execution will not continue. The problem with this model is that the user may react trying to repeat the configuration while some of the systems are still busy.

tpolecat:
Ok, it sounds like what you have is probably correct, at least operationally. All the parallel configuration tasks need to run to completion, and then once everything is idle you can inform the user. An alternative would be to report errors as they happen, but not let the user do anything until all the systems are idle. Don't know if this is any better.

As far as types are concerned, I think you can improve on what you have by using `EitherT` as Shane proposed. I wrote up some examples that might be helpful. https://gist.github.com/tpolecat/15a181568df502a03515

bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Executor.scala
jluhrs: I tried to use EitherT[Task, SeqFailure, A], but I don't see how the fail-fast behavior could work in gatherUnordered. The code completes the Task with a failure on the first task that completes with a -\/, but if my code ends a configuration with a SeqFailure, the task code will see a \/-(f: -\/[SeqFailure]) and go on the success path.

tpolecat: You need a custom `Nondeterminism[EitherT[Task, SeqFailure, ?]]` .. see the gist I posted below.

PR #345, opened 2015-04-30
REL-2232 Update NIFS Templates for 2015B

tpolecat:
This updates the selection of acquisitions based on example target brightness in K. It also adds two notes to all NIFS observations.


cquiroz:
Seems very good, I wish we could add tests to all the templates, this may be the only one that has them?

tpolecat:
Yes, it's the only one with tests, and the tests only cover these specific changes. I would like to rewrite all of these things with a better DSL now that I have a better idea of what I'm doing, but it's hard to justify when it all seems to work as-is.

PR #346, opened 2015-05-01
OCSADV-316 fixed broken delete button

tpolecat:
I messed up a logic simplification that caused the modified target environment to not be reassigned when a target was deleted. This PR fixes that. If you view it with `?w=1` on the URL you will see that it's a minor change (most changed lines are due to indentation).


cquiroz:
That's because `TargetEnvironment` always returns copies, right?
In any case :+1: and nice tip about removing the whitespaces

tpolecat:
You can also `git diff -w` to ignore whitespace locally.

PR #347, opened 2015-05-02
OCSADV-308: Support for spectroscopy for GMOS, F2 and NIRI

fnussber:
Probably (hopefully?) one of the last major refactorings for ITC. This PR aims at providing a "unified" spectroscopy calculation result that can be used in the OT as well as for creating chart images and spectra text files by the web application when requests for them are coming in. Once this was finished, hooking up the ITC service for spectroscopy for GMOS/N/S, F2 and NIRI was easy and I decided to plow forward and do that, too.

The original ITC was creating the charts and text files for each web request as part of the recipe calculations and stored those files in the ```/tmp``` folder of the server and then served those files to the web browser when requests for them were coming in. This is (a) slow because of all the disk IO and (b) clogs up the machine's disk over time (I had at some point 75'000 files in my ```/tmp``` folder). I changed that so that now a reduced set of information (see ```ItcSpectroscopyResult``` in ```ItcService```) is created as part of the recipe calculations which has all the information that is needed to create the charts and contains the spectra text files as strings.

These objects are cached (LRU) by the web server and kept in memory up to a limit that is currently (for no particular reason) set to 600. Incoming image and file requests retrieve that information and create the charts and data files from it (see ```ImageServlet```) which reduces file I/O considerably. (Overall and including previous changes the time to run tests for 5500 requests has come down from 28 minutes to 6 minutes.) Note that this cache is only needed for web requests in order to serve images and text file links on the result HTML page, not for TRCP service requests.

The same ```ItcSpectroscopyResult``` objects are now also used as result objects in the TRCP service calls of the ITC service. The charts are created in the OT from the raw data in those objects. This allows for support of plotting details limits (min/max wavelength to be displayed) and also allows to get min/max values etc from the data which can be displayed in the result table.

On top of all this infrastructure work additional UI elements were needed to display the charts in the OT.

Apologies for the size of this PR but I wanted these changes to be part of the first test release so we have plenty of time to test all of this.

image


swalker2m:
I think the charts look good.  What happens when the editor window is smaller?  It typically isn't stretched out so big, I'm guessing.  Maybe it is just me, but I think it might look even better if the white and the grey were swapped so that the charts had a white background.

swalker2m:
We should probably merge this soon to get it in the next test release.  I don't understand the cache thing but I'm assuming that when it blows up on a lookup (because the data is so old that it has been removed) something appropriate happens?

fnussber:
The caching has changed a bit in later check ins. To make it clearer what happens: The web client will hit the server with a request with a lot of form data that is needed for the calculations. The result is a HTML page with several link to images and text files. While the image links will be loaded immediately by the client the text file links may be followed some time later by the user. In order to create those chart images and text files the results are cached for a while. If they can not be found anymore in the cache a 408 is returned to the web client (session timeout) which I think is appropriate behavior for this kind of web app.

fnussber:
Regarding the charts: Currently - lazy bastard I am - I am just using the same coloring etc as in the web client. The images are a fixed size and put on a scroll pane. I will probably change that and have the images resize according to available space and I might use background colors as in QV (striped). I'll ask Andy what he would like to have there.

PR #348, opened 2015-05-05
OCSADV-319: Bug fix: Issues with nano- vs micro-meters in wavelengths.

fnussber:
This is a bug fix that deals with problems that stem from different units being used for wavelengths throughout the code (nanometers vs micrometers). It seemed appropriate to introduce a dedicated type that allows to deal with different units in a graceful way.

### DIFF
--- a/bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/Wavelength.scala
+++ b/bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/Wavelength.scala
@@ -1,5 +1,7 @@
 package edu.gemini.spModel.core
 
+import scalaz.{Order, Monoid}
+
 /** Representation of wavelengths. */
 sealed trait Wavelength extends Serializable {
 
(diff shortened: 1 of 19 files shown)


bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/Wavelength.scala
```diff
@@ -0,0 +1,66 @@
...
+      case a: Wavelength => a.toNanometers == this.toNanometers
+      case _             => false
+    }
+
+  /** @group Overrides */
+  final override def hashCode =
+    toNanometers.hashCode
+}
+
+object Wavelength {
```
cquiroz: I guess you could consider adding Monoid and Order definitions in here even though there is no use right away

cquiroz:
Looks very good :+1: 
Nice to see a good use of encapsulation to avoid the easy to miss multiplication to change units

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/Wavelength.scala
fnussber: Ok, added. In particular the order definitions might be useful in many places.

PR #352, opened 2015-05-05
OCSINF-437 OCSINF-438 revert to original SwingWorker logic

tpolecat:
This resolves (I think) some threading issues ([OCSINF-437](http://swgserv01.cl.gemini.edu:8080/browse/OCSINF-437) and [OCSINF-438](http://swgserv01.cl.gemini.edu:8080/browse/OCSINF-438)) and  that can cause races and deadlocks when doing a SIMBAD lookup. Basically I restored the old `SwingWorker` logic, in a slightly saner Scala wrapper.

Recommend viewing the diff with `?w=1` on the URL; most of the diff lines are due to a loss of a level of indentation.


swalker2m:
Looks good. :+1:

PR #353, opened 2015-05-05
OCSADV-309: magnitude sorting fixes

swalker2m:
This PR fixes an issue in which the default `Magnitude` ordering (which considers magnitude band) does not produce expected results for AGS. 

````scala
  // by system name, band name, value and error (in that order)
  implicit val MagnitudeOrdering: scala.math.Ordering[Magnitude] =
    scala.math.Ordering.by(m => (m.system.name, m.band.name, m.value, m.error))
````

AGS works with "any" R-like band (r, R, and UC) and uses the magnitude value of the first one that it finds.

The bulk of the updates are to the test cases.  The only substantive changes are those in `SingleProbeStrategy` itself. Along the way I also deleted an unused GMOS type called `PosAngleMode` which has been superseded by `PosAngleConstraint`.

### DIFF
--- a/bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/Magnitude.scala
+++ b/bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/Magnitude.scala
@@ -1,5 +1,7 @@
 package edu.gemini.spModel.core
 
+import scalaz._
+import Scalaz._
 
 case class Magnitude(value: Double, band: MagnitudeBand, error: Option[Double], system: MagnitudeSystem) {
 
@@ -40,6 +42,12 @@ object Magnitude {
   }
 
   /** @group Typeclass Instances */
-  implicit val equals = scalaz.Equal.equalA[Magnitude]
+  implicit val equals = Equal.equalA[Magnitude]
 
+  /** group Typeclass Instances */
+  implicit val MagnitudeShow: Show[Magnitude] =
+    scalaz.Show.shows { mag =>
+      val errStr = mag.error.map { e => f" e$e%.2f " } | " "
+      f"${mag.band.name}${mag.value}%.2f$errStr(${mag.system.name})"
+    }
 }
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/Magnitude.scala [hunk 1]
```diff
@@ -1,5 +1,7 @@
 package edu.gemini.spModel.core
 
+import scalaz._
+import Scalaz._
 
 case class Magnitude(value: Double, band: MagnitudeBand, error: Option[Double], system: MagnitudeSystem) {
```
cquiroz: Not in your code but should we remove the old `Ordering`, it is misleading as you can't really sort Magnitudes of different bands

project/OcsBundle.scala
```diff
@@ -355,7 +355,7 @@ trait OcsBundle {
   lazy val bundle_edu_gemini_ags = 
     project.in(file("bundle/edu.gemini.ags")).dependsOn(
       bundle_edu_gemini_shared_skyobject,
-      bundle_edu_gemini_pot,
+      bundle_edu_gemini_pot          % "test->test;compile->compile",
```
cquiroz: 30 mins :smile:

cquiroz:
would the Magnitude ordering affect the non-vignetting branch? I guess we are sorting by mag there

project/OcsBundle.scala
tpolecat: I assume you mean it took 30 minutes to reload? Did you type `reload` or restart sbt?

cquiroz: I wonder what's the difference, I type `reload` but I see the same with a restart, do you think some old caches are affecting this? I should try the `reload return` you mentioned a while ago

tpolecat: When it gets slow I usually do:

```
> reload plugins
> clean
> ^D
```

cquiroz:
Thanks for taking the time for this and for the nice tests, the code seems very good

swalker2m:
> would the Magnitude ordering affect the non-vignetting branch? I guess we are sorting by mag there

It appears to use the reference magnitude in `SingleProbeParams.brightest` so presumably it is okay.

PR #354, opened 2015-05-06
REL-2298: disappearing P2 check bug fix.

swalker2m:
Fixes a mysterious disappearing P2 check problem.  The issue is that the `Problem`s were being stored in a `TreeSet` and yet had a `Comparable` implementation wasn't consistent with `equals`.  So when you copy and paste an observation with problems and it recalculates the combined problem set, the problems in the new observation counted as already existing in the `TreeSet`.  The solution is to remove the bogus `Comparable` implementation and store the problems in a `HashSet` instead.

To keep the problem list from being displayed in random order, they are now sorted by severity and description before being listed.

This bug falls firmly into the category, "how did this ever work?".


tpolecat:
This is terrible. :+1:

cquiroz:
Good :+1: 
I believe Java `TreeSet` and Scala `TreeSet` have a slight different behaviour with respect to `equals`, at least I found a problem translating from Java to Scala in this respect

PR #356, opened 2015-05-06
REL-1842 updates to Graces blueprint

tpolecat:
This PR updates the `SpGracesBlueprint` to include the read and fiber modes from the phase-1 blueprint, adds machinery for P1/P2 translation, and then updates the group initializer logic.


bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/template/factory/impl/graces/Graces.scala
```diff
@@ -1,14 +1,69 @@
...
-  def initialize(group:ISPGroup, db:TemplateDb): Maybe[Unit] = Right(())
+  def instCompType: SPComponentType =
+    VisitorInstrument.SP_TYPE
+
+  def seqConfigCompType: SPComponentType =
+    sys.error("No sequence component type for visitor instrument GRACES")
+
+  // This seems to be necessary, sorry.
+  var db:Option[TemplateDb] = None
+  override def initialize(db:TemplateDb):Maybe[ISPGroup] =
```
cquiroz: What is `Maybe`?

cquiroz:
Looks good :+1: 
I wonder how much work would it be (**And whether it is worth it**) to have end-to-end tests starting with a phase 1 proposal xml submitted and integrated as template

bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/template/factory/impl/graces/Graces.scala
tpolecat: It's defined as `Either[String,A]` in the template code. Not a great idea in hindsight but I want to be consistent with the other stuff in that package.

tpolecat:
Yes, I actually started work on this the other day and did figure out how to move stuff all the way through in-memory (without all the auxfile crap) ... but it looks hard enough in some other ways that I set it aside so I could actually close some bugs. There is some information lost in the process that makes it hard to line up the proposal, the blueprint program, and the skeleton program. Might have time to revisit, don't know.

swalker2m:
:+1:

andrewwstephens:
This looks good for all cases except when the target magnitude in undefined, when we should include all acquisition templates.

tpolecat:
@andrewwstephens hmm, did you build the SPDB locally and try it? The testcase for undefined R-mag works and shows all acqs scheduled to be added to the program. Are they not showing up?

andrewwstephens:
@tpolecat  Nope.  I *think* I've done everything correctly...  I put additional comments and a sample phase-1.xml in REL-1842.

tpolecat:
Ok I found the problem. There is code far, far away that partitions targets by brightness and this wasn't happening for Graces.

PR #359, opened 2015-05-08
OCSADV-323: Add source profile and distribution to (I)Target

fnussber:
This is the first of a series of PRs that add additional source information to the targets, makes that information persistent and finally editable in the target editor.

Step one is the addition of the spatial profile and spectral distribution to ```ITarget``` including XML serialization through ```SPTargetPio```.

I guess in the long run these new source attributes will also have to become parts of the new ```Target``` in spModel core. However I did not dare to pollute that code with my certainly terrible attempts at getting things done and made ```SpectralDistribution``` and ```SpatialProfile``` parts of ```edu.gemini.pot``` for now.

Next step will be additional UI capabilities in the target editor in order to be able to edit those values.


bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/SpectralDistribution.scala
```diff
@@ -0,0 +1,192 @@
...
+  case object OCet              extends LibraryNonStar("O Cet (M7IIIa Star, silicate emission)",  "ocet")
+  case object OrionBar          extends LibraryNonStar("Orion Bar (Dusty HII region)",            "orionbar")
+  case object Rscl              extends LibraryNonStar("rscl (N-type Dusty Carbon Star, SiC em.)","rscl")
+  case object Txpsc             extends LibraryNonStar("txpsc (N-type Visible Carbon Star)",      "txpsc")
+  case object Wr104             extends LibraryNonStar("WR 104 (Wolf-Rayet Star + dust)",         "wr104")
+  case object Wr34              extends LibraryNonStar("WR 34 (Wolf-Rayet Star)",                 "wr34")
+
+  /** All available non-star spectras. */
+  val values = List(
+    EllipticalGalaxy, SpiralGalaxy, QS0, QS02, OrionNebula, PlanetaryNebula, PlanetaryNebula2, PlanetaryNebula3,
```
fnussber: Is there a better way to do this? Manually listing all the values that belong to a set seems like something the compiler should do for me.

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/shared/EmissionLineSpectrum.java
```diff
@@ -1,71 +1,53 @@
...
-        double start = 300;//0.2*_wavelength;//.8
-        double end = 30000;//1.8*_wavelength;//1.2
-        //System.out.println("Start: "+ start +"End: " +end);
+    public EmissionLineSpectrum(final Wavelength wavelength, final double width, final EmissionLine.Flux flux,
+                                final EmissionLine.Continuum continuum, final double z, final double interval) {
 
         //shift start and end depending on redshift
-        start /= (1 + z);
-        end /= (1 + z);
+        final double start = 300 / (1 + z);
```
cquiroz: Maybe you could preserve (adapt) the comments from the deleted lines as the formula is not that obvious

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/system/ITarget.java
```diff
@@ -28,7 +26,7 @@
  */
 public abstract class ITarget implements Cloneable, Serializable {
```
cquiroz: It has been my impression that we want to get away from the old Target model rather than extend it. @swalker2m do you have an opinion?

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/SpectralDistribution.scala
cquiroz: Macros?

cquiroz:
I guess my only question is about expanding `ITarget`, otherwise it looks good

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/system/ITarget.java
fnussber: The problem is that the target editor is based on ```SPTarget``` and ```ITarget```, also when dealing with the sequence model and getting parameters from the current environment this is all based on the old model. Since the target editor is my primary target here I think there's no way around doing this on the old model first.

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/shared/EmissionLineSpectrum.java
fnussber: That's a difficult one. I am afraid that these comments (like many others) are totally obsolete/wrong/useless. However I re-added them to preserve them. Maybe someone else can make sense out of them ;)

cquiroz: Totally agree, better no comment than a bad one

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/system/ITarget.java
swalker2m: I think @fnussber's approach makes sense.  Currently we have a mutable `SPTarget` which contains a mutable `ITarget`.  The end goal is to replace `ITarget` with the core immutable `Target` so this stuff will have to migrate to `Target` at some point.  In the meantime though we need it to exist in `ITarget` anyway.  @tpolecat is target king though, so I'd like to know if he has anything to add.

I know we do a number of model conversions but I think just in AGS calculations?  We never, say, take the science target and convert it from old to new and back to my knowledge.  If we did there would obviously be information loss.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/SpectralDistribution.scala
swalker2m: What is this used for?  Perhaps you could get away with just a `List[LibraryNonStar]` and forget the `case object`s altogether?

fnussber: That's what I had to start with. I switched to ```case object```s because it makes it straight forward to reference a particular object for example in test fixtures. However, I have to admit, that's the only place (right now) where the ```case object```s are of use.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/system/ITarget.java
fnussber: I think right now we only have ```Target``` => ```SPTarget``` conversions in the AGS part. So for now we get away with that shortcut.

fnussber: @tpolecat Any comments on this? I would like to merge this asap. A next update for the target editor based on this PR is coming soon.

tpolecat: Sorry for the delay. Yes, the `ITarget` modificaton looks fine.

PR #360, opened 2015-05-08
OCSADV-272: GEMS AGS Strategy

cquiroz:
This PR contains multiple fixes to get Gems AGS to work using several bands as R-like, (including r', R and UC). The implication of the band selection change touched quite a few pieces of Gems AGS and created some regressions with a simpler implementation. Multiple tests were added to ensure the algorithm didn't fundamentally change.

This PR includes:
* Fixes Gems AGS to work with R-Like magnitudes including multiple tests
* Fixes bogus magnitude sorting
* Fix for regressions and their respective test using UCAC3 canned data
* Fixes Mascot to properly extract the appropriate R-Like magnitude
* Restore proper order of cwfs targets ensuring that cwfs3 gets the brightest of the stars based on R-like magnitudes
* Fix for the analyze method for Gems and proper linking to the Guiding Feedback shown in the OT
* Conversion of most of the code to Scala
* Replaces the control flow of interruptible `analyzeGoodEnough` using exceptions to a more proper flow using callbacks
* Removes debugging when getting the catalog data

The screenshot below shows the result of a search with the feedback magnitudes properly displayed

screenshot 2015-05-07 14 14 48

Sorry for the size of the PR though quite a bit is unit tests


bundle/edu.gemini.ags/src/main/java/edu/gemini/ags/gems/GemsGuideStarSearchOptions.java
```diff
@@ -64,9 +65,9 @@ public String toString() {
 
 
     public enum NirBandChoice {
-        J(GemsUtils4Java.toNewBand(edu.gemini.shared.skyobject.Magnitude.Band.J)),
-        H(GemsUtils4Java.toNewBand(edu.gemini.shared.skyobject.Magnitude.Band.H)),
-        K(GemsUtils4Java.toNewBand(edu.gemini.shared.skyobject.Magnitude.Band.K)),
+        J(ModelConverters.toNewBand(edu.gemini.shared.skyobject.Magnitude.Band.J)),
+        H(ModelConverters.toNewBand(edu.gemini.shared.skyobject.Magnitude.Band.H)),
+        K(ModelConverters.toNewBand(edu.gemini.shared.skyobject.Magnitude.Band.K)),
```
swalker2m: I guess we could just directly reference the new bands?  Or maybe you're just trying to avoid the awkward `$MODULE` business?

krumpi: I couldn't find a way to reference case objects, even with `$MODULE`

cquiroz: I couldn't find a way to reference case objects, even with $MODULE

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsGuideStars.scala
```diff
@@ -0,0 +1,150 @@
...
+  private def getRLikeMag(gp: Option[GuideProbeTargets]): Double = {
+    def magnitudeExtractor(t: SPTarget): Option[Magnitude] =
+      RLikeBands.flatMap(b => t.getTarget.getMagnitude(b.toOldModel).asScalaOpt.map(_.toNewModel)).headOption
+
+    val r = for {
+      g <- gp
+      p <- g.getPrimary.asScalaOpt
+      m <- magnitudeExtractor(p)
+    } yield m.value
+    r.getOrElse(99.0)
```
swalker2m: What about just returning the `Option` directly and dealing with it upstream?  Otherwise you've got to know to look for `99.0`?

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/mascot/MagLimits.scala
```diff
@@ -15,13 +17,15 @@ case class MagLimits private (bmag: Double,
...
+  private lazy val filters = List(new Magnitude(bmag, MagnitudeBand.B), new Magnitude(vmag, MagnitudeBand.V), new Magnitude(rmag, MagnitudeBand.R), new Magnitude(rmag, MagnitudeBand.UC), new Magnitude(jmag, MagnitudeBand.J), new Magnitude(hmag, MagnitudeBand.H), new Magnitude(kmag, MagnitudeBand.K))
+  private lazy val RLikeFilters = RLikeBands.map(new Magnitude(rmag, _))
 
   /**
    * Returns true if the given star is within the mag limits
    */
   def filter(star: Star): Boolean =
-    filters.forall { m => star.target.magnitudeIn(m.band).exists(_.value <= m.value) }
+    // requires that B, V, J, H and K are within limits and at least one of r, R, UC are within limits
+    filters.forall { m => star.target.magnitudeIn(m.band).exists(_.value <= m.value) } && RLikeFilters.exists { m => star.target.magnitudeIn(m.band).exists(_.value <= m.value) }
```
swalker2m: Did you mean for `filters` to also contain `R` and `UC`?

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsGuideStars.scala
cquiroz: It is only called from the method above, if I don't do it here I need 3 `getOrElse(99.0)` at line 115

swalker2m:
> Conversion of most of the code to Scala

We're better off for it, but that was quite a lot to bite off!  Was it just too overwhelming to work directly with the old implementation?  You're definitely the Mascot expert now.

We should get this merged in for the test release today.  :+1:

cquiroz:
I mostly wanted to flip to scala trying to understand what the `analyze` method was doing, the Java code was cryptic and I just couldn't parse it to check for regression. On the other hand I underestimated how long would it take to do it, that's true.
I'm also much happier now with a real cancelable `analyzeGoodEnough` method which relied on throwing an `CancellationException` to work properly
I wasn't too happy with so many conversions back and forth between java and scala collection either

PR #361, opened 2015-05-08
OCSADV-312: Catalog PPMXL contains some bad magnitudes

cquiroz:
This PR takes care of filtering out bad magnitude values that are sometimes present on the catalogs, in particular for PPMXL


fnussber:
:+1:

PR #362, opened 2015-05-08
Changes in itc build file and credentials template

spakzad:
Changes in itc build file and credentials template that allow to build and run ITC as a standalone app.

(This is Florian talking on behalf of Sabrina ;): I've fixed a bug in the itc app build file and also added (empty) configurations for everybody to the credentials template file. This makes it easier to do a build without the credentials because the template file can be used as-is as a stand in for the "real" credentials file and allows to compile the code.


fnussber:
Woo-hoo. The very first PR from an outside person. Oh boy!

swalker2m:
:+1:

PR #363, opened 2015-05-08
Simplification of caching in ImageServlet.

fnussber:
Following Shane's suggestion I am replacing the home made LRU cache based on a ```ConcurrentHashMap``` with one based on ```LinkedHashMap```. Note that ```LinkedHashMap``` is not concurrent and has to be synchronized, the resulting speed penalty however seems to be small, at least for only a handful of parallel requests. And since the expected load on the web app is more on the order of requests per hour than requests per second I guess it doesn't really matter anyway.

### DIFF
--- a/bundle/edu.gemini.itc.web/src/main/java/edu/gemini/itc/web/servlets/ImageServlet.java
+++ b/bundle/edu.gemini.itc.web/src/main/java/edu/gemini/itc/web/servlets/ImageServlet.java
@@ -2,17 +2,13 @@ package edu.gemini.itc.web.servlets;
 
 import edu.gemini.itc.shared.*;
 import org.jfree.chart.ChartUtilities;
-import scala.Tuple2;
 
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import java.awt.image.BufferedImage;
 import java.io.IOException;
-import java.util.Calendar;
-import java.util.Map;
-import java.util.UUID;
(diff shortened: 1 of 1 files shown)


bundle/edu.gemini.itc.web/src/main/java/edu/gemini/itc/web/servlets/ImageServlet.java
```diff
@@ -42,43 +38,28 @@
 
     public static class IdTimedOutException extends RuntimeException {}
 
-    private static final int UpperLimit = 600;
+    private static class LRU extends LinkedHashMap<UUID, ItcSpectroscopyResult> {
+        private static final int CacheLimit = 300;
+        @Override protected boolean removeEldestEntry(final Map.Entry<UUID, ItcSpectroscopyResult> eldest) {
```
cquiroz: When and who calls this? Is it called when you add new ones keeping the max size under control?

swalker2m: RIght, from the docs:

> Returns true if this map should remove its eldest entry. This method is invoked by put and putAll after inserting a new entry into the map. It provides the implementor with the opportunity to remove the eldest entry each time a new one is added. This is useful if the map represents a cache: it allows the map to reduce memory consumption by deleting stale entries.

cquiroz: :+1:

PR #364, opened 2015-05-08
OCSADV-311: Manual Search with UCAC4 not working on https

cquiroz:
UCAC4 use the `VoTable2Skycat` servlet proxy running on SPDB, Access is via https and this requires supporting the self-signed certificate and skip hostname validation like TRPC does


swalker2m:
:+1:

PR #365, opened 2015-05-08
REL-2314: horizons bug fixes

swalker2m:
This PR offers fixes for a null pointer exception reported by Andy in REL-2314 along with a couple of other issues that fell out of testing.  First, when canceling a lookup of a "multiple answer" result we were showing a popup that read "Internal error: MultipleResults".  This was because the canceled result was being taken for a valid answer from Horizons (see `HorizonsService`).  Second, if we change the name of a conic target, the horizons ID was not being reset which resulted in unintended cache hits on the lookup as well as misleading information in the UI (see `ConicNameEditor`).

For the actual null pointer exception, I just assume that a `HorizonsReply` can return `null` for any of its fields, which is a decent assumption given the state of the old Java horizons code.  One thing that is still confusing though is that the "object id" is set to the name of the target.  For example, "ceres" which corresponds to minor body 1.  Unless the user does a lookup on "ceres" and then goes to the trouble of actually renaming the target "1" there is a cache miss on each subsequent lookup because we compare "ceres" to the `HorizonReply`'s object id `1`.  Perhaps that intended behavior I don't know.

This is all terrible.

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details/ConicNameEditor.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details/ConicNameEditor.scala
@@ -41,6 +41,10 @@ final class ConicNameEditor(date: HorizonsIO[Date]) extends JPanel with Telescop
       override def textBoxKeyPress(tbwe: TextBoxWidget): Unit =
         nonreentrant {
           spt.getTarget.setName(tbwe.getValue)
+          // editing the name so invalidate the horizons id.
+          ct.setHorizonsObjectId(null) //
+          ct.setHorizonsObjectTypeOrdinal(-1)
+          hid.setText(hidText(ct))
           spt.notifyOfGenericUpdate()
         }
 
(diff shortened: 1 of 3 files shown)


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details/ConicNameEditor.scala
```diff
@@ -41,6 +41,10 @@ final class ConicNameEditor(date: HorizonsIO[Date]) extends JPanel with Telescop
       override def textBoxKeyPress(tbwe: TextBoxWidget): Unit =
         nonreentrant {
           spt.getTarget.setName(tbwe.getValue)
+          // editing the name so invalidate the horizons id.
```
cquiroz: What was the role of `nonreentrant`?

cquiroz:
:+1:

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details/ConicNameEditor.scala
swalker2m: My understanding is that it is about avoiding firing events while in the middle of handling events.  It serves as a nice way of marking code that should be executed with event handlers off as opposed to explicitly keeping up with event registrations and removing them and then re-adding them at the end.

tpolecat: Yeah it's a hack that says "I'm pushing model changes into the controls (or pushing control data into the model) so ignore any feedback look that might result". As Shane said it's easier than detaching and reattaching all the listeners.

swalker2m: One thing to be a bit careful about is to avoid calling a method with a `nonreentrant` section from within a `nonreentrant` section.  Otherwise when it bubbles back up the editing flag will be off before the end of the outer `nonreentrant` block.

PR #367, opened 2015-05-11
OCSADV-323B: Source profile and distribution addition to targets

fnussber:
The target editor now sports a flashy new source editor for the base/science targets which allows to set the source profile and distribution characteristics. The ITC calculations are now based on the spatial profile and spectral distribution from the target. For sidereal targets (```HmsDegTarget```s) the redshift from the science target is now taken into account, too.

I also managed to will the layout of the target editor components into using up all the horizontal space. I therefore declare myself Master of ```GridBagPanel``` and will have a couple of :beers:  tonight.

The makeshift source editor has vanished from the ITC panels for good. The layout of the ITC panels is still very poor, this will be addressed in a next PR.

I attached some screenshots, that show how the target editor now looks like. Note that the source editor part is only shown for the base target and that by default those components are not set and grayed out. A click on the source editor will make the editor available.

image

image

image

image


cquiroz:
Congratulations on your GirdBagPanelLayout mastery :smile:

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcParametersProvider.scala
```diff
@@ -46,10 +47,16 @@ object ItcParametersProvider {
       itcPanel.conditions.right
 
     def spatialProfile: String \/ SpatialProfile =
-      itcPanel.spatialProfile.fold("Spatial profile not available".left[SpatialProfile])(_.right)
+      for {
+        tEnv <- targetEnvironment
+        sp   <- tEnv.getBase.getTarget.getSpatialProfile.fold("Spatial profile not available".left[SpatialProfile])(_.right)
```
cquiroz: Can base, ever be null?

cquiroz:
The UI looks really good, would the source fields be stored in the model?

swalker2m:
It's looking good but I think it could use a few improvements.

* I worry about how wide the target editor requires the OT to be now.  Maybe we should at least see about making a minimum window size since important elements like the + button tend to disappear when you shrink it down a bit.

* I think when the pull downs are stretched out they look odd.  The target type selector needn't be wider than the target name entry field.

* I wonder if we could remove the "Spatial Profile" and "Spectral Distribution" labels altogether since they take up so much space and leave a lot of empty real estate around them.

* I don't think we need to spell out "Full Width at Half Max" especially since it is already abbreviated on the same line. I believe it is something every astronomer is familiar with.

* A trivial detail but the guiding feedback shouldn't be wider than the content above it.  We should trim it down so it lines up perfectly with the content.

* The orbital elements labels are getting truncated despite all the empty space.

* We might move the Type Tag over beside the icons to the left.  It looks odd in the middle I think.

* For conic targets we label the name "Name" and for sidereal targets "Target Name".  They should probably be the same.

swalker2m:
For the spatial profile and spectral distribution, I see in an earlier version you made them separate little editor boxes labeled accordingly.  That takes up less space and seems like a good alternative to me.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcParametersProvider.scala
swalker2m: Target environment is immutable and throws an exception if you try to create it with a `null` target so it should be fine to assume it is set to some value.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details/SourceDetailsEditor.scala
```diff
@@ -0,0 +1,266 @@
...
+    if (isBase) {
+      visible = true
+
+      profilePanels.foreach(_.panel.setVisible(false))
+      profiles.selection.item.panel.setVisible(true)
+
+      distributionPanels.foreach(_.panel.setVisible(false))
+      distributions.selection.item.panel.setVisible(true)
+
+      enableAll(spt.getTarget.getSpatialProfile.isDefined)
```
swalker2m: What should happen if the observation is observed?  Should it be editable anyway?  If not this will be problematic.  I think we may need special-case `updateEnabledState` behavior here.

fnussber:
@cquiroz Yes, the source fields are stored in the model as a part of ```ITarget``` and are also made persistent. They will have to become part of the new ```Target``` at some point.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details/SourceDetailsEditor.scala
fnussber: I didn't think about that. I guess they should behave like all the other values on this editor, which - I assume - are not editable anymore after observation. I'll add some additional behavior for that.

fnussber:
@swalker2m I went away from the two separate boxes because I could not align them with the boxes on the left side since they are not sharing a ```GridBagLayout``` and at some point it just seemed easier to do it easier this way. However, I agree that space consumption is an issue here. I give it another go, probably having two separate boxes is the way to go here.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details/SourceDetailsEditor.scala
swalker2m: There's a general issue with enabled state right now. Perhaps it would be best to just check in the version as it is now and then I can take a look at it as part of OCSADV-330.

PR #368, opened 2015-05-11
OCSADV-330: bug fix for enabled state setting

swalker2m:
The `OtItemEditor` `updateEnabledState` methods are charged with enabling and disabling widgets in the various editors.  Unfortunately it's a bit of a flawed mechanism and getting it to work correctly is very tricky.  This PR is a bit of a hack that addresses a problem with enabled state in the target editors.  Namely, visiting an executed observation's target before looking at an editable target will incorrectly record the "previous" edit state of the widgets. If that makes no sense to you, well, don't worry it shouldn't.  Ideally we would rework the enabled state setting altogether but obviously not just before the release.

Unfortunately the changes introduced to the `TargetDetailPanel` will conflict with @fnussber's current PR (OCSADV-323B) so I will wait for that one to be merged and incorporate the new widgets it introduces before committing the changes.


swalker2m:
Closing in favor of OCSADV-330-B which is merged with ITC changes.

PR #370, opened 2015-05-11
Bugfix: Get rid of split package "edu.gemini.itc.shared".

fnussber:
This deals with the aftermath of an error I made when initially setting up the bundle/package structure for ITC which resulted in having the "edu.gemini.itc.shared" package split across two bundles, which is a no-no in OSGi.

Unfortunately renaming one of the packages made it necessary to change pretty much all imports throughout the ITC code base which is why this ginormous PR resulted.

I *think* (hope) that this PR also fixes an issue that we currently have with the test installations on gnodbtest2/gsodbtest2 on which the ITC web app and service don't work; the web app for example throws ```java.lang.NoClassDefFoundError: edu/gemini/itc/shared/ServerInfo```. However I am a bit puzzled I never came across this issue on the local development installation, which was always working despite this issue. (Potentially this bug only shows depending on class loader order or some such thing?)


swalker2m:
I had a similar problem with splitting `edu.gemini.sp.vcs`.  :+1:

PR #371, opened 2015-05-12
OCSADV-330-B: target editor enabled state fixes

swalker2m:
The `OtItemEditor` `updateEnabledState` methods are charged with enabling and disabling widgets in the various editors. Unfortunately it's a bit of a flawed mechanism and getting it to work correctly is very tricky. This PR is a bit of a hack that addresses a problem with enabled state in the target editors. Namely, visiting an executed observation's target before looking at an editable target will incorrectly record the "previous" edit state of the widgets. If that makes no sense to you, well, don't worry it shouldn't. Ideally we would rework the enabled state setting altogether but obviously not just before the release.

There are two commits.  The first fixes the enabled state issue for all non-ITC widgets and the second are the changes required for the ITC widgets.

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details/SourceDetailsEditor.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details/SourceDetailsEditor.scala
@@ -9,6 +9,7 @@ import edu.gemini.spModel.core.Wavelength
 import edu.gemini.spModel.obs.context.ObsContext
 import edu.gemini.spModel.target.EmissionLine.{Continuum, Flux}
 import edu.gemini.spModel.target._
+import jsky.app.ot.OTOptions
 import jsky.app.ot.gemini.editor.targetComponent.TelescopePosEditor
 import jsky.app.ot.gemini.editor.targetComponent.details.NumericPropertySheet.Prop
 
@@ -25,6 +26,7 @@ final class SourceDetailsEditor extends GridBagPanel with TelescopePosEditor {
 
   private[this] var spt: SPTarget = new SPTarget
   private[this] var isBase: Boolean = false
+  private[this] var node: Option[ISPNode] = None
 
   private def setDistribution(sd: SpectralDistribution) = spt.getTarget.setSpectralDistribution(Some(sd))
   private def setProfile     (sp: SpatialProfile)       = spt.getTarget.setSpatialProfile(Some(sp))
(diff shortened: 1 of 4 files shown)


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details/SourceDetailsEditor.scala
```diff
@@ -248,18 +256,18 @@ final class SourceDetailsEditor extends GridBagPanel with TelescopePosEditor {
...
+    peer.getComponents.foreach(_.setEnabled(enabled))
 
-    if (b) {
-      tooltip = ""
-      border = titleBorder("Source")
-    } else {
+    if (!hasProfile && editable) {
       tooltip = "Click on title to activate this."
       border = titleBorder("Source [Click to Activate]")
+    } else {
```
swalker2m: @fnussber  I went with the flow here but I'm wondering about the "Click to Activate" idea.  There's no way to go back once you've clicked it and it's a first for this kind of interaction in the OT.   We could certainly simplify the enabled state behavior if the source widgets are just always active when the observation is editable.  That is, it wouldn't need special case handling.

If that's not desirable, another idea would be to introduce a radio button (or something) at the top to give the user the choice of defining the source information or not.  By default it would be left undefined but if they elect to define the source then the spatial profile and spectral distribution would appear?

Either way the widgets wouldn't require special case handling.  If we decide to make any changes, let's merge this PR first regardless and then update the ITC widgets.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details/SourceDetailsEditor.scala
```diff
@@ -200,13 +202,19 @@ final class SourceDetailsEditor extends GridBagPanel with TelescopePosEditor {
...
+
   // react to any kind of target change by updating all UI elements
   def edit(obsContext: GOption[ObsContext], spTarget: SPTarget, node: ISPNode): Unit = {
 
     deafTo(editElements:_*)
 
-    spt = spTarget
-    isBase = if (obsContext.isDefined) obsContext.getValue.getTargets.getBase == spTarget else false
+    spt       = spTarget
+    isBase    = if (obsContext.isDefined) obsContext.getValue.getTargets.getBase == spTarget else false
```
cquiroz: `obsContext.exists(...)` or is it a Gemini Opt?

cquiroz:
:+1:

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details/SourceDetailsEditor.scala
swalker2m: It's a Gemini `Option` which has `exists` but of course is awkward to use.

PR #373, opened 2015-05-12
REL-1842 graces partitioner

tpolecat:
Graces targets were not being partitioned, as reported by @andrewwstephens in #356.

### DIFF
--- a/bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/Phase1FolderFactory.scala
+++ b/bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/Phase1FolderFactory.scala
@@ -1,5 +1,6 @@
 package edu.gemini.phase2.skeleton.factory
 
+import edu.gemini.spModel.gemini.graces.blueprint.SpGracesBlueprint
 import edu.gemini.spModel.gemini.obscomp.SPSiteQuality
 
 import scala.collection.JavaConverters._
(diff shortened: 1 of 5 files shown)


bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/Phase1FolderFactory.scala
```diff
@@ -155,5 +157,18 @@ object F2LongslitPartitioner extends Partitioner {
...
-
+// IF          R < 6.5 INCLUDE {1}
+// ELIF 6.5 <= R < 10  INCLUDE {2}
+// ELIF 10  <= R < 21  INCLUDE {3}
+// ELIF 21  <= R       INCLUDE {4}
+// ELSE INCLUDE {1},{2},{3},{4} # No magnitude given so include all
+
+object GracesPartitioner extends Partitioner {
+  import edu.gemini.shared.skyobject.Magnitude.Band.R
+  def bucket(t:SPTarget):Int = Option(t.getTarget.getMagnitude(R).getOrNull).map(_.getBrightness).map {R =>
```
cquiroz: I think there is an `asScalaOpt` conversion that you could use in this case

cquiroz:
Just a general question, what does partitioning mean in this context?

bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/Phase1FolderFactory.scala
tpolecat: Yeah I can never remember the right import. But since you caught me I'll fix it ;-)

tpolecat:
@cquiroz in some cases the process that expands a phase 1 blueprint into a phase 2 template group is sensitive to properties of the target. However there is no "the" target ... there can be any number of targets associated with a blueprint in phase 1. So what we do is partition the targets into buckets such that all targets in a given bucket are equivalent in terms of the template expansion, and create a template group for each bucket. If I recall correctly this was a hack; we didn't anticipate that this would be necessary and the solution is bad because you have to ensure that the partitioning logic is correct with respect to the target attributes that are relevant to the blueprint expansion routine. This needs to be revisited.

... but at least we can test it now.

swalker2m:
:+1:

swalker2m:
:+1:

PR #374, opened 2015-05-12
REL-2322: missing observing conditions prevents plotting GMOS

swalker2m:
A bit of an ad-hoc fix to a bug which prevented plotting the science area and probe range if the observation is missing the observing conditions.


cquiroz:
:+1:

PR #375, opened 2015-05-12
OCSADV-337: NPE creating observation context

swalker2m:
`EdCompTargetList`, responding to an event after the target component it is editing has been cut, makes a call to get the `ObsContext` passing in the target component's "context observation". See `OtItemEditor.getContextObservation` (which really should return an `Option` instead of `null`).  Of course having been cut there is no context observation.

In general you cannot construct an `ObsContext` from a `null` `ISPObservation` so this little hack seems somewhat defendable.  All apologies.


cquiroz:
apologized :+1:

tpolecat:
:cry:

PR #376, opened 2015-05-12
OCSADV-338: added missing guiding feedback update

swalker2m:
Somehow in all the changes the guiding feedback is not updated when the selected strategy changes.  This PR takes care of that little issue.


cquiroz:
When does this happen? If you change instrument?

swalker2m:
1. Create a new GMOS observation. 
2. Select the target component. 
3. Observe that the Guide with defaults to GMOS OIWFS. 
4. Observe that the feedback says "No GMOS OIWFS guide star selected." as it should. 
5. Using the "Guide with:" menu, choose to guide with PWFS1. 
6. Observe that the feedback still says "No GMOS OIWFS guide star selected."

cquiroz:
I see, thanks for the fix :+1:

PR #377, opened 2015-05-12
OCSADV-317: PPMXL Catalog contains more than one Magnitude measurement

cquiroz:
Update the VoTable parser to be smarter on ppmxl.
ppmxl may contain more than one measurement for R and B bands, the rule is to pick the first value over the second if possible

### DIFF
--- a/bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableParser.scala
+++ b/bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableParser.scala
@@ -209,10 +209,10 @@ trait VoTableParser {
     }
 
     def combineWithErrorsAndFilter(m: List[(FieldId, MagnitudeBand, Double)], e: List[(FieldId, MagnitudeBand, Double)]): List[Magnitude] = {
-      val mags = m.collect {
+      val mags = m.map {
           case (f, b, d) => f -> new Magnitude(d, b)
         }
-      val magErrors = e.collect {
+      val magErrors = e.map {
           case (_, b, d) => b -> d
         }.toMap
       // Filter magnitudes as a whole
--- a/bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableParser.scala
+++ b/bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableParser.scala
@@ -57,7 +57,7 @@ sealed trait MagnitudesFilter {
   // Find what band the field descriptor should represent
   def findBand(id: FieldId, band: String): Option[MagnitudeBand] = MagnitudeBand.all.find(_.name == band)
   // filter magnitudes as a whole
-  def filterMagnitudeFields(magnitudeFields: List[(FieldId, Magnitude)]): List[Magnitude] = magnitudeFields.map(_._2).filter(validMagnitude)
+  def filterMagnitudeFields(magnitudeFields: List[(FieldId, Magnitude)]): List[Magnitude] = magnitudeFields.collect { case (_, mag) if validMagnitude(mag) => mag }
 }
 
 case object UCAC4Filter extends MagnitudesFilter {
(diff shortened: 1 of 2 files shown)


bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableParser.scala
```diff
@@ -52,8 +52,12 @@ object VoTableParser extends VoTableParser {
 sealed trait MagnitudesFilter {
   // Indicates if a field should be ignored
   def ignoredMagnitudeField(v: FieldId): Boolean = false
+  // Indicates if a magnitude is valid
   def validMagnitude(m: Magnitude): Boolean = !(m.value.isNaN || m.error.exists(_.isNaN))
-  def findBand(id: FieldId, band: String): Option[MagnitudeBand] = None
+  // Find what band the field descriptor should represent
+  def findBand(id: FieldId, band: String): Option[MagnitudeBand] = MagnitudeBand.all.find(_.name == band)
+  // filter magnitudes as a whole
+  def filterMagnitudeFields(magnitudeFields: List[(FieldId, Magnitude)]): List[Magnitude] = magnitudeFields.map(_._2).filter(validMagnitude)
```
swalker2m: Admittedly not as compact but I suppose you could make one pass with `collect`, if you wanted to.

````scala
magnitudeFields.collect { case (_, mag) if validMagnitude(mag) => mag }
````

Comment (cquiroz) [inline bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableParser.scala]: I like it :smile:

File: bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableParser.scala [hunk 2]
```diff
@@ -73,6 +77,34 @@ case object UCAC4Filter extends MagnitudesFilter {
...
+    val m = primaryMagnitudes.filter {
+      case f @ (FieldId(i, _), _) if alternateMagnitudesIds.contains(i) => false
+      case _                                                            => true
+    }
+    // In case the primary isn't present use the alternative
+    idsMapping.foldLeft(m) {
+      case (mags, (id1, id2)) => if (mags.exists(_._1.id === id1)) mags else mags ++ alternateMagnitudes.get(id2)
+    }.map(_._2)
+  }
+}
```
swalker2m: Could you make one pass through the `magnitudeFields` upfront to produce a `Map[String, Magnitude]` for all valid `Magnitude` values in the list.  Then do something like the `idMapping` fold at the end to fix it up?  Does the order matter?  Is this more or less the same result?:

````scala
val magMap1 = (Map.empty[String, Magnitude]/:magnitudeFields) { case (m, (FieldId(id, _), mag)) =>
   if (validMagnitude(mag)) m + (id -> mag) else m
}
// Now magMap1 might have double entries for R and B.  Just get rid of the alternative if so.
val magMap2 = (magMap1/:idsMapping) { case (m, (id1, id2)) =>
  if (magMap1.contains(id1)) m - id2 else m
}
magMap2.values.toList  // we've scrambled the order of course
````

Comment (cquiroz) [inline bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableParser.scala]: Nice, the order may not matter as we do a full `sorted` at the end
BTW, I should replace the Ordering used for that `sorted` as the default typeclass may not be correct

**Comment by cquiroz** [comment]
I updated the code as suggested and also change the ordering so that magnitudes are sorted by band

File: bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableParser.scala [hunk 3]
```diff
@@ -178,19 +208,29 @@ trait VoTableParser {
...
 
-    def combineWithErrorsAndFilter(m: List[Magnitude], e: Map[MagnitudeBand, Double]): List[Magnitude] =
-      m.map(i => i.copy(error = e.get(i.band))).filter(magnitudesFilter.validMagnitude)
+    def combineWithErrorsAndFilter(m: List[(FieldId, MagnitudeBand, Double)], e: List[(FieldId, MagnitudeBand, Double)]): List[Magnitude] = {
+      val mags = m.collect {
+          case (f, b, d) => f -> new Magnitude(d, b)
+        }
+      val magErrors = e.collect {
+          case (_, b, d) => b -> d
+        }.toMap
```
tpolecat: in both cases above could you not just use `map`?

cquiroz: Right

PR #378, opened 2015-05-14
OCSADV-332: brightness string parser improvements

swalker2m:
This PR incorporates a number of requested changes to the parsing of the old "brightness" string:

* Parses "mJy" as milli-Jy
* Allows bands to be specified with an optional prime (') as in z'
* Parses UC, UCmag, etc.
* Allows explicit specification of the Vega system, e.g. "16.9 J (Vega)" 
* Allows "in" as a separator between a value and band, e.g. "7.39 in V" 

The only request that hasn't yet been incorporated is:

> Don't fail if there is an unknown filter, e.g. "M=5.38, N=3.11, Qa=0.96"

but I wanted to get this PR out ahead of the test release tomorrow.


bundle/edu.gemini.spModel.io/src/main/scala/edu/gemini/spModel/io/impl/migration/to2015B/BrightnessParser.scala
```diff
@@ -19,35 +19,49 @@ object BrightnessParser extends RegexParsers {
...
 
   private val band: Parser[Band] = {
     def bandParser(b: Band): Parser[Band] = {
-      val p = b.toString.r <~ """\s*(_mag|mag|-band|band)?""".r ^^^ b
+      val p = (b.toString + """'?\s*(_mag|mag|-band|band)?""").r ^^^ b
       p | paren(p) | paren(p, '<', '>')
     }
-    Band.values.map(bandParser).foldRight[Parser[Band]](failure("expected band"))(_ | _)
+
+    Band.values.map(bandParser).foldRight[Parser[Band]](failure("expected band"))(_ ||| _)
```
cquiroz: So '|||' is the trick to get the UC to work?

bundle/edu.gemini.spModel.io/src/main/scala/edu/gemini/spModel/io/impl/migration/to2015B/BrightnessParser.scala
```diff
@@ -19,35 +19,49 @@ object BrightnessParser extends RegexParsers {
     """\s*[~=,]?\s*""".r ^^^ { () }
 
   private val at: Parser[Unit] =
-    """\s*(nucleus at|at|@|/|:)?\s*""".r ^^^ { () }
+    """\s*(nucleus at|at|in|@|/|:)?\s*""".r ^^^ { () }
 
-  private val num: Parser[Double] =
-    """-?\d+(\.\d*)?""".r <~ """\s*(mag)?""".r ^^ (_.toDouble)
+  private val num: Parser[BigDecimal] =
+    """-?\d+(\.\d*)?""".r <~ """\s*(mag)?""".r ^^ (s => BigDecimal(new java.math.BigDecimal(s))) // BigDecimal.exact(s)
```
cquiroz: Why `BigDecimal`? I'd think we don't need so much precision for Magnitude values

cquiroz:
Looks good, nice work to get the band parser to work, I should try a similar trick on `VoTableParser`

bundle/edu.gemini.spModel.io/src/main/scala/edu/gemini/spModel/io/impl/migration/to2015B/BrightnessParser.scala
swalker2m: Apparently, though it would be worth playing with "reluctant" versions of the regex.  The issue seems to be partially matching the U band.  If you put UC first in the band list it also works.

bundle/edu.gemini.spModel.io/src/main/scala/edu/gemini/spModel/io/impl/migration/to2015B/BrightnessParser.scala
swalker2m: I felt a bit sheepish about hauling in `BigDecimal` but I wanted a more accurate way to convert, say, 200 mJy to 0.2 Jy.  Normal floating point math left values with tiny errors that break the test cases at any rate.

tpolecat:
Looks good. :+1:

PR #379, opened 2015-05-14
OCSADV-323C: Usability and UI improvements on ITC panels

fnussber:
Lots of UI changes.

* Added a feedback panel at the bottom for error messages (similar to guider feedback)
* Adapt table column widths to actual content (use less space horizontally)
* Charts adapt to available vertical space
* Fixed plot limits input fields
* Use ```NumberBoxWidget``` for input (same widget as in ```TargetEditor```)
* Keep current row selected, select row 0 as default
* Condition values deviating from program conditions are shown in red, identical values in black
* Mark rows with problems with x icon in first column
* Fixed rendering for special values, e.g. ```DisplayableSpType```
* and some other tiny bits and pieces 

image

image

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcPanel.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcPanel.scala
@@ -21,7 +21,6 @@ import scala.util.{Failure, Success}
 
 object ItcPanel {
 
-  val OkIcon      = new ImageIcon(classOf[ItcTableModel].getResource("/resources/images/check.png"))
   val ErrorIcon   = new ImageIcon(classOf[ItcTableModel].getResource("/resources/images/error_tsk.gif"))
   val SpinnerIcon = new ImageIcon(classOf[ItcTableModel].getResource("/resources/images/spinner16.gif"))
 
@@ -153,8 +152,7 @@ class ItcSpectroscopyPanel(val owner: EdIteratorFolder) extends ItcPanel {
   }
 }
 
-/** Panel holding some feedback.
-  * Mainly important in case something went wrong with the ITC calculations. */
+/** Panel holding some feedback, mainly in case something went wrong with the ITC calculations. */
 private class ItcFeedbackPanel(table: ItcTable) extends Label {
 
   import ItcPanel._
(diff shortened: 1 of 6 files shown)


swalker2m:
The example error feedback shows: 
* Missing data file: /gmos/gmos_n_r_G0303_RG610_G0307.dat
Is this something the user would know what to do with?

swalker2m:
I think this is looking pretty good.

As a user though, I'm not sure it would occur to me the difference between the observation's real conditions and the conditions used for the calculations.  Might a user be led to think that she is actually *changing* the required conditions when she manipulates those widgets?  The tooltip obviously helps tough and the red color indicates that something is amiss so I'm not sure what else to suggest.  Maybe we could show the actual observation conditions in fixed labels above the drop-downs to underscore the fact that they aren't being changed?  We could call them "Observation Conditions" then we could label the editable ones "ITC Plot Conditions" or something.  Perhaps Andy has an opinion.

Otherwise, it all looks good.  Could we try the plots with the white and grey reversed though?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcPanel.scala
```diff
@@ -273,6 +152,54 @@ class ItcSpectroscopyPanel(val owner: EdIteratorFolder) extends ItcPanel {
...
+        visible = false               // no feedback for this row, don't show feedback panel
+      } { case (ico, msg, col) =>
+        icon        = ico             // feedback available, show it
+        text        = msg
+        background  = col
+        visible     = true
+      }
+    }
+    revalidate()
+  }
```
tpolecat: I think you could say `table.selected.flatMap(feedback).fold ...` to remove the nesting here.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcPanel.scala
```diff
@@ -403,3 +445,37 @@ protected class PlotDetailsPanel extends GridBagPanel {
...
+    addWatcher(new TextBoxWidgetWatcher {
+      override def textBoxKeyPress(tbwe: TextBoxWidget): Unit = textBoxAction(tbwe)
+      override def textBoxAction(tbwe: TextBoxWidget): Unit =
+        try {
+          publish(new ValueChanged(NumberEdit.this))
+          tbwe.requestFocus()
+        }
+        catch {
+          case _: NumberFormatException =>
+        }
```
tpolecat: I realize this is insanely picky but the newline before `catch` is making me crazy.

tpolecat:
Code looks good (or as good as you can hope for since this is Swing). :+1:

fnussber:
* Error message was changed in different PR to include "Unsupported configuration: ..". That's not great but unfortunately it's difficult to be more specific because in the OT I currently can't track back which component (filter, grating..?) has caused the problem. Showing the name of the data file however will in most cases give you a hint which component it is since those names mostly reflect the name of the component. So, yeah,  this is not awesome, but I can't think of anything better, for now at least.

* The conditions panel (as pretty much all of the current UI) has been discussed with Andy and reflects his input. I actually started out exactly with the idea @swalker2m is mentioning but then tried to make it more compact and less cluttered and reduced it to one UI element with the colors. I think adding some additional tooltips etc might help. I was hoping to collect more user feedback for this with the release.

* Changing the charts colors was  not one of Andy's priorities and there is also the point that changing them will change them in the web app, which is not a bad thing per se, but  I was trying to avoid that for now. Since this seems to be important I am going to work on this in a separate PR and introduce the coloring that we've been using for QV charts. Andy approved that change.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcPanel.scala
fnussber: That's a good one. I sometimes get confused when there is too much (flat)mapping&folding and other stuff going on...

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcPanel.scala
fnussber: Yes. This *is* insanely picky.

PR #380, opened 2015-05-14
Log fine tuning.

fnussber:
Not all available components in the OT are supported by the ITC, for some filters and other components we don't have the necessary data files. While the options in the web app are restricted to the available components now with the OT this originally severe problem became a regular occurrence, which is why I turned down the log level in order to keep it from flooding the logs.


cquiroz:
:+1:

PR #382, opened 2015-05-14
REL-2306: Fix "AsA comet" import issue

swalker2m:
This PR contains updates to the migration code for semesters 2015A and 2015B.

* __2015A__ - In 2015A the representation of template parameters changed. Part of the conversion process required copying PIO `ParamSet`s that represent targets and conditions.  This was done by reading the `ParamSet` into an `SPTarget` (or `SPSiteQuality`) and then exporting it back to a new `ParamSet`.  This is problematic because the target `ParamSet` format has changed.  Instead of going through the model classes, the update has been changed to simply copy the XML snippets into their proper places for the 2015A template parameter representation.  Note, this was only an issue when importing pre-2015A XML.

* __2015B__ - The 2015B migration was working for targets inside observations but ignoring template parameters.  The code has been updated to include template parameters.  The magnitude conversion note is added to the template group itself in the case of template parameters.

### DIFF
--- a/bundle/edu.gemini.spModel.io/src/main/scala/edu/gemini/spModel/io/impl/migration/to2015A/To2015A.scala
+++ b/bundle/edu.gemini.spModel.io/src/main/scala/edu/gemini/spModel/io/impl/migration/to2015A/To2015A.scala
@@ -2,7 +2,7 @@ package edu.gemini.spModel.io.impl.migration.to2015A
 
 import edu.gemini.spModel.gemini.obscomp.SPSiteQuality
 import edu.gemini.spModel.io.impl.SpIOTags
-import edu.gemini.spModel.pio.xml.PioXmlFactory
+import edu.gemini.spModel.pio.xml.{PioXmlUtil, PioXmlFactory}
 import edu.gemini.spModel.pio._
 import edu.gemini.spModel.target.{SPTargetPio, SPTarget}
 import edu.gemini.spModel.template.{TemplateParameters, TemplateGroup, TemplateFolder}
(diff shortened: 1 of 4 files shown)


bundle/edu.gemini.spModel.io/src/main/scala/edu/gemini/spModel/io/impl/migration/to2015B/To2015B.scala
```diff
@@ -90,15 +97,31 @@ object To2015B {
...
   }
 
+  // Extract all the target paramsets, be they part of an observation or
+  // template parameters.
+  private def allTargets(d: Document): List[ParamSet] = {
+    val names = Set(PARAMSET_BASE, PARAMSET_TARGET)
+
+    val templateTargets = for {
+      cs  <- d.findContainers(SPComponentType.TEMPLATE_PARAMETERS)
+      tps <- cs.allParamSets if tps.getName == PARAMSET_TEMPLATE_PARAMETERS
```
cquiroz: I didn't know you could put the filter on the some line, nice

bundle/edu.gemini.spModel.io/src/main/scala/edu/gemini/spModel/io/impl/migration/to2015A/To2015A.scala
```diff
@@ -91,8 +81,8 @@ object To2015A {
 
         // Add them back as children of this template parameters
         // data object ParamSet.
-        tp.addParamSet(targetMap(targetId).getParamSet(fact))
-        tp.addParamSet(condsMap(condsId).getParamSet(fact))
+        tp.addParamSet(PioXmlUtil.read(targetMap(targetId)).asInstanceOf[ParamSet])
```
cquiroz: How do you know the cast is safe? Could `PioXmlUtil.read` return directly a `ParamSet`?

bundle/edu.gemini.spModel.io/src/test/scala/edu/gemini/spModel/io/impl/migration/to2015B/TemplateTargetConversionTest.scala
```diff
@@ -0,0 +1,98 @@
...
+
+import org.junit.Test
+import org.junit.Assert._
+
+import scala.collection.JavaConverters._
+
+// a rudimentary test to make sure it doesn't blow up
+
+class TemplateTargetConversionTest {
+  private def withTestOdb(block: IDBDatabaseService => Unit): Unit = {
```
cquiroz: I bet we are doing this in many places on the codebase, do you think it should go to a common testing package?

cquiroz:
:+1:

bundle/edu.gemini.spModel.io/src/main/scala/edu/gemini/spModel/io/impl/migration/to2015A/To2015A.scala
swalker2m: Unfortunately `PioXmlUtil.read` could return any `PioNode` subclass.  I only know it is safe here because in the `targetMap` I've stuffed `ParamSet` strings in the `mapParamSets` function above.  This is all suboptimal.

bundle/edu.gemini.spModel.io/src/test/scala/edu/gemini/spModel/io/impl/migration/to2015B/TemplateTargetConversionTest.scala
swalker2m: Sounds like a good idea yes.

PR #383, opened 2015-05-14
OCSADV-339: MASCOT error when no CWFS guide stars found

cquiroz:
Bug fix and test case for a particular target that broke the matching of guide probe targets
Note that the target is on the northern hemisphere and AGS will dutifully find targets even though you can't see the target with GEMS


swalker2m:
:+1:

PR #387, opened 2015-05-15
Add "undefined" option to source profile and distribution editors.

fnussber:
Yes, ok. The "Click to Activate" idea was not going to win any awards. It has been replaced by "«undefined»" options for both pull down menus.

@swalker2m : I simplified the ```updateEnabledState()```. However I am wondering about the ```editable``` method and the additional ```node```. ```editable``` doesn't seem to be called anywhere. Can we delete this or am I missing something?

image

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details/SourceDetailsEditor.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details/SourceDetailsEditor.scala
@@ -15,7 +15,7 @@ import jsky.app.ot.gemini.editor.targetComponent.details.NumericPropertySheet.Pr
 
 import scala.swing.GridBagPanel.{Anchor, Fill}
 import scala.swing.ListView.Renderer
-import scala.swing.event.{MouseClicked, SelectionChanged}
+import scala.swing.event.SelectionChanged
 import scala.swing.{ComboBox, GridBagPanel, Label, Swing}
 import scalaz.Scalaz._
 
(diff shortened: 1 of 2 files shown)


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details/SourceDetailsEditor.scala
```diff
@@ -216,19 +211,21 @@ final class SourceDetailsEditor extends GridBagPanel with TelescopePosEditor {
     isBase    = if (obsContext.isDefined) obsContext.getValue.getTargets.getBase == spTarget else false
     this.node = Option(node)
 
-    spt.getTarget.getSpatialProfile.getOrElse(defaultProfile) match {
-      case s: PointSource => profiles.selection.item = profilePanels.head;
-      case s: GaussianSource => profiles.selection.item = profilePanels(1); profilePanels(1).panel.asInstanceOf[NumericPropertySheet[GaussianSource]].edit(obsContext, spTarget, node)
-      case s: UniformSource => profiles.selection.item = profilePanels(2);
+    spt.getTarget.getSpatialProfile match {
+      case None                     => profiles.selection.item = profilePanels.head
+      case Some(s: PointSource)     => profiles.selection.item = profilePanels(1);
```
cquiroz: You could skip the semicolon

cquiroz:
Seems fine code-wise, I can't really comment on the functionality

swalker2m:
> @swalker2m : I simplified the updateEnabledState(). However I am wondering about the editable method and the additional node. editable doesn't seem to be called anywhere. Can we delete this or am I missing something?

The `editable` method was used in the special-case logic of `updateEnabledState` which you simplified.  With this change we shouldn't need the `editable` method nor even the simplified `updateEnabledState` method.  It looks like you could delete them both (and remove the reference to it in `EdCompTargetList`) since the default behavior should work now.  Confirm by explicitly setting the observation status to prepared / observed.

swalker2m:
I like this idea better than "Click to Activate".  Let's remove `editable` and `updateEnabledState` and assuming it works :+1:

fnussber:
Ok, I verified this and it works now with the standard handling that all the other components have.

PR #388, opened 2015-05-15
OCSADV-329: Target I/O Bug

swalker2m:
This PR introduces fixes for a couple of bugs in target I/O.

* In this release we precess B1950 targets in a migration step.  The migration code writes the precessed RA and Dec back into the `ParamSet` to be later read by `SPTargetPio`.  It takes the B1950 RA `String` in HH:MM:SS format, does the math to convert it to J2000 and then writes the value in degrees as a `String`.  This `String` value in degrees was being interpreted by `SPTargetIO` as hours.  The fix is to first try parsing the value as a `Double` degrees and, failing that, fallback to HH:MM:SS (or DD:MM:SS for declination).

* `To2015B` was setting the proper motion value and units incorrectly in the updated `ParamSet`.  The result coming back from the precession code is in degrees per year which was being converted to arcsec/year but stored with the units seconds/year.  The PR updates this to take degrees/year to mas/year and use the proper units.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/SPTargetPio.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/SPTargetPio.java
@@ -143,6 +143,18 @@ public class SPTargetPio {
         return paramSet;
     }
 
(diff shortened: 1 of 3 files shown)


tpolecat:
Ugh. :+1:

PR #391, opened 2015-05-16
OCSADV-328/346: ITC on-source fraction and coadds.

fnussber:
In order to calculate the resulting S/N it is important to know the fraction of "on-source" vs "off-source" images. ITC uses a guesstimation algorithm as follows: take the FOV / science area of the instrument and check if  ```p-offset > 0.5*width or q-offset > 0.5*height```, if so, the image is considered to be off-source and on-source otherwise.

The science area returned by the instruments already takes into account that the FOV for spectroscopy is actually reduced to the slit width of the mask, which is precisely what we need here.

Note that I added a 85"x85" science area to the GSAOI instrument, replacing the default of -1"x-1". Tests are all still ok.

I also piggy-backed a small change on top of this one which deals with taking co-adds into account for NIRI and GSAOI (OCSADV-346).

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcTable.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcTable.scala
@@ -5,7 +5,6 @@ import javax.swing.{Icon, ListSelectionModel}
 
 import edu.gemini.ags.api.AgsRegistrar
 import edu.gemini.itc.shared._
-import edu.gemini.pot.sp.SPComponentType
 import edu.gemini.pot.sp.SPComponentType._
 import edu.gemini.shared.skyobject.Magnitude
 import edu.gemini.spModel.`type`.DisplayableSpType
@@ -13,6 +12,7 @@ import edu.gemini.spModel.config2.{Config, ConfigSequence, ItemKey}
 import edu.gemini.spModel.core.{Peer, Wavelength}
 import edu.gemini.spModel.guide.GuideProbe
 import edu.gemini.spModel.obs.context.ObsContext
+import edu.gemini.spModel.obscomp.SPInstObsComp
 import edu.gemini.spModel.rich.shared.immutable.asScalaOpt
 import edu.gemini.spModel.target.system.ITarget
 import jsky.app.ot.userprefs.observer.ObservingPeer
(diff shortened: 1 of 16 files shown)


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcTable.scala
```diff
@@ -243,6 +248,30 @@ trait ItcTable extends Table {
...
+  // Makes an educated guesstimate of the fraction of images of the given set that are on source.
+  // For imaging the field of view is the science area, for spectroscopy the slit width of the mask is used as width.
+  // Then, for every image we check if the p or q offsets are more than 0.5 times the horizontal or vertical
+  // field of view, if so, we assume that image is off-source, otherwise we assume it is on-source.
+  private def extractSourceFraction(uc: ItcUniqueConfig, instrument: SPInstObsComp): String \/ Double = {
+
+    // get science area / field of view in arcsecs of the instrument
+    // this already takes care to replace the width with the slit width of the mask for spectroscopy
+    val scienceArea = instrument.getScienceArea
+    val (w, h)      = (scienceArea(0), scienceArea(1))
```
swalker2m: The `getScienceArea` method should probably be deprecated.  It assumes that the science area is always rectangular which clearly isn't the case.  Until we get all the precise science areas into the model though, I guess there's nothing else to go on.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcUniqueConfig.scala
```diff
@@ -19,7 +19,9 @@ import scalaz.Scalaz._
  * Unique configurations represent sets of observes in a sequence which are done with the same
  * instrument configuration and exposure times.
  */
-case class ItcUniqueConfig(count: Int, labels: String, config: Config) {
+case class ItcUniqueConfig(count: Int, labels: String, configs: Seq[Config]) {
+
+  def config = configs.head
```
swalker2m: I imagine you only create these when you have a `configs` sequence with values but a `NonEmptyList` might  be a good idea here to have the compiler enforce that.

tpolecat: +1 for `NonEmptyList` ... also please avoid using`Seq` because it's impossible to reason about; you must assume that any `Seq` is both infinite and mutable, which drastically limits the safe operations you can perform.

PR #392, opened 2015-05-17
Prettify charts.

fnussber:
A facelift of the ITC charts in the OT and the web app. I also made the charts in the web app a tad bit bigger and did some minor refactoring. No actual changes in functionality, this are purely visual improvements.

I recycled the color schema and general look of the QV charts, I think it makes sense if charts throughout the OT use the same "visual language".

image

image


bundle/edu.gemini.itc.web/src/main/java/edu/gemini/itc/web/html/PrinterBase.java
```diff
@@ -77,14 +77,13 @@ protected void _printImageLink(final UUID id, final SpcChartType type, final Plo
     }
 
     protected void _printImageLink(final UUID id, final SpcChartType type, final int index, final PlottingDetails pd) {
-        _print("<IMG alt=\"" + toImgAlt(type, index) + "\" height=500 src=\"" + ServerInfo.getServerURL() +
+        _print("<img alt=\"" + toImgAlt(type) + "\" src=\"" + ServerInfo.getServerURL() +
```
cquiroz: No `width` or `height` attributes on ``? Or does `toPlotLimits` fill those attributes?

cquiroz:
This maybe a stupid question since I don't know anything about the ITC. Why on the first graphs there are lines with value 0, like `Single Exp S/N`?

bundle/edu.gemini.itc.web/src/main/java/edu/gemini/itc/web/html/PrinterBase.java
fnussber: ```width``` and ```height``` are  not mandatory, so I though as long as I just want to display the chart in its original size without any further layout restrictions I can just as well omit them.

fnussber:
This is the effect of a filter that blocks light above ~560 nm.

bundle/edu.gemini.itc.shared/src/main/java/edu/gemini/itc/shared/ITCChart.java
```diff
@@ -4,29 +4,89 @@
...
 public final class ITCChart {
-    private final XYSeriesCollection seriesData = new XYSeriesCollection();
-    private final JFreeChart chart;
+
+    // ==========================================================================
+    // Static
+
+    // Define a set of colors for charts.
+    // This are the same colors as used for Queue Visualization, maybe at some point we want
+    // to make the QV color schematas available OT wide?
```
swalker2m: That makes sense to me.

swalker2m:
:+1:

PR #393, opened 2015-05-17
Avoid unnecessary updates when panels are shuffled around.

fnussber:
A tiny bug fix that removes the change listener for the duration it takes to shuffle around the tabbed panes. This avoids updates on the panels being called when they shouldn't (e.g. updating ITC panels when there is no ITC data available).


cquiroz:
General question, is `init` being called inside the Swing Event thread?

tpolecat:
@cquiroz I would be very hesitant to make any assumptions about this.

fnussber:
There's probably no good way to know for sure, but usually you notice it very quickly when methods that interact with Swing elements are not called on the event thread.

swalker2m:
I think we pretty much have to assume all GUI work is on the event thread and explicitly run later any cases where that isn't the case.

:+1:

PR #394, opened 2015-05-18
REL-2327: Remove OT "Archives" menu

cquiroz:
Small PR that removes the "Archives" menu from the Position Editor
A bunch of code could be removed right away, but it will be easier to do when we do the full replacement of the skycat queries


swalker2m:
Presumably there is still code associated with this somewhere in the OT?  I'm not sure what it did but I wonder if it is okay to get rid of it now.

cquiroz:
Yes the code is there but it is intertwined with the rest of the catalog code, I'd rather remove it all at once when we do all catalog queries with VoTable

swalker2m:
:+1:

PR #395, opened 2015-05-18
REL-2328: Remove OT catalog "Digitized Sky at CADC (2nd generation)"

cquiroz:
Tiny PR to remove the non-working CADC catalog


swalker2m:
:+1:

PR #396, opened 2015-05-18
REL-2310: Remote RPM building

sraaphorst:
Now builds OCS RPMs remotely.
NOTE: this requires an update to your OcsCredentials file, which must be checked out of the Gemini svn repository.

### DIFF
--- a/project/src/main/scala/edu/gemini/osgi/tools/app/RPMDistHandler.scala
+++ b/project/src/main/scala/edu/gemini/osgi/tools/app/RPMDistHandler.scala
@@ -83,9 +83,9 @@ class RPMDistHandler(jre: Option[String]) extends DistHandler {
     val specPath = new File(wd, "rpm.spec")
 
     // Set up an SSH connection to the Linux VM where rpmbuild will be run.
-    if (config.remoteBuildInfo.isEmpty)
-      log.error("no remote build info: update OcsCredentials?")
-    val remoteBuildInfo = config.remoteBuildInfo.get
+    val remoteBuildInfo = config.remoteBuildInfo.getOrElse {
+      sys.error("no remote build info: update OcsCredentials?")
+    }
 
     log.info(s"establishing ssh session to ${remoteBuildInfo.hostname}")
     JSch.setConfig("StrictHostKeyChecking", "no")
@@ -101,7 +101,7 @@ class RPMDistHandler(jre: Option[String]) extends DistHandler {
     sftp.connect(remoteBuildInfo.timeout)
 
     // ChannelSftp.mkdir fails with an SftpException if a directory already exists: handle this by allowing this exception.
-    def rmkdir(dir: String): Unit =
+    def remoteMkdir(dir: String): Unit =
       try {
 	sftp.mkdir(dir)
       } catch {
@@ -109,9 +109,9 @@ class RPMDistHandler(jre: Option[String]) extends DistHandler {
       }
 
     log.info("creating remote directories")
-    rmkdir(rpmBuildDir)
+    remoteMkdir(rpmBuildDir)
     sftp.cd(rpmBuildDir)
-    rmkdir("SOURCES")
+    remoteMkdir("SOURCES")
 
     log.info("copying tarball")
     sftp.put(archiveName.getAbsolutePath, s"SOURCES/${name}.tar.gz")
--- a/project/src/main/scala/edu/gemini/osgi/tools/app/RPMDistHandler.scala
+++ b/project/src/main/scala/edu/gemini/osgi/tools/app/RPMDistHandler.scala
@@ -1,5 +1,9 @@
 package edu.gemini.osgi.tools.app
 
+//import OcsCredentials._
+
+import com.jcraft.jsch.{JSch, Session, ChannelExec, ChannelSftp, SftpException}
+
 import java.io.{ File, PrintWriter, InputStream, FileInputStream, FilterInputStream }
 import java.nio.file.{Paths, Files, StandardCopyOption}
 import java.nio.charset.StandardCharsets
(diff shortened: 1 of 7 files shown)


project/src/main/scala/edu/gemini/osgi/tools/app/RPMDistHandler.scala
```diff
@@ -82,17 +79,70 @@ class RPMDistHandler(jre: Option[String]) extends DistHandler {
...
+
+    // Set up an SSH connection to the Linux VM where rpmbuild will be run.
+    if (config.remoteBuildInfo.isEmpty)
+      log.error("no remote build info: update OcsCredentials?")
+    val remoteBuildInfo = config.remoteBuildInfo.get
+
+    log.info(s"establishing ssh session to ${remoteBuildInfo.hostname}")
+    JSch.setConfig("StrictHostKeyChecking", "no")
+    val session = new JSch().getSession(remoteBuildInfo.username, remoteBuildInfo.hostname, remoteBuildInfo.port)
+    session.setPassword(remoteBuildInfo.password)
```
cquiroz: Do you think we could allow ssh-key auth instead/besides pwd based auth?

cquiroz:
Great that you took the time to do this, it probably was very labor intensive
Do we expect to execute on the same VM running on the build machine? Could we execute on any VM (e.g. locally on the dev machines?)

project/src/main/scala/edu/gemini/osgi/tools/app/RPMDistHandler.scala
swalker2m: I often find I have to edit my known hosts files as they make changes to the VMs.  I have the vague worry that ssh-key authentication would be a bit fragile?

project/src/main/scala/edu/gemini/osgi/tools/app/RPMDistHandler.scala
```diff
@@ -82,17 +79,70 @@ class RPMDistHandler(jre: Option[String]) extends DistHandler {
...
     val specPath = new File(wd, "rpm.spec")
-    val rpmArgs = Array("rpmbuild", "-bb", s"${specPath.getAbsolutePath}")
-    val result = Runtime.getRuntime.exec(rpmArgs).waitFor()
-    if (result != 0) {
-      println("*** " + rpmArgs.mkString(" "))
-      println("*** rpmbuild returned " + result)
+
+    // Set up an SSH connection to the Linux VM where rpmbuild will be run.
+    if (config.remoteBuildInfo.isEmpty)
+      log.error("no remote build info: update OcsCredentials?")
```
swalker2m: Does `log.error` stop the command?

project/src/main/scala/edu/gemini/osgi/tools/app/RPMDistHandler.scala
```diff
@@ -82,17 +79,70 @@ class RPMDistHandler(jre: Option[String]) extends DistHandler {
...
+    val sftp = session.openChannel("sftp").asInstanceOf[ChannelSftp]
+    sftp.connect(remoteBuildInfo.timeout)
+
+    // ChannelSftp.mkdir fails with an SftpException if a directory already exists: handle this by allowing this exception.
+    def rmkdir(dir: String): Unit =
+      try {
+	sftp.mkdir(dir)
+      } catch {
+	case e: SftpException =>
+      }
```
swalker2m: I guess the `r` in `rmkdir` is for remote but I can't help but read it as "remove dir".  Maybe `remoteMkdir`?

Can we tell by the exception that this is what happened?  It seems like this could mask other problems besides "directory already exists".

sraaphorst: Unfortunately, the documentation is extremely unhelpful in this regard. The exception comes back with an `id` of `4`, which, according to the documentation, is an `SSH_FX_FAILURE`, which is used in the case of a "general failure." I suppose we could test for this specifically, but the only other two cases that come up in an `SftpChannel` operation are `SSH_FX_NO_SUCH_FILE` and `SSH_FX_OP_UNSUPPORTED`, which both seem like they wouldn't come up in an `mkdir` operation, so I'm not sure how useful that would be.

As for your other suggestion, absolutely. That's much clearer, I think.

project/src/main/scala/edu/gemini/osgi/tools/app/RPMDistHandler.scala
sraaphorst: No, it won't, but I was counting on the exception being thrown in the next line via the `None.get` call to do that. Do you think this should be handled differently? It seems that allowing exceptions to bubble up is often the way that things are terminated in the build process.

project/src/main/scala/edu/gemini/osgi/tools/app/RPMDistHandler.scala
sraaphorst: It is definitely possible to do ssh-key authentication instead of password-based. I guess that this would remove the need to modify `OcsCredentials.scala` and just store the key on the build machine? (I've already checked the new `OcsCredentials.scala` into svn, but that could be undone.)

sraaphorst:
@cquiroz I think the worst part indeed was waiting for `sbt` to reload plugins or restart :-).
Right now, I guess this wouldn't allow for local RPM building. This is certainly an option I can add, though: perhaps a `LocalHost` `HostInfo` and match on that, and then change `remoteBuildInfo` to `rpmHostInfo` and match on that and take the appropriate steps.

Anyone think that this is worth doing? I'll certainly implement it if so.

project/src/main/scala/edu/gemini/osgi/tools/app/RPMDistHandler.scala
tpolecat: I would probably just do something like this, and terminate the build with an exception.

```scala
val remoteBuildInfo = config.remoteBuildInfo.getOrElse {
  sys.error("no remote build info: update OcsCredentials?")
}
```

cquiroz:
I think what you have done is more than enough, I'd rather try to get more automation and via ssh/sftp push all the installers to `gsconfig` after an `ocsDistAll` :smile:

PR #397, opened 2015-05-18
OCSADV-349: change target node labels in the tree view

swalker2m:
Target node title changes per request in OCSADV-349.


cquiroz:
:+1:

PR #398, opened 2015-05-18
REL-2322: always show the pos angle triangle

swalker2m:
This is a bug fix for a problem in which the position angle triangle plotted in the TPE would disappear unless the observation has site quality / observing conditions.  Ordinarily if an important part of an observation is missing we cannot construct an `ObsContext`.  In this case though the observing conditions are irrelevant so we simply provide them to allow the construction of `ObsContext` even without the site quality node.


cquiroz:
Just out of curiosity, how can you have an obs context without conditions? It seems the OT assigns one as soon as you create a new program? Or does it happen for imported xmls?

swalker2m:
The user can explicitly delete them but a more common case is for template observations which come without target and conditions.

This is the second time I've reached for this trick so it probably needs better support in `TpeContext`.  I just wasn't sure how it should work.

cquiroz:
I remember a PR with a similar usage, anyway this looks good

PR #399, opened 2015-05-18
OCSADV-350: target conversion updates

swalker2m:
This PR tweaks the B1950 conversion process in a couple of ways:

1. Sets the epoch to 2000.0.
2. Adds a note with the original B1950 coordinates.


tpolecat:
Looks good. :+1:

PR #400, opened 2015-05-18
REL-815: Fixes in layout and event handling.

fnussber:
Depending on their values some of the text input fields in the "Edit Timing Window" shrunk to a very small size which made it impossible to see the value that was entered. Also, the event handling was somehow weird and often the OK button had to be clicked twice.

Changing the event handling and adapting the layout seems to solve these issues, I also increased the size of the table so that there is more room for wide entries.

NEW:
image

PREVIOUS:
image

Use ?w=1 , I converted the tabs to spaces.


swalker2m:
Looks fine to me.  :+1:

PR #401, opened 2015-05-19
OCSADV-345: Don't allow users to edit sky aperture F2, NIRI, GSAOI

fnussber:
This is a small change to disable the sky aperture value of the analysis method for certain instruments for which this feature is not supported and a default value of 1.0 has to be used instead.

Unrelated to this I am also adding a bugfix to the event handling in case of updates that avoids an update causing more updates and therefore creating more work than needed.

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcPanel.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcPanel.scala
@@ -6,6 +6,9 @@ import javax.swing._
 import edu.gemini.itc.shared.PlottingDetails.PlotLimits
 import edu.gemini.itc.shared._
 import edu.gemini.pot.sp.SPComponentType
+import edu.gemini.spModel.gemini.flamingos2.Flamingos2
+import edu.gemini.spModel.gemini.gsaoi.Gsaoi
+import edu.gemini.spModel.gemini.niri.InstNIRI
 import edu.gemini.spModel.gemini.obscomp.SPSiteQuality.{CloudCover, ImageQuality, SkyBackground, WaterVapor}
 import jsky.app.ot.util.OtColor
 import jsky.util.gui.{NumberBoxWidget, TextBoxWidget, TextBoxWidgetWatcher}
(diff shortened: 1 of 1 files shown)


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcPanel.scala
```diff
@@ -388,6 +393,14 @@ private class AnalysisMethodPanel extends GridBagPanel {
     case ValueChanged(_)                => publish(new SelectionChanged(this))
   }
 
+  def update() = {
+    // OCSADV-345: Don't allow users to change sky aperture for NIRI, F2 and GSAOI, this functionality has not been verified for these instruments.
+    Option(owner.getContextInstrumentDataObject).foreach { _.getType match {
+      case InstNIRI.SP_TYPE | Flamingos2.SP_TYPE | Gsaoi.SP_TYPE  => sky.enabled = false; skyLabel.enabled = true; sky.peer.setValue(1.0)
+      case _                                                      => sky.enabled = true;
+    }}
+  }
```
swalker2m: The code in this PR seems fine and this disabling for NIRI etc. is probably fine as well.  I'm always reluctant though to disable widgets when there's no clear (to the user) reason why and no mechanism for enabling them. Perhaps we could switch them to labels or set them invisible?

````scala
sky.visible = Option(...).exists { _.getType match {
  case InstNIRI.SP_TYPE | ... => false
  case _                      => true
}}
````

Speaking of enabled, all of these widgets are enabled even when the observation is observed, which I assume is okay since it doesn't really edit anything in the observation itself correct?

fnussber: I've discussed this with Andy and he prefers to disable the components over having them made invisible because like that people see which value is used for the calculations. We could maybe add a tooltip that gives an explanation for why this value is disabled for those three instruments.

Those values are not (yet) made persistent. Which is actually something Andy has a task for, however it might be pushing it a bit to add those to the observation for this release. For now, they are not backed by anything in the observation and are just used on-the-fly for the calculations.

swalker2m: Sounds fine. :+1:

PR #402, opened 2015-05-19
REL-2330: Fix IllegalComponentStateException

swalker2m:
The sync command will pop up a dialog for selecting the ODB to associate with a new program.  It tries to place this dialog near the UI element that triggered the sync.  When this is the command shortcut, an `IllegalComponentStateException` is thrown because it cannot get the screen location for a component that isn't visible.  This PR just updates the code to also try the `SPViewer` instance and, failing that, use the center of the screen.

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/vcs/VcsPeerSelectionDialog.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/vcs/VcsPeerSelectionDialog.scala
@@ -2,10 +2,12 @@ package jsky.app.ot.vcs
 
 import edu.gemini.spModel.core.{SPProgramID, Peer}
 import jsky.app.ot.userprefs.observer.ObservingPeer
+import jsky.app.ot.viewer.SPViewer
 
-import java.awt.Color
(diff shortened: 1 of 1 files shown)


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/vcs/VcsPeerSelectionDialog.scala
```diff
@@ -52,7 +54,24 @@ class VcsPeerSelectionDialog(c: Option[Component], pid: SPProgramID, allPeers: L
...
+      SPViewer.instances().asScala.find { v =>
+        Option(v.getProgram).exists(_.getProgramID == pid)
+      }
+
+    def center: Point =
+      Toolkit.getDefaultToolkit.getScreenSize |> { dim =>
+        new Point(dim.getWidth.toInt/2, dim.getHeight.toInt/2)
+      }
+
+    (comp.flatMap(safeLocation) orElse viewer.flatMap(safeLocation)) | center
```
cquiroz: Very elegant

cquiroz:
:+1:

PR #403, opened 2015-05-19
REL-2232 fix obs renumbering issue

tpolecat:
The obs re-ordering was incomplete, leaving an incorrect reference in the LGS expansion. This PR fixes that bug and adds Andy's failing program as a testcase.


swalker2m:
:+1:

PR #404, opened 2015-05-19
REL-2331: disappearing target detail panel

swalker2m:
Apparently the `TargetDetailPanel` disappears permanently from the target editor in some cases and requires an OT restart to reenable it.  I've been unable to reproduce the issue and unable to spot any obvious issues with the code.  However past experience has shown that when adding and removing components from a panel in response to events, a `revalidate` is often needed.  I threw in a `repaint` for good measure.  I propose that we continue testing with these changes and reevaluate if necessary.  Sorry, I realize this is wholly unsatisfactory.


fnussber:
I've had similar issues and revalidate (+repaint) usually solved it. :+1:

tpolecat:
:pray:

PR #405, opened 2015-05-19
OCSADV-314 add j2000 label

tpolecat:
In response to Andy's comment I added `(J2000)` to the right of the RA/DEC boxes. It would have pushed too many things around to put it on the left.

image


swalker2m:
:+1:

tpolecat:
I'll wait and see what Andy says.

andrewwstephens:
Looks good to me!  Thanks for making that change.

PR #407, opened 2015-05-19
REL-2329: Remove obsolete CADC catalogs

cquiroz:
The name says it all. Besides removing CADC catalogs, two "catalog directories" were removed as they produce NPEs shown in the OT when clicked


swalker2m:
Looks good to me. :+1:

PR #408, opened 2015-05-19
OCSADV-347 preserve mags, spectral info on horizons update

tpolecat:
This changes the nonsidereal editor to preserve magnitude and spectral info on Horizons lookup.


swalker2m:
Makes sense to me.  :+1:

PR #409, opened 2015-05-19
REL-2307: Added new F2 filter.

sraaphorst:
New filter `K-long` added for Flamingos2.

### DIFF
--- a/bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/template/factory/impl/flamingos2/Flamingos2Imaging.scala
+++ b/bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/template/factory/impl/flamingos2/Flamingos2Imaging.scala
@@ -38,6 +38,7 @@ case class Flamingos2Imaging(blueprint:SpFlamingos2BlueprintImaging) extends Fla
   val exposureTimes: Map[Flamingos2.Filter, Double] = {
     import Flamingos2.Filter._
     Map(H       -> 10.0,
+        K_LONG  -> 30.0,
         K_SHORT -> 30.0) withDefaultValue 60.0
   }
 
--- a/bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/template/factory/impl/flamingos2/Flamingos2Imaging.scala
+++ b/bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/template/factory/impl/flamingos2/Flamingos2Imaging.scala
@@ -35,11 +35,13 @@ case class Flamingos2Imaging(blueprint:SpFlamingos2BlueprintImaging) extends Fla
   val science = Seq(1, 2)
   val cal     = Seq(3)
 
-  val exposureTimes: Map[Flamingos2.Filter, Double] = {
+  def exposureTimes(filter: Flamingos2.Filter): Double = {
     import Flamingos2.Filter._
-    Map(H       -> 10.0,
-        K_LONG  -> 30.0,
-        K_SHORT -> 30.0) withDefaultValue 60.0
+    filter match {
+      case H                => 10.0
+      case K_LONG | K_SHORT => 30.0
+      case _                => 60.0
+    }
   }
 
   def initialize(grp:ISPGroup, db:TemplateDb): Maybe[Unit] = for {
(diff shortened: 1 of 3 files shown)


bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/template/factory/impl/flamingos2/Flamingos2Imaging.scala
```diff
@@ -38,6 +38,7 @@ case class Flamingos2Imaging(blueprint:SpFlamingos2BlueprintImaging) extends Fla
   val exposureTimes: Map[Flamingos2.Filter, Double] = {
     import Flamingos2.Filter._
     Map(H       -> 10.0,
+        K_LONG  -> 30.0,
         K_SHORT -> 30.0) withDefaultValue 60.0
   }
```
tpolecat: I realize this is not your fault, but just to make me feel better can you change the declared type to `Flamingos2.Filter => Double`? It will still compile but will lose its mapness. Maps that have been crazified with `withDefaultValue` stop making sense.

tpolecat: Hell, or even just make it

```scala
def exposureTimes(f: Flamingos2.Filter): Double = 
  f match {
    case H       => 10.0
    case K_LONG  => 30.0
    case K_SHORT => 30.0
    case _       => 60.0
  }
```

bundle/edu.gemini.wdba.xmlrpc.server/src/test/java/edu/gemini/wdba/tcc/Flamingos2SupportTest.java
```diff
@@ -71,14 +71,14 @@ public void testWavelength() throws Exception {
         // Use the disperser wavelength in spectroscopy mode with no filter
         flam2.setFilter(Filter.OPEN);
 
-        Pair[] dtA = new Pair[] {
-            new Pair<Disperser, String>(Disperser.R1200JH, "1.39"),
-            new Pair<Disperser, String>(Disperser.R1200HK, "1.871"),
-            new Pair<Disperser, String>(Disperser.R3000,   "1.65"),
+        final Pair[] dtA = new Pair[] {
```
cquiroz: I don't recall but can you in Java declare an array with params like `Pair<Disperser, String>[]`

bundle/edu.gemini.wdba.xmlrpc.server/src/test/java/edu/gemini/wdba/tcc/Flamingos2SupportTest.java
```diff
@@ -71,14 +71,14 @@ public void testWavelength() throws Exception {
...
+            new Pair<>(Disperser.R1200JH, "1.39"),
+            new Pair<>(Disperser.R1200HK, "1.871"),
+            new Pair<>(Disperser.R3000,   "1.65"),
         };
 
         //noinspection unchecked
-        for (Pair<Disperser, String> t : (Pair<Disperser, String>[]) dtA ) {
+        for (final Pair<Disperser, String> t : (Pair<Disperser, String>[]) dtA ) {
```
cquiroz: These casts could be gone if you can declare the array with types

bundle/edu.gemini.wdba.xmlrpc.server/src/test/java/edu/gemini/wdba/tcc/Flamingos2SupportTest.java
sraaphorst: I don't think so. I tried that, and had complaints about generics and arrays. Looks like it is not possible:
http://stackoverflow.com/questions/7131652/generic-array-creation-error

bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/template/factory/impl/flamingos2/Flamingos2Imaging.scala
sraaphorst: Much cleaner. Thanks, @tpolecat. I'll make this change now.

bundle/edu.gemini.wdba.xmlrpc.server/src/test/java/edu/gemini/wdba/tcc/Flamingos2SupportTest.java
sraaphorst: Very annoying. Even doing an `Arrays.asList(new Pair<...>(...), ...)` fails with "generic array creation" because this method relies on `Collections.toArray` to work, and cannot thus work with arrays of generics. The only solution, I think, would be to specifically create an `ArrayList<Pair<...>>` and then add the objects one at a time, which I think we'll not worry about for now.

PR #410, opened 2015-05-19
Prevent enablement of select → box

tpolecat:
This prevents the terrible creepy-crawling enablement status setter from enabling the **Select →** box in the magnitude editor.


bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/MagnitudeEditor.java
```diff
@@ -243,11 +243,16 @@ public void setTarget(SPTarget target, Mode mode) {
...
+            tf = new JTextField("Select \u2192") {
+                {
+                    setColumns(5);
+                    setEnabled(false);
+                    setMinimumSize(getPreferredSize());
+                }
+                @Override public void setEnabled(boolean enabled) {
+                    super.setEnabled(false); // never enable
+                }
+            };
```
swalker2m: Very nice.  I guess this text field is only used to suggest editing and show the "Select ->" line? That is, once the band is selected it gets swapped out for a normal text field.

tpolecat: That seems to be the case, yes.

PR #411, opened 2015-05-20
REL-2331: create target detail editor instances for each panel

swalker2m:
This is take 2 of an attempt to fix the problem with disappearing target detail editors.  The issue is that distinct`TargetDetailPanel` instances cannot share the same instances of the individual `TargetDetailEditor`s.

This is probably not the most elegant solution but it does correctly handle both the enabled state and fix the problem with shared detail panels.


cquiroz:
:+1:

PR #412, opened 2015-05-20
OCSADV-353: Calculate peak and S/N values for multiple CCDs correctly

fnussber:
For GMOS all calculations are done for each CCD individually in order to take CCD specific properties into account. When reporting max flow and S/N values we have to take all those CCDs into account.

### DIFF
--- a/bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gmos/GmosRecipe.java
+++ b/bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gmos/GmosRecipe.java
@@ -403,7 +403,7 @@ public final class GmosRecipe implements ImagingArrayRecipe, SpectroscopyArrayRe
             }
         }
 
-        return new SpcChartData(type, title, "Wavelength (nm)", yAxis, JavaConversions.asScalaBuffer(data));
+        return new SpcChartData(type, title, "Wavelength (nm)", yAxis, JavaConversions.asScalaBuffer(data).toList());
     }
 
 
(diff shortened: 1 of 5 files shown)


bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gmos/GmosRecipe.java
```diff
@@ -403,7 +403,7 @@ private static SpcChartData createGmosChart(final SpectroscopyResult[] results,
             }
         }
 
-        return new SpcChartData(type, title, "Wavelength (nm)", yAxis, JavaConversions.asScalaBuffer(data));
+        return new SpcChartData(type, title, "Wavelength (nm)", yAxis, JavaConversions.asScalaBuffer(data).toList());
```
swalker2m: For some reason I never thought to use `JavaConversions` from Java code but that seems like a good idea.  `JavaConversions` for Java code that needs to work with Scala collections and `JavaConverters` for Scala code that needs Java collections.

bundle/edu.gemini.itc/src/main/scala/edu/gemini/itc/base/Recipe.scala
```diff
@@ -4,7 +4,7 @@ import java.util
 
 import edu.gemini.itc.shared._
 
-import scala.collection.JavaConversions
+import scala.collection.JavaConversions._
```
swalker2m: Here in Scala code I think we're better off with the explicit pimping conversions of `JavaConverters`.

swalker2m:
Merging on behalf of Florian for the test release today.

PR #414, opened 2015-05-20
REL-2307: add F2 filters to QPT options

swalker2m:
This PR adds F2 filter options to the QPT (and groups FP Units and Filters for convenience).


sraaphorst:
Very nice. I didn't think it was going to be quite that easy, but glad to see that it was! :+1:

tpolecat:
:+1:

PR #415, opened 2015-05-20
REL-2246: Workaround for hdiutil issue on Mavericks.

sraaphorst:
There is a bug on Mavericks which (sometimes) prevents `hdiutil` from making `dmg` files of sizes over 100 MB. This is a workaround for that for now.

See https://discussions.apple.com/thread/5667409


cquiroz:
Thanks for digging that up :+1:

PR #416, opened 2015-05-21
OCSADV-355: SDSS filters must default to AB mag system

cquiroz:
Update to the target editor, when a magnitude band belonging to the AB system is selected, set the magnitude system to AB. Else, select the Vega system
A few code simplifications and warnings removal are also included

### DIFF
--- a/bundle/edu.gemini.shared.skyobject/src/main/java/edu/gemini/shared/skyobject/Magnitude.java
+++ b/bundle/edu.gemini.shared.skyobject/src/main/java/edu/gemini/shared/skyobject/Magnitude.java
@@ -7,7 +7,10 @@ import edu.gemini.spModel.core.Wavelength;
 import edu.gemini.spModel.core.Wavelength$;
 
 import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Comparator;
+import java.util.List;
 
 /**
  * Celestial object brightness information.  Magnitudes are relative to
@@ -45,6 +48,18 @@ public final class Magnitude implements Comparable, Serializable {
         AP(None.INTEGER, new Some<>("apparent"))
         ;
 
+        public static final List<Band> AB_BANDS;
+
+        static {
+            List<Band> bands = new ArrayList<>();
+            bands.add(u);
+            bands.add(g);
+            bands.add(r);
+            bands.add(i);
+            bands.add(z);
+            AB_BANDS = Collections.unmodifiableList(bands);
+        }
+
         /**
          * A Comparator of magnitude bands based upon the associated
          * wavelength.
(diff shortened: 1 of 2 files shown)


bundle/edu.gemini.shared.skyobject/src/main/java/edu/gemini/shared/skyobject/Magnitude.java
```diff
@@ -45,6 +48,18 @@
...
+
+        static {
+            List<Band> bands = new ArrayList<>();
+            bands.add(u);
+            bands.add(g);
+            bands.add(r);
+            bands.add(i);
+            bands.add(z);
+            AB_BANDS = Collections.unmodifiableList(bands);
+        }
```
swalker2m: What about putting the default system as a field of the band itself?  That would force you to think about it whenever you add a new band.

swalker2m:
This raises a question about AGS.  We treat r, R, and UC as interchangeable for the purpose of determining and comparing magnitudes.  If the units, as it were, aren't even the same though  can we really do that?

bundle/edu.gemini.shared.skyobject/src/main/java/edu/gemini/shared/skyobject/Magnitude.java
krumpi: I was thinking also if we should let people change the system if bands already have one

cquiroz: I was thinking also if we should let people change the system if bands already have one

cquiroz:
So a magnitude 7 in AB is not the same as a 7 in VEGA? If so you are probably right.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/MagnitudeEditor.java
```diff
@@ -98,11 +97,18 @@ public static synchronized String formatBrightness(Magnitude mag) {
...
         // new magnitude band.
         private final ActionListener changeBandAction = new ActionListener() {
+
             @Override public void actionPerformed(ActionEvent e) {
                 final Magnitude.Band newBand = (Magnitude.Band) cb.getSelectedItem();
                 if (newBand == null) return;
                 if (newBand == MagEditRow.this.band) return;
                 changeBand(MagEditRow.this.band, newBand);
+                // OCSADV-355 If the bands is one on the AB list, switch the magnitude system to AB or else to VEGA
+                if (Magnitude.Band.AB_BANDS.stream().anyMatch(b -> b.equals(newBand))) {
```
swalker2m: Hadn't notice Java `Stream` before.  Nice.

cquiroz:
I'll merge to let @swalker2m continue

PR #417, opened 2015-05-22
OCSADV-357: minor target changes

swalker2m:
This PR contains a handful of label changes

> * Change the Sidereal target Epoch units from "JD" to "years" 
> * Change the Sidereal target proper motion label from Delta to mu 
> * Change the parallax units from arcsec to mas 

and one slightly more significant change

> * For comets and asteroids, I don't think we should pre-fill the Epoch or TP with 2000, since those are in Julian days and will have values around 2.45e6 

Here the epoch is being stored in `NonSiderealTarget` and shared by the two subclasses, `ConicTarget` and `NamedTarget`.  Presumably the epoch for `NamedTarget`, if relevant at all, should continue to be 2000.0 years while the two epoch values for `ConicTarget` now default to 0.0 JD.

The old `ITarget` hierarchy is fairly poor and this change continues the tradition.  In particular, rather than move the epoch field down to the subclasses it makes a call to an abstract method in the constructor of `NamedTarget` to get the default value.  Also, there were places where we assume epoch is not `null` and places where we explicitly check for `null`.  Instead I've tried to make it impossible to set the epoch value to `null` at all.  Of course `Epoch` values themselves are mutable and I've done nothing to fix that.  In short, the new target model will be a huge improvement.

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details/SiderealDetailEditor.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details/SiderealDetailEditor.scala
@@ -26,10 +26,10 @@ final class SiderealDetailEditor extends TargetDetailEditor(ITarget.Tag.SIDEREAL
   }
 
   val props = NumericPropertySheet[HmsDegTarget](Some("Motion"), _.getTarget.asInstanceOf[HmsDegTarget],
-    Prop("∆ RA",     "mas/year", _.getPM1),
-    Prop("∆ Dec",    "mas/year", _.getPM2),
-    Prop("Epoch",    "JD",       _.getEpoch),
-    Prop("Parallax", "arcsec",   _.getParallax),
+    Prop("µ RA",     "mas/year", _.getPM1),
+    Prop("µ Dec",    "mas/year", _.getPM2),
+    Prop("Epoch",    "years",    _.getEpoch),
+    Prop("Parallax", "mas",      _.getParallax),
     Prop("RV",       "km/sec",   _.getRV)
   )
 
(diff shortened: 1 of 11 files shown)


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details/SiderealDetailEditor.scala
```diff
@@ -26,10 +26,10 @@ final class SiderealDetailEditor extends TargetDetailEditor(ITarget.Tag.SIDEREAL
   }
 
   val props = NumericPropertySheet[HmsDegTarget](Some("Motion"), _.getTarget.asInstanceOf[HmsDegTarget],
-    Prop("∆ RA",     "mas/year", _.getPM1),
-    Prop("∆ Dec",    "mas/year", _.getPM2),
-    Prop("Epoch",    "JD",       _.getEpoch),
-    Prop("Parallax", "arcsec",   _.getParallax),
+    Prop("µ RA",     "mas/year", _.getPM1),
```
cquiroz: Why the change to `µ`?

cquiroz: Oh, I see it was on the requirements

cquiroz:
Good :+1:

swalker2m:
I added a few fixes to the coordinate parameters units.  In particular

* we assume `arcsec` everywhere when working with parallax but we are now to display and store `mas`.  I updated the old target model to convert to/from arcsec as necessary

* for non-sidereal targets we would store JD for epochs but incorrectly set the units to "years".  Now we just always work with JD and I fix the units in the `To2015B` migration

:pray:

PR #418, opened 2015-05-22
OCSADV-332: default to AB for sloan filters in brightness parser

swalker2m:
> Make sure that all Sloan filters (u, g, r, i, z) are imported with System = AB.

Also added a `defaultSystem` field to the `Band` itself for convenience and removed the explicit `AB_BANDS` listing.


cquiroz:
Do you think the `Magnitude.Band` change could affect the deserialization of existing proposals?

swalker2m:
Yes, it breaks serialization but we don't worry about that between semesters.  We rely on migration via XML which should still be okay.

cquiroz:
:+1:

PR #419, opened 2015-05-22
REL-2122: GMOS-S DTA X limit check

swalker2m:
A new Phase 2 check.

> REQUIREMENT: IF Instrument == GMOS-S AND CCD Manufacturer == Hamamatsu AND YBINNING=1 AND DTA X Offset < -4 THEN ERROR: The GMOS-S Hamamatsu allowed DTA X range is -4 to +6 for Ybin=1 
> REQUIREMENT: IF Instrument == GMOS-S AND CCD Manufacturer == Hamamatsu AND YBINNING>1 AND DTA X Offset < -2 THEN ERROR: The GMOS-S Hamamatsu allowed DTA X range is -2 to +6 for Ybin=2 or 4

### DIFF
--- a/bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/gmos/GmosRule.java
+++ b/bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/gmos/GmosRule.java
@@ -16,6 +16,7 @@ import edu.gemini.skycalc.Offset;
 import edu.gemini.spModel.config2.Config;
 import edu.gemini.spModel.config2.ConfigSequence;
 import edu.gemini.spModel.config2.ItemKey;
+import edu.gemini.spModel.core.Site;
 import edu.gemini.spModel.data.YesNoType;
 import edu.gemini.spModel.gemini.gmos.GmosCommonType.*;
 import edu.gemini.spModel.gemini.gmos.GmosNorthType.DisperserNorth;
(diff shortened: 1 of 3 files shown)


bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/gmos/GmosRule.java
```diff
@@ -431,6 +431,35 @@ public IConfigMatcher getMatcher() {
...
+            } else {
+                return null;
+            }
+        }
+
+        public IConfigMatcher getMatcher() {
+            return (config, step, elems) -> {
+                final InstGmosCommon gmos = (InstGmosCommon) elems.getInstrument();
+
+                return (gmos.getSite() == Site.SET_GS) &&
```
cquiroz: Can you do `==` comparison? shouldn't `equals` be preferred in Java?

cquiroz:
:+1:

bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/gmos/GmosRule.java
swalker2m: You're right thanks for pointing that out.  I'm under the influence of Scala.  It happens to work in this case because they refer to the same constants but that wasn't the intention.  I'll fix it before I merge.

PR #420, opened 2015-05-22
OCSADV-341: Show saturation warnings in OT

fnussber:
In order to improve the usefulness of the ITC results in the OT we added some saturation warnings which previously were only shown on the web pages.

Unfortunately this PR shows, that there is still quite some repetition of code between different instruments, that could/should be addressed at some point, like e.g. some of the warnings are the same for the different instruments. So, there's still room for improvement for the ITC code, but I guess it's quite a bit better than 4 months ago.. 

image

### DIFF
--- a/bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/acqcam/AcqCamRecipe.java
+++ b/bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/acqcam/AcqCamRecipe.java
@@ -4,6 +4,10 @@ import edu.gemini.itc.base.*;
 import edu.gemini.itc.operation.*;
 import edu.gemini.itc.shared.*;
 import edu.gemini.spModel.core.Site;
+import scala.collection.JavaConversions;
+
+import java.util.ArrayList;
+import java.util.List;
 
 /**
  * This class performs the calculations for the Acquisition Camera used for imaging.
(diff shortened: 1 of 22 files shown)


bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/acqcam/AcqCamRecipe.java
```diff
@@ -105,8 +108,17 @@ private ImagingResult calculateImaging(final AcquisitionCamera instrument) {
...
-        return ImagingResult.apply(p, instrument, IQcalc, SFcalc, peak_pixel_count, IS2Ncalc);
+        final List<ItcWarning>  warnings = warningsForImaging(instrument, peak_pixel_count);
+        return ImagingResult.apply(p, instrument, IQcalc, SFcalc, peak_pixel_count, IS2Ncalc, warnings);
+
+    }
 
+    // TODO: some of these warnings are similar for different instruments and could be calculated in a central place
+    private List<ItcWarning> warningsForImaging(final AcquisitionCamera instrument, final double peakPixelCount) {
+        final double wellLimit = 0.8 * instrument.getWellDepth();
+        return new ArrayList<ItcWarning>() {{
```
cquiroz: I don't remember this notation, is it new in Java 1.8?

fnussber: This has been around for a while. I think 1.6(?). However I just noticed that this way of setting up classes comes actually with some drawbacks. I like it for it's compact notation, but it creates an additional anonymous class.

cquiroz:
I can't follow the logic that well without knowing ITC in detail but I have no issues with the code

PR #421, opened 2015-05-23
OCSADV-358: Pass magnitude system down to ITC (Vega vs AB vs Jy).

fnussber:
Taking advantage of @swalker2m latest changes in https://github.com/gemini-hlsw/ocs/pull/418 which added the magnitude system to the magnitudes made fixing this straight forward.

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcTable.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcTable.scala
@@ -194,16 +194,16 @@ trait ItcTable extends Table {
 
   private def extractSource(target: ITarget, c: ItcUniqueConfig): String \/ SourceDefinition = {
     for {
-      (mag, band)     <- extractSourceMagnitude(target, c.config)
-      srcProfile      <- parameters.spatialProfile
-      srcDistribution <- parameters.spectralDistribution
-      srcRedshift     <- parameters.redshift
+      (mag, band, system) <- extractSourceMagnitude(target, c.config)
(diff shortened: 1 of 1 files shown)


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcTable.scala
```diff
@@ -222,33 +222,39 @@ trait ItcTable extends Table {
...
-        case Magnitude.Band.Q  => (b, WavebandDefinition.Q)
+      val value  = mag.getBrightness
+      // TODO: unify band definitions from spModel core and itc shared so that we don't need this translation anymore
+      val system = mag.getSystem match {
+        case Magnitude.System.Vega  => BrightnessUnit.MAG
+        case Magnitude.System.AB    => BrightnessUnit.ABMAG
+        case Magnitude.System.Jy    => BrightnessUnit.JY
+      }
+      val band   = mag.getBand match {
+        case Magnitude.Band.u  => WavebandDefinition.U
```
cquiroz: I think we discussed this before but this looks so much like the old band to new band conversion in `ModelConverters`, do we need separate models in ITC?

cquiroz:
:+1:

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcTable.scala
fnussber: No, the magnitude in ITC should go away asap and be replaced with the one from the core model. However there were some little details which kept me from doing it yet, but those magnitudes and bands etc definitely need to be unified.

PR #422, opened 2015-05-23
OCSADV-361: IFU left/right options instead of IFU-2 in web app

fnussber:
IFU-2 is not yet supported, instead show options for IFU left/right blue/red.


cquiroz:
:+1:

PR #423, opened 2015-05-24
OCSADV-348: Use observing peer or GN or GS for ITC calls

fnussber:
Some users don't have an observing peer defined, for those we fall back to GN or GS (in this order) as the peer to be used for ITC service calls.

This is a bit lame, but turns out it is reasonably difficult to check for a "fastest to reach" peer in a more or less reliable way. I think this deserves a better and more general solution but for the June release I am afraid this is as good as it gets.

(I just had a discussion with Andy and we decided to run some load tests against gnodbtest2 on Tuesday to get  a better feel for the CPU load that we're creating with the ITC service calls. Depending on how that works out we might also distribute the calls randomly between GN and GS.)

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcTable.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcTable.scala
@@ -9,7 +9,7 @@ import edu.gemini.pot.sp.SPComponentType._
 import edu.gemini.shared.skyobject.Magnitude
 import edu.gemini.spModel.`type`.DisplayableSpType
 import edu.gemini.spModel.config2.{Config, ConfigSequence, ItemKey}
-import edu.gemini.spModel.core.{Peer, Wavelength}
+import edu.gemini.spModel.core.{Site, Peer, Wavelength}
 import edu.gemini.spModel.guide.GuideProbe
 import edu.gemini.spModel.obs.context.ObsContext
 import edu.gemini.spModel.obscomp.SPInstObsComp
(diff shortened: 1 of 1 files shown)


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcTable.scala
```diff
@@ -166,6 +166,11 @@ trait ItcTable extends Table {
 
   }
 
+  // lacking a simple way to decide on the "closest" (i.e. "fastest to reach") peer we talk to the observing peer
+  // if defined and GN or GS otherwise, if no site is defined you're out of luck and the ITC tables will stay empty
+  protected def itcServicePeer: Option[Peer] =
+    ObservingPeer.get orElse ObservingPeer.peerFor(Site.GN) orElse ObservingPeer.peerFor(Site.GS) orElse None
```
cquiroz: Is this the server to talk to? If so shouldn't we fallback to something? I presume most users are on the northern hemisphere so defaulting to GN would make sense

fnussber: Yes, this is the service peer, so to say. I *think* defaulting to GN is what we're doing here by calling ```ObservingPeer.peerFor(Site.GN)``` ? So if that doesn't work (and there also is no peer defined for GS), I don't think there's anything else we can do.

cquiroz: Ok, then I'm misunderstanding the line, :+1:

fnussber: I will wait with merging this for a final OK from @swalker2m .

swalker2m: This seems good to me though ending with `orElse None` is superfluous. That's what you get anyway if none of the others were defined. Should be harmless though. I'll go ahead and merge this one so it can be included in the test release today.

PR #424, opened 2015-05-24
OCSADV-351: Validate sky aperture >= 1 and Gaussian FWHM >= 0.1

fnussber:
This moves some general validation logic down into the server so that it is executed for web requests and OT requests alike. Also, it deals with the problem noted in OCSADV-362 caused by exceptions thrown by ```require``` in the constructor of ```GaussianSource``` when entering values for FWHM in the target editor.


cquiroz:
Looks very straightforward :+1:

PR #425, opened 2015-05-24
OCSADV-355: SDSS filters must default to AB mag system

cquiroz:
This PR builds on PR #416 adding the Magnitude system to the conversions from the old to the new model, making AGS and Manual GS aware of the `MagnitudeSystem` when creating magnitudes
Additionally a few usages of the `Magnitude` constructors without a `Magnitude.System` were fixed to explicitly set it when appropriate in the catalog, phase2 templates and the template editor bundles

### DIFF
--- a/bundle/edu.gemini.shared.skyobject/src/main/java/edu/gemini/shared/skyobject/Magnitude.java
+++ b/bundle/edu.gemini.shared.skyobject/src/main/java/edu/gemini/shared/skyobject/Magnitude.java
@@ -122,7 +122,7 @@ public final class Magnitude implements Comparable, Serializable {
      */
     public Magnitude(Band band, double brightness) {
         //noinspection unchecked
-        this(band, brightness, None.INSTANCE, System.DEFAULT);
+        this(band, brightness, None.INSTANCE, band != null ? band.defaultSystem : System.DEFAULT);
     }
 
     /**
@@ -143,7 +143,7 @@ public final class Magnitude implements Comparable, Serializable {
      * @param error error in measurement
      */
     public Magnitude(Band band, double brightness, double error) {
-        this(band, brightness, new Some<>(error), System.DEFAULT);
+        this(band, brightness, new Some<>(error), band.defaultSystem);
     }
 
     /**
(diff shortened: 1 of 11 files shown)


fnussber:
:+1:

cquiroz:
Changed `Magnitude` constructors to use `band.defaultSystem` as much as possible

bundle/edu.gemini.shared.skyobject/src/main/java/edu/gemini/shared/skyobject/Magnitude.java
```diff
@@ -124,7 +122,7 @@
      */
     public Magnitude(Band band, double brightness) {
         //noinspection unchecked
-        this(band, brightness, None.INSTANCE, System.DEFAULT);
+        this(band, brightness, None.INSTANCE, band != null ? band.defaultSystem : System.DEFAULT);
```
swalker2m: If `band` is `null` the constructor will throw an `IllegalArgumentException`which just as well could be a `NullPointerException`.  In other words, I think just `band.defaultSystem` without the `null` check is okay.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/template/TemplateParametersEditor.scala
```diff
@@ -309,7 +309,7 @@ class TemplateParametersEditor(shells: java.util.List[ISPTemplateParameters]) ex
 
     object MagnitudesPanel extends ColumnPanel {
       def magRow(band: Magnitude.Band): Row = {
-        lazy val zero = new Magnitude(band, 0.0)
+        lazy val zero = new Magnitude(band, 0.0, band.defaultSystem)
```
swalker2m: We can go back and remove the extra argument now with your update to the `Magnitude` constructors.  It's fine either way though.

cquiroz: Somehow I think is better to make the system explicit on the constructor call

swalker2m:
:+1:

PR #427, opened 2015-05-25
REL-2338: smartgcal servlet startup issue

swalker2m:
The smartgcal servlet was not starting because of a `ClassNotFound` exception.  The class in question was actually part of the smartgcal servlet bundle so I think it is just a matter of getting the bnd parameters set correctly.


cquiroz:
:+1:

PR #428, opened 2015-05-26
REL-2333: target component layout updates

swalker2m:
This PR contains updates to the target editor layout based on Andy's idea to move the magnitude table to the right of the coordinate editor.  These are all cosmetic changes whose goal is to make it it possible to see all widgets with a narrower OT window, place all coordinate information in a single column and make it possible to view more magnitude settings without scrolling. Unfortunately to achieve a nice layout I had to extract widgets from sub-editors so that they could be placed in a single grid bag.  My apologies go out especially to Rob for destroying that part of the code.  Perhaps there is a better way of achieving the same goal?

antares

jupiter

vesta


cquiroz:
The UI looks good :+1:

PR #429, opened 2015-05-26
REL-2339: make target table scroll horizontally

swalker2m:
This PR gives the OT target component a horizontal scrollbar when necessary (as opposed to shrinking all the columns to fit in the available space).

horizontal


tpolecat:
:+1:

PR #430, opened 2015-05-26
REL-2340: change visitor instrument setup time

swalker2m:
As requested, estimated visitor instrument setup time drops to 10 minutes (though unfortunately we express setup time in seconds as a double so that's 600.0 seconds).


tpolecat:
:+1:

PR #431, opened 2015-05-26
REL-2341: change shrinking order in target details panel

swalker2m:
This PR offers a small adjustment to the target detail panel layout.  In particular, when space is tight and sub-panels begin to shrink we should prefer to first conceal the "Source" panel, then "Magnitudes" and finally "General".  Before the shrinking order was exactly the opposite.


cquiroz:
:+1:

PR #432, opened 2015-05-28
OCSADV-362: Improve event handling on source editor.

fnussber:
Some mess-up with the event handling / repaint / whatever cycle caused the text edit widgets on the source editor to loose focus. An earlier "fix" to work around this by resetting the focus after every key stroke did not work as expected on Windows and made those edit boxes unusable.

I settled on a more streamlined event handling that in particular only changes the visibility of some of the data entry panels (implemented using ```NumericPropertySheet```) if necessary, this seems to avoid taking away the focus from the text edit box which avoids the need of resetting the focus on the widget and therefore should hopefully also resolve this issue on Windows.


swalker2m:
I believe the other target component editors extend `ReentrancyHack` for this reason and in the `edit` method they surround code that modifies widgets with

````scala
nonreentrant {
  widget.setText("foo")
}
````

for example.  Perhaps the same approach would work here?

fnussber:
The problem was not a loop in event calls. The problem was that hiding/showing panels caused the text edit widgets to loose their focus when people were entering text. In so far I don't see how ```ReentrancyHack``` would have helped here.

swalker2m:
Ok.  Let's merge this then.  :+1:

PR #433, opened 2015-05-28
OCSADV-344: Implement GeMS Strehl approximation for GSAOI

fnussber:
GSAOI ITC calculations have been fully functional for a while but the Strehl value was missing. I was not able to get the precise average Strehl value from the Mascot / AGS code in a simple way due to my lack of understanding the code and also because I was very hesitant to change that code even in the slightest bit at such a late stage.

Following Bryan's suggestion I implemented an approximation for the GeMS performance based on this table http://www.gemini.edu/sciops/instruments/gems/gems-performance which should be good enough for now. We can follow up with using the more precise Strehl values at a later stage.

I am aware that this change is a bit a late comer, so I am ok in case people think we should not put that into the release. Andy is going to test it and I will certainly play around with it a bit more, too. We can probably decide tomorrow if it's worth the risk or not.

*HOWEVER* , if we add this, we would have full support for imaging and the only missing bits would be GNIRS and NIFS... just saying.

image

### DIFF
--- a/bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gems/Gems.java
+++ b/bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gems/Gems.java
@@ -168,7 +168,7 @@ public class Gems implements AOSystem {
                         break;
 
                     default:
-                        throw new IllegalArgumentException("GeMS does not support band " + strehlBand);
+                        throw new IllegalArgumentException("Current ITC implementation for GeMS does not support band " + strehlBand);
                 }
                 break;
 
--- a/bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gems/Gems.java
+++ b/bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gems/Gems.java
@@ -166,6 +166,9 @@ public class Gems implements AOSystem {
                                 return 0.12;
                         }
                         break;
+
+                    default:
+                        throw new IllegalArgumentException("GeMS does not support band " + strehlBand);
                 }
                 break;
 
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcTable.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcTable.scala
@@ -130,30 +130,25 @@ trait ItcTable extends Table {
   protected def doServiceCall(peer: Peer, inputs: String \/ ItcInputs): Future[ItcService.Result] = inputs match {
 
     case -\/(err) =>
-      Future {
-        ItcError(err).left
-      }
+      Future.successful(ItcError(err).left).andThen { case _ => updateResults() }
 
     case \/-(inp) =>
       // Do the service call
-      ItcService.calculate(peer, inp).
-
-        // whenever service call is finished notify table to update its contents
-        andThen {
-        case _ => Swing.onEDT {
-
-          // notify table of data update while keeping the current selection
-          restoreSelection {
-            this.peer.getModel.asInstanceOf[AbstractTableModel].fireTableDataChanged()
-          }
-
-          // make all columns as wide as needed
-          SequenceTabUtil.resizeTableColumns(this.peer, this.model)
-        }
-
-      }
+      ItcService.calculate(peer, inp).andThen       { case _ => updateResults() }
 
     }
+
+  // whenever service call is finished notify table to update its contents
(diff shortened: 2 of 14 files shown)


bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gems/Gems.java
```diff
@@ -166,6 +166,9 @@ public double getAOCorrectedFWHM(boolean doThrow) {
                                 return 0.12;
                         }
                         break;
+
+                    default:
+                        throw new IllegalArgumentException("GeMS does not support band " + strehlBand);
```
cquiroz: What band this refers to? Gems certainly uses a few bands to select Guide Stars

cquiroz:
The GEMS AGS code is certainly hairy, thanks for not modifying it, we should go back and see how feasible is to integrate it to ITC. Not knowing much about ITC I can't comment on how easy/hard that would be

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcTableModel.scala
```diff
@@ -150,41 +151,58 @@ sealed trait ItcTableModel extends AbstractTableModel {
...
+    Column("Strehl",          (c, i, r) => gems(i),                      tooltip = "Estimated Strehl and band")
+  )
+  val results = List(
+    Column("Peak",            (c, i, r) => imgPeakPixelFlux(r),          tooltip = ItcTableModel.PeakPixelTooltip),
+    Column("S/N Single",      (c, i, r) => imgSingleSNRatio(r)),
+    Column("S/N Total",       (c, i, r) => imgTotalSNRatio (r))
+  )
+
+  def gems(i: String \/ ItcInputs): Option[String] = i.toOption.map { inputs =>
+    val gems = inputs.ins.asInstanceOf[GsaoiParameters].gems
```
cquiroz: I got confused about this `inputs` val and the `ins` method with  the class param `inputs` perhaps using different names would make the code clearer

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcTable.scala
```diff
@@ -113,57 +113,47 @@ trait ItcTable extends Table {
...
-      case -\/(err) => Future { ItcError(err).left }
-      case \/-(res) => res
-    }
+    } yield ItcInputs(src, new ObservationDetails(method(srcFrac), analysis), cond, tele, ins)
 
-  }
+  protected def doServiceCall(peer: Peer, inputs: String \/ ItcInputs): Future[ItcService.Result] = inputs match {
+
+    case -\/(err) =>
+      Future {
```
cquiroz: `Future.successful`? No need to spawn a thread I think

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gems/Gems.java
swalker2m: I don't really know the context here but since we're talking about J, H, K perhaps it should say GSAOI instead?  I'm not sure what "GeMS does not support band" means in general though.

fnussber: For reasons unknown to me the current ITC implementation only supports J, H and K bands in some of the calculations. This error should actually not be thrown since on the web page you can only select one of those bands and with the current implementation in the OT it is also guaranteed to have only one of these bands being sent down into the ITC code. However I thought it would be a good idea to add an exception for the future so that it is explicit what's going on in case somebody sends other bands down because the fall back would have been to get an exception with the message "Gems can not be used with IQ any" which is arguably even worse. I think.

fnussber:
I think doing the GeMS / ITC integration "right" wouldn't be too difficult since all the bits and pieces are there, but we would need to do some refactoring to make it more accessible if you just want to calculate a Strehl for a given set of stars. I could not figure out how to do that in a way that did not make it necessary to touch quite some code.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcTable.scala
fnussber: Right, that's wasteful here. I didn't know you can create a Future like that.

swalker2m:
:+1:

PR #434, opened 2015-05-28
OCSADV-363: Disable ITC panels for GNIRS and NIFS

fnussber:
GNIRS and NIFS ITC support is not fully finished yet and will be added at a later stage. This PR simply hides the ITC panels for those instruments.

ITC support for Michelle and TReCS has also been disabled. Potentially ITC for these legacy instruments will never be supported in the OT; they are (still) supported in the web app.

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcPanel.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcPanel.scala
@@ -120,7 +120,7 @@ class ItcImagingPanel(val owner: EdIteratorFolder) extends ItcPanel {
     case SPComponentType.INSTRUMENT_FLAMINGOS2  => true
     case SPComponentType.INSTRUMENT_GMOS        => true
     case SPComponentType.INSTRUMENT_GMOSSOUTH   => true
-    case SPComponentType.INSTRUMENT_GSAOI       => false // will add support at a later stage
+    case SPComponentType.INSTRUMENT_GSAOI       => true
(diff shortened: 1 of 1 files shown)


cquiroz:
:+1:

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcPanel.scala
```diff
@@ -152,11 +152,11 @@ class ItcSpectroscopyPanel(val owner: EdIteratorFolder) extends ItcPanel {
...
-    case SPComponentType.INSTRUMENT_NIFS        => true
+    case SPComponentType.INSTRUMENT_GNIRS       => false // will add support at later stage
+    case SPComponentType.INSTRUMENT_MICHELLE    => false // may or may not be supported in OT at some point
+    case SPComponentType.INSTRUMENT_NIFS        => false // will add support at later stage
     case SPComponentType.INSTRUMENT_NIRI        => true
-    case SPComponentType.INSTRUMENT_TRECS       => true
+    case SPComponentType.INSTRUMENT_TRECS       => false // may or may not be supported in OT at some point
     case _                                      => false
   }
 }
```
swalker2m: The "Source" information in the target component is still valid information I suppose even if ITC isn't going to run for instrument x, yet there's no point in asking for it if we aren't going to use it.  Could/should the "Source" target editor be hidden as well?  This might be a question for Andy/Bryan.

andrewwstephens: Targets may be copied between observations using different instruments, so probably best to retain and show the Source information in all cases rather than have it lost or hidden some of the time.

PR #436, opened 2015-05-28
REL-2239: Bug fix for GMOS rules

fnussber:
The rule did not take into account the detector manufacturer when looking up the limits and always returned the limits for EE2.


cquiroz:
:+1:

PR #437, opened 2015-05-29
OCSADV-365: Ignore Y and u bands for ITC calculations.

fnussber:
One of my shameful moments when integrating the ITC was that I haven't (yet) unified the wavebands used in ITC and the core. Of course - as always - this has now come back to haunt me.

For reasons unknown ITC does not support Y and u bands. In order to deal with this I have translated those bands to other, nearby bands when passing information down from the OT to the ITC. However the preferable way seems to be to ignore those bands when deciding on the magnitude and use other magnitude information.

This PR reflects this change by ignoring magnitudes defined in Y or u and give an error message if no other magnitudes are defined.

And yes, this needs to be cleaned up and those wavelength band definitions need to be unified. But there are only so many hours in a day.


swalker2m:
Merging on behalf of Florian for test release 2015B-test.1.9.1.

PR #438, opened 2015-05-29
REL-2351: target detail enabled state fix

swalker2m:
This PR fixes a subtle bug in enabled state management for the target detail panel.  Ordinarily enabled state updating is ignored when setting the enabled state to the same value as it was before in order to avoid updating the table of "previous" enabled state values.  You can see this in `OtItemEditor`:

````java
    protected void updateEnabledState(boolean enabled) {
        if (enabled != isEnabled()) {
            setEnabled(enabled);
            updateEnabledState(getWindow().getComponents(), enabled);
        }
    }
````

In the case of `EdCompTargetList` we were ignoring this step in the overridden `updateEnabledState` method.

Unfortunately this fix is not very satisfying.  Enabled state management in the OT needs to be redone from scratch.  The table of all all widget's previous enabled state values is very fragile.

### DIFF
--- a/bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
+++ b/bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
@@ -118,22 +118,25 @@ public final class EdCompTargetList extends OtItemEditor<ISPObsComponent, Target
     }
 
     @Override protected void updateEnabledState(boolean enabled) {
-        super.updateEnabledState(enabled);
+        if (enabled != isEnabled()) {
+            setEnabled(enabled);
+            updateEnabledState(getWindow().getComponents(), enabled);
+
+            // Update enabled state for all detail widgets.  The current editor
+            // will have already been updated by the super.updateEnabledState so
+            // update the others.
+            for (final TargetDetailEditor ed : _w.detailEditor.allEditorsJava()) {
+                if (_w.detailEditor.curDetailEdiorJava().forall(cur -> cur != ed)) {
+                    updateEnabledState(new Component[]{ed}, enabled);
+                }
+            }
+        }
 
         final TargetEnvironment env = getDataObject().getTargetEnvironment();
         _w.tag.setEnabled(enabled && env.getBase() != _curPos);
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java [hunk 1]
```diff
@@ -118,22 +118,25 @@ public void textBoxAction(TextBoxWidget tbwe) {
...
-        super.updateEnabledState(enabled);
+        if (enabled != isEnabled()) {
+            setEnabled(enabled);
+            updateEnabledState(getWindow().getComponents(), enabled);
+
+            // Update enabled state for all detail widgets.  The current editor
+            // will have already been updated by the super.updateEnabledState so
+            // update the others.
+            for (final TargetDetailEditor ed : _w.detailEditor.allEditorsJava()) {
+                if (_w.detailEditor.curDetailEdiorJava().forall(cur -> cur != ed)) {
```
sraaphorst: `curDetailedEdior`: we should probably refactor and fix this horrible spelling error at some point. (Should be Edi__t__or, not Edior.)

Also, is there a reason that we don't do something like:
`final Component[] comps = _w.detailEditor.allEditorsJava().filterNot(_ != _w.detailEditor.curDetailEditorJava()).toArray();`
`updateEnabledState(comps, enabled);`
instead of the `for` loop, or am I missing something?

sraaphorst: (Excuse the horrible Java / Scala mishmash there, but I didn't want to have to look up the exact details of the method names and calls.)

swalker2m: Thanks, I fixed the comment and the misspelling of "editor".  I didn't filter because it is a Java collection.  It looks like it can be easily converted to a Java 8 `Stream` and then filtered

````java
_w.detailEditor.allEditorsJava().stream().filter(ed -> _w.detailEditor.curDetailEditorJava().forall(cur != ed))
````
or something ... but I think I'll be conservative since we're nearing the release date.

sraaphorst:
:+1:

PR #439, opened 2015-05-29
OCSADV-200-R: preparing for OT integration

swalker2m:
When I started looking into OT / VCS integration I realized I would need a few changes in the new vcs code.  The intention is for the OT to mostly work with the API provided by `edu.gemini.sp.vcs2.Vcs` which has methods for performing checkouts, adds, and syncs.  Unfortunately though,

1. It was missing a `version` method to simply fetch the current `VersionMap` from the remote peer.

2. It provided no way to cancel an ongoing sync or checkout operation.  Since a `VcsAction` is simply an `EitherT[Task, VcsFailure, A]`, I could just extract the `Task`, fork it into a new thread and call `runAsyncInterruptibly`.  I wanted more control over where the interrupt would happen though to make sure that I don't somehow half-commit any updates. I used the same API as `runAsyncInterruptibly` and just check an `AtomicBoolean` at key places.

Also, this PR

* pimps in a `forkAsync` method for `VcsAction` to simplify running the action in a separate thread

* changes the `VcsServer` constructor so that there's no need to have a `VcsLog` unless you also need to make a `SecureVcsService`

This may be terrible.  Suggestions for improvement would be extremely welcome.

### DIFF
--- a/bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs2/Vcs.scala
+++ b/bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs2/Vcs.scala
@@ -3,7 +3,8 @@ package edu.gemini.sp.vcs2
 import edu.gemini.pot.sp.{SPNodeKey, ISPFactory, ISPProgram}
 import edu.gemini.pot.sp.version._
 import edu.gemini.shared.util.VersionComparison.{Conflicting, Same, Newer}
-import edu.gemini.sp.vcs2.ProgramLocation.{LocalOnly, Neither, Remote}
+import edu.gemini.sp.vcs2.ProgramLocation.Remote
+import edu.gemini.sp.vcs2.ProgramLocationSet.{LocalOnly, Neither, RemoteOnly}
 import edu.gemini.sp.vcs2.VcsFailure.{Cancelled, IdClash, NeedsUpdate}
 import edu.gemini.sp.vcs.log.VcsEventSet
 import edu.gemini.spModel.core.{Peer, SPProgramID}
@@ -86,8 +87,8 @@ class Vcs(user: VcsAction[Set[Principal]], server: VcsServer, service: Peer => V
     * and merge them with the local program if necessary.  The action returns
     * `true` if it results in an updated local program; `false` otherwise.
     */
-  def pull(id: SPProgramID, peer: Peer, cancelled: AtomicBoolean): VcsAction[Boolean] =
-    pull0(id, Client(peer), cancelled).map(_.localUpdate)
+  def pull(id: SPProgramID, peer: Peer, cancelled: AtomicBoolean): VcsAction[PullResult] =
+    pull0(id, Client(peer), cancelled).map(_.localUpdate.fold(LocalOnly, Neither))
 
   /** Provides an action that pushes local changes to the remote peer, merging
     * them with the remote version of the program if necessary.  When performed,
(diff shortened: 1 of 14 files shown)


bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs2/Vcs.scala
```diff
@@ -66,40 +78,48 @@ class Vcs(user: VcsAction[Set[Principal]], server: VcsServer, service: Peer => V
     def filter(eval: MergeEval): Boolean = eval.localUpdate
 
     def update(f: ISPFactory, p: ISPProgram, eval: MergeEval): VcsAction[Unit] =
-      eval.plan.merge(f, p)
+      checkCancel(cancelled) >> eval.plan.merge(f, p)
```
cquiroz: What's `>>`, is it a `map`?

cquiroz:
Looks good code-wise though I can't really follow the details
A general comment about the usage of MonadTransformers like `EitherT`. It is not a very intuitive concept (at least to me), what would be the benefit over using nested `for` loops?

bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs2/Vcs.scala
swalker2m: > Sequentially compose two actions, discarding any value produced by the first, like sequencing  operators (such as the semicolon) in imperative languages.


`a >> b` is used to get the effect of `a` but without needing the output for `b`.  Here I think it should be the same as `a >>= _ => b` or

````scala
for {
  _  <- a
  b0 <- b
} yield b0
````
It's especially appropriate since `checkCancel` produces a `VcsAction[Unit]`.

Comment (swalker2m) [inline bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs2/Vcs.scala]: Another way to think of it maybe is that `>>` is just `flatMap` where the function passed is setup to ignore its parameter.

**Comment by swalker2m** [comment]
The transformers allow you to avoid some boilerplate.  Think of `EitherT` as transforming a monad, in this case `Task`, to add short-circuit failure to computations that use it.  Unfortunately I'm no expert though and this may not be the best way to do this, I'm not sure.  I think Rob favors a different technique these days, using so-called free monads.

File: bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs2/Vcs.scala (hunk 1 above)
Comment (swalker2m) [inline bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs2/Vcs.scala]: For a lot of these things, by the way, the equivalent documentation like the one I quoted is usually clearer in Haskell docs than the scant (if any) doc in scalaz.  For example all you get in scalaz `BindOps` is the method signature.

Comment (tpolecat) [inline bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs2/Vcs.scala]: Just FYI the `>>` is a historical accident from Haskell. The more general operation `*>` is [necessarily] equivalent and works for all applicative functors. There is also `<*` which discards the result on the right.

Comment (cquiroz) [inline bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs2/Vcs.scala]: This documentation maybe useful
http://eed3si9n.com/learning-scalaz/scalaz-cheatsheet.html

`>>` appears there thou is not that clear what it does from the exmaple

**Comment by tpolecat** [comment]
@cquiroz here's an `EitherT` example that shows the kind of boilerplate you eliminate:

```scala
import scalaz._, Scalaz._, effect.IO
type Error = String

// An IO operation that tries to read and return an Int
val getInt: IO[Error \/ Int] = IO {
  Console.readLine match {
    case null => "EOF".left
    case s    => 
      try s.toInt.right 
      catch { case _: NumberFormatException => "Parse Error".left }
  }
}

// Read and add two ints
val addTwo: IO[Error \/ Int] =
  for {
    e1 <- getInt
    e2 <- e1.fold(e => IO(e.left), _ => getInt) // propagate first error
  } yield e1.fold(_.left, a => e2.fold(_.left, b => (a + b).right))

// Now with EitherT
val addTwoT: EitherT[IO, Error, Int] =
  for {
    a <- EitherT(getInt)
    b <- EitherT(getInt)
  } yield a + b

// Equivalent to addTwo above, but much much simpler
val addTwo2 = addTwoT.run
```

cquiroz:
Thanks
would `EitherT` work if in this example we had `b <- EitherT(getLong)` where `def getLong: IO[Error \/ Long]`

swalker2m:
Yes, seems like it should work if you change the type of `addTwoT` to `EitherT[IO, Error, Long]`.

tpolecat:
Yes. `EitherT[M,E,A]` is a monad for any fixed `M` and `E` (`IO` and `Error` in this case). The transformer  fixes `M` and `E` and pulls `\/`'s "fast-fail, don't let me observe the error" semantics up to the top.

PR #440, opened 2015-06-01
OCSADV-366: Unify ITC and core model wavelength bands.

fnussber:
This is a PR I put out there mainly for discussion.

The aim of OCSADV-366 is to get rid of the ITC specific definition of wavelength bands in ```WavebandDefinition``` and replace them with ```MagnitudeBand```s from the core model.

In order to be able to do this I would like to make the following changes, which I put out here for review:

* Bands not only have a center point but also a width. The width of the bands are for example used in ITC to check if filters overlap with source spectras.

* I would also like to make the center and width values mandatory (instead of ```Option```) since a band without center point and width doesn't really make much sense. I spoke with Andy about this, and he agrees that from a domain viewpoint this makes sense and suggested that for apparent magnitudes ```AP``` we could use the values of the visible band ```V```.

* Some of the center values have been changed to the values from ITC. Since currently those values are not used anywhere except for sorting the bands this shouldn't have any impact on the OCS part, however it will keep the results calculated by ITC unchanged.

In addition to those changes I have a question:

What are the ```G``` and ```Z``` bands in the core model ```MagnitudeBand```? When we replace the "old" with the "new" bands in the OT, this means that two additional bands will show up in the OT. Talking with Andy in order to find out the widths of those bands he wasn't quite sure what those bands would refer to and was wondering, if those should actually be ```_g``` and ```_z``` or g' and z'. The center points indicate in this direction and also the fact that when translating those bands to the old model they are mapped to g' and z'. So my question here is, what do we need those bands for and if we can replace ```G``` and ```Z``` with the prime bands ```_g``` and ```_z```? Can anyone shed some light on this?

### DIFF
--- a/bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/MagnitudeBand.scala
+++ b/bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/MagnitudeBand.scala
@@ -1,15 +1,19 @@
 package edu.gemini.spModel.core
 
-sealed abstract class MagnitudeBand private (val name: String, val wavelengthMidPointNm: Option[Int], val description: Option[String], val defaultSystem: MagnitudeSystem) extends Product with Serializable {
+sealed abstract class MagnitudeBand private (val name: String, val center: Wavelength, val width: Wavelength, val description: Option[String], val defaultSystem: MagnitudeSystem) extends Product with Serializable {
 
-  private def this(name: String, wavelengthMidPointNm: Int) =
-    this(name, Some(wavelengthMidPointNm), None, MagnitudeSystem.VEGA)
+  private def this(name: String, center: Wavelength, width: Wavelength) =
+    this(name, center, width, None, MagnitudeSystem.VEGA)
 
-  private def this(name: String, wavelengthMidPointNm: Int, description: String) =
-    this(name, Some(wavelengthMidPointNm), Some(description), MagnitudeSystem.VEGA)
+  private def this(name: String, center: Wavelength, width: Wavelength, description: String) =
+    this(name, center, width, Some(description), MagnitudeSystem.VEGA)
 
-  private def this(name: String, wavelengthMidPointNm: Int, description: String, defaultMagnitudeSystem: MagnitudeSystem) =
-    this(name, Some(wavelengthMidPointNm), Some(description), defaultMagnitudeSystem)
+  private def this(name: String, center: Wavelength, width: Wavelength, description: String, defaultMagnitudeSystem: MagnitudeSystem) =
+    this(name, center, width, Some(description), defaultMagnitudeSystem)
+
+  val start: Wavelength = center + width / 2
+
+  val end:   Wavelength = center - width / 2
 
 }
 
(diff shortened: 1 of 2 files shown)


bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/MagnitudeBand.scala
```diff
@@ -1,15 +1,19 @@
...
+    this(name, center, width, Some(description), MagnitudeSystem.VEGA)
 
-  private def this(name: String, wavelengthMidPointNm: Int, description: String, defaultMagnitudeSystem: MagnitudeSystem) =
-    this(name, Some(wavelengthMidPointNm), Some(description), defaultMagnitudeSystem)
+  private def this(name: String, center: Wavelength, width: Wavelength, description: String, defaultMagnitudeSystem: MagnitudeSystem) =
+    this(name, center, width, Some(description), defaultMagnitudeSystem)
+
+  val start: Wavelength = center + width / 2
+
+  val end:   Wavelength = center - width / 2
```
swalker2m: I think it's great to have start and end but should the width be of type `Wavelength`?  The actual `Wavelength` that that value represents has nothing to do with the band.  Perhaps it is just a raw `Double` I don't know.

I suppose we could instead define the start and end and compute the center and width.  It's just about the same thing but start and end feel more fundamental somehow.  I admit this is maybe just nit picking.

swalker2m:
Actually, I think this is exactly the kind of thing we should be doing before adding new features.  :+1: I don't know about `G` and `Z`.  I'm guessing you're right and they should just be `_g` and `_z`.  Maybe @cquiroz will have some insight from working with catalogs?  If we end up deciding they should be `_g` and `_z` we'll need migration code for 2016A.

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/MagnitudeBand.scala
andrewwstephens: I'm not sure what is meant by wavelength "type", but the width should certainly have wavelength units.  For example, the J-band filter has a center at 1.25 microns and a width of 0.18 microns.

fnussber: I am with @andrewwstephens here: the ```Wavelength``` is just a length which can (currently) be measured in microns or nanometers.

Is this a good moment to take another attempt at promoting http://www.squants.com/ (or some such  thing)? Besides from being able to define stuff like ```1.nm``` and convert between units I also think such a library would be useful for implementing some future tasks where we (I) need to support different units for values in the source definitions.

Regarding how to define the bands: They are defined as midpoint and width in the literature (at least that's true for the Sloan filters which I've looked up), so I am somewhat leaning to use that to define the bands, however if it's preferable to define them by start and end that's ok, too.

cquiroz:
I don't think `G` or `Z` are provided via VoTable from either UCAC4 or PPMXL

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/MagnitudeBand.scala
cquiroz: I'd be ok considering the use of `squants`, I like the idea of having compilation errors when adding unit-incompatible objects. I checked and it seems to depend on `spire` and `json4s`. I think we discussed about using `spire` some time ago?

tpolecat: Yeah I think it would be good to give `squants` a try. It relies on `spire` so we would be pulling that in as well. Erik (who wrote spire) suggested squants a while back.

fnussber:
Ok, I am going to close this one for now. From the discussion so far I'll take the following points away:

* I'll replace ```G``` and ```Z``` with ```_g``` and ```_z``` respectively; assuming they're indeed the same.

* I'll proceed with the changes to ```MagnitudeBand``` as outlined here.

* I'll add ```squants``` and use its length type instead of ```Wavelength``` in the ITC code. That will give us a better idea of what it does to the code and potential pros and cons.

I'll do every step in a separate PR, so it will be easier to follow the changes, which is probably in particular important for the changes due to ```squants```. More to come..

PR #442, opened 2015-06-03
REL-2362: UCAC4 at Gemini fix

sraaphorst:
Previously, UCAC4 at Gemini queried gnodbtest2, which was not the desired behaviour.

Separated UCAC4 at Gemini into two catalogs (one for north, which now queries gnodb, and one for south, which now queries gsodb).


swalker2m:
:+1:

PR #443, opened 2015-06-03
REL-2364 fix A/QR mixup; disable id lookup

tpolecat:
There were two issues reported in [REL-2364](http://jira.gemini.edu:8080/browse/REL-2364):
- The `A` element was mistakenly used for `QR` for comets. This has been fixed.
- Horizons object ids are not stable; they can change and be reassigned as the database is updated. This led to the very strange behavior seen when old targets were pasted into new programs and their (now incorrect) ids were used for "update" operations. This has been fixed by always doing a name lookup. This means that the disambiguation dialog will always appear.


swalker2m:
:+1:

PR #444, opened 2015-06-04
OCSADV-368: Replacing home grown formatter with String.format

fnussber:
Replacing a home grown number formatter with standard functionality.


arturog8m:
Good idea to replace this. We need a JIRA ticket though so whenever we get to our testing stage, we don't forget to ask for people to test this out. Could you add one?

bundle/edu.gemini.itc.web/src/main/java/edu/gemini/itc/web/html/GnirsPrinter.java
```diff
@@ -126,11 +110,6 @@ private void writeSpectroscopyOutput(final UUID id, final GnirsSpectroscopyResul
     }
 
     private String gnirsToString(final Gnirs instrument, final Parameters p) {
-        //Used to format the strings
-        final FormatStringWriter device = new FormatStringWriter();
-        device.setPrecision(3);  // Two decimal places
-        device.clear();
-
 
         String s = "Instrument configuration: \n";
```
sraaphorst: Is there a reason you build the string like this instead of using a `StringBuilder` as you do in other places? Doesn't seem efficient.

bundle/edu.gemini.itc.web/src/main/java/edu/gemini/itc/web/html/HtmlPrinter.java
```diff
@@ -67,69 +60,53 @@ public static String printParameterSummary(final SourceDefinition sdp) {
...
-        StringBuffer sb = new StringBuffer();
-
-        // This object is used to format numerical strings.
-        FormatStringWriter device = new FormatStringWriter();
-        device.setPrecision(2);  // Two decimal places
-        device.clear();
+        final StringBuilder sb = new StringBuilder();
 
         sb.append("Output:\n<LI>Spectra ");
         switch (pdp.getPlotLimits()) {
```
sraaphorst: Do we really need a `switch` when this is really just overcomplicating an `if`...`else` or even a ternary expression?

sraaphorst: Do we really need a switch when this is really just overcomplicating an `if`...`else` or even a ternary expression?

(Sorry if any of these have repeated... I thought I made a mistake.)

bundle/edu.gemini.itc.web/src/main/java/edu/gemini/itc/web/html/HtmlPrinter.java
```diff
@@ -185,13 +156,11 @@ public static String printParameterSummary(final Altair altair) {
     }
 
     public static String printSummary(final Gems gems) {
-        final FormatStringWriter device = new FormatStringWriter();
-        device.setPrecision(3);
-        String s = "r0(" + gems.getWavelength() + "nm) = " + device.toString(gems.getr0()) + " m\n";
-        s += "Average Strehl = " + device.toString(gems.getAvgStrehl() * 100) + "%\n";
+        String s = String.format("r0(" + gems.getWavelength() + "nm) = %.3f m\n", gems.getr0());
```
sraaphorst: Again, why not a `StringBuilder` here instead of all these string `+=` inefficient ops?

bundle/edu.gemini.itc.web/src/main/java/edu/gemini/itc/web/html/MichellePrinter.java
```diff
@@ -134,42 +119,35 @@ private void writeImagingOutput(final ImagingResult result) {
...
         if (!instrument.polarimetryIsUsed()) {
-            _println(CalculatablePrinter.getTextResult(result.is2nCalc(), result.observation(), device));
+            _println(CalculatablePrinter.getTextResult(result.is2nCalc(), result.observation()));
         } else {
             _println("Polarimetry mode enabled.\n");
-            final String result2 = CalculatablePrinter.getTextResult(result.is2nCalc(), result.observation(), device);
+            final String result2 = CalculatablePrinter.getTextResult(result.is2nCalc(), result.observation());
             final String delims = "[ ]+";
             final String[] tokens = result2.split(delims);
             for (int i = 0; i < tokens.length; i++) {
```
sraaphorst: Oi... this seems truly horrible, although I'm not exactly what sure what this code is doing.
Is there really no better way to do this than modifying later entries in the array being iterated over?

bundle/edu.gemini.itc.web/src/main/java/edu/gemini/itc/web/html/MichellePrinter.java
```diff
@@ -198,7 +174,7 @@ private void writeImagingOutput(final ImagingResult result) {
...
```
sraaphorst: Same as previous comment.

bundle/edu.gemini.itc.web/src/main/java/edu/gemini/itc/web/html/NifsPrinter.java
```diff
@@ -104,12 +97,6 @@ private void writeSpectroscopyOutput(final UUID id, final SpectroscopyResult res
     }
 
     private String nifsToString(final Nifs instrument) {
-        //Used to format the strings
-        final FormatStringWriter device = new FormatStringWriter();
-        device.setPrecision(3);  // Two decimal places
-        device.clear();
-
-
         String s = "Instrument configuration: \n";
```
sraaphorst: Same as previous `StringBuilder` comments.

bundle/edu.gemini.itc.web/src/main/java/edu/gemini/itc/web/html/NiriPrinter.java
```diff
@@ -52,44 +52,35 @@ private void writeSpectroscopyOutput(final UUID id, final SpectroscopyResult res
...
         if (result.aoSystem().isDefined()) {
             _println(HtmlPrinter.printSummary((Altair) result.aoSystem().get()));
         }
 
         if (!result.observation().isAutoAperture()) {
-            _println("software aperture extent along slit = "
-                    + device.toString(result.observation().getApertureDiameter()) + " arcsec");
+            _println(String.format("software aperture extent along slit = %.2f arcsec", result.observation().getApertureDiameter()));
         } else {
             switch (result.source().getProfileType()) {
```
sraaphorst: Again, do we really need a `switch` here for two cases each with one statement and no default?

bundle/edu.gemini.itc.web/src/main/java/edu/gemini/itc/web/html/TRecsPrinter.java
```diff
@@ -174,9 +150,9 @@ private String trecsToString(final TRecs instrument, final Parameters p) {
         s += "Spatial Binning: 1\n";
```
sraaphorst: Again, `s +=` instead of `StringBuilder`?

bundle/edu.gemini.itc.web/src/main/java/edu/gemini/itc/web/html/MichellePrinter.java
fnussber: Agreed, this ain't pretty, but cleaning that up was not in the focus of this PR (replacing ```FormatStringWriter``` class).

fnussber:
Regarding ```StringBuilder``` and String concatenation I refer to (one of the many) sources on the net, e.g. http://stackoverflow.com/questions/14927630/java-string-concat-vs-stringbuilder-optimised-so-what-should-i-do

Basically: The compilers since JDK1.6 do a pretty good job in optimizing these things for you, so in general you can do whatever seems more readable / simpler without suffering from a penalty hit. (Exceptions are also outlined in the referenced link, basically when it comes to concatenations in loops, which applies here in one place, I think.)

Actually, IntelliJ will even suggest *not* to use ```StringBuilder``` for example for this
```Java
    private String printTeleParametersSummary(final ImagingResult result) {
        final StringBuilder sb = new StringBuilder();
        sb.append("Telescope configuration: \n");
        sb.append("<LI>");
        sb.append(result.telescope().getMirrorCoating().displayValue());
        sb.append(" mirror coating.\n");
        sb.append("<LI>wavefront sensor: gems\n");
        return sb.toString();
    }
```
and suggest to use String concatenation instead for better readability. 

More relevant in regards to efficiency was probably replacing ```StringBuffer``` (synchronized) with ```StringBuilder``` (unsynchronized) everywhere, which I did.

I guess I am guilty of not doing it everywhere the same way though, so I might massage the code a bit more.

tpolecat:
I don't think it's worth getting too hung up on this unless it's a known hotspot.

bundle/edu.gemini.itc.web/src/main/java/edu/gemini/itc/web/html/HtmlPrinter.java
fnussber: I added the default case here, which probably makes sense.

bundle/edu.gemini.itc.web/src/main/java/edu/gemini/itc/web/html/NiriPrinter.java
fnussber: In case you're worried about performance then I don't think it makes any/a big difference. Sometimes I think ```switch``` statements with enums are just easier to read than ```if - else if - else``` cascades (close to Scala's ```match - case``` statements), even for simple cases, but I guess it doesn't really matter here. So, ```if``` it is :smile:

PR #445, opened 2015-06-04
REL-2366: Changed message as per Bryan's request.

sraaphorst:
Bryan requested some changes to the auto-GS fail dialog.


tpolecat:
:+1: (assuming the messages look ok in the UI)

PR #446, opened 2015-06-04
REL-2260, REL-2360, REL-2361: old calendar widget bugs

swalker2m:
There are a number of related bugs that appear when selecting calendar dates for non-sidereal coordinate lookups.  This code uses an ancient, complex calendar widget that should be retired.  The fundamental problem is that calling `setDate(d)` on a `JCalendar` instance immediately followed by `getDate` does not produce the value `d` unless it happens to be a date in the *same month* that is being displayed.

The "solution" was to examine the actions that take place in response to the user clicking to advance the month and doing corresponding actions in `setDate`.  Finally `getDate` was changed to simply return the date that was previously set.

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details/ValidAtEditor.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details/ValidAtEditor.scala
@@ -109,10 +109,9 @@ abstract class ValidAtEditor[A <: ITarget](empty: A) extends TelescopePosEditor 
     node = node0
     nonreentrant {
       val nst = spt.getTarget.asInstanceOf[NonSiderealTarget]
-      Option(nst.getDateForPosition).foreach { d =>
-        calendar.setDate(d)
-        timeConfig.setTime(d)
-      }
+      val d   = Option(nst.getDateForPosition) | new java.util.Date
+      calendar.setDate(d)
+      timeConfig.setTime(d)
     }
   }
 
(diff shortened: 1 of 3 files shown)


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details/ValidAtEditor.scala
```diff
@@ -109,10 +109,9 @@ abstract class ValidAtEditor[A <: ITarget](empty: A) extends TelescopePosEditor
     node = node0
     nonreentrant {
       val nst = spt.getTarget.asInstanceOf[NonSiderealTarget]
-      Option(nst.getDateForPosition).foreach { d =>
-        calendar.setDate(d)
-        timeConfig.setTime(d)
-      }
+      val d   = Option(nst.getDateForPosition) | new java.util.Date
```
swalker2m: I made this change because otherwise a newly created non-sidereal target would display the date for the last non-sidereal target that happened to be selected.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details/ValidAtEditor.scala
```diff
@@ -128,15 +127,16 @@ abstract class ValidAtEditor[A <: ITarget](empty: A) extends TelescopePosEditor
...
+  def dateTime: HorizonsIO[Date] =
+    HorizonsIO.delay {
+      val cal = Calendar.getInstance(UTC)
+      cal.setTimeZone(calendar.getTimeZone)
+      cal.setTime(calendar.getDate)
+      cal.set(Calendar.HOUR_OF_DAY, timeConfig.textDoc.getHoursField)
+      cal.set(Calendar.MINUTE, timeConfig.textDoc.getMinutesField)
+      cal.set(Calendar.SECOND, timeConfig.textDoc.getSecondsField)
+      cal.getTime
+    }
```
swalker2m: I wrapped the entire thing here or else clicking on the magnifying glass would use the date at the time that the editor panel was created instead of the date displayed in the widget.

It is a bit worrisome that we now consult the widget in a non-event-loop thread and (elsewhere) update it in a non-event-loop thread.

tpolecat: `// yolo`

tpolecat:
This is terrible. :+1:

PR #447, opened 2015-06-04
REL-2362: Make the two new UCAC4 at Gemini (North|South) catalog entries appear in old locations

sraaphorst:
Split entry in AstroCat.xml to remove the old singular UCAC4 at Gemini entry and replace it with the two UCAC4 entries.

This causes the two entries to appear in the three places required by Bryan's comment on the JIRA task, namely:
1. TPE Catalog -> Catalogs menu
2. Guide Star Selection dialog (manual GS search)
3. Catalog Navigator -> My Catalogs menu


swalker2m:
:+1:

PR #448, opened 2015-06-06
OCSADV-369: Clean up GnirsParameters

fnussber:
In preparation for integrating GNIRS ITC into the OT this PR focuses on replacing the stringly typed values in ```GnirsParameters``` into properly typed values using the relevant types from the sp model and turning the Java class into a Scala case class. This also includes reusing read noise and slit width values from the sp model. And all necessary changes to the web pages.

As a bonus I got rid of some classes that didn't do much more than put together the file names of some dat files in an overly complicated way (see ```[Long|Short]Camera[Red|Blue]Optics.java```) and did some very minor cleanup things in the ```Gnirs``` class. This class could certainly use some more clean-up, but this is as far as I wanted to go as part of this PR.


cquiroz:
Looks good :+1:

PR #449, opened 2015-06-08
REL-2372: Prepare PIT 2016A

cquiroz:
This PR contains version and semester changes for PIT 2016A. 
There are a few tab and raw types fixes to old code as well,
Also I created a PITLauncher app similar to OCS's TestLauncher to start the PIT from inside IDEA


fnussber:
:+1:

PR #451, opened 2015-06-08
REL-2308: Add Flamingos-2 K-long filter to the 16A PIT

cquiroz:
This PR includes the new flamingos2 filter in the phase1 model and decision tree


fnussber:
:+1:

PR #453, opened 2015-06-09
REL-2355: Change DSSI/TEXES instrument options in the 2016A PIT

cquiroz:
Enable Texes on the list of supported instruments and do not delete texes proposals on upconversion

### DIFF
--- a/bundle/edu.gemini.model.p1/src/test/scala/edu/gemini/model/p1/dtree/RootSpec.scala
+++ b/bundle/edu.gemini.model.p1/src/test/scala/edu/gemini/model/p1/dtree/RootSpec.scala
@@ -8,7 +8,7 @@ class RootSpec extends SpecificationWithJUnit {
 
   "The Root Spec" should {
     "include Gsaoi" in {
-      val root = new Root(Semester(2015, A))
+      val root = new Root(Semester(2016, A))
       root.choices must contain(Instrument.Gsaoi)
     }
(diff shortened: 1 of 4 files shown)


bundle/edu.gemini.model.p1/src/test/scala/edu/gemini/model/p1/dtree/RootSpec.scala
```diff
@@ -12,8 +12,8 @@ class RootSpec extends SpecificationWithJUnit {
       root.choices must contain(Instrument.Gsaoi)
     }
     "include Texes" in {
-      val root = new Root(Semester(2015, A))
-      root.choices must not contain (Instrument.Texes)
+      val root = new Root(Semester(2016, A))
```
fnussber: I wonder if all tests here should/could be changed to 2016A? I guess the aim here is to check that for the upcoming semester all relevant instruments are available (or not available).

fnussber:
:+1:

bundle/edu.gemini.model.p1/src/test/scala/edu/gemini/model/p1/dtree/RootSpec.scala
cquiroz: That's right, these test may not make too much sense these days as the way to include/hide instruments has changed, but for completness they should all use 2016A

PR #454, opened 2015-06-09
REL-2288: Update PIT keyword list

cquiroz:
Small PR with the addition of the 'Star formation' keyword to the model. A legacy file `Keywords.txt` was removed as it was not referenced anywhere


fnussber:
:+1:

PR #455, opened 2015-06-10

cquiroz:
This PR adds support for creating Phase 1 proposals that use phoenix.

It includes changes to the model, UI support and pdf creation. It also has a test linking the phase 1 model with the OCS model, tests that were desired but couldn't be easily done with the separation of ocs2 and ocs1.5

Aditionally there are set of the usual cosmetic improvements


fnussber:
I am not an expert with blueprints but I didn't see anything suspicious in the code and since this has been out there for more than 2 days now I would say: go for it.  :+1:

PR #456, opened 2015-06-10
OCSADV-366A: Unify ITC and core model bands.

fnussber:
This PR follows up the previous round of discussions and implements the first set of changes.

* Aligning the ITC and core model bands without replacing them in the ITC model yet.
* Remove the non-existing G and Z bands and replace them with Sloan g' and z' bands.

I was trying to introduce as little changes to the ```VoTableParser``` and just changed the behavior so that by default the upper case bands take precedence but if no upper case match is found we also check for lower case. I.e., the default and PPMXL case will be that "umag", "rmag" and "imag" will be translated to U, R and I bands while "gmag" and "zmag" will translate to g' and z'. @cquiroz : Please have a critical look on that. This PR also includes the bug fixes int the ```Magnitude``` constructors, which were necessary to make the tests work.

Note that for the testcase from ```votable.xml``` this leads to the behavior that u,g,r,i,z bands from the file will be translated to U,g',R,I,z'. This is a bit strange, but is consistent with the original behavior and passes all tests.

Regarding the conversion for 2015B to 2016A data that @swalker2m mentioned I think we're in luck because currently the magnitudes of the old model are stored in the xml. For the case of G and Z bands they have always been translated to g' and z' when converting the new into the old model (and then left unchanged when converting back to the new model!) Therefore I think we never made G and Z bands persistent and therefore don't need to worry about conversion (?).


bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/Magnitude.scala
```diff
@@ -5,20 +5,20 @@ import Scalaz._
 
 case class Magnitude(value: Double, band: MagnitudeBand, error: Option[Double], system: MagnitudeSystem) {
 
-  /** Secondary constructor defaulting to `MagnitudeSystem.VEGA`. */
+  /** Secondary constructor. */
   def this(value: Double, band: MagnitudeBand, error: Double) =
-    this(value, band, Some(error), MagnitudeSystem.VEGA)
+    this(value, band, Some(error), band.defaultSystem)
```
cquiroz: These changes are good, this makes them consistent with the old spmodel magnitude changes done a while ago

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/MagnitudeBand.scala
```diff
@@ -18,36 +22,40 @@ object MagnitudeBand {
...
+  case object R  extends MagnitudeBand("R", nm(  670), nm( 100), "red")
+  case object I  extends MagnitudeBand("I", nm(  870), nm( 100), "infrared")
+  case object Y  extends MagnitudeBand("Y", nm( 1020), nm( 120))
+  case object J  extends MagnitudeBand("J", nm( 1250), nm( 240))
+  case object H  extends MagnitudeBand("H", nm( 1650), nm( 300))
+  case object K  extends MagnitudeBand("K", nm( 2200), nm( 410))
+  case object L  extends MagnitudeBand("L", nm( 3760), nm( 700))
+  case object M  extends MagnitudeBand("M", nm( 4770), nm( 240))
+  case object N  extends MagnitudeBand("N", nm(10470), nm(5230))
+  case object Q  extends MagnitudeBand("Q", nm(20130), nm(1650))
```
cquiroz: So we got rid of `G` and `Z`. Is it 100% confirmed that we don't use them? Are they on the old `Magnitude.Band`?

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/MagnitudeBand.scala
```diff
@@ -1,15 +1,19 @@
...
-    this(name, Some(wavelengthMidPointNm), Some(description), MagnitudeSystem.VEGA)
+  private def this(name: String, center: Wavelength, width: Wavelength, description: String) =
+    this(name, center, width, Some(description), MagnitudeSystem.VEGA)
 
-  private def this(name: String, wavelengthMidPointNm: Int, description: String, defaultMagnitudeSystem: MagnitudeSystem) =
-    this(name, Some(wavelengthMidPointNm), Some(description), defaultMagnitudeSystem)
+  private def this(name: String, center: Wavelength, width: Wavelength, description: String, defaultMagnitudeSystem: MagnitudeSystem) =
+    this(name, center, width, Some(description), defaultMagnitudeSystem)
+
+  val start: Wavelength = center + width / 2
```
cquiroz: Are overridden operator's precedence as you would expect, meaning `/` over `+`?

tpolecat: Yes, this is part of the lexical structure of the language.

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableParser.scala
```diff
@@ -58,8 +58,14 @@ sealed trait MagnitudesFilter {
   def ignoredMagnitudeField(v: FieldId): Boolean = false
   // Indicates if a magnitude is valid
   def validMagnitude(m: Magnitude): Boolean = !(m.value.isNaN || m.error.exists(_.isNaN))
-  // Find what band the field descriptor should represent
-  def findBand(id: FieldId, band: String): Option[MagnitudeBand] = MagnitudeBand.all.find(_.name == band)
+
+  // Find what band the field descriptor should represent, in general prefer "upper case" bands over "lower case" Sloan bands.
+  // This will prefer U, R and I over u', r' and i' but will map "g" and "z" to the Sloan bands g' and z'.
+  def findBand(id: FieldId, band: String): Option[MagnitudeBand] =
```
cquiroz: Is it correct to prefer `R` over `r'`. In AGS we want it the other way around though this is at the level of VoTable, it would be great to confirm that a catalog won't expose both `R` and `r'`

tpolecat:
I think it would be worth confirming that there will be no data migration issues.

cquiroz:
If `G` and `Z` were not on the old `Magnitude.Band` I guess there is no risk of migration problems but it is worth to check
We also have code to import targets, right? Could it be affected?

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/MagnitudeBand.scala
fnussber: They are *not* part of ```Magnitude``` in the old model. I can double/triple check with Andy again, but I am pretty sure there are no ```G``` and ```Z``` bands.

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableParser.scala
fnussber: I think if a catalog has ```R``` and ```r'``` we're sort of out of luck. How would you know which one is the right one? Andy thinks that we will have to implement specific rules for mapping the magnitudes for each and every catalog. It's a good question which bands should be preferred by default, though. This implementation tries not to change the current behavior, but it's probably not the most consistent way to deal with this issue.

fnussber:
@tpolecat Would it be sufficient to (manually) export/import some proposals with an old and a new versoin? Is there any other place where magnitude bands are imported/exported that I am not aware of?

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/MagnitudeBand.scala
cquiroz: double check is enough :smile:

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableParser.scala
cquiroz: We already have catalog-specific parsing code, exactly for things like having multiple magnitudes, or bogus values, whose rules are catalog-specific

cquiroz:
The PIT has a target importer but (just checked) doesn't include either `G` or `Z`

cquiroz:
I think this PR is good, we are just worried about the possible side effects of removing `G` and `Z` but I can't really find any use for them, I'd say /dev/null them

tpolecat:
Right, you removed `G` and `Z` from the core magnitude model, so if there are any old programs with these bands then they won't deserialize. Maybe I'm missing something.

fnussber:
The old model does not have ```G``` and ```Z``` bands and since we currently use the new model "on-the-fly" only but never make it persistent anywhere (at least not that I knew of) there are no instances of those two bands anywhere being serialized. (The new->old model conversion conveniently transformed ```G``` and ```Z``` to ```g``` and ```z``` and the translation old->new model left them unchanged  as ```g``` and ```z```.) So I am pretty sure we're good here.

tpolecat:
Ok, gotcha. I was confused. :+1:

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableParser.scala
fnussber: Right, so I think the question is if we want to leave this behavior as is right now (which matches the current test cases):

* Default/PPMXL: upper case band name first, then lower case (in order to match g' and z')
* UCAC4: g, r, i map to g', r', i' then follow rules for Default/PPMXL

or if we need to change that in any way? I personally think it's a bit weird that for the default case umag, gmag, rmag, imag and zmag result in standard bands (U, R, I) and Sloan bands (z', g' because there is no G and Z band) but then I am not sure if the test data in votable.xml is actually valid, because it's not PPMXL or UCAC4?

I guess either way, we can probably leave it as is and then refine those rules once we need to support other catalogs.

fnussber: Note that I added the "match lower case band name" only in order to match the test case defined by votable.xml, so another way to deal with this would also be to adapt the test data.

cquiroz: This is a good point, you are right, votable.xml is an early example and it is not coming from either PPMXL nor UCAC4. Given that it is not possible to make a generic votable parser it doesn't make sense to have that file.
You could get rid of that test or replace the file with UCAC4
Would it break the tests done using directly xml in `VoTableParserSpec`?

fnussber:
@cquiroz Ok, I finally took the time to deal with the UCAC4 vs PPMXL vs "Default" issue. The last commit gets rid of the ```DefaultFilter``` and catalogs now must specify either the UCAC4 or PPXML attributes. If none of those attributes is defined, all parsed rows will result in an ```UnknownCatalog``` error. It would be nicer to just return a single ```UnknownCatalog``` error instead of one per row, but I did not want to change the structure of the results because that would mean to change the way catalog results are combined etc. and I wasn't sure of all the potential side effects this might have. If you're ok with this solution let me know and I'll (finally ;) merge this.

cquiroz:
I think this is good, you could just delete the `votable-unknown.xml` file, nobody's using it and it could become confusing

fnussber:
There's a test case now base on ```votable-unknown.xml``` checking that the parser detects catalogs that are not supported (i.e. not UCAC4 or PPMXL).

cquiroz:
Ok, I didn't realize that

PR #457, opened 2015-06-12
REL-2287: Do not allow proposal submission without the full PI name

cquiroz:
Small PR to forbid a common case where proposals are submitted with the default PI name 'Principal Investigator'

### DIFF
--- a/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
+++ b/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
@@ -57,7 +57,7 @@ class ProblemRobot(s: ShellAdvisor) extends Robot {
   import ProblemRobot._
 
   val MaxAttachmentSize = 30 // in megabytes
-  val MaxAttachmentSizeBytes = MaxAttachmentSize*1000*1000 // kbytes as used on the phase1 backends
+  val MaxAttachmentSizeBytes = MaxAttachmentSize * 1000 * 1000 // kbytes as used on the phase1 backends
 
   // Our state
   type State = List[Problem]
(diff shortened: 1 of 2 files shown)


bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
```diff
@@ -87,7 +87,7 @@ class ProblemRobot(s: ShellAdvisor) extends Robot {
       ps.sorted
     }
 
-    private def when[A](b: Boolean)(a: => A) = if (b) Some(a) else None
+    private def when[A](b: Boolean)(a: => A) = b option a
```
fnussber: Maybe ```when``` could be replaced with the Scalaz operator ```option``` directly? It's a detail though.

fnussber:
:+1:

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
```diff
@@ -439,6 +438,12 @@ class ProblemRobot(s: ShellAdvisor) extends Robot {
           v.edit(i)
       })
 
+    private val nonUpdatedInvestigatorName = when(p.investigators.pi.firstName === "Principal" && p.investigators.pi.lastName === "Investigator") {
```
fnussber: Shouldn't this check for empty strings now?

cquiroz: There is another check that will do that, though it will make a ToDo rather than an Error

fnussber:
:+1:

PR #458, opened 2015-06-12
REL-2272: Gemini must host all Phase I backend servers

cquiroz:
Small PR to update the urls for the Phase I backends that will be now hosted at Gemini


fnussber:
:+1:

PR #461, opened 2015-06-13
REL-2375: Classical visitors list can be edited after submission

cquiroz:
This PR fixes a bug that would let users edit the list of classical visitors after a proposal has been submitted


fnussber:
:+1:

PR #462, opened 2015-06-15
REL-2254: Editing is disabled if all submissions failed

cquiroz:
This PR handles a small border case, if all the backends reject a proposal the user should be allowed to keep editing, before the model was marked as submitted
Partially submitted proposals remain as before, they get locked after submission

### DIFF
--- a/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/view/submit/SubmitView.scala
+++ b/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/view/submit/SubmitView.scala
@@ -145,7 +145,8 @@ class SubmitView(ph: ProblemRobot, newShellHandler: (Model,Option[File]) => Unit
                       val m0 = Model.proposal.set(m, psr.proposal)
                       panel.model = Some(m0)
                       saveHandler()
-                    } else if (errors.nonEmpty) {
+                    }
+                    if (errors.nonEmpty) {
                       // Show prompt with error messages
                       SwingUtilities.invokeLater(new Runnable {
                         override def run() {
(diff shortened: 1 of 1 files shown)


bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/view/submit/SubmitView.scala
```diff
@@ -135,18 +133,20 @@ class SubmitView(ph: ProblemRobot, newShellHandler: (Model,Option[File]) => Unit
...
                     val errors = psr.results.map(_.result).filter(!_.isSuccess)
-                    // Show prompt with error messages
-                    if (errors.nonEmpty) {
+
+                    if (errors.size != psr.results.size) {
+                      val m0 = Model.proposal.set(m, psr.proposal)
+                      panel.model = Some(m0)
+                      saveHandler()
+                    } else if (errors.nonEmpty) {
+                      // Show prompt with error messages
```
swalker2m: Perhaps you don't want to string these two together with `else`?  Do you not want the prompt with the errors if there is at least one error?

swalker2m: Also, I think the intent might be clearer with:

````scala
  if (psr.results.exists(_.result.isSuccess)) {
    ...
    saveHandler()
  }
````

instead of comparing sizes?

cquiroz: Good catch, I'll update both, the first is clearly a bug and the second is clearer as you suggests

cquiroz:
Updated the code according to @swalker2m suggestions

swalker2m:
:+1:

PR #464, opened 2015-06-17
REL-2377: PIT build does not work

cquiroz:
PIT Launcher breaks the OSGi dependencies and the PIT distributions don't properly start. In this PR `PITLauncher` is move into its own package to solve this problem.


swalker2m:
The OT launcher is in its own bundle for this reason too. :+1:

PR #465, opened 2015-06-17
REL-2372: Prepare PIT 2016A

cquiroz:
Small PR to make the release version of the PIT use the test backends and update for frame icon used in Linux/Windows


swalker2m:
:+1:

PR #466, opened 2015-06-17
REL-2292: PIT target type radio buttons

cquiroz:
The combobox used to select the target type has been replaced by Radio Buttons, icons remain as before.

See a screenshot:
screenshot 2015-06-17 15 51 52

### DIFF
--- a/bundle/edu.gemini.pit/src/main/java/edu/gemini/pit/ui/util/SharedIcons.java
+++ b/bundle/edu.gemini.pit/src/main/java/edu/gemini/pit/ui/util/SharedIcons.java
@@ -56,10 +56,10 @@ public interface SharedIcons {
     Icon ICON_SPINNER = new SharedIcon("spinner.gif");
     Icon ICON_SPINNER_BLUE = new SharedIcon("spinner_blue_gray.gif");
 
-    SharedIcon ICON_SIDEREAL = new SharedIcon("sidereal.png");
+    Icon ICON_SIDEREAL = new SharedIcon("sidereal.png");
     Icon ICON_SIDEREAL_DIS = new SharedIcon("sidereal_dis.png");
-    SharedIcon ICON_NONSIDEREAL = new SharedIcon("nonsidereal.png");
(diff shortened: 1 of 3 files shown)


swalker2m:
The alignment looks a bit off maybe?  Could the "Type:" label be vertically aligned with the row of buttons?  Also, perhaps a bit more space between options would make it clearer which circle goes with which option.

cquiroz:
How about this screenshot? The red line is just on the screenshot not on the actual app :smile: 
screenshot 2015-06-17 16 50 15

bundle/edu.gemini.pit/src/main/java/edu/gemini/pit/ui/util/SharedIcons.java
```diff
@@ -52,21 +56,21 @@ public void paintIcon(Component c, Graphics g, int x, int y) { }
     Icon ICON_SPINNER = new SharedIcon("spinner.gif");
     Icon ICON_SPINNER_BLUE = new SharedIcon("spinner_blue_gray.gif");
 
-	Icon ICON_SIDEREAL = new SharedIcon("sidereal.png");
+    SharedIcon ICON_SIDEREAL = new SharedIcon("sidereal.png");
     Icon ICON_SIDEREAL_DIS = new SharedIcon("sidereal_dis.png");
-	Icon ICON_NONSIDEREAL = new SharedIcon("nonsidereal.png");
-	Icon ICON_TOO = new SharedIcon("too.png");
+    SharedIcon ICON_NONSIDEREAL = new SharedIcon("nonsidereal.png");
+    SharedIcon ICON_TOO = new SharedIcon("too.png");
```
swalker2m: The screenshot looks good.  Do these need to be specifically `SharedIcon` instead of `Icon`?

cquiroz: I think that is a leftover of a bad experiment when I was trying to use icons on the radio buttons, will be removed

PR #467, opened 2015-06-18
OCSADV-155: Replace PIT target model

cquiroz:
This is the first PR on the process to port the PIT target model with the new model on `spModel.core`
In this PR the phase1 immutable `Magnitude`, `MagnitudeBand` and `MagnitudeSystem` classes are removed and the new classes are used.
The change itself is relatively simple but impacts several parts of the OCS which are converted and tested:

* PIT
* PIT Target importer
* Phase2 Skeleton servlet/factory
* ITAC

A next step is to port Coordinates-related classes


swalker2m:
The code looks fine to me but I'm wondering about the schema and the import / export, etc.  `Target.xsd` will still define `Magnitude` and `MagnitudeBand`?  Are there still mutable types for these etc.?

swalker2m:
Also, what impact is there (if any) on ITAC?

cquiroz:
There are mutable types for those, they are generated by JAXB out of the schema, they are needed when we move to the immutable model, in this case from `spModel.core`. The mutable classes are also used by ITAC

In response to the second question, ITAC uses the mutable model directly, being it more Java friendly than the immutable. ITAC was built against the new model and it seems fine, I tested imports, queue creation, web interfaces, exports, etc

swalker2m:
Should the bands continue to be defined in the schema?  If not, I suppose it complicates validation?

swalker2m:
:+1:

cquiroz:
Good question, there is a subset of allowed bands on the p1 schema, I'll check with Andy

cquiroz:
About bands, the missing bands are the SDSS and they should be fixed as part of REL-2352

PR #468, opened 2015-06-18
REL-2255: Submit view goes black

cquiroz:
After the switch to Java 1.7 the transparency method used for the submit window stopped working. This PR fixes that as shown below

Before
screenshot 2015-02-20 10 19 49

After
screenshot 2015-06-18 15 04 01


tpolecat:
Looks great. :+1:

PR #469, opened 2015-06-18
REL-2380: fix NPEs when instrument is in conflict folder

swalker2m:
When an instrument that features a parallactic angle option finds itself in a conflict folder, bad things happen.  The `ParallacticAngleControls` assume there will always be an observation to work with, but this isn't the case when in a conflict folder.  This PR takes the fact that there may not be a containing observation into account in order to avoid exceptions and strange behavior in the OT.  It isn't ideal because it will in some cases offer widgets that do nothing (and should therefore be disabled) but it avoids the worst problems and people aren't typically editing components in conflict folders anyway.

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/parallacticangle/ParallacticAngleControls.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/parallacticangle/ParallacticAngleControls.scala
@@ -48,22 +48,29 @@ class ParallacticAngleControls extends GridBagPanel with Publisher {
 
       def rebuild(): Unit = {
         contents.clear()
-        editor.foreach(e => {
-          val obs         = e.getContextObservation
-          val inst        = e.getContextInstrumentDataObject
+
+        // menu items that don't depend on the context
+        val fixedItems = RelativeTime("Now", 0) :: incrementsInMinutes.map(m => RelativeTime(s"Now + $m min", m * 60000))
+
+        // menu items that require an observation and instrument to compute
+        val instItems = for {
+          e    <- editor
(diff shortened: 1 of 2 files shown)


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/parallacticangle/ParallacticAngleControls.scala
```diff
@@ -48,22 +48,29 @@ class ParallacticAngleControls extends GridBagPanel with Publisher {
...
-            RelativeTime("Now", 0)
-          ) ++ incrementsInMinutes.map(m => RelativeTime(s"Now + $m min", m * 60000))
-        })
+          List(
+            RelativeTime(s"Now + Setup ${formatMin(setupTimeMs)}",  setupTimeMs),
+            RelativeTime(s"Now + Reacq. ${formatMin(reacqTimeMs)}", reacqTimeMs)
+          )
+        }
+
+        contents ++= instItems.getOrElse(Nil) ++ fixedItems
```
cquiroz: Is `contents` mutable? Part of scala.swing?

cquiroz:
Good :+1:

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/parallacticangle/ParallacticAngleControls.scala
swalker2m: Right:

````scala
package scala.swing
trait SequentialContainer extends scala.AnyRef with scala.swing.Container {
  override def contents : scala.collection.mutable.Buffer[scala.swing.Component]
}
````

PR #470, opened 2015-06-18
REL-529: ToO targets must not be allowed in non-ToO proposals (PIT)

cquiroz:
This PR forbids adding ToO targets unless the ToO Option has been selected. In case the user has a ToO target and changes the ToOption to None, submissions won't be allowed via the ProblemRobot

### DIFF
--- a/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Proposal.scala
+++ b/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Proposal.scala
@@ -7,6 +7,9 @@ import scala.collection.JavaConverters._
 import java.util.logging.{Logger, Level}
 import org.xml.sax.SAXException
 
+import scalaz._
+import Scalaz._
+
 object Proposal {
 
   // Lenses
@@ -48,6 +51,14 @@ object Proposal {
       }
     } filterNot (_.isEmpty)
 
+  }
+
+  // find the toOption of the proposal
+  def toOOption(p: Option[Proposal]): Option[TooOption] = p.map(proposalClass.get) match {
+    case Some(q: QueueProposalClass)         => q.tooOption.some
+    case Some(l: LargeProgramClass)          => l.tooOption.some
+    case Some(f: FastTurnaroundProgramClass) => f.tooOption.some
+    case _                                   => none
   }
 
   private val validate = Option(System.getProperty("edu.gemini.model.p1.validate")).isDefined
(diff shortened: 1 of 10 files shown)


bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Proposal.scala
```diff
@@ -50,6 +53,14 @@ object Proposal {
 
   }
 
+  // find the target of opportunity of the proposal if possible
+  def toO(p: Option[Proposal]): Option[TooOption] = p.map(proposalClass.get) match {
+    case Some(q: QueueProposalClass)         => q.tooOption.some
+    case Some(l: LargeProgramClass)          => l.tooOption.some
+    case Some(f: FastTurnaroundProgramClass) => f.tooOption.some
+    case _                                   => none
+  }
```
swalker2m: This isn't your doing nor do I think you should change anything, but I'll comment that I wish we (probably it was me) had chosen a different name for `TooOption`.  It gets confusing when mixed with Scala `Option`.

cquiroz: Agreed, it was difficult to give it a good name, on the UI it is shown as TOO, perhaps just ToO is good enough, changing it is pretty straightforward as it is an alias for the mutable class `TooOption` which cannot be changed without altering the schema (unless xjb lets you trick it somehow)

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
```diff
@@ -161,6 +161,12 @@ class ProblemRobot(s: ShellAdvisor) extends Robot {
       msg = s"""Target "${t.name}" appears to be empty."""
     } yield new Problem(Severity.Error, msg, "Targets", s.inTargetsView(_.edit(t)))
 
+    private lazy val tooTargetsAndNoActivation = for {
+      t @ TooTarget(_, _) <- p.targets
+      if Proposal.toO(p.some).exists(_ == TooOption.None)
+      msg = "ToO targets not allowed. Please remove or set the TOO Activation"
```
swalker2m: "ToO targets not allowed." seems a bit too definitive perhaps.  Maybe add "in non ToO proposals" or something.  We should also change TOO Activation to ToO Activation.  ToO is used in the [public documentation](http://www.gemini.edu/?q=node/11005).

swalker2m:
:+1:

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
cquiroz: I'll add your suggestions and merge

PR #471, opened 2015-06-19
OCSADV-200-S: Initial OT Integration + Merge Tweaks

swalker2m:
This PR was ostensibly about integrating the new sync algorithm with the OT.  The beginning of that process is included in the form of a "New Sync" button which has been placed alongside the existing but renamed "Old Sync".  This is a temporary arrangement until the old merge algorithm and related OT code are removed.  In the meantime there is some code duplication but testers will be able to try each version side by side and see how they differ.

syncbuttons

Most of the work that went into this PR is unrelated to the OT and rather deals with bug fixes and improvements for the sync algorithm.  These include:

* A bug fix in which the children of a node would be rearranged upon sync (and yet have identical version numbers).  For this an improved sort was added in `edu.gemini.sp.vcs2.SortHeuristic` which tries to combine two independently edited lists in a way that isn't too crazy.

* A merge correction that completes resurrected observations so that they don't come back to life in an invalid state and missing descendants.

* Turning off observation resurrection conflict notes, since I can't see a good way to add them and they may not be that useful anyway.  (Will consult with Andy/Bryan.)

### DIFF
--- a/bundle/edu.gemini.sp.vcs.reg/src/main/scala/edu/gemini/sp/vcs/reg/VcsRegistrar.scala
+++ b/bundle/edu.gemini.sp.vcs.reg/src/main/scala/edu/gemini/sp/vcs/reg/VcsRegistrar.scala
@@ -7,7 +7,7 @@ import edu.gemini.spModel.core.Peer
  * A service that provides persistent/mutable mapping between program ids and
  * VCS server locations.
  */
-trait VcsRegistrar {
+trait VcsRegistrar extends scala.collection.mutable.Publisher[VcsRegistrationEvent] {
   def allRegistrations: Map[SPProgramID, Peer]
   def registration(id: SPProgramID): Option[Peer]
   def register(id: SPProgramID, loc: Peer): Unit
--- a/bundle/edu.gemini.sp.vcs.reg/src/main/scala/edu/gemini/sp/vcs/reg/osgi/Activator.scala
+++ b/bundle/edu.gemini.sp.vcs.reg/src/main/scala/edu/gemini/sp/vcs/reg/osgi/Activator.scala
@@ -1,35 +1,24 @@
 package edu.gemini.sp.vcs.reg.osgi
 
-import edu.gemini.sp.vcs.reg.{VcsRegistrationSubscriber, VcsRegistrar}
+import edu.gemini.sp.vcs.reg.VcsRegistrar
 import edu.gemini.sp.vcs.reg.impl.VcsRegistrarImpl
-import edu.gemini.util.osgi.Tracker._
 
 import org.osgi.framework.{ServiceRegistration, BundleActivator, BundleContext}
-import org.osgi.util.tracker.ServiceTracker
 import java.io.File
 import java.util.logging.Logger
 
 class Activator extends BundleActivator {
   private var reg: ServiceRegistration[VcsRegistrar] = null
-  private var tracker: ServiceTracker[_,_] = null
   private var vcsReg: VcsRegistrarImpl = null
 
   override def start(ctx: BundleContext) {
     vcsReg = new VcsRegistrarImpl(Activator.getStorageFile(ctx))
     reg = ctx.registerService(classOf[VcsRegistrar], vcsReg, null)
-
-    tracker = track[VcsRegistrationSubscriber, VcsRegistrationSubscriber](ctx) { sub =>
-      vcsReg.subscribe(sub)
-      sub
-    } { vcsReg.unsubscribe }
-    tracker.open()
   }
 
   override def stop(ctx: BundleContext) {
     reg.unregister()
     reg = null
-    tracker.close()
-    tracker = null
   }
 }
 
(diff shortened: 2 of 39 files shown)


bundle/edu.gemini.sp.vcs.reg/src/main/scala/edu/gemini/sp/vcs/reg/VcsRegistrar.scala
```diff
@@ -7,7 +7,7 @@ import edu.gemini.spModel.core.Peer
  * A service that provides persistent/mutable mapping between program ids and
  * VCS server locations.
  */
-trait VcsRegistrar {
+trait VcsRegistrar extends scala.collection.mutable.Publisher[VcsRegistrationEvent] {
```
cquiroz: I didn't know the `Publisher` class, what mechanism does it use to propagate events? Are there any problems due to OSGi classloaders, if you publish and listen in different bundles?

bundle/edu.gemini.sp.vcs.reg/src/main/scala/edu/gemini/sp/vcs/reg/osgi/Activator.scala
```diff
@@ -1,35 +1,24 @@
...
 import edu.gemini.sp.vcs.reg.impl.VcsRegistrarImpl
-import edu.gemini.util.osgi.Tracker._
 
 import org.osgi.framework.{ServiceRegistration, BundleActivator, BundleContext}
-import org.osgi.util.tracker.ServiceTracker
 import java.io.File
 import java.util.logging.Logger
 
 class Activator extends BundleActivator {
   private var reg: ServiceRegistration[VcsRegistrar] = null
```
cquiroz: I think we often use `Option[ServiceRegistration[_]]` to avoid the `null`

bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs2/MergePlan.scala
```diff
@@ -35,10 +35,14 @@ case class MergePlan(update: Tree[MergeNode], delete: Set[Missing]) {
...
-  def vm(p: ISPProgram): VersionMap = {
+  def vm(p: ISPProgram): VersionMap = vm(p.getVersions)
+
+  /** Gets the `VersionMap` with any modifications required that will be made
+    * by this merge plan. */
+  def vm(vm0: VersionMap): VersionMap = {
     // Extract the updates to the VersionMap from the MergePlan.
     val vmUpdates: VersionMap = {
-      val vm0 = update.foldRight(Map.empty[SPNodeKey, NodeVersions]) { (mn, m) =>
+      val vm0 = update.sFoldRight(Map.empty[SPNodeKey, NodeVersions]) { (mn, m) =>
```
cquiroz: `sFoldRight`?

cquiroz:
I can't really follow the sync algorithms but I trust the extensive suite of tests will cover the issues, Code-wise it looks fine :+1:

bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs2/MergePlan.scala
swalker2m: I'm glad you called this out.  I've had a couple of problems with folding a `scalaz.Tree`.  At first I worked around this by providing a version of `foldRight` that evaluates its result eagerly using this signature:
````scala
  def sFoldRight[B](z: B)(f: (A, B) => B): B
````
and simply calling the provided `foldRight`:
````scala
def foldRight[B](z: => B)(f: (A, => B) => B): B 
````
For giant programs though this was still causing a stack overflow so I switched to my own simplistic `@tailrec` implementation and the problems went away.  Still it is not very satisfying that the library-provided `foldRight` chokes on large inputs.

bundle/edu.gemini.sp.vcs.reg/src/main/scala/edu/gemini/sp/vcs/reg/VcsRegistrar.scala
swalker2m: I believe the registration bundle should just need the corresponding `Subscriber` class.  I did this to replace the previous version which had a simple but custom publish/subscribe method and it seems to work fine with an OT subscriber and the registration bundle.

cquiroz: I have used the OSGi EventAdmin in the past but it is pretty archaic as OSGI APIs are often based on ancient java versions, I'm glad that there is this choice

bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs2/MergePlan.scala
cquiroz: :+1: for `@tailrec`

bundle/edu.gemini.sp.vcs.reg/src/main/scala/edu/gemini/sp/vcs/reg/osgi/Activator.scala
swalker2m: The reference doesn't escape the `Activator` itself and is guaranteed by OSGi to be defined if the `stop` method is called but still you're right, using the `Option` there would still be a bit better.

PR #472, opened 2015-06-22
REL-2352: Add SDSS filter to target bandpass list in PIT

cquiroz:
This PR adds the SDSS bandpasses (u, g, r, i, z) to the PIT and it allows the user to enter them on the UI.
The Simbad queries were expanded to support those bands too. I couldn't find good documentation on Simbad's votable format so I followed the result of a sample

http://simbad.u-strasbg.fr/simbad/sim-id?Ident=SDSS+J082354.96%2B280621.6

The Magnitude editor now looks like this:
screenshot 2015-06-22 10 04 32

### DIFF
--- a/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/catalog/Simbad.scala
+++ b/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/catalog/Simbad.scala
@@ -33,43 +33,43 @@ class Simbad private (val host:String) extends VOTableCatalog {
 
   def decode(vot:VOTable):Seq[I.Target] = for {
 
-  // In the List monad here, eventually iterating rows
-    resource <- vot.resources
-    table@Table("simbad", _, _, _) <- resource.tables
-    row <- table.data.tableData.rows
+    // In the List monad here, eventually iterating rows
+    resource                         <- vot.resources
+    table @ Table("simbad", _, _, _) <- resource.tables
+    row                              <- table.data.tableData.rows
     kvs = table.fields.zip(row)
 
     // Local find function
-    str = (s:String) => kvs.find(_._1.ucd.exists(_.toLowerCase == s.toLowerCase)).map(_._2)
-    num = (s:String) => str(s).flatMap(_.toDoubleOption)
+    str                                = (s:String) => kvs.find(_._1.ucd.exists(_.toLowerCase == s.toLowerCase)).map(_._2)
+    num                                = (s:String) => str(s).flatMap(_.toDoubleOption)
 
     // Switch to Option here to pull out data
-    epoch <- vot.definitions.map(_.cooSys.epoch).map {
-      case "J2000" => I.CoordinatesEpoch.J_2000
-      case s       => I.CoordinatesEpoch.forName(s)
-    }
-    name <- str("meta.id;meta.main")
-    ra <- num("pos.eq.ra;meta.main")
-    dec <- num("pos.eq.dec;meta.main")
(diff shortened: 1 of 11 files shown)


bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/catalog/Simbad.scala
```diff
@@ -29,47 +29,54 @@ object Simbad extends Catalog with App {
...
-    // TODO: more passbands
-    } yield new Magnitude(v, k, MagnitudeSystem.VEGA) // TODO
+    mags                              = for {
+                                           (k, Some(v)) <- Map(
+                                             MagnitudeBand._u -> magNum(MagnitudeBand._u, "phot.mag;em.opt.U"),
+                                             MagnitudeBand._g -> magNum(MagnitudeBand._g, "phot.mag;em.opt.B"),
+                                             MagnitudeBand._r -> magNum(MagnitudeBand._r, "phot.mag;em.opt.R"),
+                                             MagnitudeBand._i -> magNum(MagnitudeBand._i, "phot.mag;em.opt.I"),
+                                             MagnitudeBand._z -> magNum(MagnitudeBand._z, "phot.mag;em.opt.I"),
+                                             MagnitudeBand.U  -> magNum(MagnitudeBand.U, "phot.mag;em.opt.U"),
```
swalker2m: Is the `String` argument what you see in the VoTable output?  If so, how does it pick between `_u` and `U` if they have the same identifying `String`?

````scala
MagnitudeBand._u -> magNum(MagnitudeBand._u, "phot.mag;em.opt.U")
MagnitudeBand.U  -> magNum(MagnitudeBand.U, "phot.mag;em.opt.U")
````

Comment (cquiroz) [inline bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/catalog/Simbad.scala]: Yeah, that's ugly. On my sample both magnitudes have the same "UCD" `phot.mag;em.opt.U` but the field id is different just in the case. for U it is `FLUX_U` and for u is `FLEX_u`. See the `magStr` method
I hope to add Simbad parsing logic to the catalog bundle and replace this part at some moment in time

Comment (swalker2m) [inline bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/catalog/Simbad.scala]: Oh, ok I see.  :+1:

File: bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/catalog/Simbad.scala [hunk 2]
```diff
@@ -29,47 +29,54 @@ object Simbad extends Catalog with App {
...
-        MagnitudeBand.R -> num("phot.mag;em.opt.R"),
-        MagnitudeBand.J -> num("phot.mag;em.ir.J"),
-        MagnitudeBand.H -> num("phot.mag;em.ir.H"),
-        MagnitudeBand.K -> num("phot.mag;em.ir.K"))
-    // TODO: more passbands
-    } yield new Magnitude(v, k, MagnitudeSystem.VEGA) // TODO
+    mags                              = for {
+                                           (k, Some(v)) <- Map(
+                                             MagnitudeBand._u -> magNum(MagnitudeBand._u, "phot.mag;em.opt.U"),
+                                             MagnitudeBand._g -> magNum(MagnitudeBand._g, "phot.mag;em.opt.B"),
```
cquiroz: Also note that for some reason `FLUX_g` uses the UCD `em.opt.B`, this is not a typo, it comes that way on the xml

PR #473, opened 2015-06-22
REL-2284: Non-sidereal targets should not produce RA visibility errors

cquiroz:
In this PR, RA errors for non-sidereal targets get reduced severity. The change is made at the calculation level so it is reflected on `ProblemRobot` and the `Visibility` indicator


swalker2m:
:+1:

PR #474, opened 2015-06-23
REL-259: ToO targets must not be allowed in non-ToO proposals (PIT)

cquiroz:
Small update with a message for the user if the ToO target option is disabled


swalker2m:
Makes sense to me. :+1:

PR #475, opened 2015-06-23
REL-2287: Do not allow proposal submission without the full PI name

cquiroz:
Per user request the initial name of the PI is set to 'Principal Investigator'. This will generate an error asking the user to set the real name


swalker2m:
Where is the error is detected?  :+1:

cquiroz:
The error is handled on the `ProblemRobot` class and will show on the Problems section:
screenshot 2015-06-23 15 06 14

arturog8m:
I think the "Please provide PI's full name" should be a TODO rather than an error, to be consistent with everything else.

cquiroz:
I tend to agree, I'll check with Andy

cquiroz:
Updated to show a ToDo instead of Error
screenshot 2015-06-23 16 28 25

arturog8m:
Nice :+1:

PR #476, opened 2015-06-23

cquiroz:
On the first RC, Phoenix can only support a single filter, but it should have support for multiple ones. This PR changes the schema to support multiple filters


swalker2m:
Is there a task to handle the skeleton creation / blueprint application side of it for phase 2?

:+1:

cquiroz:
Not as far as I know, there is a task to import to ITAC which is impacted by this PR

PR #477, opened 2015-06-23
OCSADV-200-T: finish migration to vcs2, cleanup

swalker2m:
This PR finishes the OT migration to the new merge/sync algorithm and removes the old code.  I'm sorry about the size but it is 99% deletion.

I had wanted to rename the package "vcs2" to "vcs" but since that is the package root for two other bundles I ran into problems with OSGi.  I'll leave it at "vcs2" unless someone strongly objects in which case we can choose different names.


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/vcs/VcsOtClient.scala
```diff
@@ -20,6 +20,8 @@ object VcsOtClient {
   def ref_=(c: Option[VcsOtClient]): Unit = {
     client = c
   }
+
+  def unsafeGetRegistrar: VcsRegistrar = ref.get.reg
```
tpolecat: I would prefer `ref.fold(sys.error("some useful message"))(_.reg)` since the `None.get` message isn't very useful.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/vcs/VcsPeerSelectionDialog.scala
```diff
@@ -41,8 +41,8 @@ object VcsPeerSelectionDialog {
   }
 
   def promptAndSet(c: Option[Component], pid: SPProgramID, allPeers: List[Peer]): Unit =
-    for (r <- VcsGui.registrar; p <- prompt(c, pid, allPeers)) {
-      r.register(pid, p)
+    for (client <- VcsOtClient.ref; p <- prompt(c, pid, allPeers)) {
+      client.reg.register(pid, p)
     }
 }
```
tpolecat: I prefer braces and one line per clause in `for` comprehensions. No biggie.

tpolecat:
I love the sea of red. :+1:

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/vcs/VcsOtClient.scala
swalker2m: Agreed and fixed.

PR #478, opened 2015-06-24

cquiroz:
This PR removes two NIRI filters fro the model and adds upconversion code for older proposals
A couple of bugs were fixed on the upconversion code and a few code inspections were fixed


swalker2m:
The code looks fine and I added a comment in the task about tracking the corresponding model changes that it may imply. :+1:

PR #479, opened 2015-06-24
OCSADV-305: Unify wavelength bands from core model and ITC.

fnussber:
The name says it all: This PR is replacing an ITC specific definition of wavelength bands with the one from the core model.


bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/base/ZeroMagnitudeStar.java
```diff
@@ -8,38 +8,38 @@
...
-            case I:     return 39000000;
-            case J:     return 19700000;
-            case H:     return 9600000;
-            case K:     return 4500000;
-            case L:     return 990000;
-            case M:     return 510000;
-            case N:     return 51000;
-            case Q:     return 7700;
+        // TODO: Missing support for Y band
+        if      (band.equals(MagnitudeBand.U$.MODULE$))  return 75900000;
```
cquiroz: Do you think this could go into a scala wrapper, it should be nicer in a `match` format

cquiroz:
Excellent, one less band definition :clap:

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/base/ZeroMagnitudeStar.java
fnussber: Hmm, yeah, this is probably simple enough that I can "port" it to Scala :)

bundle/edu.gemini.itc.web/src/main/scala/edu/gemini/itc/web/ITCRequest.scala
```diff
@@ -288,7 +288,9 @@ object ITCRequest {
     }
 
     // Get Normalization info
-    val normBand = r.enumParameter(classOf[WavebandDefinition])
+    val bandName = r.parameter("WavebandDefinition")
+    val normBand = MagnitudeBand.all.find(_.name == bandName)
+    if (normBand.isEmpty) throw new IllegalArgumentException(s"Unsupported wave band $bandName")
```
swalker2m: Will you be able to tackle error handling updates amongst future improvements?

swalker2m: While we're throwing exceptions, I suppose you could just shed the `Option` immediately with

````scala
val normBand = MagnitudeBand.all.find(_.name == bandName).getOrElse(sys.error(s"Unsupported wave band $bandName")
````
Then you avoid the `normBand.get` later on.  Or maybe it needs to be `IllegalArgumentException`?

swalker2m:
I always like to see duplication go away.  :+1:

PR #480, opened 2015-06-25
Launch a program from IDEA/TestLauncher

cquiroz:
This is a very simple utility to autolaunch a program when starting the OT via `TestLauncher` with lots of help from @swalker2m . It saves a few clicks when doing UI work
I configure it on IDEA with:
screenshot 2015-06-25 10 30 19


swalker2m:
Cool, looks like it should work with multiple `-program` arguments as well. :+1:

PR #481, opened 2015-06-25
OCSADV-200: small bugfix for sync history of new programs

swalker2m:
Corrects a bad assumption that programs will always have a program id, and that programs which have an ID also have a known peer with which to sync.  This was more or less harmless but could generate a confusing log message when viewing the root science program node in the OT.


tpolecat:
:+1:

PR #482, opened 2015-06-25
REL-2396: Australia a limited partner starting in 16A

cquiroz:
Tiny PR adding Australia as not allowed in joint proposals


tpolecat:
:+1:

PR #484, opened 2015-06-25
P1 Monitor smtp server

cquiroz:
Update the smtp server configuration for use with the gmail-based system

### DIFF
--- a/bundle/edu.gemini.p1monitor/src/main/resources/edu/gemini/p1monitor/config/conf.production-2015A-kr.xml
+++ b/bundle/edu.gemini.p1monitor/src/main/resources/edu/gemini/p1monitor/config/conf.production-2015A-kr.xml
@@ -1,5 +1,5 @@
 <conf>
-    <smtp>smtp.cl.gemini.edu</smtp>
+    <smtp>sbfsmtprelay-lv1.cl.gemini.edu</smtp>
 
     <bcc>swalker@gemini.edu</bcc>
     <bcc>cquiroz@gemini.edu</bcc>
--- a/bundle/edu.gemini.p1monitor/src/main/resources/edu/gemini/p1monitor/config/conf.production-2015A.xml
+++ b/bundle/edu.gemini.p1monitor/src/main/resources/edu/gemini/p1monitor/config/conf.production-2015A.xml
@@ -1,5 +1,5 @@
 <conf>
-    <smtp>smtp.cl.gemini.edu</smtp>
+    <smtp>sbfsmtprelay-lv1.cl.gemini.edu</smtp>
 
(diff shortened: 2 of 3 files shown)


bundle/edu.gemini.p1monitor/src/main/resources/edu/gemini/p1monitor/config/conf.production-2015A-kr.xml
```diff
@@ -1,5 +1,5 @@
 <conf>
-    <smtp>smtp.cl.gemini.edu</smtp>
+    <smtp>sbfsmtprelay-lv1.cl.gemini.edu</smtp>
```
fnussber: Is this for testing, or is this the smtp server we're supposed to use in the future for good?

fnussber:
:+1:

bundle/edu.gemini.p1monitor/src/main/resources/edu/gemini/p1monitor/config/conf.production-2015A-kr.xml
cquiroz: I think it is for good, it will relay emails through gmail instead of the old exchange protocol

arturog8m: smtp.cl.gemini.edu is the name we need to keep actually - it's an alias. It will point to sbfsmtprelay-lv1.cl.gemini.edu when the move to Google is done. So we shouldn't change this at all, except if we want to test the Google relay server from this application before the actual migration. I'm not sure it's necessary. It will be useful to confirm the machine this service runs is authorized to send emails to sbfsmtprelay-lv1.cl.gemini.edu - but that can be done using telnet. 

So, this PR should :-1:

cquiroz: Oh really? I miss-understood then. Much better as the config is included in a jar file thus making it much harder to change.

cquiroz: Is there a place on the wiki where these changes are documented, and a list of the hosts that are allowed to send mails?

fnussber: Ok, I retract my thumbs up, I thought that change was for testing the relay..

cquiroz:
There is no need to change the config

PR #485, opened 2015-06-26

cquiroz:
More updates to Phoenix filters. This change also required adding a filter to Phoenix's model on the pot bundle
Also cleaned up warnings and some generated code on the Phoenix model and UI on the OT


swalker2m:
:+1:

PR #486, opened 2015-06-26
REL-2400: user key program checkout failure

swalker2m:
This PR addresses a bug in which a PI with a valid user key cannot checkout his own program.  The `ImplicitPolicy` check for `PiPermission` for a user principal will open the program in the database and check whether it contains a matching email.  For a checkout, the permission check will work fine in the server because the program exists in the remote database.  The client can successfully retrieve the program from the remote database.  If we also check locally though at the point when we're trying to add the program to the local database, It will fail.

The solution is to simply stop checking for permission to add a program to a local database while retaining permission checks if you try to add a program to a remote database.  Adding a program to a remote database continues to require a key that doesn't depend on the content of the program being added which is what we want.

### DIFF
--- a/bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs2/VcsServer.scala
+++ b/bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs2/VcsServer.scala
@@ -82,7 +82,7 @@ class VcsServer(odb: IDBDatabaseService) { vs =>
     * same key or id with an existing program in the database and the user has
     * appropriate keys.
     */
-  def add(p: ISPProgram, user: Set[Principal]): VcsAction[Unit] = {
+  def add(p: ISPProgram): VcsAction[Unit] = {
     def failIfExists(id: SPProgramID, key: SPNodeKey): VcsAction[Unit] = {
       val alreadyExists = Option(odb.lookupProgramByID(id)).as(IdAlreadyExists(id)).orElse {
         Option(odb.lookupProgram(key)).as(KeyAlreadyExists(id, key))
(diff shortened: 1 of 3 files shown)


bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs2/VcsServer.scala
```diff
@@ -109,7 +107,10 @@ class VcsServer(odb: IDBDatabaseService) { vs =>
       vs.read(id, user)(_.getVersions).unsafeRun
 
     override def add(p: ISPProgram): TryVcs[Unit] =
-      vs.add(p, user).unsafeRun
+      (for {
+        id <- (Option(p.getProgramID) \/> MissingId).liftVcs
+        _  <- accessControlled(id, user) { vs.add(p) }
+      } yield ()).unsafeRun
```
swalker2m: This is the entry point in the server for committing a new file to a remote database.

bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs2/VcsServer.scala
```diff
@@ -92,10 +92,8 @@ class VcsServer(odb: IDBDatabaseService) { vs =>
...
     (Option(p.getProgramID) \/> MissingId).liftVcs >>= { id =>
-      accessControlled(id, user) {
-        locked(p.getProgramKey, writeLock, writeUnlock) {
-          failIfExists(id, p.getProgramKey) >> putProg(odb.getFactory.copyWithNewLifespanId(p)).liftVcs
-        }
+      locked(p.getProgramKey, writeLock, writeUnlock) {
+        failIfExists(id, p.getProgramKey) >> putProg(odb.getFactory.copyWithNewLifespanId(p)).liftVcs
       }
     }
   }
```
swalker2m: The `VcsServer` name may be misleading.  `VcsServer` provides low-level database and program manipulation in general.  The methods it presents are not directly accessible by remote users, who instead go through the nested `SecureVcsService`.

cquiroz:
Looks fine, are there tests?

swalker2m:
I removed a test that should no longer apply, but left the [existing test](https://github.com/gemini-hlsw/ocs/blob/develop/bundle/edu.gemini.sp.vcs/src/test/scala/edu/gemini/sp/vcs2/VcsSpec.scala#L59-L62) in `VcsSpec` that checks whether an attempt to add a program to a remote database is rejected if you don't have an appropriate key.

cquiroz:
:+1:

PR #487, opened 2015-06-26
REL-1803: cache P2 check results

swalker2m:
This PR is aimed at speeding up the time it takes to recover from synchronization of large programs.  In particular it caches P2 check results for observations and avoids recalculating them unless something in the observation actually changes.   It makes the determination that something has changed by comparing a hash of the node version information used by the sync algorithm.

This change would introduce mutable state (the cache) in `P2Checker` so I changed it to have one instance per `SPViewer` instead of a singleton shared across all viewers.  Cache cleanup happens automatically when the viewer switches to view a new program and runs a P2 check for it.

A lot more could be done to speed up P2 checking in general but for the specific case of recovering from a sync, the cache is quite effective.

### DIFF
--- a/bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/checker/P2Checker.java
+++ b/bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/checker/P2Checker.java
@@ -41,18 +41,18 @@ import edu.gemini.spModel.gemini.trecs.InstTReCS;
 import edu.gemini.spModel.gemini.flamingos2.Flamingos2;
 import edu.gemini.spModel.gemini.visitor.VisitorInstrument;
 import edu.gemini.spModel.obscomp.SPInstObsComp;
+import edu.gemini.spModel.util.NodeValueCache;
+import edu.gemini.spModel.util.NodeValueCache$;
 
 
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 
 /**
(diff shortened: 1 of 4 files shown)


bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/checker/P2Checker.java
```diff
@@ -135,16 +125,26 @@ private IRule _getRule(ObservationElements elements) {
...
+            return check(node.getContextObservation(), mt);
+
             //observations can be checked immediately
         } else if (node instanceof ISPObservation) {
-            return _checkObservation((ISPObservation) node, mt);
+            final SPNodeKey progKey = node.getProgram().getNodeKey();
+            final NodeValueCache<IP2Problems> nvc = (progKey == cacheKey) ? cache : NodeValueCache$.MODULE$.empty();
+
+            final scala.Tuple2<IP2Problems, NodeValueCache<IP2Problems>> tup;
+            tup = nvc.get(node, new scala.runtime.AbstractFunction1<ISPNode, IP2Problems>() {
```
cquiroz: Can this be replaced by a Java 8 lambda? I'm not sure how interoperable these are

tpolecat: There is SAM support in 2.11 under `-Xfuture` I think, but we're still on 2.10 so I think no.

tpolecat:
Seems fine.  :+1:

cquiroz:
Good :+1:

PR #488, opened 2015-06-29

cquiroz:
It turns out I commented out the wrong NIRI filters, this PR fixes that


swalker2m:
:+1:

PR #489, opened 2015-06-30
REL-2372: Prepare PIT 2016A

cquiroz:
Set the PIT to use non-test mode and use of the HBF based backend.

I also have in this PR moving `GlassLabel` from `edu.gemini.pit` to the bundle `edu.gemini.shared.gui`. Though this is not strictly related to REL-2372 I wanted to make sure I could build the PIT with this change


arturog8m:
other than the final machine name, everything looks good.
:+1:

cquiroz:
We are going to switch to phase1.gemini.edu at some moment, but I wanted this PR to make it easier to make Release Candidates

arturog8m:
:+1:  perfect then

PR #490, opened 2015-06-30
OCSADV-393: Add MigLayout and custom scala API

cquiroz:
On this PR I'm introducing the MigLayout and a scala API loosely based on the one used on GPI

Some notes:
* `MigLayout` is BSD/GPL Dual license, the source code is at https://github.com/mikaelgrev/miglayout and documentation at http://www.migcalendar.com/miglayout/
* As usual, it is not OSGi compliant but I submitted a PR to have OSGi headers added with the maven  build https://github.com/mikaelgrev/miglayout/pull/33
* The latest release version is 4.2 and though it is a bit old I prefer to wait until the 5.0 release is no longer a snapshot
* I placed the API on its own bundle `edu.gemini.ui.miglayout` but that maybe overkill, I could instead move it to `edu.gemini.shared.ui`. Opinions?
* `MigLayout` tends to use `Strings` as constraints though it also has a fluent API that remove smost of them. The scala API below helps to remove most of the ones still there using implicit classes
* I'm using Structural types to reduce repetition. I don't think the runtime penalty is too bad here as these are used mostly when constructing a layout
* There is a sample `MigLayoutDemo` that can be helpful to check how the layout/panel is used. The layout produces the following app:

screenshot 2015-06-30 17 37 00

I haven't done much in terms of DSLs, I'd certainly welcome comments on the scala API


bundle/edu.gemini.ui.miglayout/src/main/scala/edu/gemini/ui/miglayout/MigPanel.scala
```diff
@@ -0,0 +1,337 @@
...
+  // MigLayout horizontal align objects
+  sealed trait HMigAlign {
+    protected [constraints] def toAlign: String
+  }
+  case object RightAlign extends HMigAlign {
+    override def toAlign = "right"
+  }
+  case object LeftAlign extends HMigAlign {
+    override def toAlign = "left"
+  }
```
tpolecat: You can save a bit of work here by using a sealed class.

```scala
sealed abstract class VMigAlign(private[constraints] val toAlign: String)
case object TopAlign extends VMigAlign("top")
// etc
```

bundle/edu.gemini.ui.miglayout/src/main/scala/edu/gemini/ui/miglayout/MigPanel.scala
```diff
@@ -0,0 +1,337 @@
...
+  object CC {
+    def apply(): MigCC = new MigCC()
+  }
+
+  // Structural type for a class that support alignment
+  private type Alignable[T] = {
+    def alignX(s: String): T
+    def alignY(s: String): T
+  }
+
```
tpolecat: Not a big deal since this is internal to the implementation, but It's generally better to use a typeclass in situations like this.

tpolecat:
It would be helpful to have a README for DSL bundle with some sample code. Did you forget to include `MigLayoutDemo`?

cquiroz:
`MigLayoutDemo` is at then end of the DSL file, I considered that the documentation :smile:

bundle/edu.gemini.ui.miglayout/src/main/scala/edu/gemini/ui/miglayout/MigPanel.scala
cquiroz: Do you mean the `object CC` part or the structural type? I thought about a typeclass but I wanted to have a single implicit class, I may explore that option though, any pointers?

bundle/edu.gemini.ui.miglayout/src/main/scala/edu/gemini/ui/miglayout/MigPanel.scala
cquiroz: Good tip :+1:

bundle/edu.gemini.ui.miglayout/src/main/scala/edu/gemini/ui/miglayout/MigPanel.scala
tpolecat: I usually end up with a typeclass rather than a structural type because:

- Method calls on instances of structural types rely on reflection at runtime, which is inefficient.
- Types conform by structure rather than by some affirmative declaration on the part of the programmer, so you can accidentally sweep up things that don't really make sense as instances.

Neither of these is really an issue here, but I wanted to point it out as something to keep in mind in the future if you're considering a structural types.

You could encode it as a typeclass with something like this:

```scala
trait Alignable[A] {
  def alignX(a: A, s: String): A
  def alignY(a: A, s: String): A
}

implicit class AlignableOps[A](val a: A) extends AnyVal {
  def alignY(align: VMigAlign)(implicit ev: Alignable[A]): A = ev.alignY(a, align.toAlign)
  def alignX(align: HMigAlign)(implicit ev: Alignable[A]): A = ev.alignX(a, align.toAlign)
}

// and then declare instances for conforming types
```

tpolecat:
Ah, don't know how I missed that.

bundle/edu.gemini.ui.miglayout/src/main/scala/edu/gemini/ui/miglayout/MigPanel.scala
cquiroz: I think I wanted to have a single `AlignableOps` implicit class that would apply to any class that goes with the structural type, saving me to need to declare multiple type classes.
I'm not too worried in this case about the performance penalty of reflection but I can see a point that a structural type would add the methods to other classes that accidentally happen to have the same structure

tpolecat: Yeah it's not a big deal here.

cquiroz:
Added a README and switched to type-classes as suggested

tpolecat:
I'm inclined to merge this and see how it goes. :+1:  Maybe let @swalker2m have a look first. It's always hard to know how far to go with this kind of thing.

cquiroz:
I agree, there are a few more strings on the API but they are seldom used, if needed the API can be expanded later on

swalker2m:
From the demo, it looks like it could be more readable than the equivalent grid bag layout.  It might help if I knew what AC, CC, and LC stand for.  I'm not very excited about this though since it is a third party library and requires wrappers to be reasonable.  Once we've created a lot of UIs with it we're kind of stuck with it.   On the other hand, it is pretty small and all standard layout managers are bad. If everyone else is up for it, let's merge it I guess.

cquiroz:
AC, CC and LC are used to define constraints, shorthands for 'AxisConstraints', 'CellConstraints' and 'LayoutConstraints', being the last 2 the most used ones.

I certainly disagree about not using third-party libraries when they are useful, we have the source code and are not part of our core business.

swalker2m:
We certainly use 3rd party libraries so that's not the issue.  It's about whether you're getting enough for the long-term commitment you're making.  In this case, I suspect you can do the same thing with grid bag albeit not as concisely so let's give it a try.

PR #491, opened 2015-06-30

cquiroz:
Another NIRI filter removed, this time  CO 3-1 (bh) (2.323 um)


tpolecat:
Looks fine. :+1:

PR #492, opened 2015-07-03
REL-2401-A: remove active guider concept in target environment

swalker2m:
This PR removes the concept of "active" guiders in the `TargetEnvironment`.  Active guiders refer to the collection of guide probes which are available in a particular observation.  For example, a GMOS-S observation can use PWFS guide probes or the GMOS OIWFS but not the Flamingos2 OIWFS.  Available guiders are always computed from the content of the observation but were unfortunately also being cached in the `TargetEnvironment` itself.  I think this may have been done in the RMI days when the computation, which involves traversing an observation's components, was expensive.  At any rate, the caching would cause the target component itself to be marked as edited when unrelated components were added or removed from the observation.  This was done by a class called `GuideSync` which would monitor all edits and make updates to the `TargetEnvironment` in the target component.

The change introduced here is to stop caching "active" guiders and simply compute the available guiders when needed.  With this simplification, one of the two issues reported in REL-2401 has been addressed.  A second PR will be added for the unrelated issue with the observation validation calculating while viewing the conflict folder.


bundle/edu.gemini.ags.servlet/src/main/java/edu/gemini/ags/servlet/ToContext.java
```diff
@@ -269,20 +229,14 @@ private InstAltair getAltair(HttpServletRequest req) throws RequestException {
      * @return
      */
     private static AltairParams.Mode modeFromAgsServletParameter(String paramName) {
```
cquiroz: Would it make sense for this to return `Option` rather than null/exception?

swalker2m: Yes, I agree.  I just switched to an easier to read `switch` statement in passing but a more extensive refactoring would probably be in order at some point.

cquiroz:
Very good :+1: Quite a big simplification
Do we still have the anti-guiders, right?

swalker2m:
Yes, the calculation is the same as before (see [`GuideProbeUtil`](https://github.com/gemini-hlsw/ocs/blob/develop/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/guide/GuideProbeUtil.java#L63-L79)). The difference is that we just don't cache the value.

PR #493, opened 2015-07-03
OCSADV-395: Create a UI with a table displaying results of catalog queries

cquiroz:
This is the first PR for the new catalog search feature. It adds a button to the target environment that opens a table and start a search for the current target using the corresponding AGS parameters

The PR ended up being large because I got carried on cleaning up old code. The main things to check are:

* `QueryResultsFrame` contains the UI of the table
* `EdCompTargetList` has an extra button to launch the UI
* A few more partial lenses were added to the target model
* `AgsStrategy` exposes what queries it can use to load targets for an observation including parameters like radius, etc.

Here is a screenshot of the table

screenshot 2015-07-03 16 15 31


swalker2m:
Does it show all the magnitude bands in the results or does it select them based on the guiders?  Presumably it will scroll horizontally? 

While the search is in progress, does it give some indication of what is going on?  Is there a way to cancel the search?

What is the next step?  Will you add the search parameters in a form in the same window?

cquiroz:
It shows all the magnitudes available (Fill hide columns for bands not available) but will filter the range based on guiders. There is no provision right now to change guiding strategy but that will be added later.

There is no horizontal scrolling, I don't have the need on my monitor but perhaps I should check it on a laptop

While searching it displays a label on top of the window, it uses the same UI as the PIT when submitting, but there is no way to cancel at this stage

The next step is to improve some search functions in the code (Leftovers from AGS changes) and add query param controls (I was thinking to add them to the right). After that, or perhaps even before as is the most uncertain part, we need to link the table to the TPE

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/Target.scala
```diff
@@ -48,6 +48,23 @@ object Target {
...
     ))
 
+  val pm: Target @?> ProperMotion =
+    PLens(_.fold(
+      PLens.nil.run,
+      SiderealTarget.pm.run,
+      PLens.nil.run,
+      PLens.nil.run,
+      PLens.nil.run
+    ))
```
swalker2m: I think I could use a bit of schooling here.  What is going on in this code?  `PLens` is partial lens I take it.

cquiroz: A partial lens means it is not defined for all inputs, so the logical signature is `A => Option[B]`. In this case from a `Target` we want to get/set the `ProperMotion` but that is only valid for certain target types, specifically `SiderealTarget`.

I'm not totally familiar with the fold syntax but my understanding is that depending on the exact subtype it will run different `PLens`, being the only non-nil the one from `SiderealTarget`

swalker2m: I think I see, the `fold` is on `Target` itself not the `Option[Store[ProperMotion, Target]]`.

bundle/jsky.app.ot/src/main/java/jsky/image/gui/DivaMainImageDisplay.java
```diff
@@ -361,20 +339,9 @@ public DivaMainImageDisplay(GraphicsPane pane) {
         _fitsGraphics = new FITSGraphics(this);
 
         // try to restore the history from the previous session
```
swalker2m: I guess this comment no longer applies.

bundle/jsky.app.ot/src/main/java/jsky/image/gui/DivaMainImageDisplay.java
```diff
@@ -361,20 +339,9 @@ public DivaMainImageDisplay(GraphicsPane pane) {
...
-//        loadHistory();
-        _historyList = new LinkedList<ImageHistoryItem>();
-//        cleanupHistoryList();
-        if (_firstTime) {
-            _firstTime = false;
-            cleanupImageCache();
-        }
```
swalker2m: What was the "first time" business?

cquiroz: A code check showed that `_firstTime` was always false

bundle/jsky.app.ot/src/main/java/jsky/image/gui/DivaMainImageDisplay.java
```diff
@@ -853,67 +790,21 @@ protected ImageHistoryItem getImageHistoryItem(File file) {
...
         }
         _progressPanel.start();
     }
 
-
-    /**
-     * Load an image directly from the given URL in a separate thread, so the user can monitor progress
-     * and cancel it, if needed.
-     */
-    protected void loadImageFromURL(final URL url) {
```
swalker2m: It can still load DSS images right?

bundle/jsky.app.ot/src/main/scala/edu/gemini/catalog/ui/QueryResultsFrame.scala
```diff
@@ -0,0 +1,206 @@
...
+  val instance = this
+
+  private def showTable(q: CatalogQuery):Unit = Swing.onEDT {
+    import QueryResultsWindow.table._
+    if (frame.visible) {
+      frame.peer.toFront()
+    } else {
+      frame.visible = true
+      frame.peer.toFront()
+    }
```
swalker2m: Is this much different than eliminating the `if` statement and just always doing the `else` branch?

swalker2m:
Looks like a great start. :+1:

bundle/jsky.app.ot/src/main/scala/edu/gemini/catalog/ui/QueryResultsFrame.scala
cquiroz: I updated this removing the non `else` branch

cquiroz:
A missing class `SortedTable` was added on the last commit

PR #494, opened 2015-07-06
REL-2401-B: specify valid types for a conflict folder

swalker2m:
This PR updates the `ConflictConstraint` to specify the set of valid types it may contain.  A conflict folder can contain anything other than another conflict folder.

This will avoid the exception reported in REL-2401, which was a result of the OT attempting to determine which UI features to enable when clicking on the conflict folder.


cquiroz:
:+1:

PR #495, opened 2015-07-07
REL-2402: local deletion of unedited remote nodes

swalker2m:
This PR contains a small update to the preliminary merge logic.  The goal is to allow a locally deleted node to remain deleted even when the containing node has also been edited remotely.  The example from the issue should help to make this clear:

> Start with a folder with one observation 
> OT1: add second observation and sync 
> OT2: delete observation and sync 
> Surprise! both observations appear in OT2 

> This works as expected using the "Old Sync" button.

All the properties still pass except the one that explicitly stated that these nodes should be resurrected.  Since the point of this PR is to make that property no longer hold, I've removed it from the `MergeTest`.


cquiroz:
:+1:

PR #496, opened 2015-07-07
OCSADV-400: sync documentation

swalker2m:
Updated sync documentation and moved it into the bundle itself.


swalker2m:
This is just documentation so I'll give myself a :+1:

PR #497, opened 2015-07-07
OCSADV-402: Improve magnitude filtering in CatalogQuery

cquiroz:
This PR pays technical debt on the catalogs code, incurred when we had to support AGS looking for multiple bands on the same target.

The solution done back then, involved a two-stage filtering process and an ad-hoc API for filtering in different situations. This PR removes this and goes for a simpler API where a `CatalogQuery` just contains filters.
 
The main benefit is that we have a uniform API with a single stage filter, that will be needed for the new Catalog Navigator.

The PR is relatively large but the core API change is on the class `CatalogQuery.scala`, the rest is updates to `CatalogQuery` clients and tests


bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/SingleProbeStrategy.scala
```diff
@@ -44,13 +44,11 @@ case class SingleProbeStrategy(key: AgsStrategyKey, params: SingleProbeStrategyP
...
     val empty = Future.successful(List((params.guideProbe: GuideProbe, List.empty[SiderealTarget])))
-    def filterOnMagnitude(q: CatalogQuery, t: SiderealTarget): Boolean = {
-      params.referenceMagnitude(t).exists(q.filterOnMagnitude(t, _))
-    }
+
     // We cannot let VoTableClient to filter targets as usual, instead we provide an empty magnitude constraint and filter locally
     catalogQueries(ctx, mt).strengthR(backend).headOption.map(Function.tupled(VoTableClient.catalog)).map(_.flatMap {
         case r if r.result.containsError => Future.failed(CatalogException(r.result.problems))
-        case r                           => Future.successful(List((params.guideProbe, r.result.targets.rows.filter(t => filterOnMagnitude(r.query, t)))))
+        case r                           => Future.successful(List((params.guideProbe, r.result.targets.rows)))
```
cquiroz: This is a good example of doing double filtering, First on the call `VoTableClient.catalog` and then on the results. This is now done in one step

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/CatalogSearchCriterion.scala
```diff
@@ -16,7 +16,7 @@ import Scalaz._
  * Used to query catalogs and filter and categorize query results.
  * See OT-20
  */
-case class CatalogSearchCriterion(name: String, referenceBand: MagnitudeBand, magRange: MagnitudeRange, radiusLimits: RadiusConstraint, offset: Option[Offset] = None, posAngle: Option[Angle] = None) {
+case class CatalogSearchCriterion(name: String, referenceBands: List[MagnitudeBand], magRange: MagnitudeRange, radiusLimits: RadiusConstraint, offset: Option[Offset] = None, posAngle: Option[Angle] = None) {
```
swalker2m: The same range has to apply to all the magnitudes in `referenceBands`?

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/SingleProbeStrategy.scala
swalker2m: It seems like there will always need to be double filtering, if you will.  One thing is the query to the catalog which should be the widest possible magnitude range.  Another is the range that applies to the observation as configured with the specific observing conditions requirements.  We are still getting candidates across all the available conditions options from the catalog?

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/CatalogSearchCriterion.scala
cquiroz: This may not be that clear but it is a side effect of the multiple band selection logic. What this actually represents is that the `magRange` will be applied to the first band available on `referenceBands`, typically `[r, R, UC]` or `H`

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/SingleProbeStrategy.scala
cquiroz: VoTable does not support mag filtering so we get all targets on a given radius, then mag filtering is local.
The strategy decides based on the context (`catalogQueries(ctx, mt)`) what to filter out based on rules defined on the magnitude
I want to avoid the second filtering step so that the CatalogNavigator is agnostic to the context or observation, it just get query params, essentially, coordinates, radius and a set of filters. the second step will mean special-casing on the CatalogNavigator

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/api/CatalogQuery.scala
```diff
@@ -41,30 +49,87 @@ sealed trait CatalogQuery {
   }
 }
 
+trait CatalogQueryWithRange extends CatalogQuery { this: CatalogQuery =>
+  val magnitudeRange: MagnitudeRange
```
swalker2m: As an aside, I think I'd rather not use an abstract `val` here.  If you declare this as `def magnitudeRange: MagnitudeRange` you can still implement it in the extending class with a `val` and it avoids having it temporarily initialized to `null`.

cquiroz: Good point

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/api/CatalogQuery.scala
```diff
@@ -41,30 +49,87 @@ sealed trait CatalogQuery {
...
+  private case class RangeConstrainedCatalogQueryWithBand(base: Coordinates, radiusConstraint: RadiusConstraint, mag: MagnitudeExtractor, magnitudeRange: MagnitudeRange, catalog: CatalogName) extends CatalogQuery with CatalogQueryWithRange {
+    override val filters: List[QueryResultsFilter] = List(RadiusFilter(base, radiusConstraint), MagnitudeValueFilter(mag, magnitudeRange))
   }
 
-  // TODO Ensure that it makes sense to default to ucac4
-  def catalogQuery(base: Coordinates, radiusConstraint: RadiusConstraint, magnitudeConstraints: Option[MagnitudeConstraints], catalog: CatalogName = ucac4): CatalogQuery
-    = BandConstrainedCatalogQuery(base, radiusConstraint, magnitudeConstraints, catalog)
+  private case class CatalogQueryWithAdjustedBand(override val id: Option[Int], base: Coordinates, radiusConstraint: RadiusConstraint, mag: MagnitudeExtractor, magnitudeRange: MagnitudeRange, rangeAdjustment: MagnitudeRangeAdjustment, catalog: CatalogName) extends CatalogQuery with CatalogQueryWithRange {
+    override val filters: List[QueryResultsFilter] = List(RadiusFilter(base, radiusConstraint), AdjustedMagnitudeValueFilter(mag, magnitudeRange, rangeAdjustment))
+  }
```
swalker2m: The `CatalogQuery` classes don't seem to differ other than the filters they employ.  Maybe `CatalogQuery` could just be a final case class which you create with the list of filters that should be used?

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/CatalogSearchCriterion.scala
tpolecat: I assume `referenceBands` can be empty?

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsVoTableCatalog.scala
```diff
@@ -71,13 +71,17 @@ case class GemsVoTableCatalog(backend: VoTableBackend = RemoteBackend, catalog:
...
     val queryArgs = for {
       c <- criterions
-      q = CatalogQuery.catalogQueryWithoutBand(basePosition, c.criterion.radiusLimits, c.criterion.magRange.some, catalog)
+      q = CatalogQuery.catalogQueryWithDynamicBand(basePosition, c.criterion.radiusLimits, extractor, c.criterion.magRange, catalog)
     } yield (q, c)
 
     val qm = queryArgs.toMap
-    VoTableClient.catalogs(queryArgs.map(_._1), backend).map(l => l.map(k => GemsCatalogSearchResults(qm.get(k.query).get, k.result.targets.rows)))
+    VoTableClient.catalogs(queryArgs.map(_._1), backend).map(l => l.collect {
+      case QueryResult(q: CatalogQueryWithRange, results:CatalogQueryResult) => GemsCatalogSearchResults(qm.get(q).get, results.targets.rows)
```
tpolecat: How is it possible to match on this? I can't find an `unapply` for it.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsVoTableCatalog.scala
```diff
@@ -93,33 +97,29 @@ case class GemsVoTableCatalog(backend: VoTableBackend = RemoteBackend, catalog:
...
-              val rangeAdjustmentForConditions = (mr: Option[MagnitudeRange], mag: Magnitude) => mr.map { m =>
-                m.adjust(m => conditions.magAdjustOp().apply(mag.toOldModel).toNewModel.value)
-              }
-
-              k.result.targets.rows.filter(t => k.query.filterOnMagnitude(t, magnitudeExtractor(defaultProbeBands(b))(t), rangeAdjustmentForConditions))
-            }.getOrElse(k.result.targets.rows)
-          }.suml
+          val targets = l.map { k =>
+              k.result.targets.rows
+            }.suml
```
tpolecat: `val targets = l.foldMap(_.result.target.rows)`

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsVoTableCatalog.scala
```diff
@@ -184,7 +184,7 @@ case class GemsVoTableCatalog(backend: VoTableBackend = RemoteBackend, catalog:
         fl = m.faintnessConstraint
       } yield (b, fl)
 
-    val faintnessMap:Map[MagnitudeBand, FaintnessConstraint] = faintLimitPerBand.groupBy(_._1).map { case (_, v) => v.maxBy(_._2)(FaintnessConstraint.order.toScalaOrdering)}
+    val faintnessMap:Map[List[MagnitudeBand], FaintnessConstraint] = faintLimitPerBand.groupBy(_._1).map { case (_, v) => v.maxBy(_._2)(FaintnessConstraint.order.toScalaOrdering)}
```
tpolecat: Same question here, can this list be empty? If not, `NonEmptyList` pls.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsVoTableCatalog.scala
```diff
@@ -196,7 +196,6 @@ case class GemsVoTableCatalog(backend: VoTableBackend = RemoteBackend, catalog:
 
     (for {
       b <- faintnessMap
-      mc = if (saturationMap.contains(b._1)) MagnitudeConstraints(b._1, b._2, saturationMap.get(b._1)) else new MagnitudeConstraints(new Magnitude(b._2.brightness, b._1))
-    } yield MagnitudeConstraints(b._1, b._2, saturationMap.get(b._1))).toList
+    } yield (b._1, MagnitudeRange(b._2, saturationMap.get(b._1)))).toList
```
tpolecat: Not a bug deal, but I would prefer `map` here rather than a 1-stanza `for`.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsVoTableCatalog.scala
cquiroz: Right, that's good, it cannot/should not be empty

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsVoTableCatalog.scala
tpolecat: You don't need the type pattern on `results`.

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/api/CatalogQuery.scala
cquiroz: That is good point, the multiple classes had different behaviors but now that they are unified they could likely be removed and make `CatalogQuery` a concrete class

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/CatalogSearchCriterion.scala
cquiroz: `referenceBands` should not be empty though there is an argument that it could be just one band, either `R` (which expands internally to `r, R, UC`) and `H`
I'll explore that a bit, or else use `NonEmptyList`

cquiroz:
Thank you very much for the comments, this made me realize that I need more work.
I plan to simplify a bit the filters and made them more visible, `MagnitudeRange` and `MagnitudeConstraints` can be simplified into a single class and recognize everywhere that magnitude filters have a main band which can be expanded into other bands if needed
This is also driven by the needs of the UI
I'm closing this PR for now

PR #498, opened 2015-07-07
OCSADV-399-A

tpolecat:
Ok this is kind of massive, sorry. The goal here is to start preparing for immutable targets by removing direct mutation of `ITarget` instances. This turned out to be a prereq for 399.

- The `ICoordinate` interface is gone, in preparation for the new RA/Dec types.
- All mutators on `ITarget` are now package-private; the public API is now entirely read-only.
- For each of these mutators there is now a public method on `SPTarget` (which contains an `ITarget`). With the new model these mutators will swap out the contained target. For now the mutators continue to modify the contained target in-place, via a `TransitionalSPTarget` superclass that lives in the same package as `ITarget` for access to the package-private methods.


bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/SPTarget.java
```diff
@@ -75,40 +72,4 @@ public SPTarget clone() {
...
-            throw new IllegalArgumentException();
-        }
-    }
-
-    // I'm making this public so I can call it from an editor when I make
-    // a change to the contained target, rather than publishing all the
-    // target members through this idiotic class. All of this crap needs
-    // to be rewritten.
-    /** @deprecated */
-    public void notifyOfGenericUpdate() {
```
cquiroz: I've seen this call many times in e.g. the TPE, are all of them unnecessary?

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/system/ICoordinate.java
```diff
@@ -1,64 +0,0 @@
-// Copyright 1997-2000
```
cquiroz: Nice to have one Coordinate-like interface gone, just  a few more to go :smile:

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/system/TransitionalSPTarget.java
```diff
@@ -0,0 +1,75 @@
+package edu.gemini.spModel.target.system;
+
+import edu.gemini.shared.skyobject.Magnitude;
+import edu.gemini.shared.util.immutable.ImList;
+import edu.gemini.spModel.target.SpatialProfile;
+import edu.gemini.spModel.target.SpectralDistribution;
+import edu.gemini.spModel.target.WatchablePos;
+
+// transitional; will go away
```
cquiroz: I often mark temporal classes as `@Deprecated`, the compiler warnings motivate me to take them out

tpolecat: That's not a bad idea.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/SPTarget.java
tpolecat: They're necessary if you change the `ITarget` **directly** because these changes won't generate a property change event on the `SPTarget`. But now that all the mutators are on `SPTarget` I think we can probably get rid of `notifyOfGenericUpdate()` … I'll check and see.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/system/ICoordinate.java
tpolecat: Yep.

cquiroz:
Looks good, A general worry with this and other large refactorings is that there maybe unforseen issues downstream on untested or GUI code. I don't have a solution for this though
:+1:

bundle/edu.gemini.oodb.too.url/src/main/java/edu/gemini/spdb/rapidtoo/impl/TooUpdateServiceImpl.java
```diff
@@ -237,10 +237,9 @@ private void _setTarget(ISPObservation obs, TooUpdate update)  {
...
         SPTarget base = targetEnv.getBase();
 
-        base.getTarget().setName(tooTarget.getName());
-        base.getTarget().getRa().setAs(tooTarget.getRa(), CoordinateParam.Units.DEGREES);
-        base.getTarget().getDec().setAs(tooTarget.getDec(), CoordinateParam.Units.DEGREES);
-        base.getTarget().setMagnitudes(tooTarget.getMagnitudes());
+        base.setName(tooTarget.getName());
+        base.setRaDecDegrees(tooTarget.getRa(), tooTarget.getDec());
+        base.setMagnitudes(tooTarget.getMagnitudes());
         base.notifyOfGenericUpdate();
```
swalker2m: Before we would call `notifyOfGenericUpdate` once at the end of a string of updates like this.  Now there will be events per mutation?  I think that's probably fine but it is a difference in behavior and I know a lot happens when targets are updated.  This is just a transition step though until we simply swap out the immutable target once it is properly configured?

At any rate, it seems like we shouldn't not need the last `notifyOfGenericUpdate` call here if all the individual mutations are generating events anyway.

tpolecat: Correct, we don't need this. I have been deleting them but I missed this one. I'll do another sweep.

You're also correct that there will be more events now than before. `SPTarget` may end up with a  `.modify(f: Target => Target)` or something so we can do several updates atomically. In any case I'll try out TPE dragging and a few other things and see if the extra events are gumming things up.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details/SiderealNameEditor.scala
```diff
@@ -108,14 +107,14 @@ final class SiderealNameEditor extends TelescopePosEditor with ReentrancyHack {
     tqr.getCoordinates(0) match {
       case pos: WorldCoords =>
         try {
-          t.getRa.setValue(pos.getRA.toString)
-          t.getDec.setValue(pos.getDec.toString)
+          t.setRaString(pos.getRA.toString)
+          t.setDecString(pos.getDec.toString)
```
swalker2m: There's a `spt.notifyOfGenericUpdate()` down at line 124 which we shouldn't need now.

swalker2m:
Seems self-contained and fairly explanatory.  :+1:

tpolecat:
okie dokie :pray:

PR #499, opened 2015-07-08
OCSADV-276: Implemented execution state memory, sequence execution co…

jluhrs:
Sequences commands: start, stop, continue, getExecutionState. I share the blame with Rob for this code, although it's easy to find who wrote what (tip: the horrible parts are probably mine)

### DIFF
--- a/bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Executor.scala
+++ b/bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Executor.scala
@@ -115,11 +115,11 @@ object Executor {
       EitherT[ExecAction, NonEmptyList[SeqexecFailure], A](a)
 
     // Execute the next step, if any
-    def stepT(stateRef: AtomicReference[ExecState]): ExecActionF[StepResult] =
+    def stepT(saveState: ExecState => ExecState): ExecActionF[StepResult] =
       for {
         st <- liftA(gets(_.nextStep))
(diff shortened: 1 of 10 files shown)


tpolecat:
Cool! I'll give this a close look.

jluhrs:
Modified the code so that instead of passing around the AtomicReference used to save the execution state (which is an implementation detail), now I pass the function that will save the state.

cquiroz:
I've tried to read this PR but it seems my knowledge of scalaz is not enough to follow it. While the comments are certainly useful and give a good indication of what's going, I was not able to follow the details of the implementation.
Perhaps pointers to documentation on the relevant concepts would be useful or we might organize a presentation to go over the details.

tpolecat:
Yeah I will read this in detail today. Sorry for the delay.

tpolecat:
@jluhrs can you separate out the impure methods on `Executor`, perhaps by making `Executor` into a trait and implementing it as `object ExecutorImpl` or something? Also please mark the impure methods with `unsafe` (as in `unsafeStartCmd`)? It will be much easier to follow that way I think.

tpolecat:
:-1: there are some issues with the state management effects. I will add a longer comment. Hold off for now.

tpolecat:
@jluhrs  I made a PR to your branch with some suggested changes.

jluhrs:
I just merged Rob's suggestions.

tpolecat:
Ok if it's working to your satisfaction then :+1:

PR #500, opened 2015-07-10
OCSADV-399-B

tpolecat:
In preparation for time-based coordinates this PR makes an `Option<SchedulingBlock>` available to some of the common "context" objects that we construct and pass around when working with observations. This value is currently unused.

- `ObsContext` (used in many places)
- `ObservationElements` (used in the P2 checker)
- `TpeContext` (used in the TPE)

### DIFF
--- a/bundle/edu.gemini.ags.servlet/src/main/java/edu/gemini/ags/servlet/ToContext.java
+++ b/bundle/edu.gemini.ags.servlet/src/main/java/edu/gemini/ags/servlet/ToContext.java
@@ -6,6 +6,7 @@ import edu.gemini.skycalc.HHMMSS;
 import edu.gemini.shared.util.immutable.Option;
 
 import edu.gemini.shared.util.immutable.Some;
+import edu.gemini.shared.util.immutable.None;
 import edu.gemini.spModel.core.Site;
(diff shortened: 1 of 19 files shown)


swalker2m:
Makes sense and looks good to me.  :+1:

cquiroz:
Seems good, but quickly what is the `SchedulingBlock` the code is not too explanatory

swalker2m:
The `SchedulingBlock` describes a reference time for when the observation will take place and the expected duration.  We need it already for determining the coordinates to use for a non-sidereal target and for calculating the parallactic angle.  We are going to need it everywhere once getting coordinates requires providing a time in general.

tpolecat:
Most observations do not have this time set, but it will soon be required for observations with nonsidereal targets. This will subsume the existing "valid at" time for conic targets.

cquiroz:
Thanks, why is that the times are `String`, could this be a good place to start using the 1.8 time classes?

tpolecat:
It's a pair of `Longs`, although there is a constructor that takes a pair of `String`s. Not sure what that's all about.

swalker2m:
I think what you're referring to is just used for parsing the time value from XML.

PR #501, opened 2015-07-14
OCSADV-402: Improve magnitude filtering in CatalogQuery

cquiroz:
This PR is a second attempt to make a better API for the catalog bundle. Thanks to the comments on PR #497 I noted that the API could be simplified. The original idea was to simplify the API making `CatalogQuery` just a collection of filters. This idea remains but the way to filter magnitudes has been simplified:

* We used to have `MagnitudeRange` and `MagnitudeConstraints`, they have now been fused into one and the difference is of a `MagnitudeExtractor` to read magnitudes from the target
* `MagnitudeExtractor` was a function but now is trait, making it simpler to do equality comparisons
* Conditions adjustment was a bit convoluted since it used the old Java target classes. It has been replaced by a simple typeclass to do `conditions.adjust(MagnitudeConstraints)`
* Along the way a bug was found on adjusting conditions for GEMS. A Bug report (REL-2420) was added and broken tests are marked as pending


bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/package.scala
```diff
@@ -1,27 +1,24 @@
...
-  type MagnitudeExtractor = SiderealTarget => Option[Magnitude]
 
-  val RLikeBands = List(MagnitudeBand._r, MagnitudeBand.R, MagnitudeBand.UC)
+  val RLikeBands = NonEmptyList(MagnitudeBand._r, MagnitudeBand.R, MagnitudeBand.UC)
 
-  /**
-   * For a given target set of probe bands build a MagnitudeExtractor that returns the first magnitude on the target
-   */
-  def magnitudeExtractor(probeBands: List[MagnitudeBand]): MagnitudeExtractor = (st: SiderealTarget) => probeBands.flatMap(st.magnitudeIn).headOption // Picks the first available magnitude on the target
+  def agsBandExtractor(band: MagnitudeBand): MagnitudeExtractor = if (band === MagnitudeBand.R) FirstBandExtractor(RLikeBands) else SingleBandExtractor(band)
```
swalker2m: Do you need anything other than the `FirstBandExtractor`?  If using `H` or something then you can just provide it a singleton `NonEmptyList`.

It seems weird to ask for a "magnitude extractor" for `R` and get back something that will pick `r` and `UC` as well.  If you ask for `H` or `r` or `UC` for that matter you get just that band.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsMagnitude.scala
```diff
@@ -23,7 +24,7 @@ object AgsMagnitude {
    * Gets the widest possible range limits incorporating the given conditions
    * and speeds.
    */
-  def rangeLimits(mc: MagnitudeCalc, c1: (Conditions, GuideSpeed), c2: (Conditions, GuideSpeed)): MagnitudeRange =
+  private def rangeConstraints(mc: MagnitudeCalc, c1: (Conditions, GuideSpeed), c2: (Conditions, GuideSpeed)): Option[MagnitudeConstraints] =
```
swalker2m: Turning this into an `Option` seems like it would have ripple effects.  What changed to make this optional now?

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/package.scala
cquiroz: That is true, it could be a single class doing both, R-family and H.
At some point it has to be encoded that R expands to r, R, UC and this is the place. I'm open to better suggestions on how to encode that fact but sooner or later you need to express that R expands to the other bands

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsMagnitude.scala
cquiroz: It used to be that the operation run on `MagnitudeRange` whose union method returned `MagnitudeRange`, instead it now it returns `Option[MagnitudeConstraint]` as you cannot do a union if e.g. you have different bands. The effects downstream are actually quite limited in `ProbeLimitsCalc` and `GemsMagnitudeTable`

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsVoTableCatalog.scala
```diff
@@ -73,11 +71,13 @@ case class GemsVoTableCatalog(backend: VoTableBackend = RemoteBackend, catalog:
...
     val queryArgs = for {
       c <- criterions
-      q = CatalogQuery.catalogQueryWithoutBand(basePosition, c.criterion.radiusLimits, c.criterion.magRange.some, catalog)
+      q = CatalogQuery.catalogQuery(basePosition, c.criterion.radiusConstraint, c.criterion.magConstraint, catalog)
     } yield (q, c)
 
     val qm = queryArgs.toMap
-    VoTableClient.catalogs(queryArgs.map(_._1), backend).map(l => l.map(k => GemsCatalogSearchResults(qm.get(k.query).get, k.result.targets.rows)))
+    VoTableClient.catalogs(queryArgs.map(_._1), backend).map(l => l.collect {
+      case QueryResult(q: CatalogQueryWithMagnitudeFilters, results:CatalogQueryResult) => GemsCatalogSearchResults(qm.get(q).get, results.targets.rows)
```
swalker2m: I think Rob mentioned this before but I don't recall the answer.  Why the explicit typed pattern for `results`?  It looks like `results` is always a `CatalogQueryResults` anyway.

cquiroz: Right, `q` may vary but I could leave the type of results off

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsVoTableCatalog.scala
```diff
@@ -172,31 +158,18 @@ case class GemsVoTableCatalog(backend: VoTableBackend = RemoteBackend, catalog:
...
+    // Get max/min limits per band
+    constraintsPerBand.flatMap {
+      case (_, Nil) =>
+        None
+      case (_, h :: Nil) =>
+        Some(h)
+      case (_, h :: tail) =>
+        tail.foldLeft(h.some) { (a, b) =>
+          a >>= (_ union b)
+        }
```
swalker2m: Don't the last two cases amount to the same thing?

cquiroz: :+1:

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/api/CatalogQuery.scala
```diff
@@ -1,31 +1,31 @@
...
+}
+
+case class RadiusFilter(base: Coordinates, rc: RadiusConstraint) extends QueryResultsFilter {
+  def filter(t: SiderealTarget): Boolean = rc.targetsFilter(base)(t)
+}
+
+case class MagnitudeQueryFilter(filters: NonEmptyList[MagnitudeConstraints]) extends QueryResultsFilter {
+  def filter(t: SiderealTarget): Boolean = filters.list.forall(_.filter(t))
+}
+
```
swalker2m: I think I might have made a single `CatalogQuery` case class with base position, radius constraint, and magnitude constraints.  You can do the filtering in your `filter` method based on those.  I'm not sure I follow what is going on with the separate `QueryResultsFilter`, `RadiusFilter`, `MagnitudeQueryFilter`, `BandConstrainedCatalogQuery`, `RangeConstrainedCatalogQuery` and the various constructors.  What do they buy us?  `CatalogQuery` is `sealed` so as a user of this API there's no way to define and apply your own filters anyway?

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/api/MagnitudeConstraints.scala
```diff
@@ -45,36 +47,57 @@ object SaturationConstraint {
...
 
-    val fl = FaintnessConstraint(mappedFaintness.value)
-    val sl = mappedSaturation.map(s => SaturationConstraint(s.value))
+/**
+ * Finds the first on a list of bands
+ */
+case class FirstBandExtractor(bands: NonEmptyList[MagnitudeBand]) extends MagnitudeExtractor {
+  override def extract(t: SiderealTarget) = bands.map(t.magnitudeIn).list.find(_.isDefined).flatten
+  override def bandSupported(b: MagnitudeBand) = bands.list.contains(b)
+}
```
swalker2m: I sort of regret the "r like" band idea.  We might have just left everything defined in terms of a single band and just run the search (on the same candidates) 3 times?  At any rate, now that we have it, maybe we should push this idea up so that we aren't passing around single bands but rather something like this extractor idea that groups and handles mapping to the "contained" if you will bands.  It's just seems too weird to ask for `R` and get `UC`, etc.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/obscomp/SPSiteQuality.java
```diff
@@ -399,12 +349,14 @@ public String displayValue() {
             return _displayValue;
         }
 
+        public double magAdjustment() {
+            return _magAdjustment;
+        }
+
```
swalker2m: Perhaps we should go all or nothing on this change?  That is, maybe we just move this out into the same place the rest of the logic went?

cquiroz: That is one option, it used to be kind of peppered in a few different ways. At least in this manner, it is consistent for all conditions

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/api/MagnitudeConstraints.scala
andrewwstephens: I don't know how the filtering is done now, but note that running the filter three times would yield different results than merging the measurements and filtering once.  The idea of merging is that for each star we take the most accurate measurement available and then filter on that.  So if a star has measurements in r, R, and UC, then we take the r value (which may be higher or lower than the other two).  If only UC is available then we use that.  We then filter this merged list of stars.  However, I think running the filter three times would reject a star if any measurement (r, R, or UC) is outside the requested range.

swalker2m: You're right of course.  Carlos is looking into a change to the API that makes the multi-band concept explicit so there are never any surprises like asking for R and getting a UC result.

cquiroz: Asking for R and getting UC as a result is exactly what the change is about. We want something close to R so we check if either r' or R or UC is available
Making it more explicit on the code is certainly ok but the concept of one band being represented by one of those three choices should remain

swalker2m: Right, agreed.  I just want it to be clear that you are querying for and getting results based on one of r', R, or UC.  I think if you explicitly query for R the results should be relative to R and nothing else, just as for H or other bands. So as you say, something that explicitly represents the concept of "one of these bands, in this order of preference".

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/GemsStrategy.scala
```diff
@@ -61,38 +59,36 @@ trait GemsStrategy extends AgsStrategy {
...
-    val adjustedConstraints = catalogQueries(ctx, mt).map { constraint =>
-      // Adjust the magnitude limits for the conditions.
-      val adjustedMagConstraints = constraint.magnitudeConstraints.map(c => c.map(m => ctx.getConditions.magAdjustOp().apply(m.toOldModel).toNewModel))
-      constraint.withMagnitudeConstraints(adjustedMagConstraints)
-    }
+    // Maps from IDs to guide probe band
+    def GuideProbeBandMap = Map[Int, MagnitudeBand](
+      CanopusTipTiltId -> MagnitudeBand.R,
+      OdgwFlexureId    -> MagnitudeBand.H
+    )
```
swalker2m: Should this be `R` or the new bands list thing?  Should `r` and `UC` not work as well?

cquiroz: This should be changed to be a `BandList` too

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/api/CatalogQuery.scala
```diff
@@ -1,31 +1,30 @@
...
+/**
+ * Represents a query on a acatalog, indicating the location, radius and an optional set of magnitude constraints
+ */
+case class CatalogQuery(id: Option[Int], base: Coordinates, radiusConstraint: RadiusConstraint, magnitudeConstraints: List[MagnitudeConstraints], catalog: CatalogName) {
 
-  def withMagnitudeConstraints(magnitudeConstraints: Option[MagnitudeConstraints]): CatalogQuery = this
-  def withMagnitudeRange(magnitudeRange: Option[MagnitudeRange]): CatalogQuery = this
+  val filters: NonEmptyList[QueryResultsFilter] = NonEmptyList(RadiusFilter(base, radiusConstraint), magnitudeConstraints.map(MagnitudeQueryFilter.apply): _*)
+  def filter(t: SiderealTarget):Boolean = filters.list.forall(_.filter(t))
```
swalker2m: What about removing the `QueryResultsFilter` hierarchy altogether since they are just predicates and only used internally?

````scala
val radiusFilter = radiusConstraint.targetsFilter(base)
def filter(t: SiderealTarget): Boolean =  radiusFilter(t) && magnitudeConstraints.filter(t)
````

Comment (cquiroz) [inline bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/api/CatalogQuery.scala]: I left them there in case a user want to add custom filters, you could override the filters list and add your filters.

File: bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/api/MagnitudeConstraints.scala [hunk 10]
```diff
@@ -45,36 +47,74 @@ object SaturationConstraint {
...
+  val bands = NonEmptyList(MagnitudeBand._r, MagnitudeBand.R, MagnitudeBand.UC)
+}
 
-    val fl = FaintnessConstraint(mappedFaintness.value)
-    val sl = mappedSaturation.map(s => SaturationConstraint(s.value))
+/**
+ * Extractor for Nici containing 4 bands
+ */
+case object NiciBandsList extends BandsList {
+  val bands = NonEmptyList(MagnitudeBand._r, MagnitudeBand.R, MagnitudeBand.UC, MagnitudeBand.K)
```
swalker2m: Probably doesn't matter since NICI is retired but a question for Andy is whether R and K are really interchangeable.

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/api/MagnitudeConstraints.scala
```diff
@@ -45,36 +47,74 @@ object SaturationConstraint {
...
+}
 
-    MagnitudeConstraints(mappedFaintness.band, fl, sl)
+object BandsList {
+  implicit val equals = Equal.equal[BandsList]((a, b) => a.bands === b.bands)
+
+  def bandList(band: MagnitudeBand):BandsList = band match {
+    case MagnitudeBand.R  => RBandsList
+    case MagnitudeBand._r => RBandsList
+    case _                => SingleBand(band)
```
swalker2m: Is this used?  There is where the client should probably just specify `RBandsList` or `SingleBand(band)` I think.

As is though, I would expect UC to produce `RBandsList` as well.

cquiroz: It is used in a few places, e.g. `SingleProbeStrategy` may use `RLike` for some instrument but not for others, I tried to limit its use though
The UC is a good question that I did consider. I'm not sure if it is the same the other way around. I'd think you may want to do `UC-only` queries, though the API doesn't really forbid.

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/api/MagnitudeConstraints.scala
andrewwstephens: Not really (R and K are not interchangeable).  According to the AGS limits spreadsheet the NICI WFS requires R or V.

cquiroz:
Added a few commits expanding the use of `BandsLists` where it makes sense inside AGS and in a few other bits of the OT

swalker2m:
Otherwise I think I've held this one up long enough already!  :+1:

PR #502, opened 2015-07-22
REL-2272: Gemini must host all Phase I backend servers

cquiroz:
This PR changes the Phase I backend server's name to phase1.gemini.edu and makes the submissions use HTTPS


bundle/edu.gemini.model.p1.submit/src/main/scala/edu/gemini/model/p1/submit/package.scala
```diff
@@ -19,8 +19,8 @@ package object submit {
     Map(LargeProgram -> url("large_program"), FastTurnaroundProgram -> url("fast_turnaround"))
   }
 
-  val productionSubmissionUrls: Map[SubmitDestination, String] = submissionUrls("http://hbfphase1-lv1.hi.gemini.edu/cgi-bin/backend/production")
+  val productionSubmissionUrls: Map[SubmitDestination, String] = submissionUrls("https://phase1.gemini.edu/cgi-bin/backend/production")
```
jluhrs: I suggest to put the protocol/host name in a separate val, and then use that value to compose the full URL here and in the next line. That way if it is necessary to change it again, the change will be in only one place.

cquiroz:
Incorporated the suggestion and will merge

PR #503, opened 2015-07-22
REL-2352: Add SDSS filter to  target bandpass list in PIT

cquiroz:
Found a small bug on PIT's display of bands. It showed the SDSS bands with the class name rather than the visible name


jluhrs:
:+1:

PR #504, opened 2015-07-22
REL-2290: Update PIT submission success status information

cquiroz:
Micro PR with a description text update according to requirements


jluhrs:
:+1:

PR #505, opened 2015-07-22
REL-2375: Classical visitors list can be edited after submission

cquiroz:
Fixed a few UI controls that should be disabled when a proposal has been submitted

### DIFF
--- a/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/view/tac/TacView.scala
+++ b/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/view/tac/TacView.scala
@@ -131,6 +131,7 @@ class TacView(loc: Locale) extends BorderPanel with BoundView[ProposalClass] { v
 
       override def refresh(m:Option[ProposalClass]) = {
         val newItems = ~m.map(partners)
+        enabled = isTac && m.isDefined
         if (!(~oldItems.map(_ == newItems))) {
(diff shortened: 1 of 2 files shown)


bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/view/tac/TacView.scala
```diff
@@ -188,7 +189,7 @@ class TacView(loc: Locale) extends BorderPanel with BoundView[ProposalClass] { v
       override def refresh(m:Option[Option[SubmissionDecision]]) {
```
swalker2m: Everything looks good here :+1:.  The comment Art was referring to in his IM was one I made about the `Option[Option[SubmissionDecision]]` parameter which looks suspicious.  Perhaps it should have been joined at a higher level but of course that isn't the point of this PR.  At least you did get rid of the no-op `headOption` below.

cquiroz: I agree, that looks odd, I'll quickly review if it could/should be flattened up on the call chain

jluhrs:
:+1:

cquiroz:
Actually the `Option[Option[SubmissionDecision]]` comes naturally from the way the P1 model is handled. Changing it to be flatten would have implications in other places with a bigger impact than having it this way

PR #506, opened 2015-07-23
OCSADV-399-C

tpolecat:
Ok this is the next set of changes in preparation for time-varying coordinates.

- I added the following methods to `ITarget` which simply delegate to the old methods and always return `Some`. The idea is that you pass a time if you have one … in cases where it doesn't matter (like with extragalactic stuff) you don't need to.
  - `Option<Double> getRaDegrees(Option<Long> when)`
  - `Option<Double> getDecDegrees(Option<Long> when)`
- I began changing calls to use these new methods in cases where a `SchedulingBlock` is available (passing the `start` time) when it's clear what should happen if `None` is returned (which never happens at the moment).

The lack of `for` comprehensions or `|@|` in Java is really irritating.

### DIFF
--- a/bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/nifs/NifsRule.java
+++ b/bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/nifs/NifsRule.java
@@ -14,6 +14,7 @@ import edu.gemini.spModel.gemini.nifs.NIFSParams;
 import edu.gemini.spModel.gemini.nifs.NifsOiwfsGuideProbe;
 import edu.gemini.spModel.guide.GuideProbe;
 import edu.gemini.spModel.guide.GuideProbeUtil;
+import edu.gemini.spModel.obs.SchedulingBlock;
 import edu.gemini.spModel.obs.context.ObsContext;
 import edu.gemini.spModel.obsclass.ObsClass;
 import edu.gemini.spModel.obscomp.InstConstants;
@@ -182,6 +183,7 @@ public final class NifsRule implements IRule {
     private static IRule OI_AO_SAME_AS_SCIENCE_TARGET_RULE = new IRule() {
         private static final String MESSAGE = "For on-axis guiding, target cannot be used as OIWFS b/c of vignetting";
 
+        P2Problems prob = new P2Problems();
         public IP2Problems check(ObservationElements elements)  {
             for (ObsContext ctx : elements.getObsContext()) {
                 final SPTarget baseTarget       = ctx.getTargets().getBase();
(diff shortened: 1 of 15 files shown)


bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/nifs/NifsRule.java
```diff
@@ -191,29 +193,31 @@ public IP2Problems check(ObservationElements elements)  {
...
-                        &&
-                        Double.compare(oiC1, aoC1) == 0
-                        &&
-                        Double.compare(oiC2, aoC2) == 0) {
-                    P2Problems prob = new P2Problems();
-                    prob.addError(PREFIX + "OI_AO_SAME_AS_SCIENCE_TARGET_RULE", MESSAGE, elements.getTargetObsComponentNode().getValue());
-                    return prob;
+                aoTarget.getValue().getTarget().getRaHours(when).foreach(aoC1 ->
+                aoTarget.getValue().getTarget().getDecDegrees(when).foreach(aoC2 -> {
+                    if (Double.compare(baseC1, oiC1) == 0
```
cquiroz: How likely is that a target and its ao target would be the same but the coordinates will be slightly different so that the comparison returns a non zero. I just worry that this could lead to false negatives due to rounding error

bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/nifs/NifsRule.java
```diff
@@ -191,29 +193,31 @@ public IP2Problems check(ObservationElements elements)  {
                 if (baseTarget == null || oiTarget.isEmpty() || aoTarget.isEmpty()) return null;
                 //now, let's compare coordinates. If they are the same, raise an error
 
-                double baseC1 = baseTarget.getTarget().getRaHours();
-                double baseC2 = baseTarget.getTarget().getDecDegrees();
+                final Option<Long> when = elements.getSchedulingBlock().map(SchedulingBlock::start);
```
cquiroz: Is this the gemini `Option`?

bundle/jsky.app.ot/src/main/java/jsky/app/ot/tpe/TelescopePosEditor.java
```diff
@@ -306,7 +309,11 @@ private boolean basePosUpdated(TpeContext oldCtx, TpeContext newCtx) {
         ITarget oldBase = oldBasePos.getTarget();
         ITarget newBase = newBasePos.getTarget();
 
-        return (oldBase.getRaDegrees() != newBase.getRaDegrees() || oldBase.getDecDegrees() != newBase.getDecDegrees());
+        Option<Long> oldWhen = oldCtx.schedulingBlockJava().map(SchedulingBlock::start);
+        Option<Long> newWhen = oldCtx.schedulingBlockJava().map(SchedulingBlock::start);
+
+        return (oldBase.getRaDegrees(oldWhen) != newBase.getRaDegrees(newWhen) ||
```
cquiroz: `equals?`

bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/nifs/NifsRule.java
swalker2m: Usually targets aren't computed but rather typed in directly or more likely looked up in simbad or guide star catalogs.  Here we'll potentially be accounting for proper motion at some point but the same calculation will apply to all the targets and will compute the same values so nothing about how this check works is fundamentally changing.

You're right though and I imagine it would make sense to define some minimum angular separation that still counts as the same star if possible to do in all cases(?).  We'd need to hunt down all the places we do this kind of thing and use the same criterion everywhere as a separate issue.

swalker2m:
Except for the `equals` vs `==` bug in Java bug that Carlos found, this looks good to me.

bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/nifs/NifsRule.java
cquiroz: I've seen a few cases where the UI rounds numbers to 2 decimals but the catalog can be more precise, especially when the angles are represented as decimals.
This for example makes some queries not to hit the cache even though logically it should
I'm not sure what an appropriate delta would be

cquiroz:
Looks good too, I wonder if Java is too limiting sometimes, would it make sense to just port classes or methods to scala?

bundle/jsky.app.ot/src/main/java/jsky/app/ot/tpe/TelescopePosEditor.java
```diff
@@ -306,7 +309,11 @@ private boolean basePosUpdated(TpeContext oldCtx, TpeContext newCtx) {
         ITarget oldBase = oldBasePos.getTarget();
         ITarget newBase = newBasePos.getTarget();
 
-        return (oldBase.getRaDegrees() != newBase.getRaDegrees() || oldBase.getDecDegrees() != newBase.getDecDegrees());
+        Option<Long> oldWhen = oldCtx.schedulingBlockJava().map(SchedulingBlock::start);
+        Option<Long> newWhen = oldCtx.schedulingBlockJava().map(SchedulingBlock::start);
```
arturog8m: Shouldn't  `newWhen` be obtained from `newCtx`? As it is looks like `oldWhen` and `newWhen` are the same thing. Is that the intention?

tpolecat: Good catch, thanks.

bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/nifs/NifsRule.java
tpolecat: Yeah I decided not to question the existing logic. If it's questionable we should open another issue.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/tpe/TelescopePosEditor.java
tpolecat: Yep, good. Thanks. Java is hard.

bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/nifs/NifsRule.java
tpolecat: Yes, Gemini `Option` right now. I guess we should replace it with Java `Optional` at some point.

tpolecat:
Fixed those two bugs. Thanks!

PR #508, opened 2015-07-24
OCSADV-398: Add controls to make custom queries

cquiroz:
This PR changes the UI of the new catalog navigator letting the user defines queries. On startup it will show the last query parameters and the user can change it accordingly. Here is how the new UI looks like:

screenshot 2015-07-24 14 02 03

Some notes:
* Most of the code is just UI, setting up controls, updating the values after a query completes, etc.
* There are a few UI classes duplicated on the PIT and the OT and they were consolidated
* The Catalog query cache wasn't handle properly the new filtering style from OCSADV-402, that was fixed as well as other remediation for some cases when the cache wasn't hit. Still, it would be good to implement OCSADV-271 to improve the cache's efficiency
* Sometimes the query button doesn't work at once, requiring a second click, in particular if focus is on a text field. This seems the same issue on REL-2253	and it will be addressed separately


arturog8m:
The UI looks good. I have a couple of suggestions for improvements:
1. I think the table should be aligned with the border of the "Query Params" widget. As it is now it looks weird - the query results number kind of messes thing up in my opinion.
2. I would add the "Query Results: 11" at the bottom of the UI, in some sort of status field. This can be also used to show other information relevant to the query (like what Catalog is being queried, or stuff like that). I think that would be useful.

cquiroz:
I thought about aligning the border bit it is quite tricky, especially given that fonts are system dependent. If I were to remove the Query Results perhaps it is better:
screenshot 2015-07-24 14 23 42

arturog8m:
OK. At any rate, just removing the Query Results looks better in my
opinion. What do you think about a status field at the end?

Also, perhaps the query params can be put in one of those widgets that a
user can show/hide, as the Phase 2 Checks? Then the user can have more
space horizontally to look at the actual results.

cquiroz:
In my ideal de border of the Query Params (which is identical to the ones on the target editor) should align with the top part of the table

I was thinking to put a status area at the bottom, next to the Close button to show errors and results

Personally I don't like the slider controls to adjust the width, especially the swing one looks really dated (It is from the early 2000s after all)

swalker2m:
> Personally I don't like the slider controls to adjust the width, especially the swing one looks really dated (It is from the early 2000s after all)

Are you referring to a `SplitPane`?  I think Art is talking about a slide out drawer like the P2 check feedback in the OT.  That might be a good idea but should probably open up expanded and only slide in on explicit request.

What happens when you click `-` on the last magnitude value?  Perhaps it should use the same plus and x icons in a flattened button that are used elsewhere in the OT?

Why would we need more than one plus button?  Is the order of the magnitude filters important?

The range might not be intuitive?  Or maybe it is but I would expect it to be the other way around (saturation limit - faintness limit).  At any rate, a label over the columns might be a good idea?

cquiroz:
* I was referring to `SplitPane`, the slide out would be a reasonable option though

* When you remove all the (-) you get this:
screenshot 2015-07-24 15 05 45

Meaning, no filtering but you can still add filters later on
* The order of the filters is not relevant, the behave like a big AND applied to each target
* Multiple (+) buttons is debatable. Right now if you click (+) it would duplicate the row you click, Having only one (+) should copy the last or perhaps use a default with a large range
( Good suggestion about the labels and switching the order

cquiroz:
What do you think about this layout:

It makes the table use the same "group-style" as the query form, it adds Saturation and faintness constraint, and uses the same title for PM as the target editor: µ RA and µ Dec

screenshot 2015-07-28 11 17 06

andrewwstephens:
I like using the correct symbols for proper motion, and the 'Query result" count in the bottom left.  For the magnitude labels, maybe use "Bright" and "Faint" ?

cquiroz:
Screenshot with the latest updates

screenshot 2015-07-31 12 41 40

bundle/edu.gemini.shared.gui/src/main/scala/edu/gemini/shared/gui/textComponent/NumberField.scala
```diff
@@ -27,7 +27,7 @@ class NumberField(d: Option[Double]) extends FormattedTextField(NumberField.df)
   private val pink = new Color(255, 224, 224)
   private val white = background // N.B. this means we can't change the background to anything else
 
-  var valid = true
+  @volatile var valid = true
```
swalker2m: Could you explain a bit what prompted the `volatile` annotation?  On the surface it is confusing because Swing is single threaded anyway.  If it is being accessed from another thread perhaps it should be shunted into the event loop with `Swing.onEDT` or something?

cquiroz: Sorry, that was part of my ongoing investigation of why the FormattedTextField emits an extra empty event that confuses clicking on the ok button (REL-2253). As you say shouldn't be necessary as long as we are in Swing

PR #510, opened 2015-07-27
REL-2431: GPI target errors when all requested magnitudes entered

cquiroz:
Small leftover from the transition to the new target model in OCSADV-155

### DIFF
--- a/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
+++ b/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
@@ -11,13 +11,14 @@ import edu.gemini.spModel.core.MagnitudeBand
 import view.obs.ObsListGrouping
 import edu.gemini.model.p1.visibility.TargetVisibilityCalc
 import edu.gemini.pit.ui.view.tac.TacView
-import scalaz._
-import Scalaz._
 import java.text.SimpleDateFormat
 import java.io.File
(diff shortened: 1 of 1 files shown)


bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
```diff
@@ -208,7 +208,7 @@ class ProblemRobot(s: ShellAdvisor) extends Robot {
 
     private val gpiCheck = {
       def gpiMagnitudesPresent(target: SiderealTarget):List[(Severity, String)] = {
-        val requiredMagnitudes = "I" :: "Y" :: "J" :: "H" :: "K" :: Nil
+        val requiredMagnitudes = MagnitudeBand.I :: MagnitudeBand.Y :: MagnitudeBand.J :: MagnitudeBand.H :: MagnitudeBand.K :: Nil
         val obsMagnitudes = target.magnitudes.map(_.band)
         if (!requiredMagnitudes.forall(obsMagnitudes.contains)) {
```
swalker2m: We should be able to arrange it so that the compiler can verify this.  Something like:

````scala
        import MagnitudeBand.{I, Y, J, H, K}
        val requiredBands = Set(I, Y, J, H, K)
        val obsBands      = target.magnitudes.map(_.band).toSet
        if ((requiredBands & obsBands) === requiredBands) Nil
        else List((Severity.Error, "The magnitude information in the GPI target component should include the bandpasses I, Y, J, H, and K"))
````

cquiroz:
Updated according to comments

PR #511, opened 2015-07-31
REL-2436: Activate AGS for Phoenix in PIT

cquiroz:
This PR adds Phoenix support to AGS. It provides the same strategy used for Visitor instruments and multiple tests. Most of the lines are just catalog data for the tests
It was also requested to print on the PIT the query being executed on AGS calls


swalker2m:
The code looks fine. :+1:  Very little was updated compared to the amount of XML files for testing.  I wonder if we could share all these data files across all the AGS tests.

cquiroz:
The xml is different for each target, if we could test on each target we should be able to share the xml but I get different sample targets for different instruments

PR #512, opened 2015-08-01
REL-2420: Wrong AGS GEMS conditions adjustments

cquiroz:
While doing OCSADV-402 a bug on the conditions adjustments was found that affects GEMS AGS. The bug was fixed in OCSADV-402 but tests were commented out. This PR reenables and fixes the tests


swalker2m:
:+1:

PR #515, opened 2015-08-03
OCSADV-399-E

tpolecat:
More prep for time-varying targets (2/6).
- Update `[Obs]TargetDesc` to take a function to compute coordinates.
- Remove unused target table from the elevation plot ... it allows *setting* the coordinates, which we don't want to deal with right now. And lucky we don't have to since the code is unused.
See also inline comments below.


bundle/jsky.app.ot/src/main/java/jsky/app/ot/viewer/SPElevationPlotPlugin.java
```diff
@@ -136,10 +136,6 @@ public void stateChanged(ChangeEvent e) {
     private void _addMenuItems(ElevationPlotMenuBar menuBar) {
         JMenu menu = menuBar.getViewMenu();
 
-        // need to disable manually adding and deleting targets, since
-        // it conflicts with the target lists in the observations
-        menu.remove(1);
-
         menu.add(_createViewLabelMenu());
```
tpolecat: The target list menu item is always removed here, but it no longer needs to be removed because it is never added (see below).

bundle/jsky.elevation.plot/src/main/java/jsky/plot/ElevationPlotMenuBar.java
```diff
@@ -71,7 +71,6 @@ protected JMenu createFileMenu() {
     protected JMenu createViewMenu() {
         JMenu menu = new JMenu(_I18N.getString("view"));
         menu.add(_plotPanel.getDateAction());
-        menu.add(_plotPanel.getTargetsAction());
         menu.add(_showLegendMenuItem = _createViewLegendMenuItem());
```
tpolecat: The target list menu item is not added anymore, and thus doesn't need to be removed (see above).

cquiroz:
Nice, lots of red :+1:

bundle/jsky.elevation.plot/src/main/java/jsky/plot/ElevationPlotModel.java
```diff
@@ -414,10 +418,13 @@ public XYDataset getConstraintsDataset() {
                 endAlt   = ImprovedSkyCalcMethods.getAltitude(elMin);
                 break;
             case HOUR_ANGLE:
-                double dec = target.getCoordinates().getDecDeg();
-                double lat = _site.latitude;
-                startAlt = ImprovedSkyCalcMethods.altit(dec, elMin, lat);
-                endAlt   = ImprovedSkyCalcMethods.altit(dec, elMax, lat);
+                final Option<WorldCoords> coords = target.getCoordinates(None.instance());
```
swalker2m: Should we be reading the time from the scheduling block here?

tpolecat: Hmm yes, I believe so. Thanks.

bundle/jsky.elevation.plot/src/main/java/jsky/plot/ElevationPlotModel.java
```diff
@@ -426,6 +429,14 @@ public XYDataset getConstraintsDataset() {
...
 
+    // Altitude in degrees at the target's declination at the given time, the given hour angle, and
+    // the observer's latitude, or None if the target's coordinates are unknown.
+    private static Option<Double> altit(TargetDesc target, Date when, double ha, double lat) {
+        return target.getCoordinates(new Some(when.getTime())).map(coords ->
+            ImprovedSkyCalcMethods.altit(coords.getDecDeg(), ha, lat)
+        );
+    }
+
     // Returns true if the given elevation value meets the constraints, otherwise false.
```
tpolecat: So, I changed it to compute `start/endAlt` at the start and end dates. I'm not sure if that is correct but it doesn't seem any worse than what's there now.

swalker2m:
:+1:

PR #517, opened 2015-08-04
OCSADV-399-F

tpolecat:
More time-based target updates. (3/6)
- Update `SPTarget.hasDefinedTarget` logic to match current logic in effect: `true` unless it's an `HmsDegTarget` with coordinates at `(0.0, 0.0)`.
- Update `SPTargetSkyObjectTest` trivially.
- Update `TelescopePosTableWidget` to support unknown coordinates (still more to do).


swalker2m:
:+1:

PR #518, opened 2015-08-04
OCSADV-389: basic implementation of TCS configuration.

jluhrs:
This is an incomplete implementation. I created the pull request to have feedback on the design I am following.
The topic is a bit complicated, because it deals with the details of the TCS. I included a document to help understand some of those details.

### DIFF
--- a/bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Instrument.scala
+++ b/bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Instrument.scala
@@ -8,6 +8,7 @@ import scalaz.concurrent.Task
  * Created by jluhrs on 4/22/15.
  */
 trait Instrument extends System {
+  // The name used for this instrument in the science fold configuration
   val sfName: String
   def observe(config: Config): SeqAction[ObserveResult]
 }
--- a/bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Instrument.scala
+++ b/bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Instrument.scala
@@ -8,6 +8,7 @@ import scalaz.concurrent.Task
  * Created by jluhrs on 4/22/15.
  */
 trait Instrument extends System {
+  val sfName: String
   def observe(config: Config): SeqAction[ObserveResult]
 }
 
--- a/bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Instrument.scala
+++ b/bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Instrument.scala
@@ -1,6 +1,10 @@
 package edu.gemini.seqexec.server
 
+import java.util.logging.{Level, Logger}
+
 import edu.gemini.spModel.config2.Config
+import edu.gemini.spModel.gemini.gmos.InstGmosSouth._
+import edu.gemini.spModel.seqcomp.SeqConfigNames._
 
 import scalaz.concurrent.Task
 
@@ -16,3 +20,20 @@ trait Instrument extends System {
 //Placeholder for observe response
 case class ObserveResult(dataId: String)
 
+object UnknownInstrument extends Instrument {
+
+  override val name: String = "UNKNOWN"
+
+  override val sfName: String = "unknown"
+
+  var imageCount = 0
+
+  override def configure(config: Config): SeqAction[ConfigResult] = Task {
+    TrySeq(ConfigResult(this))
+  }
+
+  override def observe(config: Config): SeqAction[ObserveResult] = Task {
+    imageCount += 1
+    TrySeq(ObserveResult(f"S20150519S$imageCount%04d"))
+  }
+}
--- a/bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/TcsController.scala
+++ b/bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/TcsController.scala
@@ -1,6 +1,10 @@
 package edu.gemini.seqexec.server
 
-import edu.gemini.spModel.core.{Wavelength, Offset}
+import edu.gemini.spModel.core.Wavelength
+
+import scalaz.{\/-, \/}
+
+import squants.{Length, Angle}
 
 /**
  * Created by jluhrs on 7/30/15.
@@ -21,6 +25,8 @@ trait TcsController {
 }
 
 object TcsController {
+
+  case class Requested[T](self: T) extends AnyVal
 
   sealed trait TipTiltSource
 
@@ -45,21 +51,26 @@ object TcsController {
 
   }
 
-  sealed trait M2GuideConfig
+  sealed trait M2GuideConfig { val active: Boolean }
 
-  object M2GuideOff extends M2GuideConfig
+  object M2GuideOff extends M2GuideConfig { override val active = false}
   
-  sealed trait ComaOption
-  object ComaOn extends ComaOption
-  object ComaOff extends ComaOption
+  sealed trait ComaOption { val active: Boolean }
+  object ComaOn extends ComaOption { override val active = true}
+  object ComaOff extends ComaOption { override val active = false}
 
-  final case class M2GuideOn(coma: ComaOption, source: Set[TipTiltSource]) extends M2GuideConfig
+  final case class M2GuideOn(coma: ComaOption, source: Set[TipTiltSource]) extends M2GuideConfig {
+    override val active = true
+    def setComa(v: ComaOption) = M2GuideOn(v, source)
+    def setSource(v: Set[TipTiltSource]) = M2GuideOn(coma, v)
+  }
 
-  sealed trait M1GuideConfig
 
-  object M1GuideOff extends M1GuideConfig
+  sealed trait M1GuideConfig { val active: Boolean }
 
-  final case class M1GuideOn(source: M1Source) extends M2GuideConfig
+  object M1GuideOff extends M1GuideConfig { override val active = false }
+
+  final case class M1GuideOn(source: M1Source) extends M1GuideConfig  { override val active = true}
 
   sealed trait Beam
 
@@ -83,7 +94,7 @@ object TcsController {
     def get(nodchop: NodChop): NodChopTrackingOption
   }
 
-  // If x is of type ActiveNodChopTracking then ∃ a:NodChop ∍ x.get(a)
+  // If x is of type ActiveNodChopTracking then ∃ a:NodChop ∍ x.get(a) == NodChopTrackingOn
   // How could I reflect that in the code?
   sealed trait ActiveNodChopTracking extends NodChopTrackingConfig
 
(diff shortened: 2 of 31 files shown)


bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Instrument.scala
```diff
@@ -8,6 +8,7 @@ import scalaz.concurrent.Task
  * Created by jluhrs on 4/22/15.
  */
 trait Instrument extends System {
+  val sfName: String
```
cquiroz: Can you document what `sfName` stands for?

bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/TcsController.scala
```diff
@@ -0,0 +1,203 @@
...
+  def getConfig: SeqAction[TcsConfig]
+
+  def guide(gc: GuideConfig): SeqAction[Unit]
+
+  def applyConfig(tc: TelescopeConfig, gtc: GuidersTrackingConfig, ge: GuidersEnabled, agc: AGConfig): SeqAction[Unit]
+}
+
+object TcsController {
+
+  case class Offset(p: Double, q: Double)
```
cquiroz: We have an `Offset` class on the target model, would it make sense to reuse it?

bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/TcsController.scala
```diff
@@ -0,0 +1,203 @@
...
+
+  object Offset {
+
+    object Null extends Offset(0.0, 0.0)
+
+  }
+
+  type Wavelength = Double
```
cquiroz: There is also a `Wavelength` class in ITC I think

bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/TcsController.scala
```diff
@@ -0,0 +1,203 @@
...
+
+    object None extends NodChopTrackingConfig {
+      def get(nodchop: NodChop): Boolean = false
+    }
+
+    object Normal extends ActiveNodChopTracking {
+      def get(nodchop: NodChop): Boolean = nodchop.nod == nodchop.chop
+    }
+
+    // Must be a non-empty set
```
cquiroz: Maybe you can add a `require(s.nonEmpty)`. I couldn't find an `NonEmptySet` like the one for `List`

bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/TcsControllerSim.scala
```diff
@@ -0,0 +1,51 @@
...
+object TcsControllerSim extends TcsController {
+
+  val guideState = newTaskRef(GuideConfig(false, M1GuideOff, M2GuideOff))
+  val telescopeState = newTaskRef(TelescopeConfig(Offset.Null, Offset.Null, Offset.Null, 445e-9, 445e-9, 445e-9, Beam.A))
+  val guidersTrackState = newTaskRef(GuidersTrackingConfig(ProbeTrackingConfig.Parked, ProbeTrackingConfig.Parked, ProbeTrackingConfig.Parked, ProbeTrackingConfig.Parked))
+  val guidersActivityState = newTaskRef(GuidersEnabled(false, false, false, false))
+  val agState = newTaskRef(AGConfig(ScienceFoldPosition.Parked, HrwfsPickupPosition.Parked))
+  private val Log = Logger.getLogger(getClass.getName)
+
+  override def getConfig: SeqAction[TcsConfig] = for {
```
cquiroz: In a non-sim version, this will go to the EPICS channels to get the state, right?

bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/TcsControllerSim.scala
```diff
@@ -0,0 +1,51 @@
...
+      _ <- guidersActivityState.flatMap(_.put(ge))
+      _ <- agState.flatMap(_.put(agc))
+    } yield TrySeq(())
+
+  }
+
+  override def guide(gc: GuideConfig): SeqAction[Unit] = {
+    Log.log(Level.INFO, "Applying guiding configuration")
+
+    Thread.sleep(1000)
```
cquiroz: Where does the 1-sec sleep come from?

cquiroz:
In general it seems a good model and **great** that you did documentation. As I asked on the code I think we can reuse some classes from `spModel` to avoid duplication
The Seqexec in general has many "custom" waits that have come up along many years of experience, How would that work in the new Seqexec? Would for example `telescopeState` know that it has to wait x seconds before reading/writing?

bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/TcsController.scala
jluhrs: Yes, it makes sense.

bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/TcsController.scala
jluhrs: Yes, I had the same idea about a NonEmptySet.

bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/TcsControllerSim.scala
jluhrs: Exactly. This allows me to test all the logic without depending on external systems.

bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/TcsControllerSim.scala
jluhrs: The sleeps are there just to make execution of the commands non instantaneous.

jluhrs:
@cquiroz My goal is to put all those timing dependencies down on TcsControllerReal, the class that will deal with the real EPICS system.

tpolecat:
I just started looking at this, but I have to say that the initial doc is really helpful.

bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/TcsController.scala
```diff
@@ -0,0 +1,203 @@
...
+
+    object HRWFS extends M1Source
+
+  }
+
+  sealed trait M2GuideConfig
+
+  object M2GuideOff extends M2GuideConfig
+
+  case class M2GuideOn(coma: Boolean, source: Set[TipTiltSource]) extends M2GuideConfig
```
tpolecat: Please make all case classes `final`.

bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/TcsController.scala
tpolecat: You can use `OneAnd[Set, NodChop]`.

bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/TcsController.scala
```diff
@@ -0,0 +1,203 @@
...
+
+  sealed trait ScienceFoldPosition
+
+  object ScienceFoldPosition {
+
+    import LightSource._
+
+    object Parked extends ScienceFoldPosition
+
+    class Position(source: LightSource, sink: Instrument) extends ScienceFoldPosition {
```
tpolecat: Why not `case class`?

bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/TcsController.scala
```diff
@@ -0,0 +1,203 @@
...
+        case AO => "ao2" + sink.sfName
+        case GCAL => "gcal2" + sink.sfName
+      }
+    }
+
+  }
+
+  case class GuideConfig(mountGuide: Boolean, m1Guide: M1GuideConfig, m2Guide: M2GuideConfig)
+
+  case class TelescopeConfig(offsetA: Offset, offsetB: Offset, offsetC: Offset, wavelA: Wavelength, wavelB: Wavelength,
```
tpolecat: It's may be worth having distinct types for offsets and wavelengths in A/B/C, which will prevent them from getting mixed up. Similar with `GuidersTrackingConfig` below.

bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/TcsControllerSim.scala
```diff
@@ -0,0 +1,51 @@
...
+  val agState = newTaskRef(AGConfig(ScienceFoldPosition.Parked, HrwfsPickupPosition.Parked))
+  private val Log = Logger.getLogger(getClass.getName)
+
+  override def getConfig: SeqAction[TcsConfig] = for {
+    a <- guideState.flatMap(_.get)
+    b <- telescopeState.flatMap(_.get)
+    c <- guidersTrackState.flatMap(_.get)
+    d <- guidersActivityState.flatMap(_.get)
+    e <- agState.flatMap(_.get)
+  } yield TrySeq(TcsConfig(a, b, c, d, e))
```
tpolecat: This may not do what you think ... `getConfig` is going to construct a *new* config each time you run it.

bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/TcsControllerSim.scala
```diff
@@ -0,0 +1,51 @@
...
+  override def applyConfig(tc: TelescopeConfig, gtc: GuidersTrackingConfig, ge: GuidersEnabled, agc: AGConfig): SeqAction[Unit] = {
+    Log.log(Level.INFO, "Applying TCS configuration")
+
+    Thread.sleep(2000)
+    for {
+      _ <- telescopeState.flatMap(_.put(tc))
+      _ <- guidersTrackState.flatMap(_.put(gtc))
+      _ <- guidersActivityState.flatMap(_.put(ge))
+      _ <- agState.flatMap(_.put(agc))
+    } yield TrySeq(())
```
tpolecat: Similarly, this is going to create new reference cells, put values in them, and then throw them away. You need to pass this config stuff around.

tpolecat:
Something to consider is all the booleans ... I would prefer to have, for instance `ComaOn/Off`, `MountGuideOn/Off`, and so on. This makes it very clear what's going on and prevents boolean values from being used in the wrong place. Just something to think about.

jluhrs:
* Used types from spModel for Offset and Wavelength.
* Created special types to replace all Booleans.
* Created specific parameter types to avoid confusion.

jluhrs:
I'm working on implementing the TCS control with EPICS (using epics acm package). That will complete this task, and then I'll merge and close this PR.

jluhrs:
I updated the PR with the code to interact with the TCS EPICS interface using the library epics.acm (work still in progress). I'm trying to put all the interactions with the TCS EPICS interface in a single object at the bottom, and then "scalafy" its behavior.

tpolecat:
@jluhrs do you want me to look at this again now or wait until you're more done?

cquiroz:
Were you planning on adding tests?

jluhrs:
@tpolecat I would like to have some opinions on the code that wraps the epics.acm package, in files TcsEpics.scala and EpicsCommand.scala. The epics.acm interface uses callbacks and exceptions, and the wrapper transforms it to use conjunctions and tasks.

jluhrs:
I've just committed the code used for the demo. Some pointers to understand the code:
* TcsController offers the functions to change the TCS configuration, and a series of traits, classes and objects to represent such configuration. I tried to create a more logical representation of the TCS configuration, so it is not a one to one representation of the TCS parameters. The application of the configuration is separated in guiding, and everything else.
* TcsEpicsController is a concrete implementation of TcsController, that deals with the actual TCS system. It makes the transition between my logical representation of the TCS configuration and actual TCS parameters.

jluhrs:
Working on OCSADV-457, I realized that errors in TCS configuration not passed up, but just ignored. The problem is present in this PR, but fixed in OCSADV-457.

Please take a look at this PR, I want to merge it soon.

bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/TcsEpics.scala
```diff
@@ -0,0 +1,355 @@
...
+ * Created by jluhrs on 10/1/15.
+ */
+
+final class TcsEpics private () {
+  import TcsEpics._
+  import EpicsCommand.setParameter
+
+
+  // This is a bit ugly. Commands are triggered from the main apply record, so I just choose an arbitrary command here.
+  // Triggering that command will trigger all the marked commands.
```
jluhrs: This needs a bit more explanation. In EPICS the CAD records are used to execute commands (with or without parameters). There are also apply records, used to trigger several commands at once. Apply records can be cascaded, and in each Gemini EPICS system there is a top level apply record. The way it is done is: first we mark the CAD we want to trigger (writing a value to a parameter input counts as marking a CAD), and then trigger the top apply record. All the marked CAD records will be eventually executed. The top apply record has a CAR record associated, which indicates the state of the commands executed.
What all means is that to trigger a command and monitor its execution, we need to trigger the top apply record and monitor the top CAR record, and that is actually what the CaCommandSender post method does. But if another CAD record was marked, it also will be executed. That fact is used, for example, in the implementation of TcsControllerEpics.config. There is a series of calls that write parameters on different TCS CAD records, followed by a post that triggers the top apply. That relationship between commands and the top apply is not represented in the code.

jluhrs:
Merged as part of PR #744

PR #520, opened 2015-08-11
OCSADV-399-H

tpolecat:
Ok this fixes the guider boundary checking rules to handle the case where the coordinates of the base position are unknown. In this case boundary warnings don't apply and no errors are generated.


bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/obs/context/ObsContext.java
```diff
@@ -273,6 +273,16 @@ public Coordinates getBaseCoordinates() {
...
+    public Option<Coordinates> getBaseCoordinatesOpt() {
+        final Option<Long> when = getSchedulingBlock().map(SchedulingBlock::start);
+        SPTarget target = targets.getBase();
+        return
+            target.getTarget().getRaDegrees(when).flatMap(raDeg ->
+            target.getTarget().getDecDegrees(when).map(decDeg ->
+                new Coordinates(raDeg, decDeg)
+            ));
+    }
+
```
tpolecat: The non-`Option` version will be removed in a later PR and this one will be renamed simply to `getBaseCoordinates`.

swalker2m:
:+1:

PR #521, opened 2015-08-12
OCSADV-381: F2 OI vignetting implemented.

sraaphorst:
Implementation of the vignetting feature for the F2 OI guide probe: arm angle calculation fixed to properly account for offsets.
TPE feature code changed to incorporate moving of F2 science area and OI probe arm to the model.


bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/flamingos2/F2OiwfsProbeArm.scala
```diff
@@ -58,17 +59,31 @@ object F2OiwfsProbeArm extends ProbeArmGeometry {
...
+   * @return          probe arm adjustments for this data
+   */
+  override def armAdjustment(ctx: ObsContext, guideStar: Coordinates, offset: Offset): Option[ArmAdjustment] = {
     import ProbeArmGeometry._
 
     for {
-      ctx      <- Option(ctx0)
-      offset   <- Option(offset0)
+      ctx      <- Option(ctx)
+      offset   <- Option(offset)
```
swalker2m: The GMOS implementation doesn't do any `null` checks for the context or offset and it seems to work out okay.  Do we have reason to believe that `null`s may be passed in in the case of F2?

swalker2m:
It looks fine as far as I can tell by inspecting the code.  There are no new test cases though.  GMOS vignetting testing is scant, but maybe we can parameterize what is there so it works for GMOS or F2?  See `VignettingCalcSpec` and `SingleProbeVignettingTest`.

Also, could you add a sentence or two in the description field?

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/flamingos2/F2OiwfsProbeArm.scala
sraaphorst: I guess not; I was just trying to be better safe than sorry with the Java calling code.
I'll eliminate these, then, and just return `Some(angle...)`.

swalker2m:
It's great that you re-enabled `VignettingTest`, which I'd left turned off after the updates to GMOS before the June release.  Still, I'd have more confidence in that test if it worked with a base position other than (0,0) or close to (0,0).   Something at a higher dec would require all the calculations to be taking into account the impact of declination on the delta RA computation.  For example, targets at some random non-zero RA (why not) like 5:00:00 and a high dec like 85:00:00.

`VignettingCalcSpec` could also be modified to be parameterized by science area and probe arm and thereby work with both GMOS or F2.

I think it's worth doing that, but we should probably go ahead and make a test release with what you've got so far. :+1:

PR #522, opened 2015-08-13
OCSADV-399-I

tpolecat:
This PR changes guidestar validation from a `Boolean` to a ternary `Valid/Invalid/Undefined` to reflect the case where base or guidestar coordinates are unknown for the context scheduling block.


swalker2m:
This seems fine to me.  We could probably get away with mapping `Undefined` to `Invalid` but it is better to make the distinction.

PR #523, opened 2015-08-13
Added empty settings for OT

fnussber:
Adding empty settings for OT credentials to credentials template file. This allows to compile the code with the template file without access to the actual credentials. (This is only relevant for people outside of the software group that need to work on some parts of the code base, like e.g. Sabrina working on ITC code.)


swalker2m:
:+1:

PR #524, opened 2015-08-14
OCSADV-371: Support for GNRS ITC

fnussber:
Extraction of GNIRS parameters for ITC calculations and some other changes to allow ITC calculations for GNIRS in the OT.

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcTable.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcTable.scala
@@ -11,6 +11,7 @@ import edu.gemini.shared.skyobject.Magnitude
 import edu.gemini.spModel.`type`.DisplayableSpType
 import edu.gemini.spModel.config2.{Config, ConfigSequence, ItemKey}
 import edu.gemini.spModel.core.{MagnitudeBand, Peer, Site, Wavelength}
+import edu.gemini.spModel.gemini.gnirs.InstGNIRS
 import edu.gemini.spModel.guide.GuideProbe
 import edu.gemini.spModel.obs.context.ObsContext
 import edu.gemini.spModel.obscomp.SPInstObsComp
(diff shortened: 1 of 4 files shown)


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcTable.scala
```diff
@@ -270,7 +271,7 @@ class ItcImagingTable(val parameters: ItcParametersProvider) extends ItcTable {
         case INSTRUMENT_GSAOI  =>
           new ItcGsaoiImagingTableModel(keys, uniqueConfigs, inputs, results)
 
-        case INSTRUMENT_GNIRS | INSTRUMENT_NIFS | INSTRUMENT_NIRI =>
+        case INSTRUMENT_NIRI =>
```
swalker2m: With this change the `showCoadds` parameter value changes from `true` to `false` for NIFS as well as GNIRS.  I imagine that is what you intended but I thought I'd check since this change was mostly about GNIRS.

swalker2m:
:+1:

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcTable.scala
fnussber: NIFS doesn't support imaging (except for acquisition, which currently is not supported by ITC). This is why I removed it (having NIFS listed here was an error). However, I now think it would be better to explicitly list all instruments that support imaging and throw an error in the `case _` statement, or remove it, to make clear that the caller of this method should never call this code for unsupported instruments. This case is already handled further up in the call chain that's why unsupported instruments are not handled anymore down here, but I guess the assumption should be that somebody can still call this code for unsupported instruments.

fnussber:
Ok, this is not a clean solution because currently there is no clear concept of which instruments support which ITC calculations (imaging, spectroscopy). I'll come up with something more explicit for this, but I'll do it in a separate PR.

PR #525, opened 2015-08-18
OCSADV-372: ITC support for NIFS

fnussber:
A first iteration to support ITC for NIFS. Currently only one IFU mode is supported (single with offset 0, same as for GMOS).

Most of the work went into replacing the String based parameter object `NifsParameter` with a Scala case class that uses the typed enums etc. from `NIFSParams`.


swalker2m:
Good to see string parameters get replaced.  :+1:

PR #526, opened 2015-08-18
OCSADV-410: Request for feedback on Background AGS and Guide Star Status type

sraaphorst:
This is a quick initial PR to begin the project and gather feedback regarding the initial planned changed for Background AGS. @tpolecat - looking at you in particular :-).

In order to properly implement the background AGS (BAGS) lookups, we need to be able to mark guide stars as either manually added / selected or automatically (by AGS) added / selected: this is to preserve manual guide star selections and not override them by a background AGS lookup when an observation changes.

Right now, the `GuideProbeTargets` class holds `SPTargets`, which need to be marked as either manually added / selected or automatically added / selected. The mechanism we propose to do this is to add a new immutable class, `GuideStar`, which will simply contain an `SPTarget` and a `GuideStarStatus` (as per this PR), and then modify the `GuideProbeTargets` class and the affected code accordingly.

Does anyone see any issues that may arise because of this, or have any questions / comments?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,6 +40,7 @@ project/OcsCredentials.scala
 # Reinclude packages named "target" and bundles that end in ".log"
 !/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/view/target
 !/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target
+!/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target
 !/bundle/edu.gemini.pot/src/main/resources/edu/gemini/spModel/target
 !/bundle/edu.gemini.pot/src/test/java/edu/gemini/spModel/target
 !/bundle/edu.gemini.pot/src/test/scala/edu/gemini/spModel/target
--- a/build.sbt
+++ b/build.sbt
@@ -2,7 +2,7 @@ import OcsKeys._
 
 name := "ocs"
 
-ocsVersion in ThisBuild := OcsVersion("2015B", true, 1, 1, 1)
+ocsVersion in ThisBuild := OcsVersion("2015B", true, 1, 2, 1)
 
 pitVersion in ThisBuild := OcsVersion("2016A", false, 1, 1, 0)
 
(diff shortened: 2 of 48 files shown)


swalker2m:
I'll just contribute a bit of additional information on the request from science.  As the user edits an observation, we'll eventually be performing new background AGS lookups which can potentially differ from any current guide stars in the target list.  The idea is to keep the number of guide stars in the target list from continually growing as parameters to the AGS lookup change and different guide stars are selected. So the request is that BAGS automatically replace any automatically added guide stars. In other words, there will only ever be at most one automatically added guide star.

At the same time, any guide star that a user has gone to the trouble of manually selecting should always appear in the list (and I think remain the active guide star), even if it isn't the current best candidate discovered by BAGS.

So maybe a different way of doing this, rather than the Automatic/Manual attribute, would be an `Option[SPTarget]` that represents the automatic candidate found by BAGS in the `GuideProbeTargets`.  Any other guide star would be manually added and never messed with.

tpolecat:
I agree with @swalker2m that we should reflect the cardinality in the type if possible. If every `GuideProbeTargets` has zero or one automatically added guide star then it seems like the `Option` member makes more sense. If I understand what you're saying.

arturog8m:
What's the state of this PR? It would be great to start merging this work, as I understand we are making good progress...so, @sraaphorst could you comment on whether this PR is still valid? Otherwise, make a new one with your latest changes for the model updates that got completed yesterday to support BAGS.

swalker2m:
My take is that the model changes along with the tweak to the target list view are probably enough to warrant a task and PR on their own.  I assume we'll get the BAGS lookup working too in time for December.  My doubt is whether, if we don't have all the UI features that were requested done, would they still want this in the release?  If not, we might want to hold off on starting to merge the predecesor tasks until we are 100% certain?

I think it is a good idea to go ahead and commit the changes to the model and push them up though.

sraaphorst:
This will be done in just a moment! I was just cleaning up the code to take
out some comments and to segregate the changes for the background process
so as to not include something incomplete.

cquiroz:
Without checking the code I wonder if this can accommodate GEMS/Mascot that generates multiple GS each time AGS is called, typically from 0 to 4

PR #527, opened 2015-08-24
REL-2464: change DSSI label

swalker2m:
A trivial change to the display name of DSSI in the PIT instrument list.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -4,7 +4,7 @@ name := "ocs"
 
 ocsVersion in ThisBuild := OcsVersion("2015B", true, 1, 1, 1)
 
-pitVersion in ThisBuild := OcsVersion("2016A", false, 1, 2, 0)
+pitVersion in ThisBuild := OcsVersion("2016A", false, 1, 1, 0)
 
 // Bundles by default use the ocsVersion; this is overridden in bundles used only by the PIT
 version in ThisBuild := ocsVersion.value.toOsgiVersion
--- a/build.sbt
+++ b/build.sbt
@@ -4,7 +4,7 @@ name := "ocs"
 
 ocsVersion in ThisBuild := OcsVersion("2015B", true, 1, 1, 1)
 
-pitVersion in ThisBuild := OcsVersion("2016A", false, 1, 1, 1)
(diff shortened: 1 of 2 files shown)


sraaphorst:
Why was the PIT version renamed as 1.1.1? I was under the impression from discussions with Carlos that all PIT versions were always only two numbers (e.g. 1.1), hence the 0 in the third position.

swalker2m:
I didn't know about the two-number convention, thanks for pointing that out!  I've changed it to 2016A.1.2.0.

It's a bit confusing because the version number actually *has* 3 digits and they all appear in the distribution files.  Nevertheless I see the app icon itself just says "Gemini PIT 2016A.1.1".  Perhaps you could update [the documentation](https://github.com/gemini-hlsw/ocs/wiki/Building-Phase-I-tools)?

swalker2m:
Okay, it turns out that if you change the version number you have to update the ODB and probably ITAC.  I've returned to version 1.1.0 which is confusing since the app actually differs slightly, but we don't want to have to do an ODB update nor touch ITAC for this.

It would be useful to document this.  I put a placeholder in the [wiki page](https://github.com/gemini-hlsw/ocs/wiki/Building-Phase-I-tools#version-numbers) but someone who knows more should provide the information. ( @cquiroz  @sraaphorst ?)

arturog8m:
I added REL-2475 so @cquiroz can update the documentation when he gets back. This should be merged :+1:

cquiroz:
The pattern we agreed some time ago about versioning is year.semester.minor hence 2016.1.1 and decided to discard the last digit as we don't do PIT releases without changing the model.
About the change of DSSI I think it should produce a model change and we need to update the PDF templates and support upconversion. The change proposed will update the UI which may be enough for the users though
Was PIT 2016.1.1 released?

bryanmiller:
Yes, PIT 2016A.1.1 is released.

cquiroz:
Do we want to make an interim release for 2016A? If it is only a UI change I'd recommend keeping the current version otherwise it would mess the backends. If we need a full model change the we must switch to 2016A.1.2, update the backends and obsolete the previous client

bryanmiller:
No, I don't think that we need an interim release. The current version is fine and I don't want to induce a lot of changes in other components.

PR #528, opened 2015-08-25
OCSADV-270: Detect instrument mode from configuration (imaging vs. spectroscopy)

fnussber:
A small but really annoying usability nuisance that plagued users for a long time was, that users had to explicitly select a mode (imaging or spectroscopy) on the ITC web pages, although the mode is given by the instrument configuration, i.e. the absence or presence of a grating element.

I promised Andy and others (including myself) to deal with this but never really got around to do it until now: the radio buttons for spectroscopy/imaging are gone on the HTML pages and the mode is deduced from the configuration.

I also addressed some other minor issues:
* One servlet to rule them all: the instrument specific servlets were not really useful, so I replaced them with a more generic calculation servlet that covers all instruments.
* The servlet version numbers on the output pages (which were hardly ever updated and therefore not very useful) are gone; if a version number is needed, we should use the current ODB release version as the version number.
* The spacer.gifs on the HTML pages showed on Chrome, a small change to the HTML makes sure they are always hidden.


bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/InstrumentDetails.scala
```diff
@@ -82,3 +82,28 @@ final case class GemsParameters(
                      avgStrehl:           Double,
                      strehlBand:          String)
 
+
+object InstrumentDetails {
+
+  // NOTE: This is similar to the code in ItcUniqueConfig which decides on imaging or spectroscopy setup
+  // on Config elements. There should be a way to share this?
```
swalker2m: Unfortunately at the moment the inputs to the function are all different.  Here they are `InstrumentDetails` implementations while in Phase 2 checking they are `Config`s and elsewhere we work directly with `SPInstObsComp`s.  Even the thing being matched upon differs.  Here it is, for example for Michelle, a `String` value whereas in the case of Phase 2 checking we look for one of the `MichelleParams.Disperser` enum values.

In the new sequence model, the sequence/config and static component will be united into one immutable object.  For example see [the prototype F2 definition](https://github.com/swalker2m/ocs/blob/seqmod/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/sequence/F2.scala). That could conceivably be used in ITC instead of the various `*Parameters` classes (?).  If so, it then it seems like it becomes possible to define these rules in one place.  Or are these initialized as servlet parameters?

Unrelated to this, I'm curious about `MichelleParameters`.   It implements, in Java, `InstrumentDetails` which is supposedly a `sealed` trait.  I'm surprised that compiles but since it surely does, I wonder if it at least issues a big warning.  If you leave `MichelleParameters` out of a match statement does it warn you about a missing case then?  I'm guessing no.

fnussber: Regarding this see also my comments in: http://jira.gemini.edu:8080/browse/OCSADV-267 . I agree that a next refactoring step with ITC could be to use the definition of a step from the SequenceModel NG as the input values.

Regarding sealed: yeah, that surprised me, too. It compiles and works. The compiler produces a warning that the `case` statements with `MichelleParameters`/`TrecsParameters` can not be reached (because they are not defined in the `InstrumentDetails` code file, but everything works as expected. I will clean that up at some point, however Michelle and TReCS have no priority because they are not in use anymore.

fnussber: (I should probably mention that the usage of Strings and the definition of `MichelleParameters`and `TrecsParameters` as Java classes is historical, this is how all these objects were originally implemented, I've changed this instrument by instrument, now only Michelle and TReCS are left; the idea is to turn them into typed `InstrumentDetail` case classes.)

swalker2m:
Will this also be reflected in the OT sequence node?  It too shows spectroscopy and imaging tabs.

Basically the ITC works with the static instrument configuration only?  I ask because of course the disperser value could change over the course of the sequence.  I don't know but I doubt it changes from imaging to spectroscopy in the course of a single observation but it is possible.

Otherwise this seems like a good improvement.  :+1:

bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/InstrumentDetails.scala
swalker2m: I suppose we could define an `ObservingMode` (or something) type class that provides the imaging/spectroscopy distinction.  It would still be implemented differently for all the different kinds of instrument/config representations but at least it would be easier to track them down.

fnussber:
The decision imaging / spectroscopy is done for each step of a sequence individually in ITC, so there is no danger that this gets messed up.

PR #529, opened 2015-08-25
Typed MichelleParameters object.

fnussber:
This is a follow up for OCSADV-270 and deals with one of the last remaining String typed parameter objects: `MichelleParameters`. This class has been replaced by a typed Scala case class.

On the side I've simplified how the instrument name is read from the HTML. Check out the last commit only, this is the actual relevant work for this PR.


swalker2m:
:+1:

PR #530, opened 2015-08-26
Typed TReCSParameters object

fnussber:
This finishes the refactoring of all stringly typed input parameter objects currently used by ITC to typed ones using the types defined in `spModel`. No new functionality, this is pure code hygiene.

This also removes the weirdness that a sealed Scala trait is extended by external Java classes. Yes, this works.

Note that TReCS and Michelle are not really supported anymore, this is why I waited with this work until now. However for further refactoring steps we will need to bring everything up-to-date.


swalker2m:
Excellent. :+1:

PR #531, opened 2015-08-27
REL-1852: Flamingos-2 ITC updates

fnussber:
Implementation of the following update requests to F2:
* Change of gain on HTML page
* Change of read noise for bright objects
* Change of full well depth value
* Additional warning in case linearity limit is reached

I took the liberty to fold some cleanups/todos into this task:
* F2 was using the sky background values for MK; it now uses the values for CP
* F2 was the only instrument using old sky background dat files; it now uses the HI-Res ones like all other IR instruments
* These changes allowed to delete the IR sky background files outside the HI-Res folder
* F2 was using a slightly different calculation order for the spectral energy distribution for no obvious reasons; it now uses the same calculation order as all other instruments (change in values is <1%)

Notes:
1) `SEDFactory.calculate()` still has instrument specific calculation orders which stem from my attempts to unify as much ITC code as possible. I am convinced, that these differences should/could be removed, too. However we need to do this instrument by instrument.
2) A more prudent approach would be to implement each of those changes separately and verify the differences in the results individually. However Andy and I decided to fold a series of changes into this PR because the changes are obviously needed (e.g. using CP data files instead of MK) and the differences in the calculations are <5% in most cases up to no more than 10% which seems reasonable for the changes I did. Ruben will verify the new values with the next test release.


swalker2m:
:+1: Thanks for the enum cleanups too.

PR #532, opened 2015-08-28
OCSADV-399-J

tpolecat:
Updates Canopus to handle guidestars with unknown coordinates. Probe arm geometry computes `Option<Area>` now.

Note that if you add `?w=1` to the URL you'll see a whitespace-neutral diff that is less dramatic.


tpolecat:
N.B. Tests pass.

swalker2m:
:+1:

PR #533, opened 2015-08-29
OCSADV-335: Squants example: ITC EmissonLine.

fnussber:
PR only for discussion.

This is a very cheap example of how using [squants](http://www.squants.com) could look like. Note that I was able to get rid of the `Flux` type, which is currently used to represent W/m^2, with the type `Irradiance` from squants; unfortunately there is no definition for [spectral flux densities](https://en.wikipedia.org/wiki/Spectral_flux_density) (W/m^3), but you get the idea.

SI units are implemented including lots of conversions and I think this could be a very useful library. We might want to add some aliases (e.g. `Wavelength` for `Length`) and we might want to add some stuff to the lib (see above).

Obviously I only want to start investing time into doing substantial work (including potential additions to squants library) once people agree that we want to go forward with squants.

Oppinions?

### DIFF
--- a/bundle/edu.gemini.itc.shared/src/main/java/edu/gemini/itc/shared/SourceDefinition.java
+++ b/bundle/edu.gemini.itc.shared/src/main/java/edu/gemini/itc/shared/SourceDefinition.java
@@ -3,6 +3,9 @@ package edu.gemini.itc.shared;
 import edu.gemini.spModel.core.MagnitudeBand;
 import edu.gemini.spModel.core.Wavelength;
 import edu.gemini.spModel.target.*;
+import squants.motion.Velocity;
+import squants.radio.Irradiance;
+import squants.space.Length;
 
 import java.io.Serializable;
 
@@ -117,15 +120,15 @@ public final class SourceDefinition implements Serializable {
         return ((BlackBody) distribution).temperature();
     }
 
-    public Wavelength getELineWavelength() {
+    public Length getELineWavelength() {
         return ((EmissionLine) distribution).wavelength();
     }
 
-    public double getELineWidth() {
+    public Velocity getELineWidth() {
         return ((EmissionLine) distribution).width();
     }
 
-    public EmissionLine.Flux getELineFlux() {
+    public Irradiance getELineFlux() {
         return ((EmissionLine) distribution).flux();
     }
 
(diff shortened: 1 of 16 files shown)


swalker2m:
I think it looks useful.  I'm not sure about replacing `Wavelength` with `Length` in particular.  It seems odd to be able to define a wavelength in miles or lightyears which would then be possible.  More importantly, it feels like a more fundamental type that you wouldn't want to confuse with the length of a probe arm or something. I guess a simple value class wrapper would solve that problem though.

But yeah in general I think it is worth seeing about adding new types that we need.  Anybody else have an opinion to share?

arturog8m:
Would it make sense to hide the library behind an edu.gemini.units package
of something like that and create wrappers for the kind of stuff we are
interested on? That way we ensure everyone uses this library consistently.
Also, if the library API changes later, we only have one place to update.
That might be an overkill though. Thoughts?

tpolecat:
I think it would make sense to make type aliases in `core

bundle/edu.gemini.itc.shared/src/main/java/edu/gemini/itc/shared/SourceDefinition.java
```diff
@@ -117,15 +120,15 @@ public double getBBTemp() {
         return ((BlackBody) distribution).temperature();
     }
 
-    public Wavelength getELineWavelength() {
+    public Length getELineWavelength() {
         return ((EmissionLine) distribution).wavelength();
     }
```
tpolecat: As @swalker2m notes, this is kind of odd; we might at least want to say `type Wavelength = Length` somewhere.

tpolecat:
Deleted my comments and starting over, sorry. 

I think this is great, but we do need to recognize that model semantics and units are orthogonal concerns. It's true that a wavelength is just a length, but it should be distinct at the type level from other lengths as @swalker2m pointed out, just as `RA` and `Dec` are now distinct types. So after some thought I so think it does make sense to at least have a bunch of type aliases in `core`, and then add wrapping or tagging or something as necessary to distinguish quantities that have the same dimensions but different meanings in the model.

Anyway this seems great to me.

tpolecat:
I also confirmed with Erik Osheim that this seems to be the right lib to use. There are some others but this one has been around for a while and is active.

fnussber:
Replacing this PR with next iteration.

PR #535, opened 2015-09-01
OCSADV-399-L

tpolecat:
This updates the Scala uses of `ObsContext.getBaseCoordinates` to use the variant returning an `Option`, which will be renamed and become the only such method in a later PR.


swalker2m:
:+1:

PR #536, opened 2015-09-02
OCSADV-335A: Turning Wavelength into squants based value class.

fnussber:
Presenting the current state of work as another PR for discussion, needs more work before merging.

Following up on the previous discussion I thought I start with the `Wavelength` values. I turned the `Wavelength` trait into a value class in order to keep it as a completely distinctive type. 

As mentioned, just defining an alias has - apart from not really being a distinctive type - also the disadvantage of not being usable in the Java code, which bit me here. Although it involves a bit more work, I guess it makes sense to define a wrapper class for types which are used often and should be distinct from others, `Wavelength` is probably a good example.

If people are ok with this so far, I am going to introduce more Squants based values in the ITC code and keep going...


bundle/edu.gemini.itc.shared/src/main/java/edu/gemini/itc/shared/SourceDefinition.java
```diff
@@ -118,7 +118,9 @@ public double getBBTemp() {
     }
 
     public Wavelength getELineWavelength() {
-        return ((EmissionLine) distribution).wavelength();
+        // Wavelength is a value class, on the Java side wavelength() therefore returns just a Length
+        // and not a Wavelength object, so we need to repackage it to make it a wavelength. Blech.
+        return new Wavelength(((EmissionLine) distribution).wavelength());
```
fnussber: Yo! This is a funny one where Java doesn't know what Scala is doing. The remedy will be to turn `SourceDefinition` into a Scala case class sometime in the future. Baby steps, baby steps...

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/Wavelength.scala
```diff
@@ -103,6 +70,10 @@ object Wavelength {
   implicit val WavelengthOrdering: scala.Ordering[Wavelength] =
     scala.Ordering.by(_.toNanometers)
 
-
+  /** @group Implicit Conversions */
+  implicit def fromLength(l: Length): Wavelength = {
```
fnussber: In theory this breaks a little bit with type safety because it allows people to pass an arbitrary `Length` when a `Wavelength` is needed but it allows for pretty code (see for example all the definitions in the test fixtures etc) and if someone really passes a light year as a wavelength then so be it. At least we still keep people from mixing `Wavelength`s with other things in the code.

swalker2m: I'd rather that we add pimping to explicitly go from `Length`to `Wavelength` when desired (think `JavaConverters` instead of `JavaConversions`).  As you point out, this type of implicit conversion can be tricky.  It seems like we'd be losing a lot of the benefit of having an explicit `Wavelength` type.

swalker2m: Can we can add our own `Int` pimping to get `.nm` and `.microns` that convert to `Wavelength` and still have the concise code?

swalker2m: For example, see core `ToDegrees` and `AngleSyntax.ToDegreesOps`.

fnussber: Ok, I'll look into this. However, as long as `Wavelength` is a value class there's not much we can do to keep people from calling `Wavelength(crazyLength)` directly (?). But I guess we can live with that.

fnussber:
Following up on comments from @swalker2m I introduced new implicit conversions in `WavelengthConversions` that allow creation of `Wavelength` values from microns and nanometers like so: `<x>.nm`, `<x>.nanometers` and `<x>.microns`. I think that makes it all pretty nice to use.

fnussber:
Retracting again in favor of the final PR.

PR #539, opened 2015-09-04
OCSADV-355B: Replace EmissionLine.Flux with Irradiance from squants.

fnussber:
Taking advantage of squants..


swalker2m:
Looks good. :+1:

PR #540, opened 2015-09-05
ITC: Add GMOS gaps at end of GMOS recipe.

fnussber:
ITC has a mechanism that allows to represent gaps between CCDs and will be taken into account in spectroscopy calculations to "pull" the signal down to zero in places where the signal falls on a gap and will therefore not be available in the science data.

While GMOS is currently the only instrument where we have to deal with this, many instruments have this capability implemented (either by over-engineering or copy-pasting) and apply those gaps at the very beginning of the calculation.

For GMOS IFU-2 calculations Sabrina needs a way to represent those gaps in different places depending on the actual configuration, which means the gaps should not be applied at the beginning of the calculations (recipe) but at the very end, possibly only when representing the data in charts / tables.

This refactoring is aimed at making Sabrina's life easier by 
* pulling the application of the gap mechanism out of the general calculations and
* applying the gaps at the end of the calculations for GMOS only.

Sabrina should be able to take it from here and apply the gaps in a way that is suitable for her purposes without having to deal with other instruments and unneeded generic mechanisms.

As part of this refactoring I also tossed out some dead code.


fnussber:
Merging this right now so that Sabrina can keep working on her stuff.

PR #541, opened 2015-09-07
Readme Update based on recent setup

arturog8m:
Just some minor updates to the `README.md` file based on my recent exercise of setting things up on my laptop. I'm calling this project the **OCS** rather than **OCS 2.1**. Probably we should also remove references to the old ocs 1.5, but I'll leave that for later.


swalker2m:
Looks good. :+1:

PR #542, opened 2015-09-08
REL-2473: Updated LICENSE file to BSD-3

arturog8m:
Updated LICENSE file to BSD-style as approved by Markus. I had to update a zillion files to refer to this file when appropriate and kept the (other) original copyright notes in place. You probably will need to update your local copies to get these small changes. 

There was a problem with the script that I used to do this that ended up adding a new line to several files that didn't have one... sorry about that, so ignore those.


fnussber:
Is it legally relevant whether every single source file has a header referring to the license or not? I am wondering because we do have quite some files without headers and I stopped adding headers to new files I have created in the past. Should we add a standard header to new files, and if so, how should it look like?

arturog8m:
The BSD-3 license we are using does not say anything about that, so I'm not going to push for it. I was in fact tempted to delete the existing headers that have the AURA copyright on it, but it was too much trouble, so I just update those to say "look at the LICENSE file for more details" or something along those lines, but for the files that didn't have headers I just checked they were in fact ours and moved on. 

We will need of course keep existing headers that refer to other licenses; I did my best to keep those of course.

swalker2m:
:+1:

PR #543, opened 2015-09-08
OCSADV-399-N

tpolecat:
Next up, this adds a time parameter to `getRa/DecString`.

### DIFF
--- a/bundle/jsky.app.ot/src/main/java/jsky/app/ot/tpe/gems/CandidateAsterismsTreeTableModel.java
+++ b/bundle/jsky.app.ot/src/main/java/jsky/app/ot/tpe/gems/CandidateAsterismsTreeTableModel.java
@@ -6,6 +6,7 @@ import edu.gemini.shared.util.immutable.DefaultImList;
 import edu.gemini.shared.util.immutable.ImList;
 import edu.gemini.ags.gems.GemsGuideStars;
 import edu.gemini.ags.gems.GemsStrehl;
+import edu.gemini.shared.util.immutable.Option;
 import edu.gemini.spModel.guide.GuideProbe;
 import edu.gemini.spModel.target.SPTarget;
 import edu.gemini.spModel.target.env.GuideProbeTargets;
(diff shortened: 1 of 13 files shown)


swalker2m:
:+1:

bundle/jsky.app.ot/src/main/java/jsky/app/ot/tpe/gems/CandidateAsterismsTreeTableModel.java
```diff
@@ -95,9 +96,11 @@ public Object getValue(Row row) {
         // (XXX There might be another way to get the parent node)
         private Row _parent;
 
+        private final Option<Long> _when;
+
         // A row corresponding to a GemsGuideStars object
         // (top level node, displays the Strehl info and checkbox)
-        Row(GemsGuideStars gemsGuideStars, List<Row> children) {
+        Row(GemsGuideStars gemsGuideStars, List<Row> children, Option<Long> when) {
```
cquiroz: What this is saying is let's display the asterism with ra/dec at the time set on the scheduling block right?
Shouldn't this be done when AGS calculates the targets instead and then ra/dec of each target would already be corrected? I may be miss-understanding this though

tpolecat: The point in time used by AGS and the one we use when displaying guide-star coordinates are not necessarily the same, but in practice they will be because we're currently using the start of the scheduling block everywhere. Does this answer your question?

cquiroz: Is the scheduling block defined at the observation level right?
I think AGS is not using the scheduling block, is that planned?

tpolecat: Yes, scheduling block is a value on the observation. I don't think it's used in AGS search but it's used when asking targets for their coordinates. I don't know much about AGS so I don't know what the plans are.

cquiroz: I assume AGS needs to get the coordinates of the targets at the given time, both the one in the observation and the ones from the catalog
Which takes me to another point, we should likely correct on the catalog side too, right?

tpolecat: Right. AGS has to pass a time to targets (otherwise it won't compile). But maybe the catalogs can do a better job if we pass the time to them as well. Don't know.

cquiroz: Did you check if the changes break compilation on other modules? Passing a time will certainly break a bunch of tests :cry:

tpolecat: Yes, tests pass.

tpolecat:
Rebased. Tests pass, other than those addressed by #551.

PR #544, opened 2015-09-09
OCSADV-419 remove conic targets from new model

tpolecat:
This removes conic targets from the new target model.


swalker2m:
It will be great to see this go.  Do we need to first assure that the TCS integration, as you mention, is ready?

tpolecat:
I don't think we need to wait; there is no real way to support the new model with conic targets unless we solve the conic equations (which I gave up on) or add an ephemeris, which is kind of pointless. So I don't see a situation where these types will ever be used.

cquiroz:
Looks good

PR #545, opened 2015-09-10
OCSADV-410A: Data model changes and changes to Target Editor for BAGS

sraaphorst:
Apologies for the nightmare of a PR this is.
This comprises the first two milestones for BAGS (background AGS, i.e. AGS lookups done in the background).

The main elements changed here:
* The `GuideProbeTargets` class has essentially been rewritten from the ground up to allow for both a primary guide star and a BAGS guide star. The old `GuideProbeTargets` used an `OptionsList` to contain a list of `SPTarget` with one designated optionally as the primary. Since this behaviour no longer makes sense as we need two "special" `SPTargets` that can significantly interact and place restrictions on each other, the entire class has been revamped.

* The change has been propagated through the data model and through the test cases.

* The Target Editor has been changed to accommodate the idea of an automatic guide star as selected by BAGS. This guide star should always appear at the top of the target list and not be able to be deleted or moved. Classes included in this change as `TelescopePosWidget` and `EdCompTargetList`, primarily.


andrewwstephens:
Will it be possible to override the guide star selected by BAGS?  (you mention the BAGS guide star cannot be deleted)

sraaphorst:
@andrewwstephens Oh yes, of course. You can select any guide star you wish, and that will override the BAGS selection (and keep it overridden until a point where no primary guide star is selected).

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/guide/OptimizableGuideProbeGroup.java
```diff
@@ -20,13 +20,16 @@
...
+     * AGS (true) or manually selected (false)*
+     *
      * @param ctx the context of the observation, which contains information
      * needed to assign a guider to the new guide star
      *
      * @return a newly optimized TargetEnvironment which incorporates the given
      * <code>guideStar</code>
      */
-    TargetEnvironment add(SPTarget guideStar, ObsContext ctx);
+    TargetEnvironment add(SPTarget guideStar, boolean isBAGS, ObsContext ctx);
```
cquiroz: Perhapsthe `isBAGS` boolean could be made a type to make it a stronger assertion

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/GuideGroup.java
```diff
@@ -155,7 +155,7 @@ public GuideGroup setName(String name) {
      * instance; <code>false</code> otherwise
      */
     public boolean contains(GuideProbe guider) {
-        return guideTargets.exists(GuideProbeTargets.match(guider));
+        return guideTargets.exists(gpt -> gpt.getGuider() == guider);
```
cquiroz: is `==` correct here? Is `Guider` an enum?

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/GuideProbeTargets.java
```diff
@@ -104,87 +65,104 @@
...
 
 
     private final GuideProbe guider;
-    private final OptionsList<SPTarget> targetOptions;
+    private final ImList<SPTarget> manualTargets;
+
+    // The targets selected as the primary target and the target chosen by bags.
+    // Note that these can be the same.
+    private final Option<SPTarget> primaryTarget;
+    private final Option<SPTarget> bagsTarget;
```
cquiroz: How does this work with GEMS when you can have up to 4 guide stars?

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/GuideProbeTargets.java
```diff
@@ -194,147 +172,392 @@ public GuideProbe getGuider() {
...
-    @Override
-    public Option<Integer> getPrimaryIndex() {
-        return targetOptions.getPrimaryIndex();
+    /**
+     * Set the BAGS target. If the primary is the old BAGS target, then the primary is set to the
+     * new BAGS target automatically; otherwise, it is maintained.
+     * @param target the new BAGS target
+     * @return a new GuideProbeTargets with the primary target and BAGS target as described
+     */
+    public GuideProbeTargets setBagsTarget(final SPTarget target) {
```
cquiroz: Somehow the `setXXX` methods that actually create a copy throw me off. Could it be clearer if they were called `withBagsTarget` then you know it is a copy with an update?

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/GuideProbeTargets.java
```diff
@@ -194,147 +172,392 @@ public GuideProbe getGuider() {
...
+
+        return setPrimaryIndex(indexOption.getValue());
+    }
+
+    /**
+     * Set the primary target by index.
+     * If there is a BAGS star, we begin counting it as 0, and then the manual targets at index 1.
+     * Otherwise, we begin counting the manual targets at index 0.
+     * @param index the index to use as the primary
+     * @throws java.lang.IndexOutOfBoundsException if the index does not indicate a valid guide star
```
cquiroz: Could you instead return `Option[GuideProbeTargets]`

cquiroz:
Given that `GuideProbeTargets` is a core class and seems fairly complex, could we have ported it to scala? I think the tests for that class are the same as before even if we have significant changes, perhaps new tests would be a good addition.
I'm concerned that there is a single BAGS target, it would seem to me that you may need to support multiple, at least for gems?

geminihlswadmin:
Each GuideProbeTargets is associated with a single guide probe so there can be one BAGS target for each guider involved.

By the way, we're planning on making a few model changes that will use the type system to avoid some of the selection errors that the API permits today.

tpolecat:
Hm who is logged in as @geminihlswadmin? Probably shouldn't do that.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/GuideProbeTargets.java
sraaphorst: This is handled with multiple `GuideProbeTargets` classes, one for each guider.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/GuideProbeTargets.java
sraaphorst: Sure, this makes sense. I'll make this change.

sraaphorst:
@cquiroz This is a possible future improvement that we are discussing. (Porting `GuideProbeTargets` to scala, and making it more type-safe.) As it stands now, though, due to time constraints, this will likely be pushed back.

PR #547, opened 2015-09-11
REL-2399: Obsolete the NIRI filters Jcont(1.065) and Jcont(1.122) 

arturog8m:
I took what I thought it was going to be a small task (<30m) to play around with GitHub and our development process. Unfortunately this took a bit longer than planned, as I was working on this thing I found a few *special* features that distracted me for a while. 

First, the filters were stored as a collection of `String` in the combo, instead of real `Filters` as God mandates. 

Second, they were separated by narrowband/broadband when they were displayed in the Combo (using of course a special String - `-----------`) which looked hideous, and had code to figure out how to map the Strings in the combo to the real Filters. I still believe the UI is confusing as there is no indication as to why the filters are separated, you must know that the ones at the top are broadband, the ones below are narrowband. I actually experimented with adding a small icon before each element to make it clear what they were, but since I didn't have a good icon and this wasn't really being asked, I dropped that idea... oh well. 

Finally, they were not using their display value in the Combo, but the description of the Filters instead. 

So, I ended up creating a custom `ComboBoxModel` and a custom `BasicComboBoxRenderer` to deal with these particularities, and get rid of all those strings and replaced the original `String` separator with a `JSeparator`.

So, the first commit 58b06b4 in this PR deals with making the filters obsolete as requested in the task. The second 4fadf2a was some code clean up to make most of the code warnings go away. The third commit cf777ab was to add the new combobox models and renderer. Finally, the last commit 6cbf9b5
was to update the usage of our `Option` classes to the `Optional` class in Java 8. 

So here you go. Thanks @swalker2m for the help on using some functional features of Java 8 for some of the work done in cf777ab

### DIFF
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/niri/Niri.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/niri/Niri.java
@@ -27,7 +27,7 @@ public final class Niri {
     /**
      * Cameras
      */
-    public static enum Camera implements StandardSpType {
+    public enum Camera implements StandardSpType {
 
         F6("f/6", "(0.12 arcsec/pix)", "f6", Size.F6),
         F14("f/14", "(0.05 arcsec/pix)", "f14", Size.F14),
@@ -58,14 +58,14 @@ public final class Niri {
          */
         public static final Camera DEFAULT = F6;
 
-        private String _displayValue;
-        private String _description;
-        private String _logValue;
+        private final String _displayValue;
+        private final String _description;
+        private final String _logValue;
 
         // The default science area size for the camera.  Values is "height".
-        private double _height;
+        private final double _height;
 
-        private Camera(String displayVal, String desc, String logValue, double height) {
+        Camera(final String displayVal, final String desc, final String logValue, final double height) {
             _displayValue = displayVal;
             _description = desc;
             _logValue = logValue;
@@ -130,7 +130,7 @@ public final class Niri {
     /**
      * Beam Splitter
      */
-    public static enum BeamSplitter implements DisplayableSpType, SequenceableSpType {
+    public enum BeamSplitter implements DisplayableSpType, SequenceableSpType {
 
         same_as_camera("same as camera", 0.),
         f6("f/6", Camera.Size.F6),
@@ -139,10 +139,10 @@ public final class Niri {
 
         public static final BeamSplitter DEFAULT = same_as_camera;
 
-        private String _displayValue;
-        private double _height;
+        private final String _displayValue;
+        private final double _height;
 
-        private BeamSplitter(String displayVal, double height) {
+        BeamSplitter(final String displayVal, final double height) {
             _displayValue = displayVal;
             _height = height;
         }
@@ -161,14 +161,6 @@ public final class Niri {
         public static BeamSplitter getBeamSplitterByIndex(int index) {
             return SpTypeUtil.valueOf(BeamSplitter.class, index, DEFAULT);
         }
-
-        /**
-         * Return a BeamSplitter by name
-         */
-        public static BeamSplitter getBeamSplitter(String name) {
-            return getBeamSplitter(name, DEFAULT);
-        }
-
 
         /**
          * Return a BeamSplitter by name giving a value to return upon error
@@ -209,7 +201,7 @@ public final class Niri {
     /**
      * Dispersers
      */
-    public static enum Disperser implements DisplayableSpType, SequenceableSpType, LoggableSpType {
+    public enum Disperser implements DisplayableSpType, SequenceableSpType, LoggableSpType {
 
         NONE("none", 0.0, "none"),
         J("J-grism f/6", 1.20, "Jgrism"),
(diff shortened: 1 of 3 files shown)


bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/niri/Niri.java
```diff
@@ -550,12 +536,12 @@ public String getRecommendedMinExp() {
         public static final Mask DEFAULT = MASK_IMAGING;
 
         // The internal values of height and width for this instance of Mask
-        private double _width = Size.IMAGING_SIZE;
-        private double _height = Size.IMAGING_SIZE;
-        private String _displayValue;
-        private String _logValue;
+        private final double _width;
```
sraaphorst: Since the default values here for `_width` and `_height` were rightfully changed from `Size.IMAGING_SIZE` to not being assigned until the constructor (which requires values for these parameters so one would assume they were always created assigned to whatever was specified) to allow the members to be marked as `final`, I am guessing that the defaults were never actually used in any way?

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/niri/EdCompInstNIRI.java
```diff
@@ -397,4 +336,77 @@ public void textBoxKeyPress(TextBoxWidget tbwe) {
...
+     * purposes in the ComboBox
+     * @return a Vector of optional filters.
+     */
+    private static Vector<Optional<Filter>> buildElements() {
+        final java.util.List<Filter> v = new ArrayList<>(SpTypeUtil.getSelectableItems(Niri.Filter.class));
+
+        final List<Optional<Filter>> bbFilters = new ArrayList<>();
+        final List<Optional<Niri.Filter>> nbFilters = new ArrayList<>();
+
+        for (Niri.Filter filter: v) {
```
sraaphorst: Seems this could be done with lambdas instead as:
`v.filter(broadband::equals).foreach(f -> bbFilters.add(Optional.of(f));`
`v.filter(narrowband::equals).foreach(f -> nbFilters.add(Optional.of(f));`
to avoid the loop and switch, although this does require more linear passes through `v`, though I doubt efficiency is of a concern here.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/niri/NiriForm.java
```diff
@@ -2,53 +2,51 @@
```
sraaphorst: Nice cleanup of this messy class, with the elimination of unnecessary stored variables and the action listener redundancy.

sraaphorst:
:+1:

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/niri/Niri.java
arturog8m: Correct. There wasn't a use of the default values for these fields, the constructor of `Mask` forces you to initialize them. I actually realized that just by marking the items final and seeing what the compiler said.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/niri/EdCompInstNIRI.java
arturog8m: Thanks, I'll try this. I looked at a way of getting the two lists that I needed filtering directly the original list, but the code looked so convoluted that I gave up. This is more readable, so I'll incorporate it. I think it will be a little bit more verbose in Java, but still worth it.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/niri/Niri.java
```diff
@@ -797,20 +793,6 @@ public ROIDescription(ROIDescription roid) {
...
-        }
-
-        /**
-         * Return the y start pixel.
-         */
-        public int getYStart() {
-            return _yStart;
-        }
-
-        /**
```
arturog8m: I should have asked... I removed these 2 accessors as they aren't being used in the OCS code tree. However, now  I wonder if there is any other software that is relying on these things? LCH? ITC? Or maybe something using reflection  (oh my, I hope not!) that could be accessing these? Maybe I should just leave these in to be safe, but it's kind of annoying to leave stuff in not knowing if it is really necessary. It's kind of strange that this class has these fields defined but nobody seems to be using them - I'm of the idea of removing as much code as we can that we don't need. What do you guys do normally in these situations?

fnussber: Remove. If it was used in ITC you would immediately notice it when doing a full recompile. The only dependency LCH/LTTS has is the XML interface. And I think the usage of reflection is limited to the initialization of the UI components in the OT.

tpolecat: :+1: if the full build compiles and tests pass, call it good. There is very little reflection left in OCS, thankfully.

swalker2m: The only exception I can recall is for the [TPE features](https://github.com/gemini-hlsw/ocs/blob/develop/bundle/jsky.app.ot/src/main/java/jsky/app/ot/OT.java#L223-L237), which are added via reflection.  If there are others it's probably better to find out now and deal with it anyway.

swalker2m: In this particular case, if we aren't using the (x,y) start we might as well remove the fields altogether? Also, the class seems to want to be immutable but isn't.  Seems like we could nudge it there without much work.

PR #548, opened 2015-09-14
REL-2479: add proprietary header items to sequence

swalker2m:
This grim little PR adds information about the program proprietary period to the sequence so that the seqexec can add the corresponding FITS headers to the datasets as they are created.  This is part of a larger task (REL-2407) in which the Data Manager is being relieved of all FITS file manipulation.

Specifically, the following sequence keys are added

*  `observe:headerVisibility` with values `PUBLIC` or `PRIVATE`
* `observe:proprietaryMonths` with non-zero integer value

The first item is static across the entire sequence but the second may change from step to step depending on the observation class of the step.  For example, daytime calibrations are not charged and the datasets are not kept private regardless of the program type.


bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/dataflow/GsaAspect.java
```diff
@@ -46,28 +67,50 @@ public static GsaAspect getDefaultAspect(ProgramType type) {
...
+    private final int _proprietaryMonths;
+    private final Visibility _headerVisibility;
 
     public GsaAspect(boolean sendToGsa, int months) {
-        this(sendToGsa, months, false);
+        this(sendToGsa, months, Visibility.PUBLIC);
     }
 
-    public GsaAspect(boolean sendToGsa, int months, boolean headerPrivate) {
+    public GsaAspect(boolean sendToGsa, int months, Visibility visibility) {
```
swalker2m: I switched the boolean `headerPrivate` flag for an enum but left `sendToGsa` alone.  The `sendToGsa` flag will be removed altogether in a subsequent PR.  Now all datasets are archived so it is irrelevant.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/obscomp/GemObservationCB.java
```diff
@@ -99,8 +99,6 @@ public Object clone() {
         private final Option<Integer> _scienceBand;
         private final SPInstObsComp _inst;
         private final ObsExecRecord _obsExecRecord;
-        private final Set<GuideProbe> _availableGuiders;
```
cquiroz: Were these used on the GSA logic? Or is it just an opportunity to cleanup?

bundle/edu.gemini.pot/src/test/scala/edu/gemini/spModel/dataflow/ProprietaryTest.scala
```diff
@@ -0,0 +1,76 @@
...
+
+class ProprietaryTest extends InstrumentSequenceTestBase[Flamingos2, SeqConfigFlamingos2] {
+  override protected def getObsCompSpType = Flamingos2.SP_TYPE
+  override protected def getSeqCompSpType = SeqConfigFlamingos2.SP_TYPE
+
+  override def setUp(): Unit = {
+    super.setUp()
+
+    // We're not really interested in the instrument sequence but we have to put
+    // at least thing there in order to have steps to iterate.
```
cquiroz: 'one thing'?

cquiroz:
Looks good :+1:

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/obscomp/GemObservationCB.java
swalker2m: Just cleanup.  I went through the class before I started adding my changes and removed unused code and did a light cleanup in general.

 It used to be that there was logic here for instructing the seqexec when and when not to guide.  That got moved but some of the logic was apparently left behind.

PR #549, opened 2015-09-14
OCSADV-335C: Use squants Velocity to represent emission line widths (km/s)

fnussber:
This turned into a much bigger refactoring task than anticipated because there was some code that checked for every instrument individually if the emission line width was in a certain relation to some "fudge" factor we didn't really understand in the beginning what it was standing for. My main goal was to centralize this validation check and reuse it for all instruments, leaving the fudge factor as is.

However Andy figured out that this factor represents the sampling resolution of the data files representing the atmospheric properties which (currently) depends on the band range covered by the instrument (Visible vs Near IR vs Mid IR).

Once that was clear it seemed to make sense to represent this in a way in the code that makes it clearer where the values used in the validation check come from, so I did refactor things a bit more to make the band range covered by the instrument available for the checks.


bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/base/Instrument.java
```diff
@@ -17,8 +18,28 @@
  */
 public abstract class Instrument {
 
+    public enum Bands {
```
cquiroz: Can this be represented with the existing `MagnitudeBand` class?

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/operation/WaterTransmissionVisitor.java
```diff
@@ -14,112 +15,114 @@
...
-                default:
-                    throw new IllegalArgumentException("unknown WV value");
-            }
-
-            name = "/HI-Res/" + abbrForSite(site) + wavelenRange + ITCConstants.TRANSMISSION_LIB + "/"
-                    + file_name + _transmissionCategory + _airmassCategory + ITCConstants.DATA_SUFFIX;
+        switch (instrument.getBands()) {
+            case VISIBLE:
+                switch (skyTransparencyWater) {
+                    case 1:
```
cquiroz: Magic constants?

cquiroz:
I know little about ITC or squants but this seems fine to me
Why are we including the squants binary again?

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/base/Instrument.java
fnussber: This is more of a representation of a number of bands an instrument is covering or the instrument type, i.e. visible light, vs near IR vs far IR. I do wonder however if we could use something like that in other places (or already have something similar)?

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/operation/WaterTransmissionVisitor.java
fnussber: This is a left-over I haven't cleaned up (yet). These values correspond to the WV bands. Same issue with CC and SB and IQ. Now that we have introduced those values into ITC we can replace these magic numbers at some point. It's on the still pretty long list for future refactorings..

fnussber:
Squants allows to represent a quantity AND a unit, convert between different units and on top of that quantities also represent a dimension like time, length, velocitiy, area etc and when multiplying / dividing quantities the result has the correct dimensions, ie. multiplying a length with a length will result in an area. See http://www.squants.com/

PR #551, opened 2015-09-15
REL-2399: Removing obsolete Niri filters from smartgcal lookup tables

arturog8m:
One test case wasn't passing in my previous PR as the look up tables for SmartGCAL we had were referring to these obsolete filters. I'm not sure how we handle this in operations. @fnussber could you review this PR and let me know what the procedure should be when we put this in operations?

I also added more detail info to the error message that is generated when this happens, as suggested by @tpolecat

### DIFF
--- a/bundle/edu.gemini.spModel.smartgcal/src/test/java/edu/gemini/spModel/smartgcal/CalibrationMapTest.java
+++ b/bundle/edu.gemini.spModel.smartgcal/src/test/java/edu/gemini/spModel/smartgcal/CalibrationMapTest.java
@@ -46,17 +46,17 @@ public class CalibrationMapTest {
         Assert.assertEquals(6, keys.size());
 
         // wildcard that matches all available values
-        // this also tests that obsoleted values are not taken into account (13 are defined, 2 of them are obsolete)
+        // Note that obsolete values are allowed!
         properties.setProperty(ConfigKeyGnirs.Values.SLIT_WIDTH.toString(), "*");
         keys = map.createConfig(properties);
-        Assert.assertEquals(11, keys.size());
+        Assert.assertEquals(13, keys.size());
 
-        // combination of wildcards: 11 slit widths, 2 pixel scales, 2 well depths = 44 keys...
+        // combination of wildcards: 13 slit widths (2 obsolete), 2 pixel scales, 2 well depths = 52 keys...
         properties.setProperty(ConfigKeyGnirs.Values.SLIT_WIDTH.toString(), "*");
         properties.setProperty(ConfigKeyGnirs.Values.PIXEL_SCALE.toString(), "*");
         properties.setProperty(ConfigKeyGnirs.Values.WELL_DEPTH.toString(), "*");
         keys = map.createConfig(properties);
-        Assert.assertEquals(44, keys.size());
+        Assert.assertEquals(52, keys.size());
     }
 
     @Test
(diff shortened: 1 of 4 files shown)


fnussber:
All obsolete enum values for filters, dispersers etc are ignored when creating the lookup maps for the smartgcal calibrations. This will result in the errors we've seen here when parsing the config files. All other calibrations in the config files will be correctly processed regardless.

When obsoleting elements we should make sure that smartgcal entries for those elements are removed from the config files before the new code base with the freshly obsoleted elements goes into production.

Once the new code base is in production science folks won't be able to upload files with obsoleted elements anymore because files are validated before every upload.

I think the best way forward is to let science know that obsoleted elements should be removed from the config files and have them do it asap, if possible before the new code goes into production.

BUT: We should also make sure that we really don't need those smartgcal configs anymore. E.g. because they need to be displayed in observation sequences that have been executed in the past and reference those filters? If this is the case we should allow obsoleted elements (which is a very small code change), however this will also mean that config lines for obsoleted elements will probably stay in the config files forever..

Is there an issue with this from technical @swalker2m  or science @andrewwstephens side?

andrewwstephens:
This sounds reasonable as long as executed SmartCals retain their configuration information.

fnussber:
@andrewwstephens : "as long as executed SmartCals retain their configuration information" that basically tells me we need to allow obsoleted values for smartgcal and keep the configurations. Doing it that way we are sure not to introduce any unwanted side effects. This is also in sync with what we do everywhere else, there is a reason why we don't remove elements but just obsolete them. So probably we should do the same for smart gcal configurations.

arturog8m:
Alright, I'll go ahead and revert the change in the NIRI lookup table we have in OCS code, and will *not* filter obsolete items. I'll update this PR once that is done.

arturog8m:
Ok,  the changes are now in place for allowing obsolete items in smart gcal tables. @fnussber could you look at this and see if this looks correct to you? I re-ran all tests and everything passes now.

bundle/edu.gemini.spModel.smartgcal/src/test/java/edu/gemini/spModel/smartgcal/CalibrationMapTest.java
```diff
@@ -46,17 +46,17 @@ public void doesSupportGNIRSWildcards() {
         Assert.assertEquals(6, keys.size());
 
         // wildcard that matches all available values
-        // this also tests that obsoleted values are not taken into account (13 are defined, 2 of them are obsolete)
+        // Note that obsolete values are allowed!
         properties.setProperty(ConfigKeyGnirs.Values.SLIT_WIDTH.toString(), "*");
         keys = map.createConfig(properties);
-        Assert.assertEquals(11, keys.size());
+        Assert.assertEquals(13, keys.size());
```
cquiroz: Why did the number change? because the removed `isObsoletable` calls?

arturog8m: Right - the previous test was considering that 2 elements in the test set were obsolete, so it was accounting for that - now that we are allowing obsolete items, these 2 elements will be there, increasing the expected count by 2. Same case in a subsequent test.

cquiroz:
Good :+1:

PR #552, opened 2015-09-15
OCSADV-399-O

tpolecat:
This PR removes `getRaDegrees` and fixes a few remaining uses (mostly tests).

### DIFF
--- a/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/SingleProbeStrategy.scala
+++ b/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/SingleProbeStrategy.scala
@@ -11,6 +11,7 @@ import edu.gemini.spModel.core.{Coordinates, Angle}
 import edu.gemini.spModel.core.Target.SiderealTarget
 import edu.gemini.spModel.guide.{ValidatableGuideProbe, VignettingGuideProbe, GuideProbe}
 import edu.gemini.spModel.obs.context.ObsContext
+import edu.gemini.spModel.target.system.CoordinateParam.Units
 import edu.gemini.spModel.target.system.HmsDegTarget
 import edu.gemini.spModel.telescope.PosAngleConstraint._
 import edu.gemini.shared.util.immutable.ScalaConverters._
@@ -250,8 +251,8 @@ object SingleProbeStrategy {
     val ra1    = st.coordinates.ra.toAngle.toRadians
     val dec1   = st.coordinates.dec.toAngle.toRadians
     val target = HmsDegTarget.fromSkyObject(st.toOldModel)
-    val ra2    = Angle.fromDegrees(target.getRaDegrees).toRadians
-    val dec2   = Angle.fromDegrees(target.getDecDegrees).toRadians
+    val ra2    = Angle.fromDegrees(target.getRa.getAs(Units.DEGREES)).toRadians
+    val dec2   = Angle.fromDegrees(target.getDec.getAs(Units.DEGREES)).toRadians
     val raDiff = ra2 - ra1
     val angle  = atan2(sin(raDiff), cos(dec1) * tan(dec2) - sin(dec1) * cos(raDiff))
     Angle.fromRadians(angle)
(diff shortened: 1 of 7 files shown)


bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/SingleProbeStrategy.scala
```diff
@@ -250,8 +251,8 @@ object SingleProbeStrategy {
     val ra1    = st.coordinates.ra.toAngle.toRadians
     val dec1   = st.coordinates.dec.toAngle.toRadians
     val target = HmsDegTarget.fromSkyObject(st.toOldModel)
-    val ra2    = Angle.fromDegrees(target.getRaDegrees).toRadians
-    val dec2   = Angle.fromDegrees(target.getDecDegrees).toRadians
+    val ra2    = Angle.fromDegrees(target.getRa.getAs(Units.DEGREES)).toRadians
+    val dec2   = Angle.fromDegrees(target.getDec.getAs(Units.DEGREES)).toRadians
     val raDiff = ra2 - ra1
```
tpolecat: Here we know we have a sidereal target so we can get the coordinates directly. Once we start correcting for proper motion this will need a time as well.

cquiroz: Could we use `ModelConverter` here?

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/system/ITarget.java
```diff
@@ -75,10 +75,6 @@ public static ITarget forTag(Tag tag) {
         return new Some(getRa().getAs(CoordinateParam.Units.DEGREES));
     }
 
-    public double getRaDegrees() {
```
cquiroz: Should getDecDecrees be deleted too?

bundle/edu.gemini.spdb.rollover.servlet/src/main/scala/edu/gemini/rollover/servlet/RolloverObservation.scala
```diff
@@ -48,9 +50,25 @@ object RolloverObservation {
       targetEnv  <- Option(dataObj.getTargetEnvironment)
       science    <- Option(targetEnv.getBase)
       name       <- Option(science.getTarget.getName)
-      ra         <- Option(science.getTarget.getRaDegrees)
-      dec        <- Option(science.getTarget.getDecDegrees)
-    } yield RolloverTarget(name, new Angle(ra, Angle.Unit.DEGREES), new Angle(dec, Angle.Unit.DEGREES))
+    } yield {
+
+      // Amazingly this is easier in Java
+      val when = o.getDataObject
```
cquiroz: I wonder if this has any impact on ITAC, it calls the `RolloverServlet` to load targets
I'd think this is inconsequential for ITAC but I cannot say for sure

cquiroz:
Seems fine to me, I have to wonder about the impact on ITAC, I believe it has no impact but we should do a sanity check with the experts

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/SingleProbeStrategy.scala
tpolecat: Oh, possibly. I haven't seen that before. I will take a look.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/system/ITarget.java
tpolecat: Yes, in a later PR

bundle/edu.gemini.spdb.rollover.servlet/src/main/scala/edu/gemini/rollover/servlet/RolloverObservation.scala
tpolecat: Well, the `RolloverTarget` now has an optional coordinates pair, which shows up as `<ra>0.0</ra>` in the response. We could change ITAC to handle the option in a better way but I don't see a better immediate solution.

arturog8m:
What is the state of this PR? In addition to resolving the merge conflicts, is there any additional review that has to take place?

cquiroz:
On the ITAC side we need to do a sanity check once the updated rollover servlet is in place. Task REL-2486 was added to keep track of this issue

tpolecat:
Yeah this is fine I think. I just need to rebase. I'm sick today so running at quarter-speed.

tpolecat:
Rebased, waiting for resolution of test failure on `develop`.

PR #553, opened 2015-09-16
OCSADV-335D: Replace eline continuum with squants SpectralIrradiance.

fnussber:
Introduction of squants `SpectralIrradiance` quantity to represent the emission line continuum.

Note that `SpectralIrradiance` is a new `Dimension` that I have added to squants to represent W/m3, or more specifically W/m2/wavelength values. I am also working on getting this new Dimension and some additional units like `erg` (Energy) `erg/s` (Power) and `Angstrom` (Length) into the squants library. For now those values only exist in our separate squants fork and therefore an update of the squants lib is needed here, too.

(Note: I'll deal with the version number of the squants lib later on, for now I just keep the current version number 0.5.3.)


cquiroz:
Looks fine :+1: 
If the squants project wouldn't want to accept the new `Dimensions` for some reason, would we move that code into the ocs?

fnussber:
`SpectralIrradiance` should be general enough that I am pretty sure it will make it into the squants library. If not, then yes, we should moce those additional `Dimension`s and other extensions to squants to the OCS source instead of having them in our squants fork.

PR #554, opened 2015-09-17
ITC Cleanup: Replace magic numbers with IQ, CC, WV, SB enums

fnussber:
A quick ITC cleanup task: as @cquiroz pointed out in one of my last PRs there were places in the code where IQ, CC, WV and SB values were represented by int values instead of an enum. This is a leftover from the old code I haven't bothered to clean up yet. Since we are now using the core model enums for those values I have followed up on that and replaced the magic numbers in switch statements and other places with the corresponding enum values. I also changed some filenames in order to make them easier identifiable and turned `ObservingConditions` into a Scala case class.


bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/base/SEDFactory.java
```diff
@@ -338,5 +338,13 @@ public SourceResult(final VisitableSampledSpectrum sed, final VisitableSampledSp
         }
     }
 
+    private static String airmassCategory(final double airmass) {
```
cquiroz: Should you return an object/enum/type rather than a `String`?

cquiroz:
Looks good, nice to see code reuse :smile:

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/base/SEDFactory.java
fnussber: Probably yes. Although here it really is only about creating a string representation, I know that this concept of an airmass category is used in at least one more place in the ITC. But let me tackle that in a separate PR.

fnussber:
Yes, being able to use the core model stuff in ITC really simplifies a lot of things. This is an ongoing process though that will continue for quite some while. Currently ITC still defines a few things which could probably be recycled from the core model or should be added to the core model. To be continued..

PR #555, opened 2015-09-18
OCSADV-414: Display observation information on the Catalog Navigator

cquiroz:
On this PR, the Catalog Navigator is enhanced to display information about the observation that triggered the manual guide star search. This includes instrument, guider and conditions. A screenshot is included below:

screenshot 2015-09-18 00 22 16

Note that this is a subtask of the overall *Manual Guide Search* task. In this subtask I want to display the observation context info and in later subtasks you could change the conditions to refine the queries.
Observation info is displayed in controls instead of labels, even though those controls don't do anything, as it is duplicated effort to do labels and then change them to combo boxes. In future PRs, these boxes will be used to update the queries

### DIFF
--- a/bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/AgsStrategyCombo.java
+++ b/bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/AgsStrategyCombo.java
@@ -29,8 +29,7 @@ public final class AgsStrategyCombo extends AgsSelectorControl implements Action
             if (o == null || getClass() != o.getClass()) return false;
 
             final ComboEntry that = (ComboEntry) o;
-            if (!name.equals(that.name)) return false;
-            return strategy.equals(that.strategy);
+            return name.equals(that.name) && strategy.equals(that.strategy);
         }
 
         @Override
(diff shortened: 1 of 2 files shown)


swalker2m:
The guider and conditions are over in the query parameters.  Is the idea that they modify the radial and magnitude ranges for you?

cquiroz:
This PR only displays the obs context info but the current query uses the default strategy to define the radius/magnitude constraints. A second step will let the user change the guider and conditions to change the magnitude ranges refining the constraints

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/AgsStrategyCombo.java
```diff
@@ -59,36 +58,27 @@ public void setAgsOptions(final AgsContext opts) {
...
-                }
-            });
-
-        sel.foreach(new ApplyOp<ComboEntry>() {
-            @Override public void apply(ComboEntry comboEntry) {
-                model.setSelectedItem(comboEntry);
+        validEntries.foreach(new ApplyOp<ComboEntry>() {
+            @Override
+            public void apply(ComboEntry comboEntry) {
+                System.out.println("ENT " + comboEntry);
```
swalker2m: Is this just left over debugging?

cquiroz: Right, thanks for noticing it :disappointed:

swalker2m:
Since we present the controls, perhaps we should wait until they are hooked up before merging?

cquiroz:
That's right, we could do that, I didn't want to make an interim PR with just the labels and then switch to controls. I have separate tickets for adapting the controls

cquiroz:
I removed the extra debugging code.
In this sub-task we only want to display the values of the obs context without interaction yet. That will come on future tasks like OCSADV-416 and OCSADV-403. It didn't make sense to do more work than necessary to display the info in labels when it is known that they will be switched to combo boxes
If there are no objections I'll merge soon though I agree that perhaps the subtasks should be more self-contained

arturog8m:
This looks sensible to me :+1:

PR #556, opened 2015-09-21
OCSADV-419-B remove named target from core model

tpolecat:
This removes `NamedTarget` from the new model. It is subsumed by `NonSiderealTarget`.


cquiroz:
:+1:

PR #557, opened 2015-09-21
OCSADV-419-C update ephemeris model

tpolecat:
This PR replaces the ephemeris model with a more efficient and general representation, as a map from `Long` to `Coordinates`, with some machinery for interpolating missing points.
- The `Ephemeris` type itself is an alias for `Long ==>> Coordinates`, where `==>>` is the type constructor for the scalaz tree map.
- Given an instance of `Interpolate[K,V]` we have syntax allowing interpolated lookups of "missing" values in `K ==>> V`. This allows us to make the ephemeris appear to be as continuous as we like, within its actual min/max boundaries.

~~the `Coordinates` interpolator is wrong when crossing the pole, which shouldn't be an issue in practice but I'd like to make it work.~~ **Edit:** fixed


bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/Interpolate.scala
```diff
@@ -0,0 +1,76 @@
...
+
+  implicit val LongDeclinationInterpolation: Interpolate[Long, Declination] =
+    LongDoubleInterpolation.xmap(
+      d => Declination.fromAngle(Angle.fromDegrees(d)).getOrElse(sys.error("Declination out of range.")),
+      _.toDegrees) // sigh
+
+  implicit val InterpolateCoordinates: Interpolate[Long, Coordinates] =
+    (LongRightAscensionInterpolation zip LongDeclinationInterpolation).xmap(
+      Coordinates.tupled, c => (c.ra, c.dec))
+
```
tpolecat: You can't really do this, because when the declination hits a pole you have to flip the whole thing around. So this needs to be better.

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/package.scala
```diff
@@ -8,4 +11,30 @@ package object core {
   type Dec = Declination
   val  Dec = Declination
 
-}
\ No newline at end of file
+  type Ephemeris = Long ==>> Coordinates
```
cquiroz: Would it make sense to us an actual `Date` (please the Java 1.8 date) instead of `Long`?

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/package.scala
```diff
@@ -8,4 +11,30 @@ package object core {
...
+  /** Operations for maps of types that we can interpolate.  We use these for Ephemerides. */ 
+  implicit class MoreMapOps[K, V](m: K ==>> V)(implicit O: Order[K], I: Interpolate[K,V]) {
+
+    /** Perform an exact or interpolated lookup. */
+    def iLookup(k: K): Option[V] =
+      m.lookup(k) match {
+        case Some(v) => Some(v)
+        case None    => 
+          val (lt, gt) = m.split(k)
+          ^(lt.findMax, gt.findMin)(I.interpolate(_, _, k))
```
cquiroz: What does the `^` operator do?

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/package.scala
```diff
@@ -8,4 +11,30 @@ package object core {
...
+
+    /** Construct an exact or interpolated slice. */
+    def iSlice(lo: K, hi: K): Option[K ==>> V] =
+      ^(iLookup(lo), iLookup(hi)) { (lov, hiv) => 
+        m.trim(O(_, lo), O(_, hi)) + (lo -> lov) + (hi -> hiv)
+      }
+
+    /** Construct a table of (K, V) values on the given interval. */
+    def iTable(lo: K, hi: K, step: K)(implicit ev: Integral[K]): Option[List[(K, V)]] =
+      NumericRange.inclusive(lo, hi, step).toList.traverse(k => iLookup(k).strengthL(k))
```
cquiroz: Is this equivalent to `lo to hi`? Is this better is some way? Due to `step` perhaps?

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/Interpolate.scala
```diff
@@ -0,0 +1,73 @@
...
+  implicit def zipInterpolate[T]: Zip[({ type l[b] = Interpolate[T, b] })#l] =
+    new Zip[({ type l[b] = Interpolate[T, b] })#l] {
+      def zip[A, B](a: => Interpolate[T,A], b: => Interpolate[T,B]): Interpolate[T,(A, B)] = a.zip(b)
+    }
+
+  implicit val LongDoubleInterpolation: Interpolate[Long, Double] =
+    new Interpolate[Long, Double] {
+      def interpolate(a: (Long, Double), b: (Long, Double), c: Long): Double = {
+        val ((n1, d1), (n2, d2)) = (a, b)
+        d1 + (d2 - d1) * (c.toDouble - n1.toDouble) / (n2.toDouble - n1.toDouble)
```
cquiroz: I guess we should consider the case that `n2.toDouble - n1.toDouble` is zero? Do the tests cover that case?

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/Interpolate.scala
```diff
@@ -0,0 +1,73 @@
...
+      }
+    }
+
+  implicit val InterpolateCoordinates: Interpolate[Long, Coordinates] =
+    new Interpolate[Long, Coordinates] {
+      def interpolate(a: (Long, Coordinates), b: (Long, Coordinates), c: Long): Coordinates = {
+        val ((n1, c1), (n2, c2)) = (a, b)
+        val f =  ((c.toDouble - n1.toDouble) / (n2.toDouble - n1.toDouble))
```
cquiroz: Same as above, what about the degenerate case?

cquiroz:
Looks very nice, I started wondering why we are using `Long` for dates rather than an actual `Date` type, the new 1.8 Date API should be really usable. This is not only for this PR but in general to check coordinates at a given date
You say it works crossing the pole, is there a test that explicitly checks that?

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/package.scala
tpolecat: It's syntax for `lift2` ... `^(a, b)(f)` is equivalent to `(a |@| b)(f)`. I'll switch it to the latter since we use that one more.

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/Interpolate.scala
tpolecat: Yeah I'll check the zero cases. Looks like it won't work.

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/package.scala
tpolecat: I had not considered that, but I guess it makes sense now that we're using Java 8. I'll see how it looks.

tpolecat:
Yes, it works going across the poles but I'll add an explicit test. It was tricky.

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/package.scala
tpolecat: You can't say `lo to hi by step` with arbitrary `K: Integral` .. it only works for concrete types. Probably an oversight.

tpolecat:
:-1: yeah I need to do this differently. You have to find the shortest path on a great circle and interpolate along that. To be continued.

tpolecat:
Formulae [here](http://www.movable-type.co.uk/scripts/latlong.html) should work, assuming ra/dec ~ lon/lat.

tpolecat:
Ok I added the angular separation and great-circle interpolation code and tests thereof.

tpolecat:
I'm leaving the time representation as `Long` for now. Changing it has implications throughout OCS so I'd prefer to consider it some other time.

This PR needs another look. Thanks.

tpolecat:
Rebased. Tests pass. Any more comments?

cquiroz:
Seems fine to me

PR #558, opened 2015-09-23
REL-2382-A: Update Simbad name resolvers

cquiroz:
This is the first PR to support Simbad catalogs in the catalog code. This PR contains substantial changes to `VoTableParser` to support simbad VoTable format and initial support to make remote queries no simbad in `VoTableClient`

These changes are contained to the catalog bundle, while next PR will include updates to the target model to capture extra metadata about the targets (Radial Velocity, etc) and integration with the OT

As usual I include sample votable queries for tests

### DIFF
--- a/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/GemsStrategy.scala
+++ b/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/GemsStrategy.scala
@@ -78,8 +78,8 @@ trait GemsStrategy extends AgsStrategy {
         for {
           result                                                    <- agsCatalogResults
           angle                                                     <- anglesToTry
-          q @ ConeSearchCatalogQuery(_, _, radiusConstraint, mc, _) = result.query
         } yield {
+          val ConeSearchCatalogQuery(_, _, radiusConstraint, mc, _) = result.query
           val catalogSearchCriterion     = CatalogSearchCriterion("ags", radiusConstraint, mc.head, None, angle.some)
           val gemsCatalogSearchCriterion = new GemsCatalogSearchCriterion(result.searchKey, catalogSearchCriterion)
           new GemsCatalogSearchResults(gemsCatalogSearchCriterion, result.catalogResult.targets.rows)
--- a/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/GemsStrategy.scala
+++ b/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/GemsStrategy.scala
@@ -75,17 +75,15 @@ trait GemsStrategy extends AgsStrategy {
     val anglesToTry = (0 until 360 by 45).map(Angle.fromDegrees(_))
 
     futureAgsCatalogResults.map { agsCatalogResults =>
-      for {
-        result  <- agsCatalogResults
-        angle   <- anglesToTry
-      } yield {
-        val query                      = result.query
-        val radiusConstraint           = query.radiusConstraint
-        val mc                         = query.magnitudeConstraints
-        val catalogSearchCriterion     = CatalogSearchCriterion("ags", radiusConstraint, mc.head, None, angle.some)
-        val gemsCatalogSearchCriterion = new GemsCatalogSearchCriterion(result.searchKey, catalogSearchCriterion)
-        new GemsCatalogSearchResults(gemsCatalogSearchCriterion, result.catalogResult.targets.rows)
-      }
+        for {
+          result                                                    <- agsCatalogResults
+          angle                                                     <- anglesToTry
+          q @ ConeSearchCatalogQuery(_, _, radiusConstraint, mc, _) = result.query
+        } yield {
+          val catalogSearchCriterion     = CatalogSearchCriterion("ags", radiusConstraint, mc.head, None, angle.some)
+          val gemsCatalogSearchCriterion = new GemsCatalogSearchCriterion(result.searchKey, catalogSearchCriterion)
+          new GemsCatalogSearchResults(gemsCatalogSearchCriterion, result.catalogResult.targets.rows)
+        }
     }
   }
 
@@ -220,8 +218,8 @@ trait GemsStrategy extends AgsStrategy {
         (ml |@| lim(can))(_ union _).flatten
       }
 
-      val canopusConstraint = canMagLimits.map(c => CatalogQuery(CanopusTipTiltId.some, base.toNewModel, RadiusConstraint.between(Angle.zero, Canopus.Wfs.Group.instance.getRadiusLimits.toNewModel), List(ctx.getConditions.adjust(c)), ucac4))
-      val odgwConstraint    = odgwMagLimits.map(c => CatalogQuery(OdgwFlexureId.some,   base.toNewModel, RadiusConstraint.between(Angle.zero, GsaoiOdgw.Group.instance.getRadiusLimits.toNewModel), List(ctx.getConditions.adjust(c)), ucac4))
+      val canopusConstraint = canMagLimits.map(c => CatalogQuery(CanopusTipTiltId, base.toNewModel, RadiusConstraint.between(Angle.zero, Canopus.Wfs.Group.instance.getRadiusLimits.toNewModel), List(ctx.getConditions.adjust(c)), ucac4))
+      val odgwConstraint    = odgwMagLimits.map(c => CatalogQuery(OdgwFlexureId,   base.toNewModel, RadiusConstraint.between(Angle.zero, GsaoiOdgw.Group.instance.getRadiusLimits.toNewModel), List(ctx.getConditions.adjust(c)), ucac4))
       List(canopusConstraint, odgwConstraint).flatten
     }
 
--- a/bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableClient.scala
+++ b/bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableClient.scala
@@ -157,11 +157,47 @@ case object RemoteBackend extends CachedBackend {
       new NameValuePair("RA", format(qs.base.ra.toAngle)),
       new NameValuePair("DEC", f"${qs.base.dec.toDegrees}%4.03f"),
       new NameValuePair("SR", format(qs.radiusConstraint.maxLimit)))
-    case _ => Array.empty
+    case _                          => Array.empty
   }
 
   override protected def query(e: SearchKey): QueryResult = {
     val method = new GetMethod(s"${e.url}/cgi-bin/conesearch.py")
+    val qs = queryParams(e.query)
+    method.setQueryString(qs)
+    Log.info(s"Catalog query to ${method.getURI}")
+
+    val client = new HttpClient
+    client.setConnectionTimeout(timeout)
+
+    try {
+      client.executeMethod(method)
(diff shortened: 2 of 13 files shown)


bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableClient.scala
```diff
@@ -101,8 +102,9 @@ trait CachedBackend extends VoTableBackend {
       @tailrec
       def go(pos: Int):Option[(Int, QueryResult)] =
         a.lift(pos) match {
-          case None    => None
-          case Some(x) => if (x.k.query.isSuperSetOf(k.query)) Some((pos, x.v)) else go(pos + 1)
+          case None                                                            => None
+          case Some(CacheEntry(SearchKey(query:ConeSearchCatalogQuery, r), v)) => if (query.isSuperSetOf(k.query)) Some((pos, v)) else go(pos + 1)
+          case Some(_)                                                         => None // Not caching named queries so far
```
sraaphorst: Just out of curiosity, is there a reason you are matching on `Some(_)` here instead of just `_`?
It shouldn't make a difference, I wouldn't think.

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableParser.scala
```diff
@@ -48,53 +49,114 @@ object VoTableParser extends VoTableParser {
...
+  // filter magnitudes as a whole, removing invalid values and duplicates
+  def filterAndDeduplicateMagnitudes(magnitudeFields: List[(FieldId, Magnitude)]): List[Magnitude] = magnitudeFields.collect { case (_, mag) if validMagnitude(mag) => mag }
+  // Attempts to extract a band and value for a magnitude from a pair of field and value
+  def parseMagnitude(p: (FieldId, String)): CatalogProblem \/ (FieldId, MagnitudeBand, Double) = {
+    val (fieldId: FieldId, value: String) = p
+
+    val band = fieldToBand(fieldId)
+
+    for {
+      b <- band \/> UnmatchedField(fieldId.ucd)
```
sraaphorst: A random question: does anyone have a good resource to learn all of this scalaz magic?

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableClient.scala
cquiroz: I thought about adding caching to `NamedQuery` right away but then decided against it so it would have used `Some(CacheEntry(SearchKey(query:NamedQuery, r), v))`
So, you are right, it can be replaced by `_`

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableParser.scala
cquiroz: IMHO there are no good  resources to learn scalaz, you need to pick bits and pieces from videos and blogs, or ask rob/shane
A useful resource though are the tutorials below, though they are still in my opinion a bit cryptic
http://eed3si9n.com/learning-scalaz/

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableParser.scala
```diff
@@ -220,36 +286,37 @@ trait VoTableParser {
...
-
-    result.getOrElse(\/.left(MissingValues(missing)))
+      adapter <- catalogAdapter
+      id      <- entries.get(adapter.idField) \/> MissingValue(adapter.idField)
+      ra      <- entries.get(adapter.raField) \/> MissingValue(adapter.raField)
+      dec     <- entries.get(adapter.decField) \/> MissingValue(adapter.decField)
+      (pmRa, pmDec)   = (entries.get(adapter.pmRaField), entries.get(adapter.pmDecField))
+    } yield toSiderealTarget(id, ra, dec, (pmRa, pmDec))
+
+    result.flatMap(identity)
```
sraaphorst: What is the difference between `result.flatMap(identity)` and `result.flatten`?

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableParser.scala
sraaphorst: Thanks! I'll take a look. A lot of scalaz is making its way into the code base and I feel at a disadvantage for not knowing more!

sraaphorst:
Even though I won't claim to fully understand every detail of what has been done here, looking through it, my impression is :+1:

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableParser.scala
cquiroz: `result` is a disjunction `CatalogProblem \/ (CatalogProblem \/ Target.SiderealTarget`). There is no `flatten` on the scalaz disjunction but you can do `flatMap` to extract the inner CatalogProblem if it exists.
PS I bet there is a more elegant way to do this :smile:

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableParser.scala
cquiroz: BTW the `\/>` operator takes an `Option` like in the case above and if it `None` i will convert it into a left disjunction with content `UnmatchedField`
In this case it basically tells us that `fieldToBand` tried to extract the band from the fieldId but couldn't, for example, due to parsing errors

cquiroz:
Updated with Sebastian's comments and fixed some conflicts with upstream

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableParser.scala
tpolecat: Yeah the `flatten` method is called `.join` in scalaz in order to not conflict with stdlib (and `join` is the common name for this operation in FP generally). But you are @sraaphorst you are correct:

1. `ma.flatMap(f)` = `ma.map(f).join` // monad law
2. `ma.flatMap(identity)` = `ma.map(identity).join` // substitute `identity` for `f`
3. `ma.flatMap(identity)` = `ma.join` // functor identity

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableParser.scala
tpolecat: @sraaphorst yeah Eugene's tutorial is pretty good (but slightly out of date I think). You can also just ask me, or get on the `#scala` IRC channel on FreeNode and ask questions.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/GemsStrategy.scala
```diff
@@ -75,17 +75,15 @@ trait GemsStrategy extends AgsStrategy {
...
-      }
+        for {
+          result                                                    <- agsCatalogResults
+          angle                                                     <- anglesToTry
+          q @ ConeSearchCatalogQuery(_, _, radiusConstraint, mc, _) = result.query
+        } yield {
+          val catalogSearchCriterion     = CatalogSearchCriterion("ags", radiusConstraint, mc.head, None, angle.some)
+          val gemsCatalogSearchCriterion = new GemsCatalogSearchCriterion(result.searchKey, catalogSearchCriterion)
+          new GemsCatalogSearchResults(gemsCatalogSearchCriterion, result.catalogResult.targets.rows)
+        }
```
tpolecat: Can you move that let-binding into the `yield`?

```scala
val ConeSearchCatalogQuery(_, _, radiusConstraint, mc, _) = result.query
```

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableClient.scala
```diff
@@ -148,11 +150,14 @@ case object RemoteBackend extends CachedBackend {
...
-    new NameValuePair("DEC", f"${qs.base.dec.toDegrees}%4.03f"),
-    new NameValuePair("SR", format(qs.radiusConstraint.maxLimit)))
+  protected [votable] def queryParams(q: CatalogQuery): Array[NameValuePair] = q match {
+    case qs: ConeSearchCatalogQuery => Array(
+      new NameValuePair("CATALOG", qs.catalog.id),
+      new NameValuePair("RA", format(qs.base.ra.toAngle)),
+      new NameValuePair("DEC", f"${qs.base.dec.toDegrees}%4.03f"),
+      new NameValuePair("SR", format(qs.radiusConstraint.maxLimit)))
+    case _                          => Array.empty
+  }
```
tpolecat: Why is this an `Array`? Is it needed for some old API?

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableClient.scala
```diff
@@ -173,8 +178,44 @@ case object RemoteBackend extends CachedBackend {
...
+    val qs = queryParams(e.query)
+    method.setQueryString(qs)
+    Log.info(s"Catalog query to ${method.getURI}")
+
+    val client = new HttpClient
+    client.setConnectionTimeout(timeout)
+
+    try {
+      client.executeMethod(method)
+      VoTableParser.parse(e.url, method.getResponseBodyAsStream).fold(p => QueryResult(e.query, CatalogQueryResult(TargetsTable.Zero, List(p))), y => QueryResult(e.query, CatalogQueryResult(y)))
```
tpolecat: For things like this I prefer a match to a fold because it allows you to compare the cases visually (you could also do this by adding newlines but I don't really like newlines inside parens). Not a huge deal but I think this reads better.

```scala
VoTableParser.parse(e.url, method.getResponseBodyAsStream) match {
  case -\/(p) => QueryResult(e.query, CatalogQueryResult(TargetsTable.Zero, List(p)))
  case \/-(y) => QueryResult(e.query, CatalogQueryResult(y)))
}
```

tpolecat:
Looks great. I can't imagine writing this in Java. :+1:

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableParser.scala
cquiroz: Good to know, I'll switch to `ma.join`

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/GemsStrategy.scala
cquiroz: Nice, I didn't know you could do that

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableClient.scala
cquiroz: Yes, we have a very old version of `http-client` from apache commands that uses java Arrays

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableClient.scala
cquiroz: Cool idea, :+1:

cquiroz:
I incorporated the last comments and will merge now

PR #559, opened 2015-09-25
OCSADV-403: Let the user choose the selected AGS strategy on manual search

cquiroz:
This is the next step on the catalog navigator. With this PR the user can select a different guider and re-query with updated limits. By user request the initial load will read all the available rows without magnitude limits.
Note that the limits for different guide speeds for each guider are displayed too

screenshot 2015-09-25 10 11 47


tpolecat:
:+1:

PR #560, opened 2015-09-25
Broken Gems AGS test

cquiroz:
A filtering of invalid magnitudes was removed by mistake and that let some extra targets through breaking a few Gems AGS test

This PR fixes  this, tests in catalog and ags pass


tpolecat:
:+1:

PR #561, opened 2015-09-28
OCSADV-399-P

tpolecat:
This PR addresses the last few uses of non-time-based coordinate access on `ITarget`. This doesn't affect the behavior of the template parameter editor yet, but the UI will need to be revisited when the model is swapped out because "set RA" for example only makes sense for sidereal targets.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/system/ITarget.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/system/ITarget.java
@@ -79,10 +79,6 @@ public abstract class ITarget implements Cloneable, Serializable {
         return new Some(getRa().getAs(CoordinateParam.Units.HMS));
     }
 
-    public double getRaHours() {
-        return getRa().getAs(CoordinateParam.Units.HMS);
-    }
-
     public Option<String> getRaString(Option<Long> time) {
         return new Some(getRa().toString());
     }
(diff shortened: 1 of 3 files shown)


arturog8m:
Having these two methods `SPTarget.getTarget()` and `TemplateParameters.getTarget():SPTarget` creates calls like `getTarget.getTarget` which look confusing - so probably we should look at improving that API in the future. However, this is not this PR's business. All looks good to me :+1:

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/system/ITarget.java
```diff
@@ -79,10 +79,6 @@ public static ITarget forTag(Tag tag) {
         return new Some(getRa().getAs(CoordinateParam.Units.HMS));
     }
 
-    public double getRaHours() {
-        return getRa().getAs(CoordinateParam.Units.HMS);
-    }
-
     public Option<String> getRaString(Option<Long> time) {
```
cquiroz: Why isn't the `time` param used?

tpolecat: In the old model the coordinates are not time-varying. This extended refactoring exercise is simply to introduce the parameter into the API, to make it possible to introduce the new model in place of `ITarget`.

cquiroz:
Looks good :+1:

tpolecat:
Rebased again. Tests pass.

PR #562, opened 2015-09-28
REL-2482: Enable second Altair dichroic, update labels

arturog8m:
This is a PR to enable the second Altair dichroic, which was never been used. In the model, these are represented by the `AltairParams.Wavelength` type. Although the changes could have been made just in the GUI, I decided to update this type to use names that don't have the actual wavelength encoded - original names were `WAVELENGTH_850NM` and `WAVELENGTH_1UM`, which now wouldn't make sense. So I changed them to `WAVELENGTH_A` and `WAVELENGTH_B`, just to impress you with my creativity. I also change the `sequenceValue` to use the enumerated `name()` rather than its display value. This is going to require a change in the TCC, but I felt it was better to separate the display business from the names we use to communicate with other subsystems. @jluhrs will take care of the TCC business, it's a one line change. 

Conversion from old programs will work as the old dichroic was never used, so when we import existing programs with `WAVELENGTH_850NM`, we will get a warning in the import and we will set up the value to the new default, `WAVELENGTH_A`, so things should just work.


bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/altair/AltairParams.java
```diff
@@ -44,14 +44,7 @@ public String displayValue() {
...
         public String sequenceValue() {
-            return _displayValue;
-        }
-
-        /**
-         * Return a Wavelength by index *
-         */
-        static public Wavelength getWavelengthByIndex(int index) {
-            return SpTypeUtil.valueOf(Wavelength.class, index, DEFAULT);
+            return name();
```
tpolecat: I really don't like this convention because it means that we can no longer rename-refactor the identifiers without changing the meaning of the program. It is in effect a very subtle kind of reflection. So I would prefer a distinct string field in this case and others like it.

I realize that there's lots of this kind of thing in OCS, but we should fix it when we encounter it.

Does this make sense?

arturog8m: Makes sense, you are right. I'll create a separate string field for the sequence value that you will need to provide at construction.

arturog8m:
Ok. Commit 360c343 uses a separate string field for the sequence value so we can keep a distinction between what is used for display and the sequence in this type.

tpolecat:
:+1:

PR #563, opened 2015-09-30
OCSADV-335E: Allow selection of unit for emission line flux and continuum

fnussber:
Reaping the fruits of all the preparatory tasks to introduce squants: Finally we can take advantage of the additional unit that comes with all squants quantities and make it editable by the user for the flux and continuum in emission lines in the spectral distribution.

This PR adds squants quantities support to the `NumericPropertySheet` in order to make it simple to do similar things in the future for other values. See also attached screenshot.

There are also new `addQuantity()` and `getQuantity()` methods on `Pio.java`. I took a shortcut and followed the current scheme on how errors are handled by ignoring them and returning a default value (for quantities: 0 of the primary unit of the given quantity). Of course this could be done better, but it didn't seem to make sense to introduce a different way (e.g. `Try` or `\/`) of how errors are handled for quantities(?). 

Note that a weakness of our current approach of implementing `Wavelength` as  a value class instead of a "real" squants quantity with all the bells and whistles is that we can not use it the same way we can use other quantities, which currently keeps us from allowing users to select different units. Probably it would make sense to turn `Wavelength` into a real quantity (i.e. it extends the appropriate traits) so that we can use it equally to all other quantities (like e.g. velocities, lengths etc). If necessary this should probably be done as a separate PR and not as part of this one. OR: we back down and use a squants `Length` instead of our own `Wavelength` type.

This also comes with a new squants library because I had to fix a bug, a PR for fixing this is pending with the squants project...

This PR also contains the necessary migration code for adding units to the values that have been changed to squants quantities and now need a unit in the XML representation.

image


cquiroz:
This looks very good, I think the discussion about converting `Wavelength` to a squants quantity is worthwhile
I'm facing a similar decision when adding `RadialVelocity` to the target model

fnussber:
How about adding a separate PR asap where I follow up with turning `Wavelength` into a squants quantitiy? This would look something like: 
```
final class Wavelength private (val value: Double, val unit: LengthUnit) extends Quantity[Wavelength] {
   ......... crazy code ..........
}
```
Then we can look at the code and decide what would be the best way forward. Turning `Wavelength` into a quantity shouldn't be too difficult I think and once we have this, we have a "template" for other core model quantities (like e.g. Carlo's `RadialVelocity`).

cquiroz:
That sounds good to me :smile:

PR #564, opened 2015-09-30
OCSADV-419-D horizons designation

tpolecat:
This PR introduces `HorizonsDesignation` which finally gives us a way to have repeatable Horizons queries with unique results. It also moves the Horizons information from `Target` to `NonSiderealTarget` where it belongs (duh).


bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/HorizonsDesignation.scala
```diff
@@ -0,0 +1,25 @@
...
+/** 
+ * Unique Horizons designation, which must be formatted differently depending on the object type.
+ * The `queryString` should yield a unique result.
+ */
+sealed abstract class HorizonsDesignation(val designation: String, val queryString: String) 
+  extends Product with Serializable
+
+object HorizonsDesignation {
+
+  final case class Comet(des: String)     extends HorizonsDesignation(des, s"DES=$des;CAP")
```
arturog8m: Just to reduce a bit my ignorance: What are these `s"DES=$des;CAP"`  doing? what kind of query is that? where does it go?

tpolecat: This is the literal search string you send to Horizons. Andy and I talked to the JPL guy and figured it out after a lot of experimentation.
- For comets you need `DES=...;CAP` which says use the provided designation, with the current apparition (i.e., the newest ephemeris). So the current orbit of Halley's comet is `DES=1P;CAP`.
- For asteroids you must have `DES=` but cannot have `;CAP`. So 599 Luisa is `DES=1906 UJ`.
- And for major bodies like planets and their moons you cannot have either. So Mercury is just `1`.

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/Target.scala
```diff
@@ -11,9 +11,6 @@ sealed trait Target {
   /** Coordinates (if known) for this target at the specified UNIX time. */
   def coords(time: Long): Option[Coordinates]
 
-  /** Horizons information for this target, if known. */
-  def horizonsInfo: Option[Target.HorizonsInfo]
```
cquiroz: I guess this have some effects downstream but I'd assume you have checked them already

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/HorizonsDesignation.scala
arturog8m: Thanks. Probably this is not documented anywhere in the Horizons documentation. So probably it's worth some words in this code or somewhere else?

cquiroz:
Looks fine, I'd do a full rebuild to check if the removed field on `SiderealTarget` has some effects, e.g. on AGS, I don't think we use it but it may fail to compile

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/Target.scala
tpolecat: Yeah nobody is using it.

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/HorizonsDesignation.scala
tpolecat: Not documented. Horizons is terrible. I'll add some comments.

tpolecat: Actually it turns out that the largest asteroids don't fit this pattern (Ceres and Vesta don't have primary designations) so we may need another case. Ugh.

tpolecat:
@cquiroz good catch, I was working locally and hadn't propagated the changes. Compiles clean. Tests pass.

cquiroz:
Good :+1:

tpolecat:
Ok @arturog8m I added a markdown doc at the `spModel.core` root explaining what we know about HORIZONS identifiers. Here's a [direct link](https://github.com/tpolecat/ocs/blob/new-model-4/bundle/edu.gemini.spModel.core/HORIZONS.md).

arturog8m:
Looks good! :+1:

PR #565, opened 2015-09-30
OCSADV-420: Search by name to import targets

cquiroz:
This PR uses the support for Simbad queries added on PR #558 with the Catalog Navigator. It will do named queries and take the first result and replace the coordinates on the query form:

screenshot 2015-09-30 15 45 51

Unlike the target editor, errors are displayed without a popup, instead a message is displayed on the form and at the window's bottom:
screenshot 2015-09-30 17 13 15

Simbad returns non-valid votable documents if no targets are found, this PR handles that case too

### DIFF
--- a/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsVoTableCatalog.scala
+++ b/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsVoTableCatalog.scala
@@ -24,7 +24,7 @@ import jsky.util.gui.StatusLogger
  * The catalog search will provide the inputs to the analysis phase, which actually assigns guide stars to guiders.
  * See OT-26
  */
-case class GemsVoTableCatalog(backend: VoTableBackend = RemoteBackend, catalog: CatalogName = ucac4) {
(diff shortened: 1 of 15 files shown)


fnussber:
As a general comment: Would it be possible to display the error feedback the same way we are displaying such messages in other places like e.g. `GuidingFeedback`, `VcsStatusPanel` and `ItcFeedbackPanel` as a colored label at the bottom of the UI element (see screenshots)?

image

image

This would use the same UI convention across different parts of the OT. It might even make sense to try to pull the common code out of all these places into a reusable UI component. But of course that is going far beyond this PR and should be done in a separate PR.

cquiroz:
Consistency does make sense, we should attempt to unify the error display for sure, IMHO at least we should get rid of the popups

cquiroz:
I added a new task OCSADV-421 to unify the error message display with the OT, When we integrate the new simbad code into the main window we should revisit error displaying

cquiroz:
Any comments on the code or GUI? If not I'll merge this by tomorrow morning

tpolecat:
This looks good. :+1:

PR #567, opened 2015-10-02
OCSADV-406: Support changing the gemini-hosted catalog (e.g. to PPMXL)

cquiroz:
This is a small update to the Catalog Navigator adding  a combo box to choose among the available catalogs (Right now just UCAC4 and PPMXL hosted at Gemini)

see the screenshot below:

screenshot 2015-10-02 10 20 59

### DIFF
--- a/bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/api/CatalogQuery.scala
+++ b/bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/api/CatalogQuery.scala
@@ -18,6 +18,21 @@ case class MagnitudeQueryFilter(mc: MagnitudeConstraints) extends QueryResultsFi
   def filter(t: SiderealTarget): Boolean = mc.filter(t)
 }
 
+sealed abstract class CatalogName(val id: String, val displayName: String)
+
+case object sdss extends CatalogName("sdss9", "SDSS9 @ Gemini")
+case object gsc234 extends CatalogName("gsc234", "GSC234 @ Gemini")
+case object ppmxl extends CatalogName("ppmxl", "PPMXL @ Gemini")
+case object ucac4 extends CatalogName("ucac4", "UCAC4 @ Gemini")
+case object twomass_psc extends CatalogName("twomass_psc", "TwoMass PSC @ Gemini")
+case object twomass_xsc extends CatalogName("twomass_xsc", "TwoMass XSC @ Gemini")
+case object simbad extends CatalogName("simbad", "Simbad")
+
+object CatalogName {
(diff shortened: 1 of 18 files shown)


bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/api/CatalogQuery.scala
```diff
@@ -18,6 +18,21 @@ case class MagnitudeQueryFilter(mc: MagnitudeConstraints) extends QueryResultsFi
...
+sealed abstract class CatalogName(val id: String, val displayName: String)
+
+case object sdss extends CatalogName("sdss9", "SDSS9 @ Gemini")
+case object gsc234 extends CatalogName("gsc234", "GSC234 @ Gemini")
+case object ppmxl extends CatalogName("ppmxl", "PPMXL @ Gemini")
+case object ucac4 extends CatalogName("ucac4", "UCAC4 @ Gemini")
+case object twomass_psc extends CatalogName("twomass_psc", "TwoMass PSC @ Gemini")
+case object twomass_xsc extends CatalogName("twomass_xsc", "TwoMass XSC @ Gemini")
+case object simbad extends CatalogName("simbad", "Simbad")
+
```
tpolecat: It's important to capitalize the names of case objects because otherwise you cannot pattern-match on them without some trickery. For example, this method will always return `0` because lowercase identifiers in patterns match everything.

```scala
def foo(cn: CatalogName): Int = 
  cn match {
    case sdss        => 0
    case gsc234      => 1
    case ppmxl       => 2
    case ucac4       => 3
    case twomass_psc => 4
    case twomass_xsc => 5
    case simbad      => 6
  }
```

So unless you have an *extremely* good reason for keeping these lower-case, I think you should switch them,

cquiroz: Good reminder, I have the feeling I was already hit by that issue and I switched them to uppercase but clearly I reverted

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/api/CatalogQuery.scala
```diff
@@ -18,6 +18,21 @@ case class MagnitudeQueryFilter(mc: MagnitudeConstraints) extends QueryResultsFi
...
+object CatalogName {
+  implicit val equals = Equal.equal[CatalogName]((a, b) => a.id === b.id)
+
```
tpolecat: This is fine, but just fyi you can express this as 

```scala
Equal.equalBy((_ : CatalogName).id)
```

tpolecat:
Other than the `case object` thing above, this looks great. :+1:

PR #568, opened 2015-10-02
OCSADV-410: Bags of BAGS

sraaphorst:
Hi all,

I'm sorry for the incredibly massive PR. I closed the previous one as there were some issues in that one regarding GeMS handling and some logic, and opened a new mammoth one that you will likely all hate me for :-).

Little has changed in the `GuideProbeTargets` from the previous PR except some method names, and @cquiroz already reviewed that code previously in the earlier PR.

The main addition here is the `BagsManager` and related classes, which handle the background lookups. Any feedback on this would be highly appreciated.

@cquiroz: If you could look at the GeMS changes I made, especially to `GemsGuideStarWorker`, and let me know if you see any issues or have suggestions, that would be great. I cleaned up some things such as unused method parameters and made `static` methods in `GemsGuideStarWorker` to encapsulate code that could be performed without GUI interaction so that this class could be used by both manual and auto GS searches.

@tpolecat: If you could review the general quality of the scala code and make any suggestions, that would be much appreciated!

I realize that this is likely to be quite time consuming, unfortunately, so I'm not sure if I will be able to have a build today. I found a bug today in imported programs with existing guide stars (and no BAGS guide star) that I fixed, and improved the code so as to minimize any unnecessary searches.


tpolecat:
:open_mouth: okay, i better eat some lunch first

cquiroz:
At least there is more deleted code than added :smile:

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsVoTableCatalog.scala
```diff
@@ -67,7 +67,7 @@ case class GemsVoTableCatalog(backend: VoTableBackend = ConeSearchBackend, catal
     }))
   }
 
-  private def searchCatalog(basePosition: Coordinates, criterions: List[GemsCatalogSearchCriterion], statusLogger: StatusLogger): Future[List[GemsCatalogSearchResults]] = {
+  private def searchCatalog(basePosition: Coordinates, criterions: List[GemsCatalogSearchCriterion]): Future[List[GemsCatalogSearchResults]] = {
     val queryArgs = for {
       c <- criterions
       q = CatalogQuery(basePosition, c.criterion.radiusConstraint, c.criterion.magConstraint, catalog)
```
tpolecat: Not strictly part of your PR, but this can be written as a simple call to `map`, which I prefer. 

```scala
val queryArgs = criterions.map { c =>
  val q = CatalogQuery(basePosition, c.criterion.radiusConstraint, c.criterion.magConstraint, catalog)
  (q, c)
}
```

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsVoTableCatalog.scala
```diff
@@ -38,8 +38,8 @@ case class GemsVoTableCatalog(backend: VoTableBackend = ConeSearchBackend, catal
    * @param timeout      Timeout in seconds
    * @return list of search results
    */
-  def search4Java(obsContext: ObsContext, basePosition: Coordinates, options: GemsGuideStarSearchOptions, nirBand: Option[MagnitudeBand], statusLogger: StatusLogger, timeout: Int = 10): java.util.List[GemsCatalogSearchResults] =
-    Await.result(search(obsContext, basePosition, options, nirBand, statusLogger), timeout.seconds).asJava
```
cquiroz: Good riddance!

bundle/edu.gemini.oodb.too.url/src/main/java/edu/gemini/spdb/rapidtoo/impl/TooUpdateServiceImpl.java
```diff
@@ -220,78 +218,77 @@ private ISPObsComponent _findComponent(ISPObservation obs, SPComponentType type)
...
-                            probe = getOiwfs(obs);
-                            break;
-                        case PWFS1:
-                            probe = PwfsGuideProbe.pwfs1;
-                            break;
-                        case PWFS2:
-                            probe = PwfsGuideProbe.pwfs2;
-                            break;
-                    }
+                // damn
```
cquiroz: damn?

sraaphorst: Haha! I think that was already there regarding the cases, and just got moved or realigned. I'm really not sure, but I think we can probably take it out :-).

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsVoTableCatalog.scala
sraaphorst: May as well clean it up while we're here. I'll make the change.

tpolecat:
@sraaphorst the Scala looks good. The `BagsManager` is very side-effecty but this is pretty much unavoidable in the OT. I did not see any obvious issues. :+1:

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/GuideGroup.java
```diff
@@ -155,7 +155,7 @@ public GuideGroup setName(String name) {
      * instance; <code>false</code> otherwise
      */
     public boolean contains(GuideProbe guider) {
-        return guideTargets.exists(GuideProbeTargets.match(guider));
+        return guideTargets.exists(gpt -> gpt.getGuider() == guider);
```
cquiroz: Guider is an Enum right?

sraaphorst: Yes, that's right. The static `match` method, I believe, on `GuideProbeTargets` was doing exactly that, but seemed redundant with lambda expressions.

sraaphorst:
@tpolecat It is rather unfortunate to have so much mutability and side effects, but I couldn't figure out how else to handle this, unfortunately.

bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/TccNames.java
```diff
@@ -23,6 +23,7 @@ private TccNames() {
     public static final String AZEL = "AzEl";
     public static final String BASE = "Base";
 //    public static final String BASE = "sciencetarget";
+    public static final String BAGSTARGET = "bagsTarget";
```
cquiroz: Have we tested/considered any effects onto the TCC?

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/TelescopePosTableWidget.java
```diff
@@ -193,8 +183,44 @@ public String formatMagnitude(Magnitude.Band band) {
...
+                this.isActiveGuideProbe = isActiveGuideProbe;
+                this.quality = quality;
+            }
+
+            private static final Icon bagsEnabled  = bagsIcon(true);
+            private static final Icon bagsDisabled = bagsIcon(false);
+            private static final int  bagsGap      = 5;
+
+            private static Icon bagsIcon(boolean enabled) {
+                final BufferedImage img = new BufferedImage(16, 16, BufferedImage.TYPE_INT_ARGB);
```
cquiroz: Why do we render the icon instead of using a static image?
Should the icon's size be fixed?

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/TelescopePosTableWidget.java
```diff
@@ -262,41 +288,56 @@ public String formatMagnitude(Magnitude.Band band) {
             tmp.add(new BaseTargetRow(base, when));
 
             // Add all the guide groups and targets.
-            final Option<Tuple2<ObsContext, AgsMagnitude.MagnitudeTable>> ags = ctx.map(new MapOp<ObsContext, Tuple2<ObsContext, AgsMagnitude.MagnitudeTable>>() {
-                @Override public Tuple2<ObsContext, AgsMagnitude.MagnitudeTable> apply(ObsContext oc) {
-                    return new Pair<>(oc, OT.getMagnitudeTable());
-                }
-            });
-
+            final Option<Tuple2<ObsContext, AgsMagnitude.MagnitudeTable>> ags = ctx.map(oc -> new Pair<>(oc, OT.getMagnitudeTable()));
```
cquiroz: I've noted that you can do either `OT.getMagnitudeTable` or `ProbeLimitsTable.loadOrThrow()` to get the magnitude table, it isn't clear to me which one is best but the latter works for standalone applications.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/tpe/GemsGuideStarWorker.java
```diff
@@ -276,6 +330,18 @@ private GemsGuideStars findGuideStars(ObsContext obsContext, Set<edu.gemini.spMo
         }
     }
 
+    public static Option<GemsGuideStars> findGuideStars(final ObsContext obsContext) throws Exception {
+        try {
+            final Set<edu.gemini.spModel.core.Angle> posAngles = getPosAngles(obsContext);
+            final List<GemsCatalogSearchResults> results = unloggedSearch(GemsGuideStarSearchOptions.DEFAULT, GemsTipTiltMode.canopus, obsContext, posAngles,
+                    scala.Option.empty());
+            final List<GemsGuideStars> gemsResults = GemsResultsAnalyzer.instance().analyze(obsContext, posAngles, results, scala.Option.empty());
```
cquiroz: This seems fine, I'm still at loss why this is on the demo so quick, it normally takes a while being quadratic (I think) on the amount of results. Could we test on a more populated field?

bundle/jsky.app.ot/src/main/java/jsky/app/ot/tpe/TpeImageDisplayToolBar.java
```diff
@@ -96,7 +77,6 @@ protected JButton makeImageButton() {
     public void update() {
         super.update();
         makeManualGuideStarButton();
-        makeAutoGuideStarButton();
```
cquiroz: Do we have any other way to manually trigger AGS?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/tpe/AgsClient.scala
```diff
@@ -1,245 +0,0 @@
...
-    mainPanel.statusIcon.icon      = FailureIcon
-    mainPanel.statusMessage.text   = "No guide star was found."
-
-    import mainPanel.buttonPanel.acceptButton
-    acceptButton.text = "Manual Search"
-    acceptButton.enabled = true
-    listenTo(acceptButton)
-    reactions += {
-      case ButtonClicked(`acceptButton`) =>
-        TpeGuideStarDialog.showDialog(TpeManager.open().getImageWidget)
```
cquiroz: Can `TpeGuideStarDialog` be deleted?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -0,0 +1,321 @@
...
+          (oldBags.isEmpty && newBags.isEmpty) || oldBags.exists(oldTarget => newBags.exists(_.getTarget.equals(oldTarget.getTarget)))
+        }
+      }
+
+  private val propertyChangeListener = new PropertyChangeListener {
+    override def propertyChange(evt: PropertyChangeEvent): Unit = {
+      evt.getSource match {
+        case obsComp: ISPObsComponent =>
+          // Check to see if anything has changed.
+          val sameEnv = evt.getOldValue.isInstanceOf[TargetObsComp] && evt.getNewValue.isInstanceOf[TargetObsComp] && {
```
cquiroz: Can you put the `TargetObsComp` checks on the match/case to avoid the casts?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -0,0 +1,321 @@
...
+
+    def newContext(ctxOpt: Option[ObsContext]): BagsTask =
+      BagsTask(observation, curCtx, ctxOpt)
+
+    def performLookup(): Unit = {
+      LOG.info(s"Performing BAGS lookup for observation=${observation.getObservationID}")
+
+
+      def resultString[A](r: Option[A]): String =
+        s"Results=${r.fold("No")(_ => "Yes")}."
```
cquiroz: In scalaz you could do r ? "No" | "Yes"

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -0,0 +1,321 @@
...
+
+      // We must handle GeMS separately from other strategies, and delegate the work to the GemsGuideStarWorker.
+      def gemsLookup(obsCtx: ObsContext): Unit = {
+        LOG.info(s"BAGS GeMS lookup for observation=${observation.getObservationID} starting...")
+        val ggsTry = Try { GemsGuideStarWorker.findGuideStars(obsCtx) }
```
cquiroz: @tpolecat Are we rather using `\/` than `Try`? In any case you could do patter matching on the try to make it a bit cleaner
```
ggsTry match {
    case Failure(x) => LOG.warning("....")
    case Sucess(gss) => Swing.onEDT {...
}
```

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -0,0 +1,321 @@
...
+        def showTpeFeatures(selOpt: Option[AgsStrategy.Selection]): Unit =
+          selOpt.foreach { sel =>
+            Option(TpeManager.get()).filter(_.isVisible).foreach { tpe =>
+              sel.assignments.foreach { ass =>
+                val clazz = ass.guideProbe.getType match {
+                  case GuideProbe.Type.AOWFS => classOf[Altair_WFS_Feature]
+                  case GuideProbe.Type.OIWFS => classOf[OIWFS_Feature]
+                  case GuideProbe.Type.PWFS => classOf[TpePWFSFeature]
+                }
+                Option(tpe.getFeature(clazz)).foreach {
```
cquiroz: I wish we could get away from the reflection based properties of the TPE, probably it doesn't belong to this PR though

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -0,0 +1,321 @@
...
+import jsky.app.ot.gemini.tpe.TpePWFSFeature
+import jsky.app.ot.tpe.{GemsGuideStarWorker, GuideStarSupport, TpeContext, TpeManager}
+import scala.collection.JavaConverters._
+import scala.collection.immutable.HashMap
+import scala.swing.Swing
+import scala.util.{Try, Failure, Success}
+import scala.concurrent.ExecutionContext.Implicits.global
+
+
+object BagsManager {
```
cquiroz: Can we separate the UI related code from the pure BAGS code? I see in the same class code to handle async AGS calculation with UI updates. It maybe difficult on the OT to separate them though, but having multiple responsibilities makes it harder to follow and test

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -0,0 +1,321 @@
...
+
+      // Execute the task.
+      task.foreach(_.performLookup())
+
+      // Resubmit self for execution.
+      executor.submit(this)
+    }
+  }
+
+  private val NumThreads = 5
```
cquiroz: Should the number of threads be related to the CPU count?
`Runtime.getRuntime().availableProcessors()`

cquiroz:
First of all, congratulations on getting this through, it is a mean task and thanks for the general cleanup of the Java code.
I guess the meat is in `BagsManager`, as already mentioned is side-effecty but that maybe just the nature of things.
I'd like to separate a bit concerns, I see `BagsManager` doing task handling, calling AGS and doing UI. Perhaps those could be divided on different classes. In particular I'd like to see more tests on those classes and that gets pretty difficult with multiple tasks
Perhaps there is a way to do task processing with `Future` or `Task` instead of a blocking queue?  Does @tpolecat knows any?

sraaphorst:
BTW, thanks so much to both of you for looking this over and spending so much time on a Saturday doing it. I really appreciate it! We have guests tonight so I won't be able to make these changes today, but will address it all early tomorrow and make the recommended changes then.

bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/TccNames.java
sraaphorst: Hmmm... we haven't, and I'm not sure what the best way is to go about doing that.
Anyone know who I should talk to?

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/TelescopePosTableWidget.java
sraaphorst: This is sort of a placeholder for now, since we might (I suspect will) change this at some point in the future.
I was using the same size as the AGS feedback icons since they are displayed beside them in the table, which where the size came from, but this is probably something we can avoid worrying about until now until a better solution arises, since the icons are not in the optimal place anyway.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/TelescopePosTableWidget.java
sraaphorst: Interesting and good to know! Thanks for pointing this out! I was previously unaware.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/tpe/GemsGuideStarWorker.java
sraaphorst: I'll ask Andy and Bryan to test this thoroughly, but the preliminary results seem to be consistent with what we expect from what I can tell.

Thanks for checking this for me: it is appreciated!

bundle/jsky.app.ot/src/main/java/jsky/app/ot/tpe/TpeImageDisplayToolBar.java
sraaphorst: Not any more: they have all been removed.
I'm not sure if this could cause problems if the BAGS code is working properly, but it may be worthwhile to consider a (nonobvious - used only in case of serious issues) mechanism to force a lookup.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/tpe/AgsClient.scala
sraaphorst: It looks like it is still being used in the manual search?
`TpeImageWidget.manualGuideStarSearch`, for example, seems to use it.

bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/TccNames.java
arturog8m: @jluhrs needs to review this. I believe we don't need to expose the `bagsTargets` to the TCC.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/tpe/AgsClient.scala
sraaphorst: I am wondering why it has been deprecated, though. Any ideas?
Is there another way to do this?
I just removed the no longer used `TpeGuideStarDialogInternalFrame` (I suspect this class has been redundant for some time).

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
sraaphorst: I'm not sure how to do this or if it's possible? Since these are of type `AnyRef`, I think there may be complaints from the compiler about type patterns being unchecked / eliminated by erasure?

Or maybe I have no idea what I am talking about :D.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
sraaphorst: Nice. I still need to learn more scalaz, and any opportunity to pick up new tricks is appreciated!

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
sraaphorst: Much better with the match. Thanks for the recommendation.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
sraaphorst: Yes, this is better.

bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/TccNames.java
sraaphorst: You are absolutely right. Since just the names are being passed, there is absolutely zero reason to expose the BAGS target to the TCC. I will instead simply include it with the list of all targets by using `GuideProbeTargets.getTargets` instead of `GuideProbeTargets.getManualTargets` and processing the BAGS target separately.

In that case, might be good to still test with TCC, but the behaviour with regards to `TargetGroupConfig` should be unchanged, essentially.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
sraaphorst: Okay, this is a much better idea and this was bothering me.
I created a new class to contain the code for non-GeMS strategies to apply the selections to observations and to update the TPE, as it being done in `GemsGuideStarWorker`, and it is much clearer and cleaner now, and the GeMS and non-GeMS cases in `BagsManager` are now much more similar, as they should be.

Thanks. I am much happier with the state of affairs now, so I appreciate that.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/tpe/AgsClient.scala
cquiroz: We may remove the `GuideStarDialog` when the new catalog navigator is ready

PR #569, opened 2015-10-05
REL-2382-B: Update Simbad name resolvers

cquiroz:
This PR is the next step on updating support for simbad. It changes the target model moving radial velocity and parallax from `ProperMotion` to `SiderealTarget` and adds redshift
These three values can be read independently from a simbad query
I'd appreciate feedback especially from @fnussber and @tpolecat on the model changes and the use of squants


fnussber:
Implementing `RadialVelocity` as `case class RadialVelocity(velocity: Velocity)` has the disadvantage that we can not take direct advantage of the arithmetic operations `+`, `-`, `*` and `/` that are defined on squants quantities and take care of keeping the dimensions straight (e.g. velocity * times = length). The question is, if this matters for this particular qantity or not.

cquiroz:
Right, that was a question I wasn't sure, whether `RadialVelocity` should be a squants quantity. I doubt we'll use this class on arithmetics though a `Monoid` was defined.
I could have just made `radialVelocity: Velocity` on `SiderealTarget` I think

fnussber:
:+1:

PR #570, opened 2015-10-05
REL-2459: Updated OCS version and OT and splash screen icons.

sraaphorst:
Very simple update: just changed the splash screen, image, and icons for OT.


cquiroz:
Did we go back to some previous icon? That looks oddly familiar :smile: 
Anyway, :+1:

sraaphorst:
I think it is different, but it is reminding me of what we had 1-2 times ago, too :-).

andrewwstephens:
2008B was the same field but with a different telescope and different instrument.  :-)
http://internal.gemini.edu/science/softdevel/splash/splash.html

PR #572, opened 2015-10-06
OCSADV-379: Compress ITC output table.

fnussber:
Requirement: Allow multiple-word titles like "Exposure Time" and "Source Fraction" in ITC tables to be split into two rows, and those columns could be made narrower.

This is implemented by taking advantage of the little known fact that cell renderers in Swing tables support HTML. To simplify the definition of `Column`s in the code the HTML for the column header is created from a string that can contain newline characters `'\n'` to mark line breaks.

I also had to tweak the util method `SequenceTabUtil.resizeTableColumns`, which optimizes the column width of a table given the column headers and contents and needs to take multi line column headers into account properly.

[**Edit:** In an additional commit the multi line headers have been extended to the sequence table columns.]

image

image

### DIFF
--- a/bundle/jsky.app.ot/src/main/java/jsky/app/ot/editor/seq/DynamicSequenceCellRenderer.java
+++ b/bundle/jsky.app.ot/src/main/java/jsky/app/ot/editor/seq/DynamicSequenceCellRenderer.java
@@ -1,13 +1,10 @@
-//
-// $Id$
-//
-
 package jsky.app.ot.editor.seq;
 
 import edu.gemini.spModel.config2.ItemKey;
 import jsky.app.ot.util.OtColor;
 
 import java.awt.*;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.Set;
 
@@ -17,28 +14,24 @@ import java.util.Set;
 final class DynamicSequenceCellRenderer extends SequenceCellRenderer {
     private static final Color VERY_LIGHT_GREY = new Color(247, 243, 239);
 
-    private static final Set<ItemKey> FIXED_ITEMS = new HashSet<ItemKey>();
-
-    static {
-        for (ItemKey key : DynamicSequenceTableModel.SORT_ORDER) {
-            FIXED_ITEMS.add(key);
-        }
-    }
+    private static final Set<ItemKey> FIXED_ITEMS = new HashSet<ItemKey>() {{
+        Collections.addAll(this, DynamicSequenceTableModel.SORT_ORDER);
+    }};
 
     private DynamicSequenceTableModel _model;
 
-    DynamicSequenceCellRenderer(DynamicSequenceTableModel model) {
+    DynamicSequenceCellRenderer(final DynamicSequenceTableModel model) {
         _model  = model;
     }
 
-    protected Color lookupTextColor(int row, int column) {
+    protected Color lookupTextColor(final int row, final int column) {
         return _model.matchesNodeId(row) ?
                 super.lookupTextColor(row, column) :
                 Color.LIGHT_GRAY;
     }
 
-    protected Color lookupColor(int row, int column) {
-        ItemKey key = _model.getItemKeyAt(column);
+    protected Color lookupColor(final int row, final int column) {
+        final ItemKey key = _model.getItemKeyAt(column);
         Color color = VERY_LIGHT_GREY;
         if (!FIXED_ITEMS.contains(key)) color = lookupColor(key);
         if (_model.isError(row)) {
--- a/bundle/jsky.app.ot/src/main/java/jsky/app/ot/editor/seq/DynamicSequenceCellRenderer.java
+++ b/bundle/jsky.app.ot/src/main/java/jsky/app/ot/editor/seq/DynamicSequenceCellRenderer.java
@@ -18,7 +18,7 @@ final class DynamicSequenceCellRenderer extends SequenceCellRenderer {
         Collections.addAll(this, DynamicSequenceTableModel.SORT_ORDER);
     }};
 
-    private DynamicSequenceTableModel _model;
+    private final DynamicSequenceTableModel _model;
 
     DynamicSequenceCellRenderer(final DynamicSequenceTableModel model) {
         _model  = model;
--- a/bundle/jsky.app.ot/src/main/java/jsky/app/ot/editor/seq/DynamicSequenceTableModel.java
+++ b/bundle/jsky.app.ot/src/main/java/jsky/app/ot/editor/seq/DynamicSequenceTableModel.java
@@ -1,7 +1,3 @@
-//
-// $Id$
-//
-
 package jsky.app.ot.editor.seq;
 
 import edu.gemini.pot.sp.SPNodeKey;
@@ -25,7 +21,7 @@ import static jsky.app.ot.editor.seq.Keys.*;
  * Table model that holds the values in the configuration that change over its
  * course.
  */
-public class DynamicSequenceTableModel extends AbstractTableModel {
+public final class DynamicSequenceTableModel extends AbstractTableModel {
     public static final ItemKey[] SORT_ORDER = new ItemKey[] {
             DATALABEL_KEY,
             OBS_CLASS_KEY,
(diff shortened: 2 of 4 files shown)


andrewwstephens:
Looks good!  Could we also split "Observing Wavelength"?

fnussber:
@andrewwstephens I knew this was coming ;) Those columns are provided from the code that deals with the sequence table. If we want the column headers for those to have several lines, it might be easier to have them displayed the same way in the sequence table, too. @andrewwstephens tells me, that it would actually be desirable to make the sequence table more compact anyway, so I'll look into that and see if there is any craziness I am not aware of. If it's easy to do, I'll go ahead.

fnussber:
Following up on @andrewwstephens suggestion I have added the multiline column headers to the sequence table.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/editor/seq/DynamicSequenceTableModel.java
```diff
@@ -40,22 +36,22 @@
...
     private SPNodeKey _nodeKey;
 
-    public void setSequence(ConfigSequence sequence, SPNodeKey key, List<ItemKey> alwaysShow) {
+    public void setSequence(final ConfigSequence sequence, final SPNodeKey key, final List<ItemKey> alwaysShow) {
         _sequence     = sequence;
         _nodeKey      = key;
 
         final ItemKey[] newIteratedKeys = sort(getIteratedKeys(sequence, alwaysShow));
-        if (newIteratedKeys.equals(_iteratedKeys)) {
+        if (Arrays.equals(newIteratedKeys, _iteratedKeys)) {
```
fnussber: I don't think the `true` branch of this `if` statement has ever been executed. I *think* this is now working as it was meant to (updating the data only if the table structure hasn't changed), but I would appreciate if somebody could double check this one.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/editor/seq/DynamicSequenceTableModel.java
```diff
@@ -40,22 +36,22 @@
...
             fireTableDataChanged();
         } else {
-            _iteratedKeys = sort(getIteratedKeys(sequence, alwaysShow));
+            _iteratedKeys = newIteratedKeys;
```
fnussber: `sort(getIteratedKeys(sequence, alwaysShow))` seems to be side-effect free, so I think we don't need to evaluate this twice.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/editor/seq/DynamicSequenceTableModel.java
cquiroz: Right `equals` on Array does an instance equality check not a deep chcek on the array's contents

bundle/jsky.app.ot/src/main/java/jsky/app/ot/editor/seq/DynamicSequenceTableModel.java
```diff
@@ -166,7 +162,7 @@ public void setSequence(ConfigSequence sequence, SPNodeKey key, List<ItemKey> al
             res.addAll(groupedItems.get(category));
         }
 
-        return res.toArray(new ItemKey[0]);
```
cquiroz: How did that work with an array of size 0?

bundle/jsky.app.ot/src/main/java/jsky/app/ot/editor/seq/SequenceTabUtil.java
```diff
@@ -122,6 +123,22 @@ static void resizeTableColumns(JTable table, TableModel model) {
...
 
+    /**
+     * Gets the (longest) column header line.
+     * In case of html headers which may contain several lines separated by <br> tags we need to ignore
+     * all html tags and return the longest of the given lines.
+     */
+    private static String getLongestHeaderLine(final TableModel model, final int col) {
+        final String title = model.getColumnName(col);
+        if (title.startsWith("<html>")) {
+            return Arrays.stream(title.split("<br/?>")).                    // split around html breaks
```
cquiroz: nice

cquiroz:
Looks good, the html support in swing is quite limited and non-documented but if it gets the job done :+1:

arturog8m:
Nice. We should test this on Linux & Windows to ensure the behavior of Swing + HTML is as expected.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/editor/seq/DynamicSequenceCellRenderer.java
```diff
@@ -17,28 +14,24 @@
...
-    static {
-        for (ItemKey key : DynamicSequenceTableModel.SORT_ORDER) {
-            FIXED_ITEMS.add(key);
-        }
-    }
+    private static final Set<ItemKey> FIXED_ITEMS = new HashSet<ItemKey>() {{
+        Collections.addAll(this, DynamicSequenceTableModel.SORT_ORDER);
+    }};
 
     private DynamicSequenceTableModel _model;
```
arturog8m: can we make this `final`?

bundle/jsky.app.ot/src/main/java/jsky/app/ot/editor/seq/DynamicSequenceTableModel.java
```diff
@@ -179,18 +175,18 @@ public int getColumnCount() {
...
-    public Object getValueAt(int rowIndex, int columnIndex) {
+    public Object getValueAt(final int rowIndex, final int columnIndex) {
         if (_sequence == null) return null;
         if (_sequence.size() == 0) return null;
 
-        Config config = _sequence.getStep(rowIndex);
-        ItemKey   key = _iteratedKeys[columnIndex];
+        final Config config = _sequence.getStep(rowIndex);
+        final ItemKey   key = _iteratedKeys[columnIndex];
         if (!isGcal(config) && "calibration".equals(key.getParent().toString())) return "";
```
arturog8m: Not part of your changes, but looks like `key.getParent().toString()` can yield a NPE.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/editor/seq/DynamicSequenceCellRenderer.java
```diff
@@ -17,28 +14,24 @@
...
         return _model.matchesNodeId(row) ?
                 super.lookupTextColor(row, column) :
                 Color.LIGHT_GRAY;
     }
 
-    protected Color lookupColor(int row, int column) {
-        ItemKey key = _model.getItemKeyAt(column);
+    protected Color lookupColor(final int row, final int column) {
+        final ItemKey key = _model.getItemKeyAt(column);
         Color color = VERY_LIGHT_GREY;
```
sraaphorst: This could also be made final with a ternary expression or by breaking the `if` statement into an `if`...`else` and not assigning to `color` initially.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/editor/seq/DynamicSequenceTableModel.java
```diff
@@ -40,22 +36,22 @@
...
 
-    private Set<ItemKey> getIteratedKeys(ConfigSequence seq, List<ItemKey> alwaysShow) {
-        Set<ItemKey> res = new HashSet<ItemKey>();
-        ItemKey[]   keys = seq.getIteratedKeys();
+    private Set<ItemKey> getIteratedKeys(final ConfigSequence seq, final List<ItemKey> alwaysShow) {
+        final Set<ItemKey> res = new HashSet<>();
+        final ItemKey[]   keys = seq.getIteratedKeys();
         for (ItemKey key : keys) {
             // Strip out all the smart gcal metadata.
             if (MetaDataConfig.MATCHER.matches(key) || SmartgcalSysConfig.MATCHER.matches(key)) continue;
```
sraaphorst: Seems like this could be done with a `filter` and `foreach`. Not sure if that is more readable, though.
`Arrays.stream(seq.getIteratedKeys())
.filter(key -> !(MetaDataConfig.MATCHER.matches(key) && SmartgcalSysConfig.MATCHER.matches(key)) && templ.matches(key))
.foreach(res::add);`

bundle/jsky.app.ot/src/main/java/jsky/app/ot/editor/seq/DynamicSequenceTableModel.java
```diff
@@ -143,7 +139,7 @@ public void setSequence(ConfigSequence sequence, SPNodeKey key, List<ItemKey> al
             // Add it to the list for that parent.
             List<ItemKey> lst = groupedItems.get(parent.getName());
```
sraaphorst: I know this is nitpicky, but `final` could be used here with a second list, easily.
`final List<ItemKey> lst = groupedItems.get(parent.getName());`
`final List<ItemKey> finalLst;`
`if (lst == null) { finalList = new ArrayList<>(); groupedItems.put(parent.getName(), finalList); }`
`else finalList = lst;`

bundle/jsky.app.ot/src/main/java/jsky/app/ot/editor/seq/DynamicSequenceTableModel.java
fnussber: The warning IntelliJ gave me said that in this case reflection was used to create an array of the appropriate type and size which should be avoided for efficiency reasons.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/editor/seq/DynamicSequenceTableModel.java
fnussber: Yeah, I saw that warning. Didn't really know what to do about it, because it doesn't seem to be an actual problem. So I silently ignored it.. ;) I can give it another look.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/editor/seq/DynamicSequenceCellRenderer.java
fnussber: Sure, it's a bit of a question how much time we want to invest in these things, I usually try to focus on the quick wins and things I think are causing problems. But I can give it a second pass.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/editor/seq/DynamicSequenceCellRenderer.java
fnussber: yes, will do

fnussber:
I'll check with Andy if he can test it on Linux/Windows.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/editor/seq/SequenceTabUtil.java
fnussber: @cquiroz They also call it: "functional programming for the paupers" :smile:

PR #573, opened 2015-10-06
OCSADV-417: Display Guide Quality icon on the catalog navigator

cquiroz:
On this PR a column is added to the Catalog navigator displaying ax guiding quality indicator as below (By default targets are sorted by quality):

screenshot 2015-10-06 15 12 19

This is a rather large code update, The Catalog Navigator was getting too big and had to be separated on a few files, additionally changes were made to support custom sorting and displaying of the drawn guiding quality icons.

Finally, for guiding quality calculations, we need to have a way to create an `ObsContext` from the information on the query form. I'm not completely happy with the way it is done in this PR but I have marked those to be reviewed later with @swalker2m 

For now I want to get these changes in to move forward with TPE integration

Several old java classes were updated with modern code while doing this PR

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/edu/gemini/catalog/ui/TableColumnsAdjuster.scala
+++ b/bundle/jsky.app.ot/src/main/scala/edu/gemini/catalog/ui/TableColumnsAdjuster.scala
@@ -18,10 +18,8 @@ trait TableColumnsAdjuster { this: Table =>
     import scala.math.max
 
     def updateTableColumn(column: TableColumn, width: Int):Unit = {
-      if (column.getResizable) {
-        this.peer.getTableHeader.setResizingColumn(column)
-        column <| {_.setPreferredWidth(width)} <| {_.setWidth(width)}
-      }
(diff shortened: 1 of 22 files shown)


andrewwstephens:
Beautiful!

bundle/jsky.app.ot/src/main/scala/edu/gemini/catalog/ui/TableColumnsAdjuster.scala
```diff
@@ -0,0 +1,84 @@
...
+import javax.swing.table.TableColumn
+import scala.swing.Table
+
+import scalaz._
+import Scalaz._
+
+/**
+ * Support calculating column widths and can adjust them to the outer width
+ */
+trait TableColumnsAdjuster { this: Table =>
```
fnussber: This is very similar to `resizeTableColumns` in `SequenceTabUtil`. Maybe we can bring those together at some point (not saying as part of this PR ;).

fnussber:
This is a big one, phew. Didn't see anything that looked suspect, so I guess :+1:

PR #576, opened 2015-10-07
OCSADV-1852B: Display correct read noise for F2; use unicode characters.

fnussber:
A quick correction of a read noise value for F2 on the HTML page.

image

Under cover of this change I also decided to introduce unicode characters where appropriate to display units.

image


cquiroz:
:+1:

PR #577, opened 2015-10-07
OCSADV-421: Convert non-gems ManualGS to use VoTable

cquiroz:
Small PR that integrates the New Catalog Navigator instead of the old one. The old one is still accessible from the TPE
The PR also includes a couple of small bug fixes
With this PR in we should be able to make a release to let science staff check the new navigator


fnussber:
:+1:

PR #578, opened 2015-10-08
OCSADV-386: Spectroscopy ITC Peak counts = signal + background.

fnussber:
Small bug fix that calculates the peak value for spectroscopy as the maximum value of the sum of the signal and background functions over wavelength.

**GNIRS**
image

**GNIRS Cross Dispersed**
image

*As a side note:*
If you wonder about the structuring of the data: ITC calculation results contain several series of resulting data, e.g. `SignalData` and `BackgroundData` (available types depend on instrument and calculation modes). For each of these types there can be several arrays, e.g. for GNIRS in cross dispersion mode there is one of each data arrays (signal, background) for each order (6).

The modeling of those result values is a bit cryptic, but I have to work with what the underlying ITC code provides. However by now I also have a better feeling for how this data looks like for different instruments and potentially I'll restructure those result objects a bit at some point. For now it is what it is I am afraid..


cquiroz:
I cannot claim to understand the calculations but codewise looks good :+1:

fnussber:
Hmm, I was hoping my attempt at explaining what's going on would be successful. But it seems I failed miserably.

PR #580, opened 2015-10-08
REL-2167: Added P2 check for GNIRS as requested.

sraaphorst:
New GNIRS P2 warning for when AcqCam is OUT and Decker is set to AcqCam.

Also some general small cleanup of existing code.


fnussber:
:+1:

PR #581, opened 2015-10-08
REL-2382-C: Update Simbad name resolvers

cquiroz:
Here is the next step on Simbad search support. The PR includes integration into the UI, replacing the current name search on the target editor. With the new search many more fields are filled including coordinates, proper motion, radial velocity, parallax and redshift. Additionally it will set the magnitudes from the search result

The screenshot below shows how it looks like when the search values are retrieved:
screenshot 2015-10-08 13 32 06

A new `Redshift` property was added to `HmsDegTarget`. Note that I'm diverging from the usual and I put the scala `Redshift` class directly as a member of `HMSDegTarget`

While searching an overlay is put in front of the main window
screenshot 2015-10-08 14 56 54

Errors are displayed as before
screenshot 2015-10-08 14 14 52

As usual the PR contains a bunch of improvements to legacy code

Comments on the UI are totally welcome


fnussber:
I like the `GlassLabel`, looks all good, I'll soon take it up from here and integrate RV / redshift into the ITC calculations. :+1:

cquiroz:
In answer to @fnussber question, once you do query the values already on the details, e.g. the magnitudes will be removed and replaced by the response from the catalog

cquiroz:
Due to @andrewwstephens request, radial velocity was removed and we only story redshift in the model, the updates take care of the fixes to do this including sending the correct values to the tcc, serialization and model conversions. We are also capable of transforming redshift to its alternative representations
The UI also changes and the user can choose how to display redshift in either Redshift, RadialVelocity or ApparentRadialVelocity format. See the screenshot below:

screenshot 2015-10-16 14 36 27

andrewwstephens:
Excellent!
Will the units update (km/s for RV and cz, and none for z)?

cquiroz:
Yes, km/s for cz and RV and none for z

cquiroz:
Note that I also left-aligned the labels of each item and units

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/Target.scala
```diff
@@ -126,13 +125,14 @@ object Target {
   }
 
   object SiderealTarget {
+
+    def apply(name: String, coordinates: Coordinates, magnitudes: List[Magnitude]): SiderealTarget = SiderealTarget(name, coordinates, None, None, None, magnitudes)
```
tpolecat: I'm not sure about this secondary factory method. I know you need it for some of your tests but you could define it locally (I have done this on my branch actually ... which will have a bunch of conflicts with this one).

tpolecat:
... and you're right, this is going to conflict a bunch with #593 and the next two things I have staged. So I guess I should vote :-1: so mine gets merged first, right? :smile_cat:

tpolecat:
This is a lot of stuff but it looks good to me. :+1:

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/Target.scala
cquiroz: Yes, I did it to simplify the test, I guess I could move it to the tests as you mention

tpolecat: Yeah I just find that default arguments (which is basically what this does) tend to lead to bugs. You might implement it in terms of `SiderealTarget.empty.copy(...)` and then you'll be insulated from the other changes I'm making.

fnussber:
Nice, I like the extensions you made to the `NumericPropertySheet`, including the left alignment of the various elements.

cquiroz:
Removed the constructor, updated the tests and ready to merge

PR #583, opened 2015-10-08
OCSADV-336A: Unify magnitude systems

fnussber:
This is some prep work for the introduction of additional (integrated) magnitude units and additional surface brightness magnitude units and values for ITC.

For those changes it seemed to be easier to first unify the `MagnitudeSystem` from the new model and the `Magnitude.System` from the old model in order to not have to deal with implementing conversions between the two.

Tests are working and I'll do some additional testing with XML import/export etc. I think I will also wait for @cquiroz to merge https://github.com/gemini-hlsw/ocs/pull/581 and test again once those changes are incorporated. However I would like to get some feedback on this asap.

As far as I understand the code I think these changes should be transparent to everything, but I might overlook something important. 

@jluhrs : Are magnitudes used by the TCC for anything other than being displayed? This PR is not going to change anything, but would additional units for the magnitudes have an impact on something outside the ODB/OT universe?


bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/MagnitudeSystem.scala
```diff
@@ -1,14 +1,25 @@
...
 
-  case object VEGA extends MagnitudeSystem("Vega")
-  case object AB   extends MagnitudeSystem("AB")
-  case object JY   extends MagnitudeSystem("Jy")
+  case object VEGA   extends MagnitudeSystem("Vega")
+  case object AB     extends MagnitudeSystem("AB")
+  case object JY     extends MagnitudeSystem("Jy")
+//  case object WATTS  extends MagnitudeSystem("W/m²/µm")
+//  case object ERGSA  extends MagnitudeSystem("erg/s/cm²/Å")
+//  case object ERGSHZ extends MagnitudeSystem("erg/s/cm²/Hz")
```
fnussber: This are the new units I am planning to add.

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/MagnitudeSystem.scala
```diff
@@ -18,4 +29,11 @@ object MagnitudeSystem {
   implicit val MagnitudeSystemEqual: Equal[MagnitudeSystem] =
     Equal.equalA
 
+
+  // ===== LEGACY JAVA SUPPORT =====
+  val allAsJava = new java.util.Vector[MagnitudeSystem](all)
+  val DEFAULT = default
+
+  def valueOf(s: String) = all.find(_.name == s).get
```
cquiroz: Is it ok to throw an exception if the value isn't found?

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/MagnitudeSystem.scala
```diff
@@ -1,14 +1,25 @@
...
+import scala.collection.JavaConversions._
 import scalaz.Equal
 
-sealed abstract class MagnitudeSystem(val name: String) extends Product with Serializable
+sealed abstract class MagnitudeSystem(val name: String) extends Product with Comparable[MagnitudeSystem] with Serializable {
+
+  // ===== LEGACY JAVA SUPPORT =====
+  import MagnitudeSystem._
+
+  def compareTo(other: MagnitudeSystem) = all.indexOf(other) - all.indexOf(this)
```
cquiroz: this is kind of strange, is there a natural way to order `MagnitudeSystems`? Is it done anywhere?

cquiroz:
Certainly it seems a good idea to eliminate duplicate classes as much as possible. Do you think there could be any serialization problem with legacy programs?

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/MagnitudeSystem.scala
fnussber: This takes the place of the `valueOf()` method on the Java enum which also throws an exception if the value can not be found. This is only used in the `PIOXml` code. Probably I should move this there so this is clear.

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/MagnitudeSystem.scala
fnussber: Again, this mimics the Java enum behavior by replacing the `ordinal()` method which is used for `compareTo` on enums, which by default is based  on the order the values are defined. There isn't really a natural order. The only place this is relevant is in the `compareTo()` method of the legacy `Magnitude` class. My main objective was to not change the sort order because I don't really know where it is relevant, but I am open for suggestions how to do this in a better way. Maybe I should move that code directly into the `compareTo()` method though, so that it is clear this is not really meant to be used for anything other than that.

fnussber:
I'll do a few more tests, but the strings used for XML serialization are the same, so this shouldn't cause any problems.

fnussber:
@cquiroz : I moved the code that basically mimics old Java enum behavior to the places where it is used so it doesn't clutter up the `MagnitudeSystem` class in the core model.

jluhrs:
@fnussber TCC sends the magnitude of bands H and I to GPI. It passes the value received from WDBA as it is. So we either change TCC to have it understand units, or make sure WDBA sends the raw value in the same units as before.

fnussber:
Closing this one after some substantial changes have been made, will reopen as a new PR to reset discussion.

PR #584, opened 2015-10-09
REL-2407: add argonaut.io and instructions for wrapping bundle jars in general

swalker2m:
This tiny PR has two goals:

1) Add (admittedly scant) documentation and tools for wrapping 3rd party jar files for use with OSGi
2) Introduce `argonaut` as a Data Manager dependency

The first goal is accomplished by adding `lib/tools/biz.aQute.bnd-3.1.0.jar` so that it can be executed from the command line and a `lib/bundle/README.md` file with brief instructions on how to use it.  Does this seem okay?

[Argonaut](http://argonaut.io) is a functional library for working with JSON.  I wanted to add a PR for this now in case anybody has any doubts or comments before I spend too much time with it.  The Data Manager is being reworked to use the new in-house archive which provides results and commands in JSON documents.  Our needs are fairly trivial at the moment but the built-in JSON support in Scala is lacking.  Argonaut seems to be the choice that best matches our emerging codebase and philosophy, and I understand it is being updated to use [cats](https://non.github.io/cats/) so that it will be ready when we eventually make that switch.

The bulk of the PR is `FileList.scala` which is just throwaway code that exercises argonaut in a trivial way to make sure it is hooked-up correctly.  It will be removed in subsequent PRs.


tpolecat:
:+1:

cquiroz:
:scream: I only read `cats` on the description

swalker2m:
Sorry about that.  I didn't mean to frighten anybody.  Cats isn't ready yet so it's a non-issue for now.  We can cross that bridge, or not, later.  It's worth checking anything that uses scalaz though in order to keep our options open.

lib/bundle/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+though some of the options don't seem to actually work.
+
+With a bit of luck though, it may be quicker than manually creating
+the missing MANIFEST information.  The procedure that worked for me
+is:
+
+````
+cd <path-to-ocs-dist>/lib/bundle
+java -jar ../tools/biz.aQute.bnd-3.1.0.jar wrap <path-to-raw-jar-file>
```
cquiroz: This is very good! I don't remember how many times I have searched for this same command

cquiroz:
I did a quick check on argonaut's [build] (https://github.com/argonaut-io/argonaut/blob/master/project/build.scala) and there are several dependencies besides scalaz, like monocle and lift-json, do you need to import those too?

swalker2m:
Good question.  I'm actually using the released stable version [`6.0.4`](https://github.com/argonaut-io/argonaut/blob/series/6.0.x/project/build.scala) which doesn't have those dependencies.   We need that version to match our version of scalaz.  At any rate, I was able to get the tiny sample app to work without adding anything else.

fnussber:
This looks nice, JSON support will be very useful in the future.

PR #585, opened 2015-10-13
OCSADV-422: Fix to results output for BAGS logging.

sraaphorst:
Previously, the `Yes` and `No` labels were accidentally reversed when I changed the code to use scalaz.


swalker2m:
:+1:

PR #586, opened 2015-10-13
OCSADV-431: Removed BAGS label from type tag.

sraaphorst:
Since we have an icon indicating BAGS guide stars, having the type tag include `(BAGS)` is redundant, as pointed out by Andy. Thus, we are simply removing it.


arturog8m:
After a thorough revision and much debate, I think this is okay :+1:

PR #587, opened 2015-10-13
OCSADV-430: Fixed issue where observations in groups were being ignored.

sraaphorst:
This was just a stupid lack of knowledge on my part: instead of iterating over ALL observations of a program, I was only iterating over the observations not contained in groups. By changing the method call from `getObservations` to `getAllObservations`, now observations in groups are handled.

Additionally, I added a case clause to handle changes to `ISPGroup`, since otherwise this would fall through to `ISPNode` and the `getContextObservation` method would not handle all observations in the group as should be done.


cquiroz:
:+1:

PR #588, opened 2015-10-13
OCSADV-425: Fixed issue with PA.

sraaphorst:
The position angle was not always being set properly, as I was only handling it in an `if` statement when auto guide stars changed. It is now examined and handled independently of whether or not auto guide stars change, and fixes the issue with the parallactic angle where the position angle was not being properly flipped by 180.


tpolecat:
:+1:

PR #589, opened 2015-10-13
OCSADV-433: BAGS thread reprioritizing and bugfix

sraaphorst:
Two things in this PR:
1. BAGS threads are now prioritized to be below normal thread priority to make sure they run at a lower priority than other OT functionality.
2. A bug was fixed in the calculation of the number of threads to use (ooops).

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
@@ -93,7 +93,9 @@ object BagsManager {
     }
   }
 
-  private class BagsThread(id: Int) extends Runnable {
+  private class BagsThread(id: Int) extends Thread {
+    setPriority(Thread.NORM_PRIORITY - 1)
+
     override def run() {
       // Get the next task to run. This blocks on the queue.
       val task = nextTask()
@@ -106,7 +108,7 @@ object BagsManager {
     }
   }
 
-  private val NumThreads = math.min(1, Runtime.getRuntime.availableProcessors())
+  private val NumThreads = math.max(1, Runtime.getRuntime.availableProcessors()-1)
   private val RequeueDelay = 3000
   private val executor = Executors.newFixedThreadPool(NumThreads)
   private var taskMap = new HashMap[SPNodeKey, BagsTask]


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -93,7 +93,9 @@ object BagsManager {
     }
   }
 
-  private class BagsThread(id: Int) extends Runnable {
+  private class BagsThread(id: Int) extends Thread {
+    setPriority(Thread.NORM_PRIORITY - 1)
```
cquiroz: I'm not sure if there is a noticeable effect setting a thread's priority, it is also system dependent.
Why do you want to have lower priority? Are you seeing performance troubles?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -106,7 +108,7 @@ object BagsManager {
     }
   }
 
-  private val NumThreads = math.min(1, Runtime.getRuntime.availableProcessors())
+  private val NumThreads = math.max(1, Runtime.getRuntime.availableProcessors()-1)
```
tpolecat: Since these threads are blocking on IO you could probably have a bunch of them.

cquiroz: There is IO contention on the catalog queries but GEMS is CPU bounded too

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
sraaphorst: @swalker2m  had suggested as much, just to place the BAGS threads at a lower priority than the rest of the OT since they should be background tasks. I'm not sure if it will make a difference, and I haven't seen any performance issues, but I suspect there could potentially be performance issues on an older machine with a single CPU?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
sraaphorst: Sorry: are you suggesting having more than the number of available processors?
At first, we had gone with 5 fixed, but I wasn't sure how to best proceed.
Even with a single thread, things proved to be generally "speedy enough", it seemed to me, when importing and working on an observation, so I'm not sure the best way to handle this.

tpolecat: Ok if there is significant CPU load then what you have here seems fine.

swalker2m:
I think this looks fine. :+1:

PR #590, opened 2015-10-13
OCSADV-435: BAGS improved event handling

sraaphorst:
Improvements to event handling to BAGS to cut down on having to handle specific types of nodes and to reduce the number of unnecessary lookups considerably.

Also fixed an issue where when a program was closed, failed lookups would continue to be performed. This is done by maintaining a set of active program node keys.

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
@@ -93,9 +93,7 @@ object BagsManager {
     }
   }
 
-  private class BagsThread(id: Int) extends Thread {
-    setPriority(Thread.NORM_PRIORITY - 1)
-
+  private class BagsThread(id: Int) extends Runnable {
     override def run() {
       // Get the next task to run. This blocks on the queue.
       val task = nextTask()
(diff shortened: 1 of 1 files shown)


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -108,18 +106,25 @@ object BagsManager {
     }
   }
 
-  private val NumThreads = math.max(1, Runtime.getRuntime.availableProcessors()-1)
+  private val NumThreads = math.min(1, Runtime.getRuntime.availableProcessors())
```
swalker2m: There might be a problem with this PR?  It seems to be undoing work that you did previously?

sraaphorst: Huh. That's odd. I'll go ahead and fix that.
Thanks for pointing it out.

sraaphorst:
Okay, there we go. Sanity (largely) restored!

cquiroz:
This seems fine though I have to worry a bit about the synchronization calls, do you think you should sync `taskMap` at the same time as `taskQueue`, the code seems to point that they have to be updated atomically

tpolecat:
Agree, I would like to see a comment somewhere explaining the synchronization strategy.

sraaphorst:
I am going to merge this one, and go into more detail on the synchronization strategy in a separate issue, where I will address the concerns listed by @cquiroz and @tpolecat, since they aren't specific to this PR.

PR #591, opened 2015-10-14
OCSADV-396-A: Draw selected targets on the TPE

cquiroz:
This PR is the first step to integrate the TPE with the Catalog Navigator. It supports the following actions:

* When opening the navigator the TPE is opened too
* If a query is done on the navigator, the TPE is updated to be centered on the searched coordinates

I'm not totally happy with this code but given time constraints I'm pressing ahead. The main difficulty is that TPE is designed around the current tree of nodes (`ISPNode` and children) via the `TpeContext` class. However, that tree does not exist on the context of the Catalog Navigator so we need to produce an alternative of `TpeContext` that extracts target information from an `ObsContext`. As `ObsContext` is quite unrelated to the nodes trees, adapters are built to let the TPE work against it in a read-only mode.

In this PR most of the code deals with updated legacy code, but the relevant changes are concentrated in [`TpeContext`](ocs/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/tpe/TpeContext.scala)

### DIFF
--- a/bundle/jsky.app.ot/src/main/java/jsky/app/jskycat/JSkyCat.java
+++ b/bundle/jsky.app.ot/src/main/java/jsky/app/jskycat/JSkyCat.java
@@ -7,11 +7,8 @@ import jsky.navigator.NavigatorImageDisplayInternalFrame;
 import jsky.navigator.NavigatorInternalFrame;
 import jsky.util.I18N;
 import jsky.util.Preferences;
-import jsky.util.Resources;
 import jsky.util.gui.*;
 
-import javax.media.jai.JAI;
-import javax.media.jai.TileCache;
 import javax.swing.*;
 import java.awt.*;
 import java.awt.event.WindowAdapter;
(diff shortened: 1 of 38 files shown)


bundle/jsky.app.ot/src/main/java/jsky/app/jskycat/JSkyCat.java
```diff
@@ -405,28 +395,35 @@ public static void main(String args[]) {
         boolean ok = true;
         int tilecache = 64;
 
+        label:
```
fnussber: Is the enclosing `main()` method still used anywhere? OT has it's own starter method, no?

tpolecat: Almost certainly unused.

cquiroz: I assume it was used for testing JskyCat back in the day, it could be well removed

fnussber:
I think the adapter objects are a good way to contain the issue for now. :+1:

bundle/jsky.app.ot/src/main/scala/edu/gemini/catalog/ui/package.scala
```diff
@@ -83,17 +85,20 @@ case class ObservationInfo(ctx: Option[ObsContext], objectName: Option[String],
...
-      ObsContext.create(env, i, new JSome(s), conditions.get, null, null, JNone.instance())
+      ObsContext.create(env, i, new JSome(s), cond, null, null, JNone.instance())
     }
   }
 }
 
 object ObservationInfo {
+  // Observation context loaded initially with default parameters
+  // TODO don't hardcode to Gmos South
+  val zero = new ObservationInfo(None, "".some, Coordinates.zero.some, SPComponentType.INSTRUMENT_GMOSSOUTH.some, None, Nil, SPSiteQuality.Conditions.BEST.some, UCAC4, ProbeLimitsTable.loadOrThrow())
```
swalker2m: On the surface of it, I would expect a `zero` to have no instrument.

swalker2m:
So I know from experience that it is difficult to make any change and not break the TPE. I very much sympathize with you.  All the cleanup is fantastic.

There's always a selected tree node, so I don't quite understand why the TPE needs to be updated to possibly work without one. I worry about it showing an observation without the underlying model elements because it seems like things you'd expect to work may no longer work. For example, dragging the position angle and seeing it update in the program.  It's also a bit awkward to have two hierarchies of TPE context nodes as you point out.

I guess I could use more information on why we need to do this and see if there are other alternatives first.  What does the catalog navigator need to do with the TPE?  Is it just about plotting the candidates and showing the selected row in the table as well as in the TPE?

bundle/jsky.app.ot/src/main/scala/edu/gemini/catalog/ui/package.scala
cquiroz: That will be addressed on OCSADV-436

cquiroz:
The navigator needs to to plot the candidates but given that you can search in any location, you'd need to center the TPE on that location. I couldn't find any other way to do that without providing a `TpeContext` via the `reset` method.
You can test that on the old CN plotting works, but it will display something only if you have centered the TPE before on the given location. Let's say you start empty so you are at 0,0. Then you search for vega, get your results and plot. Nothing will show because the TPE is still at 0,0
The way to center the TPE is via a `reset` call. I'd rather have the new CN not depend on the nodes only for this
Given that the alternative `TpeContext` doesn't have all the features of the real one certain stuff won't work like dragging but that makes sense since there is no context attached
I share your concern though, we could end up with some difficulties linking back to the program

cquiroz:
I reverted the changes to `TpeContext` but I'm keeping the updates to legacy classes, I assume those were reviewed so I can merge them

PR #592, opened 2015-10-14
OCSADV-426: Added checks for executed observations.

sraaphorst:
Previously, BAGS was selecting guide stars for observations marked as "executed".
We now leave these observations alone.

@swalker2m, as per your suggestion due to the computation-heavy nature of calls to `ObservationStatus.computeFor`, I tried to push this call as far as possible into the BAGS process, but I am not sure if this is the best place to put it (i.e. after AGS is called but before results are applied), and the calls to AGS are redundant in this case.

I would greatly appreciate any feedback or suggestions as to if you think there is a preferable place to put this, as the calls to AGS are unnecessary with the results thrown away as things currently stand.

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
@@ -12,7 +12,6 @@ import edu.gemini.spModel.obs.ObservationStatus
 import edu.gemini.spModel.obs.context.ObsContext
 import edu.gemini.spModel.rich.shared.immutable._
 import edu.gemini.spModel.target.env.TargetEnvironment
-import edu.gemini.spModel.target.obsComp.TargetObsComp
 import edu.gemini.shared.util.immutable.{Option => GOption}
 import jsky.app.ot.OT
 import jsky.app.ot.tpe._
@@ -60,12 +59,12 @@ object BagsManager {
           // This is reported only as a GenericError in a CatalogException, unfortunately.
           case Failure(CatalogException((e: GenericError) :: _)) =>
             LOG.warning(s"BAGS lookup for observation=${observation.getObservationID} failed: ${e.msg}")
-            taskComplete (this, success = false)
+            taskComplete(this, success = false)
 
           // For all other exceptions, print the full stack trace.
           case Failure(ex) =>
             LOG.log(Level.WARNING, s"BAGS lookup for observation=${observation.getObservationID} failed.", ex)
-            taskComplete (this, success = false)
+            taskComplete(this, success = false)
         }
       }
 
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
@@ -2,10 +2,11 @@ package jsky.app.ot.ags
 
 import java.beans.{PropertyChangeEvent, PropertyChangeListener}
 import java.util.concurrent.{LinkedBlockingQueue, Executors}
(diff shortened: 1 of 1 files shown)


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -50,10 +51,12 @@ object BagsManager {
         Try { GemsGuideStarWorker.findGuideStars(obsCtx) } match {
           case Success(ggs) =>
             LOG.info(s"BAGS GeMS lookup for observation=${observation.getObservationID} successful. ${resultString(ggs.asScalaOpt)}")
-            Swing.onEDT {
-              muteObservation(observation)
-              GemsGuideStarWorker.applyResults(TpeContext(observation), ggs, true)
-              unmuteObservation(observation)
+            if (ObservationStatus.computeFor(observation) != ObservationStatus.OBSERVED) {
```
cquiroz: Both code fragments look very similar, could they be refactored?

sraaphorst:
I made the refactor as suggested by @cquiroz (which I think is a very good idea).
I'm not entirely happy with the way that the code looks, as there is ugliness from having to mix the Gemini `Option` (which is used by `GemsGuideStarWorker`) with the Scala `Option` (which is used by `GuideStarWorker`). If anyone has any suggestions to get around this without mucking things up further, I would very much be happy to hear! @swalker2m or @tpolecat ?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -39,54 +42,33 @@ object BagsManager {
...
+              Swing.onEDT {
+                muteObservation(observation)
+                worker(TpeContext(observation), selOpt)
+                unmuteObservation(observation)
+              }
             }
             taskComplete(this, success = true)
+
           case Failure(ex) =>
             LOG.warning(s"BAGS lookup for observation=${observation.getObservationID} failed. Exception=$ex")
```
tpolecat: Not part of this PR, but would you mind including the stacktrace via `Log.log(Level.WARNING, ..., ex)`?

tpolecat:
Yeah the `Option` thing is a pain. Seems fine.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
sraaphorst: I don't mind at all. My only concern is that this log message is output if BAGS is attempting to run and the network is out, which could get very verbose.

Perhaps I should match on the exception and output a separate message if that is the case. The only problem I see with this is that a `GenericError` is what is reported if the catalog host is unreachable, and from what I can tell, this may be issued in other circumstances.

For now, for a `GenericError`, I will simply output the message, and for any other error type, I will output the stack trace.

cquiroz: Now that you mention the catalog calls, there could be a concern on affecting the scalability of the catalog servers if we have all the OTs doing queries in the background. I don't know if that is a real concern though
What if the user drags a target, would it do a query at the end or on multiple steps?

sraaphorst: It will only do a query at the end of the drag, as I believe that the `TargetEnvironment` is only changed when the drag is complete.

sraaphorst:
Okay, all concerns have been addressed, so I am going to merge this one.
Added the stack trace for exceptions in logging (trying to avoid network outages) as per the request by @tpolecat

PR #593, opened 2015-10-14
Typeclass-based PIO and impl for new target model

tpolecat:
This PR does three things:
- It adds `PioParamCodec` and `PioParamSetCodec` typeclasses for PIO encoding. The main advantage of this approach is that encoding and decoding are defined together via lensing and cannot get out of sync.
- It adds instances for the atomic and composite types in `spModel.core` to allow serialization of the new target model. Tests are included.
- It straightens out the `Lens` situation for the new `Target` model.

:-1: need Jira issue, rebase

### DIFF
--- a/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/TargetParamCodecs.scala
+++ b/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/TargetParamCodecs.scala
@@ -12,19 +12,16 @@ import scalaz._, Scalaz._
 object TargetParamCodecs {
 
   implicit val RedshiftParamCodec: ParamCodec[Redshift] =
-    ParamCodec[Double].xmap(Redshift(_), _.redshift)
+    ParamCodec[Double].xmap(Redshift(_), _.z)
 
   implicit val AngleParamCodec: ParamCodec[Angle] =
     ParamCodec[Double].xmap(Angle.fromDegrees, _.toDegrees)
 
   implicit val ParallaxParamCodec: ParamCodec[Parallax] =
-    ParamCodec[Angle].xmap(Parallax(_), _.angle)
+    ParamCodec[Double].xmap(Parallax(_), _.mas)
 
   implicit val VelocityParamCodec: ParamCodec[Velocity] =
     ParamCodec[Double].xmap(KilometersPerSecond(_), _.toKilometersPerSecond)
-
-  implicit val RadialVelocityParamCodec: ParamCodec[RadialVelocity] =
-    ParamCodec[Velocity].xmap(RadialVelocity(_), _.velocity)
 
   implicit val RaParamCodec: ParamCodec[RA] =
     ParamCodec[Angle].xmap(RA.fromAngle, _.toAngle)
--- a/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/TargetParamCodecs.scala
+++ b/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/TargetParamCodecs.scala
@@ -1,0 +1,60 @@
+package edu.gemini.spModel.target
+
+import edu.gemini.spModel.core._
+import edu.gemini.spModel.core.Target._
+import edu.gemini.spModel.pio.Param
+import edu.gemini.spModel.pio.codec._
+
+import squants.motion.{ Velocity, KilometersPerSecond }
+
+import scalaz._, Scalaz._
+
+object TargetParamCodecs {
+
+  implicit val RedshiftParamCodec: ParamCodec[Redshift] =
+    ParamCodec[Double].xmap(Redshift(_), _.redshift)
+
+  implicit val AngleParamCodec: ParamCodec[Angle] =
+    ParamCodec[Double].xmap(Angle.fromDegrees, _.toDegrees)
+
+  implicit val ParallaxParamCodec: ParamCodec[Parallax] =
+    ParamCodec[Angle].xmap(Parallax(_), _.angle)
+
+  implicit val VelocityParamCodec: ParamCodec[Velocity] =
+    ParamCodec[Double].xmap(KilometersPerSecond(_), _.toKilometersPerSecond)
+
+  implicit val RadialVelocityParamCodec: ParamCodec[RadialVelocity] =
+    ParamCodec[Velocity].xmap(RadialVelocity(_), _.velocity)
+
+  implicit val RaParamCodec: ParamCodec[RA] =
+    ParamCodec[Angle].xmap(RA.fromAngle, _.toAngle)
+
+  implicit val MagnitudeBParamCodec: ParamCodec[MagnitudeBand] =
+    ParamCodec[String].xmap(MagnitudeBand.unsafeFromString, _.name)
+
+  implicit val MagnitudeSysParamCodec: ParamCodec[MagnitudeSystem] =
+    ParamCodec[String].xmap(MagnitudeSystem.unsafeFromString, _.name)
+
+  implicit val AngularVelocParamCodec: ParamCodec[AngularVelocity] =
+    ParamCodec[Double].xmap(AngularVelocity(_), _.masPerYear)
+
+  implicit val RightAscensionAngularVelocParamCodec: ParamCodec[RightAscensionAngularVelocity] =
+    ParamCodec[AngularVelocity].xmap(RightAscensionAngularVelocity(_), _.velocity)
+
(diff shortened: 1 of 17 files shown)


bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/TargetParamSetCodecs.scala
```diff
@@ -0,0 +1,124 @@
...
+      def decode(ps: ParamSet): PioError \/ HorizonsDesignation = {
+        (Option(ps.getParam("tag")).map(_.getValue) \/> MissingKey("tag")) flatMap {
+          case "comet"              => CometParamSetCodec.decode(ps)
+          case "asteroid"           => AsteroidParamSetCodec.decode(ps)
+          case "asteroid-old-style" => AsteroidOldStyleParamSetCodec.decode(ps)
+          case "major-body"         => MajorBodyParamSetCodec.decode(ps)
+          case hmm                  => UnknownTag(hmm, "HorizonsDesignation").left
+        }
+      }
+    }
```
tpolecat: The mechanism for doing tagged sum types is ugly, and needs to be factored out, at least a little.

bundle/edu.gemini.spModel.pio/src/main/scala/edu/gemini/spModel/pio/codec/ParamCodec.scala
```diff
@@ -0,0 +1,59 @@
...
+  def xmap[B](f: A => B, g: B => A): ParamCodec[B] =
+    new ParamCodec[B] {
+      def encode(key: String, a: B): Param = outer.encode(key, g(a))
+      def decode(p: Param): PioError \/ B = outer.decode(p).map(f)
+    }
+}
+
+object ParamCodec {
+  
+  private val pf = new PioXmlFactory
```
tpolecat: I'm done pretending we have more than one PIO implementation

tpolecat:
Oh and I broke the OT. :-1:

tpolecat:
@swalker2m can you glance at this?

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/TargetParamCodecs.scala
```diff
@@ -0,0 +1,60 @@
...
+  implicit val AngleParamCodec: ParamCodec[Angle] =
+    ParamCodec[Double].xmap(Angle.fromDegrees, _.toDegrees)
+
+  implicit val ParallaxParamCodec: ParamCodec[Parallax] =
+    ParamCodec[Angle].xmap(Parallax(_), _.angle)
+
+  implicit val VelocityParamCodec: ParamCodec[Velocity] =
+    ParamCodec[Double].xmap(KilometersPerSecond(_), _.toKilometersPerSecond)
+
+  implicit val RadialVelocityParamCodec: ParamCodec[RadialVelocity] =
```
cquiroz: This will conflict with PR #581 as `RadialVelocity` is gone but that's a minor detail

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/Epoch.scala
```diff
@@ -1,10 +1,17 @@
 package edu.gemini.spModel.core
 
-/** Epoch in Julian years. */
-final case class Epoch(value: Double)
+import scalaz._, Scalaz._
+
+/** Epoch in Gregorian (?) years. */
+final case class Epoch(year: Double)
```
cquiroz: Would it make sense to make it private as we only support J2000?

tpolecat: This is the epoch of observation for targets with proper motion, which is not always 2000.0.

bundle/edu.gemini.spModel.pio/src/main/scala/edu/gemini/spModel/pio/codec/ParamCodec.scala
```diff
@@ -0,0 +1,59 @@
...
+  def encode(key: String, a: A): Param
+  def decode(p: Param): PioError \/ A
+  def xmap[B](f: A => B, g: B => A): ParamCodec[B] =
+    new ParamCodec[B] {
+      def encode(key: String, a: B): Param = outer.encode(key, g(a))
+      def decode(p: Param): PioError \/ B = outer.decode(p).map(f)
+    }
+}
+
+object ParamCodec {
```
cquiroz: Is it the equivalent of `SPTargetPio` for the new model?

tpolecat: `ParamCodec` is a typeclass for things that can be represented as a `Param`; ditto with `ParamSetCodec` and `ParamSet`. I created instances for the new target model but it will work with anything (although it's intended for data types written in Scala).

tpolecat: So I guess `TargetParamSetCodecs.scala` is the equivalent of `SPTargetPio`.

cquiroz:
This looks very good to me, definitely a huge improvement over the old `SPTargetPio` style of implementation.
Perhaps @swalker2m should check it as it may have side effects. On my side it certainly gets a :+1:

bundle/edu.gemini.spModel.pio/src/main/scala/edu/gemini/spModel/pio/codec/ParamCodec.scala
```diff
@@ -0,0 +1,59 @@
...
+  
+  private val pf = new PioXmlFactory
+
+  def apply[A](implicit ev: ParamCodec[A]) = ev
+
+  implicit val StringParamCodec: ParamCodec[String] =
+    new ParamCodec[String] {
+      def encode(key: String, a: String): Param = pf.createParam(key) <| (_.setValue(a))
+      def decode(p: Param): PioError \/ String = Option(p.getValue) \/> NullValue(p.getName)
```
swalker2m: It might not be the case for targets, but sometimes parameters are optional and the absence of a particular parameter isn't an error.  I suppose instead of `String` params in that case they should be `Option[String]` params.

bundle/edu.gemini.spModel.pio/src/main/scala/edu/gemini/spModel/pio/codec/ParamCodec.scala
```diff
@@ -0,0 +1,59 @@
...
+        case \/-(s) => s.parseDouble.disjunction.leftMap(_ => ParseError(p.getName, s, "Double"))
+        case -\/(e) => -\/(e)
+      }          
+    }
+
+  implicit val IntParamCodec: ParamCodec[Int] =
+    new ParamCodec[Int] {
+      def encode(key: String, a: Int): Param = StringParamCodec.encode(key, a.toString)
+      def decode(p: Param): PioError \/ Int = StringParamCodec.decode(p) match {
+        case \/-(s) => s.parseInt.disjunction.leftMap(_ => ParseError(p.getName, s, "Int"))
```
swalker2m: May be a style difference or (likely) I'm just missing something, but I would have been tempted to write:  `StringParamCodec.decode(p).flatMap { s =>  s.parseInt .... }`

bundle/edu.gemini.spModel.pio/src/main/scala/edu/gemini/spModel/pio/codec/ParamSetCodec.scala
```diff
@@ -0,0 +1,113 @@
...
+      }
+      def decode(ps: ParamSet): PioError \/ A =
+        for {
+          a <- outer.decode(ps)
+          p <- Option(ps.getParam(key)) \/> MissingKey("withParam: " + key)
+          b <- pc.decode(p)
+        } yield lens.set(a, b)
+    }
+
+  def withOptionalParam[B](key: String, lens: A @> Option[B])(implicit pc: ParamCodec[B]): ParamSetCodec[A] =
```
swalker2m: Ok, I see optional parameters are accounted for here.

swalker2m:
I think I'd have to use it to really get the feel for it, but the concept seems right to me. :+1:

bundle/edu.gemini.spModel.pio/src/main/scala/edu/gemini/spModel/pio/codec/ParamCodec.scala
tpolecat: No, you're right. I'll fix that.

PR #594, opened 2015-10-14
OCSADV-336A: Unify magnitude systems

fnussber:
Reopened this one for a second round of feedback and to share some insights I had from talking to @jluhrs , @cquiroz  and Angelic.

We have several places where magnitudes are used by systems outside the OCS universe, in particular by the TCC and seqexec, and further downstream GPI and Altair. As far as I know in all those places the magnitudes are sent as bare values without units. The assumption is that the unit for those values is always Vega Mag.

This "works" because it seems that we only use magnitudes for bands for which the default value is indeed Vega Mag (e.g. H and I vor GPI, others for Altair). BUT, in theory, if ever we start sending AB or Jy values, something is most probably not going to work as expected, and Andy found out that Jy is actually not used at all in the current semester.

One of the goals of this PR was to introduce additional magnitude units (actually, they should be called brightness units, because only Vega and AB are magnitudes) for ITC. However when I noticed that I was about to open Pandora's box by doing that I decided with Andy to be a bit less ambitious for now: I am currently only aiming at introducing surface brightness values, which will be handled separately. This PR is still needed, because it facilitates some of the changes I need for that.

We will probably still want to introduce additional integrated brightness units that go along with Vega, AB and Jy (like e.g. Watts/m2), but, as I pointed out above, this will involve a closer look at other systems and is a bit more involved than what we want to do right now. I thought this is interesting information that might interest others, so this is why I went into a second review round with this PR.

@swalker2m Can you give this a quick check, I am interested if you can think of any side-effects I am overlooking here. Thanks.


bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/MagnitudeSystem.scala
```diff
@@ -1,21 +1,27 @@
...
 import scalaz.Equal
 
 sealed abstract class MagnitudeSystem(val name: String) extends Product with Serializable
 
 object MagnitudeSystem {
 
-  case object VEGA extends MagnitudeSystem("Vega")
-  case object AB   extends MagnitudeSystem("AB")
-  case object JY   extends MagnitudeSystem("Jy")
+  case object Vega   extends MagnitudeSystem("Vega")
```
fnussber: Renamed in order to make sure that `toString()` yields the same value as before.

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/MagnitudeSystem.scala
```diff
@@ -1,21 +1,27 @@
...
+  case object AB     extends MagnitudeSystem("AB")
+  case object Jy     extends MagnitudeSystem("Jy")
```
fnussber: Renamed in order to make sure that `toString()` yields the same value as before.

bundle/edu.gemini.shared.skyobject/src/main/java/edu/gemini/shared/skyobject/Magnitude.java
```diff
@@ -235,7 +228,7 @@ public int hashCode() {
      */
     @Override
     public int compareTo(Magnitude that) {
-        int res = system.compareTo(that.system);
+        int res = MagnitudeSystem.all().indexOf(that.system) - MagnitudeSystem.all().indexOf(this.system);
```
cquiroz: I think the sorting of magnitudes in general is problematic, we need it for  display I guess but is there really a way to in general say this magnitude is "before" the other one if they have different systems?

cquiroz:
It looks very similar to the previous PR so :+1: 
However, we should discuss the effects of not caring about magnitude systems in many places, just for starter AGS does numeric comparison and it even uses either r or R which by default have different systems.
GPI also uses just the number internally

bundle/edu.gemini.shared.skyobject/src/main/java/edu/gemini/shared/skyobject/Magnitude.java
fnussber: I agree that there is no meaningful order for the systems and this is entirely arbitrary. My focus here was on mimicking the old behavior in order to avoid unintended side-effects if the order suddenly changes.

fnussber:
Yes, no big changes, but I wanted to bring the problem with the magnitudes without units to everybody's attention.

PR #595, opened 2015-10-14
OCSADV-401 ITC F2 R3000 K-band support

fnussber:
Tiny PR that adds support for an additional filter and filter/disperser combinations for F2 to ITC.


cquiroz:
Not much to comment :+1:

PR #596, opened 2015-10-15
OCSADV-437: BAGS synchronization and threading

sraaphorst:
This PR is to address several things, notably:

1. During refactoring, I found an issue in the way that I was using scala `Future`s inside of a `Thread`, which we have to do to work with the thread pool and GeMS as well as single probes due to the differences in the way these function. This has been fixed with `Await` to wait for the `Future` to complete, whereas previously the thread just merrily kept on letting the `Futures` run and completed, restarting to make more of them, which is exactly what we do not want.

2. Minor refactoring of methods and check for `TimeoutException` during lookups to handle it differently.

3. Synchronization. @tpolecat and @cquiroz both stated some concerns with the way I was using synchronization before, namely due to a nested synchronization. I have managed to work out the kinks and avoid this now so as to avoid possible deadlocking risks. I have also added some comments to the code to provide more detail about how the two data structures used to manage the observation IDs / associated tasks are handled.

I hope that this is sufficient documentation, but I can provide more if there is still any concern.

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
@@ -1,7 +1,7 @@
 package jsky.app.ot.ags
 
 import java.beans.{PropertyChangeEvent, PropertyChangeListener}
-import java.util.concurrent.{LinkedBlockingQueue, Executors}
+import java.util.concurrent.{TimeoutException, LinkedBlockingQueue, Executors}
 import java.util.logging.{Level, Logger}
 
 import edu.gemini.ags.api.{AgsRegistrar, AgsStrategy}
@@ -17,8 +17,10 @@ import jsky.app.ot.OT
 import jsky.app.ot.tpe._
 import scala.collection.JavaConverters._
 import scala.collection.immutable.HashMap
+import scala.concurrent.Await
 import scala.swing.Swing
 import scala.util.{Try, Failure, Success}
+import scala.concurrent.duration._
 import scala.concurrent.ExecutionContext.Implicits.global
 
 import scalaz._, Scalaz._
@@ -41,11 +43,13 @@ object BagsManager {
     def newContext(ctxOpt: Option[ObsContext]): BagsTask =
       BagsTask(observation, curCtx, ctxOpt)
 
-    def performLookup(): Unit = {
+    def performLookup(bagsThread: BagsThread): Unit = {
+      val bagsIdMsg = s"BAGS lookup on thread=${bagsThread.id} for observation=${observation.getObservationID}"
+
       def lookup[S,T](optExtract: S => Option[T])(worker: (TpeContext, S) => Unit)(results: Try[S]): Unit = {
         results match {
           case Success(selOpt) =>
-            LOG.info(s"BAGS lookup for observation=${observation.getObservationID} successful. Results=${optExtract(selOpt) ? "Yes" | "No"}.")
+            LOG.info(s"$bagsIdMsg successful. Results=${optExtract(selOpt) ? "Yes" | "No"}.")
             if (ObservationStatus.computeFor(observation) != ObservationStatus.OBSERVED) {
               Swing.onEDT {
                 muteObservation(observation)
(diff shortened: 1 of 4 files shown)


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -53,37 +57,45 @@ object BagsManager {
...
       curCtx.foreach { obsCtx =>
         AgsRegistrar.currentStrategy(obsCtx).foreach { strategy =>
           if (GuideStarSupport.hasGemsComponent(observation)) {
             lookup((x: GOption[GemsGuideStars]) => x.asScalaOpt)(GemsGuideStarWorker.applyResults(_, _, true))(Try(GemsGuideStarWorker.findGuideStars(obsCtx)))
           } else {
+            // This future always completes, either with results or timeout of 30s, so we wait for it with a long timeout
+            // to let it do its thing.
             val fut = strategy.select(obsCtx, OT.getMagnitudeTable)
-            fut.onComplete(lookup(identity[Option[AgsStrategy.Selection]])(GuideStarWorker.applyResults))
+            val futDone = Await.ready(fut, LookupTimeout)
```
cquiroz: This is a bit strange, in general you don't want to block on a `Future`, that's the whole point of the future. Is it done to make `GuideStarWorker.applyResults` complete inside of the for each?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -91,23 +103,34 @@ object BagsManager {
...
+  private val RequeueDelay = 5000
+
+  // We use an executor to allow for a certain number of threads to run, which are created and fixed as per below.
   private val NumThreads = math.max(1, Runtime.getRuntime.availableProcessors()-1)
-  private val RequeueDelay = 3000
   private val executor = Executors.newFixedThreadPool(NumThreads)
-  private var taskMap = new HashMap[SPNodeKey, BagsTask]
+
+  // Tasks to lookup are managed in two forms: a blocking FIFO queue contains a list of observations by node key
+  // waiting for threads to perform lookups for them, and these node keys are mapped to acutal tasks to perform
```
cquiroz: actual?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -166,50 +189,61 @@ object BagsManager {
...
-            taskMap.synchronized(taskMap += ((obsKey, newTask)))
+          if (!taskQueue.contains(obsKey))
             taskQueue.add(obsKey)
-          }
         }
       }
     }
   }
 
+  // Called when a BAGS task failed to be processed properly, in which case, we must requeue it.
```
cquiroz: Is there an upper limit of retries when you give up on trying to do a BAGS task?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
sraaphorst: This is the result of trying to mix Java `Thread`s with Scala `Future`s so that we can deal with GeMS (Java) and AGS (Scala) in the same way, and I'm really not sure how else to handle it.

Essentially, we want the `Future` to complete within the Java `Thread` that executes it before the `Thread` terminates. I realize that this is not the way that `Futures` are supposed to be used, generally, and is sort of an abuse of the concept, but I don't know how else to make the thread wait on the future to complete before requeueing itself for work.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
sraaphorst: Good catch. Thanks :-).

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
sraaphorst: There isn't. It would be quite simple to add, but not ideal in any sense, as if network connectivity were to, say, come back, then the BAGS task would be dead.

Maybe a better thing to do would be to include a delay directly in BAGS tasks that fail and wait on that instead of what I am doing now. Indeed, I think that would be preferable and prevent them from being queued up again right away.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
tpolecat: According to the doc `Await.ready` throws if there's a timeout exception .. have you tested this with small timeouts just to be sure it does what you think?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -41,11 +43,13 @@ object BagsManager {
     def newContext(ctxOpt: Option[ObsContext]): BagsTask =
       BagsTask(observation, curCtx, ctxOpt)
 
-    def performLookup(): Unit = {
+    def performLookup(bagsThread: BagsThread): Unit = {
+      val bagsIdMsg = s"BAGS lookup on thread=${bagsThread.id} for observation=${observation.getObservationID}"
```
tpolecat: It seems safer to just use `Thread.currentThread` here because there is really nothing guaranteeing that it's running on `bagsThread`.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
cquiroz: Right, I can imagine a situation where the catalog is down and people are accumulating a bunch of tasks that cannot be run. then the catalog comes back and you get a storm of requests. This is probably low priority though

tpolecat:
I can't keep up with this code in my little brain, so I need to spend a couple hours with it on Monday morning. I'll get back with some concrete suggestions.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -1,266 +1,238 @@
...
 
-  private val compositePropertyChangeListener = new PropertyChangeListener {
-    override def propertyChange(evt: PropertyChangeEvent): Unit = evt.getSource match {
-      case node: ISPNode => updateObservation(node.getContextObservation)
-      case _             => // Ignore
-    }
+
+  // Check two target environments to see if the BAGS targets match exactly between them.
+  def bagsTargetsMatch(oldEnv: TargetEnvironment, newEnv: TargetEnvironment): Boolean = {
+    // Find a group with a BAGS target in it in the old and new envs.
```
swalker2m: Of course, I don't know what was discussed with Andy about this but my guess would have been that only the primary guide group matters in this comparison.  In fact I think only the primary guide group should have BAGS targets.  There's no guarantee here that you're even comparing the "same" group since there is no unique identifier for groups.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -1,266 +1,238 @@
...
+                    // We don't want to print the stack trace if the host is simply unreachable.
+                    // This is reported only as a GenericError in a CatalogException, unfortunately.
+                    case Failure(CatalogException((e: GenericError) :: _)) =>
+                      LOG.warning(s"$bagsIdMsg failed: ${e.msg}")
+                      enqueue(obs, 5000L)
+
+                    // If we timed out, we don't want to delay.
+                    case Failure(ex: TimeoutException) =>
+                      LOG.warning(s"$bagsIdMsg failed: ${ex.getMessage}")
+                      enqueue(obs, 5000L)
```
swalker2m: The comment says "we don't want to delay" but then we seem to delay 5 seconds anyway?

sraaphorst: Rob pointed that out to me, and I have removed the delay.
I will wait for any further possible issues before I create a new PR.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
sraaphorst: Right now, the BAGS target is always added to the primary guide group, but what should happen if the user changes the primary guide group? Should the BAGS target be automatically shifted over to the other guide group (or removed from the old guide group and re-looked up)?

swalker2m: I think the later.  In order to change the primary guide group you have to edit the target environment which should cause BAGS to fire up.  When BAGS runs and gets a result I say it should put it in the new primary guide group and remove any from any other groups.  It's quite possible though that I'm not thinking clearly!

PR #597, opened 2015-10-16
OCSADV-336B: Unify magnitude systems, step 2

fnussber:
Second step. This time the magnitude/brightness units of the ITC are unified with the ones of the core model. This made it necessary to introduce a common trait `BrightnessUnit` which is implemented by our good old friend `MagnitudeSystem` and a new version of brightness units `SurfaceBrightness`.

While `MagnitudeSystem` represents integrated brightness units, the new `SurfaceBrightness` is a uniform measurement for the brightness of an object per area (arcsec²).

The next step will be to extend the Target Environment editor in a way that allows to enter brightness values with any of the `SurfaceBrightness` units for ITC calculations.

Note that I did not use `squants` for these values because the refactoring effort would have been considerably bigger and riskier. I leave that as a challenge for some brave soul for the near future..


bundle/edu.gemini.itc.web/src/main/scala/edu/gemini/itc/web/ITCRequest.scala
```diff
@@ -275,19 +275,37 @@ object ITCRequest {
...
     // Get the source geometry and type
     import SourceDefinition.Profile._
+
+    def magnitudeSystemFor(s: String) = s match {
+      case "MAG"                  => MagnitudeSystem.Vega
+      case "ABMAG"                => MagnitudeSystem.AB
+      case "JY"                   => MagnitudeSystem.Jy
+      case "WATTS"                => MagnitudeSystem.Watts
+      case "ERGS_WAVELENGTH"      => MagnitudeSystem.ErgsWavelength
+      case "ERGS_FREQUENCY"       => MagnitudeSystem.ErgsFrequency
```
cquiroz: Do you need default cases in the switch?

cquiroz:
Looks fine :+1:

bundle/edu.gemini.itc.web/src/main/scala/edu/gemini/itc/web/ITCRequest.scala
fnussber: Yeah, you're right. This is input that comes from the HTML form, so it's not like a user will enter these values, but some useful error message would probably still make sense.

fnussber: Well, I guess, you could argue if the exception that the `match` command in Scala throws isn't good enough. Don't know. Anyone has a strong opinion on that? (At least Scala *does* throw in these cases, which is great and often probably good enough.)

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/base/BlackBodySpectrum.java
```diff
@@ -75,23 +74,18 @@ private double _convertToMag(final double flux, final BrightnessUnit units, fina
...
+        } else if (units.equals(MagnitudeSystem.Jy$.MODULE$) || units.equals(SurfaceBrightness.Jy$.MODULE$)) {
+            norm = flux * 1.509e7 / band.center().toNanometers();
+        } else if (units.equals(MagnitudeSystem.Watts$.MODULE$) || units.equals(SurfaceBrightness.Watts$.MODULE$)) {
+            norm = flux * band.center().toNanometers() / 1.988e-13;
+        } else if (units.equals(MagnitudeSystem.ErgsWavelength$.MODULE$) || units.equals(SurfaceBrightness.ErgsWavelength$.MODULE$)) {
+            norm = flux * band.center().toNanometers() / 1.988e-14;
+        } else if (units.equals(MagnitudeSystem.ErgsFrequency$.MODULE$) || units.equals(SurfaceBrightness.ErgsFrequency$.MODULE$)) {
+            norm = flux * 1.509e30 / band.center().toNanometers();
+        } else {
+            throw new IllegalArgumentException("invalid units " + units);
```
tpolecat: I think this would work better if it were written in Scala … this seems hard to maintain and doesn't provide exhaustiveness checking.

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/MagnitudeSystem.scala
```diff
@@ -3,16 +3,36 @@ package edu.gemini.spModel.core
...
+  // currently not supported in OT, only available in web app
+  case object Watts               extends MagnitudeSystem("W/m²/µm")
+  case object ErgsWavelength      extends MagnitudeSystem("erg/s/cm²/Å")
+  case object ErgsFrequency       extends MagnitudeSystem("erg/s/cm²/Hz")
+
 
   val default: MagnitudeSystem = Vega
 
+  // this is used in the OT, for now we only support Vega, AB and Jy in the OT
   val all: List[MagnitudeSystem] =
```
tpolecat: This name is misleading ... could we improve it somehow?

bundle/edu.gemini.itc.web/src/main/scala/edu/gemini/itc/web/ITCRequest.scala
tpolecat: If these are the canonical string representations can we put them in `fromString(s:String): Option[Whatevs]` and `unsafeFromString(s: String): Whatevs` methods on the Companions?

fnussber:
Incorporated @tpolecat 's feedback.

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/MagnitudeSystem.scala
```diff
@@ -3,25 +3,91 @@ package edu.gemini.spModel.core
...
+    case "MAG"                  => Some(MagnitudeSystem.Vega)
+    case "ABMAG"                => Some(MagnitudeSystem.AB)
+    case "JY"                   => Some(MagnitudeSystem.Jy)
+    case "WATTS"                => Some(MagnitudeSystem.Watts)
+    case "ERGS_WAVELENGTH"      => Some(MagnitudeSystem.ErgsWavelength)
+    case "ERGS_FREQUENCY"       => Some(MagnitudeSystem.ErgsFrequency)
+    case _                      => None
+  }
+
+  def unsafeFromString(s: String): MagnitudeSystem = fromString(s).get
```
tpolecat: Recommend a useful message here rather than `None.get`

```scala
.getOrElse(throw new NoSuchElementException("Unknown MagnitudeSystem: " + s))
```

tpolecat: I would expect `fromString(sys.name) == Some(sys)` to hold but this doesn't seem to be the case (`MAG` vs `Vega` for instance). If this is a non-canonical mapping then it may need a more descriptive name. I'm being slightly hard-assed about this because this kind of thing has caused all kinds of trouble in OCS (human-readable vs. TCS-grokkable names, with no explanation anywhere).

fnussber:
Next try.

cquiroz:
I think it looks good, it is much more readable in scala with patter matching than the ifs
A general question, does it make sense to make the convert java classes into case classes or do they need to follow some Java spec (JavaBeans perhaps?)

fnussber:
@cquiroz  
Cloning: Yes, most ITC operations are in-place on mutable data structures, but then there is the occasional step where a copy of the data structure is needed and this is done using `clone`. All very old-school ;) Obviously there is room for improvement and although I have changed stuff to be immutable in many places there is still much left to do.

Redshift: Straightening out redshift issues and using `Redshift` from the core is planned with OCSADV-334.

Case classes: I usually turn Java classes into a `case class` only if they are basic data containers without much logic. Otherwise I turn them into a normal `class` or `object` if possible. ITC does not rely on any JavaBean naming conventions, so we're pretty free to change/rename stuff here. And if JavaBean is needed there's also the magical `@BeanProperty` annotation.

tpolecat:
@fnussber @cquiroz I think the choice between `class` and `case class` comes down to whether the type is isomorphic to an equivalent tuple or not. That is, can you convert it into a tuple for a while and then back into the original type with no ill effect? Obviously for anything mutable (where reference equality is important) or with overridden `equals` or `toString` or `hashCode` this is not the case. So I think when converting *mechanically* from Java a plain `class` is often more appropriate. Which is to say, I think @fnussber's rule of thumb is sensible.

Also, this PR seems fine. :+1:

PR #598, opened 2015-10-16
OCSADV-354: OT ITC support for user-defined SED

fnussber:
Ok, this is good to go. I've already incorporated Rob's feedback regarding migrating some Java code to Scala. More comments welcome.

In particular: I am introducing a new `AuxFileType` for SED (Source energy distribution) files. These should be handled like `other` files except for not having email notifications sent out when they get uploaded/replaced/checked. It seemed that introducing an additional `AuxFileType` would be the right thing to do to deal with that. However there is the downside that currently for every `AuxFileType` there needs to be a specific configuration for host, directory etc. In order to simplify this, I decided to assume that for file types for which no specific configuration is available (like e.g `fits`), I just "recycle" the configuration values of the `other` type. This allows to have different aux file types, without turning this into a configuration nightmare. See also changes in `OsgiCopyConfig`.

Another point worth mentioning: In the UI the user selects an aux file by name (filtered by ending ".sed"). In order to avoid a roundtrip of the actual file Server->OT->Server the aux file content is not loaded into the OT and sent down to the ITC service but the aux file is retrieved on the server. This works as long as the ITC service used by the OT is running on the same machine as the database, which currently is the case. If this changes, we need to revisit this and change the host used in the `AuxFileClient` used to retrieve the files. See changes to `ItcServiceImpl`.


bundle/edu.gemini.itc.web/src/main/java/edu/gemini/itc/web/html/HtmlPrinter.java
```diff
@@ -23,36 +23,32 @@ public static String printParameterSummary(final SourceDefinition sdp) {
...
+                    " star in the " + sdp.normBand().name() + " band.");
+        } else if (sdp.distribution() instanceof LibraryNonStar) {
+            sb.append(" " + sdp.norm() + " " + sdp.units().displayValue() + " " + ((Library) sdp.distribution()).sedSpectrum() +
+                    " in the " + sdp.normBand().name() + " band.");
+        } else if (sdp.distribution() instanceof UserDefinedSpectrum) {
+            sb.append(" a user defined spectrum with the name: " + ((UserDefinedSpectrum) sdp.distribution()).name());
+        } else if (sdp.distribution() instanceof PowerLaw) {
+            final double index = ((PowerLaw) sdp.distribution()).index();
+            sb.append(" Power Law Spectrum, with an index of " + index
+                    + " and " + sdp.norm() + " mag in the " + sdp.normBand().name() + " band.");
```
tpolecat: As with the previous PR it seems to me that this would work better as a method written in Scala since you get no exhaustiveness checking here.

fnussber: Agreed, by moving all the parameter objects to Scala writing new code became easier, but obviously there is a downside to it on the Java side of things. The question is how much time we are going to spend on this, I was currently not planning on moving all that code to Scala (yet).

bundle/edu.gemini.itc.web/src/main/scala/edu/gemini/itc/web/html/HtmlUtil.scala
```diff
@@ -0,0 +1,42 @@
...
+ */
+object HtmlUtil {
+
+  // string representation for the spectral distribution
+  def sourceDistributionString(sd: SourceDefinition): String = sd.distribution match {
+
+    case EmissionLine(wavelength, width, flux, cont) =>
+      s"n emission line, at a wavelength of $wavelength, and with a width of $width.\n" +
+      s"It's total flux is $flux on a flat continuum of flux density $cont."
+
```
fnussber: Thanks to squants this automatically takes the units into account and produces nicely formatted output like for example: 
```
The z = 0.0 extended source is an emission line, at a wavelength of 0.76 µm, and with a width of 500.123 km/s.
It's total flux is 5.123E-17 erg/s/cm² on a flat continuum of flux density 1.0E-17 W/m²/µm.
```

cquiroz: :+1:

fnussber:
@swalker2m This is the PR I would love to include in the test release on Monday. Can you have a very quick look at the `AuxFileType` extension (including the change to `OsgiCopyConfig`) and the use of `AuxFileClient` in `ItcServiceImpl`? The `AuxFileClient` usage follows what we've discussed on Thursday, so there shouldn't be any surprises there. Thanks!

bundle/edu.gemini.auxfile.workflow/src/main/java/edu/gemini/auxfile/client/AuxFileClient.java
```diff
@@ -120,6 +122,29 @@ private static void writeChunk(AuxFileChunk chunk, File destFile) throws IOExcep
         }
     }
 
+    @Override public byte[] fetchToMemory(final SPProgramID programId, final String remoteFileName) throws AuxFileException {
+
+        try (final ByteArrayOutputStream out = new ByteArrayOutputStream(FileChunkSize)) {
```
cquiroz: This syntax is new to me, is it the Java 8 resources self-cleaning idea?

bundle/edu.gemini.auxfile.workflow/src/main/java/edu/gemini/auxfile/workflow/CheckTask.java
```diff
@@ -17,7 +17,11 @@ public CheckTask(SPProgramID progId, File file, boolean checked) {
 	@Override
 	public void execute(Workflow wf) throws Exception {
         final AuxFileType type = AuxFileType.getFileType(getFile());
-        wf.getMailer().notifyChecked(getProgId(), type, getFile().getName(), _checked);
+
+        // Try to send out emails if notification for this file type is desired.
+        if (type.sendNotification()) {
```
cquiroz: Is indentation off? Is email sending async?

cquiroz:
I don't really understand much about the role of aux files so I can't comment
Code-wise though things seems good to me

bundle/edu.gemini.auxfile.workflow/src/main/java/edu/gemini/auxfile/client/AuxFileClient.java
fnussber: Yes, it basically adds a `finally` which closes all the resources. Similar to `using` in C#.

bundle/edu.gemini.auxfile.workflow/src/main/java/edu/gemini/auxfile/workflow/CheckTask.java
fnussber: Not sure what the question is. This is not `notify()` or some such thing but just a `boolean` that tells you if emails should be sent out for this aux file type or not. Maybe the name is misleading and it should be renamed to `sendEmails()`?

bundle/edu.gemini.itc/src/main/scala/edu/gemini/itc/service/ItcServiceImpl.scala
```diff
@@ -26,8 +29,24 @@ class ItcServiceImpl extends ItcService {
...
   def calculate(source: SourceDefinition, obs: ObservationDetails, cond: ObservingConditions, tele: TelescopeDetails, ins: InstrumentDetails): Result = try {
 
-    if (obs.getMethod.isImaging)  calculateImaging(source, obs, cond, tele, ins)
-    else                          calculateSpectroscopy(source, obs, cond, tele, ins)
+    // Get the SED data from an aux file. For now we can assume that the ITC service is running on the same
+    // machine as the database (localhost). In case this setup changes, we need to change this here, too.
+    def auxFileDistribution(id: String, name: String): SpectralDistribution = {
+      val programId     = SPProgramID.toProgramID(id)
+      val spectrumBytes = new AuxFileClient("localhost", 8443).fetchToMemory(programId, name)
+      val spectrum      = new String(spectrumBytes)
```
swalker2m: The is the first auxfile type where we convert the data to/from Strings. Do we need to worry about character encodings?  How are the files created?

bundle/edu.gemini.auxfile.workflow/src/main/java/edu/gemini/auxfile/client/AuxFileClient.java
```diff
@@ -21,15 +21,17 @@
...
-    private final int chunkSize;
+    private static final int FileChunkSize = 32 * 1024;
+
     private final AuxFileServer server;
 
-    public AuxFileClient(KeyChain kc, int chunkSize, String host, int port) {
-        this.chunkSize = chunkSize;
-        this.server    = TrpcClient$.MODULE$.apply(host, port).withKeyChain(kc).proxy(AuxFileServer.class);
+    public AuxFileClient(String host, int port) {
+        this.server    = TrpcClient$.MODULE$.apply(host, port).withoutKeys().proxy(AuxFileServer.class);
```
swalker2m: To my surprise, unless I'm missing something, we don't seem to be doing any permissions checking for auxfile operations on the server side.  That seems poor, though obviously not the fault of this PR. We _should_ require a valid key to manipulate auxfiles though, at which point this constructor won't be very useful.

That raises the question of how this would work if the auxfile server were doing permission checking.  The ODB itself (or whatever process the ITC service is running in) would need a staff key.

bundle/edu.gemini.auxfile.workflow/src/main/java/edu/gemini/auxfile/workflow/CheckTask.java
swalker2m: I think there might be a spaces/tabs issue in the formatting.

`sendNotification` seems fine to me too.  It boils down to a call `javax.mail.Transport.send` which is synchronous.  The `CheckTask` itself is running in a separate worker thread though.  At any rate, none of this behavior is new in the PR it's just that the new auxfile type apparently doesn't need to trigger the email so the flag had to be added to differentiate auxfiles that do and do not trigger notifications.

swalker2m:
The new `AuxFileType` and handling all seem fine to me.  It looks to me like you covered all the bases with that.  I'm confused on the permission checking question and unsure about the character encoding issue.  Both seem potentially important but I think we shouldn't hold off on putting it into the test release though.

bundle/edu.gemini.auxfile.workflow/src/main/java/edu/gemini/auxfile/client/AuxFileClient.java
fnussber: Agreed. We need to introduce some sort of a service account that has staff privileges or some such thing.

bundle/edu.gemini.itc/src/main/scala/edu/gemini/itc/service/ItcServiceImpl.scala
fnussber: The good news is that these particular files usually are just long lists of numbers. However I agree that encoding can break things. My hope is, that `String(bytes[])` tries to do something sensible by using the system's default encoding or whatever, but of course I might be wrong about that and also if someone throws a file at us with a weird encoding then I really don't know how we can handle that gracefully. I'll do some additional testing and see what happens if we throw weird encodings at this.

PR #600, opened 2015-10-19
REL-2407-C: GSA queries

swalker2m:
This grim PR adds support for doing queries to the new GSA server.  In particular, it adds a query for obtaining GSA file records matching various constraints and a query for setting dataset QA states.  Both queries are synchronous and any background processing will be built on top.  A basic shell command is provided for executing the queries, mainly for testing and debugging.  __Note, the Data Manager itself has not yet been updated to take advantage of these queries.__

Now a moral dilemma.  The GSA server provides rudimentary security by requiring that a particular secret cookie value be sent by the client.  This isn't the product of authentication and instead is just "known". I'm aware that this is not optimal but I'm not sure how much we should push back.  It doesn't really trouble my conscience too much since it is only used internally.  We probably should at least try to get this done over https.


bundle/edu.gemini.dataman.app/src/main/java/edu/gemini/dataman/osgi/Activator.java
```diff
@@ -4,80 +4,125 @@
...
-import edu.gemini.dataman.raw.RawCopier;
-import edu.gemini.dirmon.DirListener;
-import edu.gemini.dirmon.DirLocation;
-import edu.gemini.util.security.principal.StaffPrincipal;
+//import edu.gemini.dataman.context.DatamanConfig;
+//import edu.gemini.dataman.context.DatamanContext;
+//import edu.gemini.dataman.raw.RawCopier;
+//import edu.gemini.dirmon.DirListener;
+//import edu.gemini.dirmon.DirLocation;
+//import edu.gemini.util.security.principal.StaffPrincipal;
```
swalker2m: Don't worry too much about the `Activator`.   I've commented out all the old Data Manager setup but will be coming back to it in the next PR (or a near future PR).  Some bits will be deleted and others modified and in fact I'll probably rewrite it in Scala.

bundle/edu.gemini.dataman.app/src/main/java/edu/gemini/dataman/osgi/Activator.java
```diff
@@ -4,80 +4,125 @@
...
-    }
+        final Dictionary<String, Object> dict = new Hashtable<>();
+        dict.put(COMMAND_SCOPE, "gsa");
+        dict.put(COMMAND_FUNCTION, new String[]{"gsa"});
 
-    private static RawCopier _registerRawCopier(BundleContext ctx, DatamanContext dctx) {
-        DatamanConfig conf = dctx.getConfig();
-        LOG.fine("Register DirListener for: " + conf.getRawDir());
+        // TODO: we don't have a GS GSA test server or GS data so fix to GN for now
+        ctx.registerService(GsaCommands.class, GsaCommands$.MODULE$.apply(gsaHost, Site.GN, gsaAuth), dict);
```
cquiroz: I thought there will be a single production GSA? Or do you think there will be a GS test server?

swalker2m: There's a GN test server and I've asked for a GS test server.  In production it will use production servers at GN and GS.  So probably 4 different GSA servers.

bundle/edu.gemini.dataman.app/src/main/scala/edu/gemini/dataman/gsa/query/GsaFileQuery.scala
```diff
@@ -0,0 +1,80 @@
...
+        val sitePrefix = site match {
+          case Site.GN => "N"
+          case Site.GS => "S"
+        }
+        GsaQuery.get(url(s"$sitePrefix${night.getNightString}"))
+      }
+
+      override def thisWeek: GsaResponse[List[GsaFile]] = {
+        val endNight = new ObservingNight(site)
+        val c = new GregorianCalendar(site.timezone()) <|
```
cquiroz: Can you do the same operations with the new Java DateTime API?

bundle/edu.gemini.dataman.app/src/main/scala/edu/gemini/dataman/gsa/query/GsaQaUpdateQuery.scala
```diff
@@ -0,0 +1,37 @@
...
+  def setQaState(label: DatasetLabel, qa: DatasetQaState): GsaResponse[List[GsaQaUpdate.Response]] =
+    setQaStates(List(GsaQaUpdate.Request(label, qa)))
+}
+
+object GsaQaUpdateQuery {
+  import GsaQaUpdate._
+
+  /** Constructs a QA update query for the given host and site.  The `auth`
+    * parameter is how the GSA service provides some minimal protection against
+    * inappropriate setting of QA states.  The value is sent in a cookie that
```
cquiroz: Yes, the cookie seems weak protection, Where does it come from?

bundle/edu.gemini.dataman.app/src/main/scala/edu/gemini/dataman/gsa/query/GsaQaUpdate.scala
```diff
@@ -0,0 +1,47 @@
...
+  implicit def EncodeJsonResponse: EncodeJson[Response] =
+    EncodeJson((r: Response) =>
+      ("error"  :=? r.failure)                            ->?:
+      ("result" :=  r.failure.fold("true")(_ => "false")) ->:
+      ("id"     :=  r.label.toString)                     ->:
+      jEmptyObject
+    )
+
+  // Decoder for the server response to an update request for a particular
+  // dataset.  Here we ignore the "result" flag that comes in the JSON and base
```
cquiroz: Does the HTTP response also contain an error code in case of a failure?

bundle/edu.gemini.dataman.app/src/main/scala/edu/gemini/dataman/gsa/query/GsaFileQuery.scala
swalker2m: Surely it is possible, perhaps with `ZonedDateTime`?  Observing nights are specific to sites.  The number of milliseconds over "6 days" will differ if there is a time zone switch so it requires knowledge of the time zone.  I'm probably guilty of being lazy but I just wanted to use the exact same logic as used internally in `ObservingNight`.

bundle/edu.gemini.dataman.app/src/main/scala/edu/gemini/dataman/gsa/query/GsaQaUpdateQuery.scala
swalker2m: This is what I'm referring to by "moral dilemma" in the PR description.  It's just "known".  Specifically we'll keep it in svn in `OcsCredentials` along with the other sensitive information.  I'm sure we're doing worse things in our own codebase and this doesn't concern me too much, but I thought I should highlight it for discussion just in case.

bundle/edu.gemini.dataman.app/src/main/scala/edu/gemini/dataman/gsa/query/GsaQaUpdate.scala
swalker2m: Yes, I'm reading that into a `RequestError` response in `GsaQuery.unsafeDoQuery`.

PR #601, opened 2015-10-20
OCSADV-334: Use Redshift from core model in ITC code.

fnussber:
Use `Redshift` value from core model instead of a double in ITC code.

@cquiroz Maybe we should rename the value from `redshift` to `value` or `z`? As it is now we end up with many places like `.. = redshift.redshift` in the code. Not really a big issue, just wondering if renaming would make the code more readable.


cquiroz:
Looks good :+1: 
I'm ok with renaming  it to `z` or something else readable

PR #603, opened 2015-10-20
Bugfix: Added missing export.

fnussber:
Export for codec package seems to be missing from the spModel pio bundle.


cquiroz:
:+1:

PR #604, opened 2015-10-21
Move SpatialProfile and SpectralDistribution to core

tpolecat:
This PR just moves two types from `pot` to `spModel_core` so we can add them to `Target`.


cquiroz:
Seems very straightforward :+1:

PR #605, opened 2015-10-21
OCSADV-??? Target Cleanup

tpolecat:
This PR cleans up `Target` a bit, makes the subtypes top-level, introduces a `DefinedTarget` intermediate trait, and moves the lenses into separate traits. This is just mechanical cleanup.


cquiroz:
Looks very good, I was perplexed why the target subtypes were not top-level classes :+1:

PR #606, opened 2015-10-22
OCSADV-437: BAGS threading and lookups

sraaphorst:
This fixes three minor things:

1. Thread priority of BAGS threads. (May speed things up?)

2. Initial check for whether or not observations should be queued for BAGS lookup had accidentally been moved to always, causing obs changes to fail to queue obs for BAGS, so there is a fix for this with the addition of an `initialEnqueue` parameter to `BagsManager.enqueue` to only perform the check when programs are first watched. This will go away with the upcoming feedback changes, so this should be considered a temporary fix to a temporary problem.

3. Unrelated: timeout increased for `VoTableClientSpec` since the small value for timeout was causing this test case to frequently fail.

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
@@ -143,7 +143,7 @@ final class BagsManager(executor: ScheduledThreadPoolExecutor) {
                     // If we timed out, we don't want to delay.
                     case Failure(ex: TimeoutException) =>
                       LOG.warning(s"$bagsIdMsg failed: ${ex.getMessage}")
-                      enqueue(obs, 5000L)
+                      enqueue(obs, 0L)
 
                     // For all other exceptions, print the full stack trace.
                     case Failure(ex) =>
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
@@ -97,8 +97,10 @@ final class BagsManager(executor: ScheduledThreadPoolExecutor) {
       synchronized {
         val key = obs.getNodeKey
         state += key
-        executor.schedule(new Runnable {
-          def run(): Unit =
+        executor.schedule(new Thread {
+          setPriority(Thread.NORM_PRIORITY - 1)
+
+          override def run(): Unit =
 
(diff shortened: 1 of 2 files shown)


sraaphorst:
Oh, also fixed the discrepancy between delay and the comment for requeueing failed lookups as per Shane and Rob's comments.

tpolecat:
Do we not need to do the `isEligibleForBags` check each time? I think I don't understand the logic. What if the properties of the observation change in such a way that an observation's eligibility changes?

sraaphorst:
The `isEligibleForBags` is simply a check at the time a program is opened to determine if an initial lookup should be done for an observation: if there is already a BAGS guide star when a program is opened, then there is no need to invoke BAGS.

The problem with doing this call each time is that if there is an existing BAGS target and a target designated as the primary, then this call will return `false` even when there is a change to an observation, and thus no BAGS lookup will be done despite there being a change to the observation.
(It can be forced to return `true` by deselecting the primary guide star, in which case, a lookup will be done, but it will not automatically execute as it should whenever an obs changes.)

This is all going to change, though, by switching the `GuideProbeTargets.bagsTarget` to a `BagsStatus` sealed trait with several case classes / objects, so this will be irrelevant soon enough.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -92,20 +92,22 @@ final class BagsManager(executor: ScheduledThreadPoolExecutor) {
    * Atomically enqueue a task that will consider the specified observation for BAGS lookup, after
    * a delay of at least `delay` milliseconds.
    */
-  def enqueue(observation: ISPObservation, delay: Long): Unit =
+  def enqueue(observation: ISPObservation, delay: Long, initialEnqueue: Boolean = false): Unit =
```
tpolecat: Ok if `isEligibleForBags` only happens once then I suggest you pass the eligibility check as a parameter.

```scala
def enque(observation: ISPObservation, delay: Long)(continue: ObsContext => Boolean): Unit = {
  ...
  ...  ObsContext.create(obs).asScalaOpt.filter(continue).foreach { ctx => ...
}
```

So the initial call is `enque(..., ...)(_.isEligibleForBags)` and subsequent `enque(..., ...)(_ => true)`.

Whatever you do here please update the doc to indicate what the check does and why it only happens once.

sraaphorst:
If nobody has any objections or further comments, I am going to merge this in preparation for the changes for OCSADV-443, as I will need this in place to move to the next stage of that task.

PR #607, opened 2015-10-22
OCSADV-419 add SpectralDistribution and SpatialProfile to DefinedTarget

tpolecat:
This PR adds an `Option[SpectralDistribution]` and `Option[SpatialProfile]` to `DefinedTarget` (i.e., `SiderealTarget` and `NonSiderealTarget`), which is mostly trivial but required some work for PIO and associated unit testing.


bundle/edu.gemini.pot/src/test/scala/edu/gemini/spModel/target/SpTargetPioSpec.scala
```diff
@@ -24,28 +18,6 @@ import edu.gemini.shared.util.immutable.ScalaConverters._
...
-      )
-    }
-    implicit val arbProfile = Arbitrary[SpatialProfile] {
-      Gen.oneOf(
-        PointSource,
-        UniformSource,
-        GaussianSource(0.5),
-        GaussianSource(0.75)
-      )
-    }
```
tpolecat: These were moved to `edu.gemini.spModel.core.Arbitraries` since they're needed for core unit tests now.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/pot/ModelConverters.scala
```diff
@@ -194,7 +198,13 @@ object ModelConverters {
         case t:HmsDegTarget => Some(t.getRedshift)
         case _              => None
       }
-      SiderealTarget(name, coordinates, pm, z, px, mags)
+      SiderealTarget.empty.copy(
```
cquiroz: Do `SpectralDistribution` or `SpatialProfile` have any sense on the old model
Somewhat I'd like in here to use the full constructor so if it changes like in this PR you must remember to update the conversion, otherwise this will compile cleanly

fnussber: Yeah, I am wondering, if we don't loose the `SpectralDistribution` and `SpatialProfile` if we convert back and forth between the two? I am not sure if it matters right now, since ITC currently takes those values from the old model (ITarget).

tpolecat: Yeah I missed that. I'll fix it.

tpolecat:
@cquiroz I fixed `ModelConverters` to preserve spectral distribution and spatial profile for `SPTarget`, and replaced `.copy(...)` with full constructor calls as suggested.

tpolecat:
Rebased. Tests pass.

cquiroz:
Looks good :+1:

PR #609, opened 2015-10-26
REL-2429 blueprint updates for GRACES 

tpolecat:
- Updates GRACES blueprint and template expansion logic.
- Fixes incorrect Altair wavelength constant in GSAOI_BP and NIFS_BP to get rid of warning.

Man I'm glad I added that end-to-end test support last time around. This is much easier than it used to be.


bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/template/factory/impl/graces/Graces.scala
```diff
@@ -30,40 +30,47 @@ case class Graces(blueprint: SpGracesBlueprint, exampleTarget: Option[SPTarget])
       this.db = None
     }
 
-  // R = Phase-I target R-band magnitude
+  // R = Phase-I target R-band or V-band magnitude
   val rMag: Option[Double] =
     for {
       t <- exampleTarget
-      m <- t.getTarget.getMagnitude(Band.R).asScalaOpt
+      m <- t.getTarget.getMagnitude(Band.R).asScalaOpt orElse
```
cquiroz: Unrelated to this PR but should we consider if this must follow the RLike band concept? perhaps a question for @andrewwstephens

cquiroz:
Love the tests :+1:

bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/template/factory/impl/graces/Graces.scala
andrewwstephens: We don't expect UC magnitudes, so I think we are good for now.  In the future we may want to include other filters.

PR #611, opened 2015-10-26
OCSADV-440: Bug fix in extraction of central wavelength for GNIRS ITC.

fnussber:
For GNIRS the central wavelength in a sequence can be a `String` or a `GNIRSParams.Wavelength` in case of a GNIRS sequence iterator. This has to be taken into account when extracting this value for ITC calculations.


cquiroz:
:+1:

PR #612, opened 2015-10-26
OCSADV-443: Data model change for BAGS feedback

sraaphorst:
Previously, we maintained an `Option<SPTarget>` in `GuideProbeTargets` to store a possible BAGS target; however, due to the new requirement to provide feedback for BAGS, the decision was made to instead replace this with a `BagsStatus` which comprises a number of possible statuses that can be used and represented by BAGS.

This is the initial change to the data model to accommodate this feature. The actual feedback itself and changes to `BagsManager` will be forthcoming, but the previous functionality has been maintained through this change.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/GuideProbeTargets.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/GuideProbeTargets.java
@@ -16,8 +16,9 @@ import java.util.*;
 //* context of the observation, and disabled otherwise.
 
 /**
- * An immutable pairing of a {@link GuideProbe} with a collection of {@SPTarget}.
- * One target may be marked as having been picked by background AGS (BAGS).
+ * An immutable pairing of a {@link GuideProbe} with a collection of {@link SPTarget}.
+ * One target may be marked as having been picked by background AGS (BAGS), or a status of such a background AGS lookup
+ * is stored.
  * One target may be marked as the primary target for the guider.
  *
  * <p>Note that while this object and the list of targets itself is immutable, the actual SPTargets themselves are
@@ -85,7 +86,7 @@ public final class GuideProbeTargets implements Serializable, TargetContainer, I
     /**
      * Creates a GuideTargets instance associated with the given
      * {@link GuideProbe}.
-     * It marks no BAGS guide star, and no primary guide star.
+     * It marks no BAGS guide star yet found, and no primary guide star.
      *
      * @param guider guide probe to associate with the GuideTargets
      * @param manualTargets zero or more manual targets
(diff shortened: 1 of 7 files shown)


bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/GuideProbeTargets.java
```diff
@@ -467,11 +479,11 @@ public GuideProbeTargets addManualTarget(final SPTarget target) {
      */
     public GuideProbeTargets withManualTargets(final ImList<SPTarget> manualTargetsNew) {
         // If the primary target is None, or the primary is the BAGS, then just return with the new list.
-        if (primaryTarget.forall(t -> bagsTarget.forall(t::equals)))
-            return new GuideProbeTargets(guider, bagsTarget, primaryTarget, manualTargetsNew);
+        if (bagsStatus.bagsStarAsJava().forall(bt -> primaryTarget.forall(bt::equals)))
```
cquiroz: Can you just do `Option` equality?

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/env/BagsStatus.scala
```diff
@@ -0,0 +1,78 @@
...
+import edu.gemini.shared.util.immutable.{Option => GOption}
+import edu.gemini.spModel.pio.{ParamSet, Pio, PioFactory}
+import edu.gemini.spModel.target.SPTarget
+import edu.gemini.spModel.rich.shared.immutable._
+
+// Possible BAGS statuses used by GuideProbeTargets to indicate the state of BAGS targets and lookups.
+sealed trait BagsStatus extends Cloneable {
+  val bagsStar: Option[SPTarget] = None
+  def bagsStarAsJava: GOption[SPTarget] = bagsStar.asGeminiOpt
+  override def clone: BagsStatus = this
```
cquiroz: This looks a bit funny, is `BagsSuccessWithTarget` the special case? Could you use the new model `Target` instead of the old `SPTarget`?

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/env/BagsStatus.scala
```diff
@@ -0,0 +1,78 @@
...
+//    Option(parent.getParamSet(BagsStatusParamSetName)).map { ps =>
+//      Option(ps.getParam(BagsStatusParamName).getValue) match {
+//        case Some(BagsSuccessNoTargets.id)  => BagsSuccessNoTargets
+//        case Some(BagsSuccessWithTarget.id) =>
+//          Option(ps.getParamSet(BagsTargetParamSetName)).map(bps => BagsSuccessWithTarget(SPTarget.fromParamSet(bps))).getOrElse(BagsLookupPending)
+//        case _ => BagsLookupPending
+//      }
+//    }.getOrElse(BagsLookupPending)
+//  }
+  def fromParamSet(parent: ParamSet): BagsStatus = {
```
cquiroz: Do we have tests for serialization/deserialization?

cquiroz:
It looks in general fine, though for model changes it would be good to have tests. It seems there is a `PIO...Test` somewhere, maybe that covers storing/reading the model with `BagsStatus`?

sraaphorst:
Thanks for the feedback, @cquiroz. I'll look at this when I get in in a few minutes.
I also thought of a couple bits of logic I would like to change that I had originally planned to put in `BagsManager`, but that I believe would be better suited to `GuideProbeTargets` instead, so expect a commit changing things early this morning.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/env/BagsStatus.scala
```diff
@@ -0,0 +1,78 @@
...
+// TODO: For case class, we want to return an instance.
+
+// Note that we use BagsLookupPending as a default state when BAGS hasn't run or should be rerun, and the BagsManager
+// determines based on the observation if BAGS actually should be run.
+case object BagsLookupPending extends BagsStatus {
+  override val id = "BagsLookupPending"
+}
+case object BagsLookupRunning extends BagsStatus {
+  override val id = "BagsLookupRunning"
+}
```
swalker2m: I hate to bring this up, but in light of Andy's findings about BAGS and program synchronization ...  Perhaps it isn't a good idea to store transient states in the model, if that's what we're doing (?).  Certainly the end result of the BAGS search is necessary in the model but updating to store intermediate results will mark the target node as modified even if in the end the search comes up with the same answer.  That in turn only increases the chance of bogus conflict reports.  Would it be possible to manage the state outside of the model and only store the final result, and only if it differs from what is already stored?  You may be doing it that way already?

PR #613, opened 2015-10-26
REL-2498 Update GNIRS blueprint

tpolecat:
This adds the new XML file which has minor changes and did not require any logic changes. I added a sanity test just to ensure that template expansion works but didn't try to test the existing expansion logic (which is quite complex).

### DIFF
--- a/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Altair.scala
+++ b/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Altair.scala
@@ -45,6 +45,7 @@ case object AltairNone extends Altair {
 }
 
 case class AltairLGS(pwfs1: Boolean, aowfs: Boolean = false, oiwfs: Boolean = false) extends AltairAO {
+  // At most one can be true
   require((!pwfs1 && !aowfs && !oiwfs) || (pwfs1 ^ aowfs ^ oiwfs))
   require(!(pwfs1 && aowfs && oiwfs))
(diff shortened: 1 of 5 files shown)


cquiroz:
Why does the p1 model changes? Is the change reflected on the schema?

tpolecat:
P1 changes are trivial and don't affect the model.

If you have a case class with no explicit companion you can use the *compiler-generated companion* as a function value because it extends a type that matches the constructor.

```scala
case class Foo(n: Int, s: String)

// compiler-generated companion object looks like this
object Foo extends ((Int, String) => Foo) {
  def apply(n: Int, s: String): Foo = new Foo(n, s)
  def unapply ...
}
```

But if you have an *explicit* companion object you have to extend the function type manually, otherwise you can't use it as a function value. You can *call* it because of its `apply` method but you can't pass it to a higher-order function for instance.

```scala
case class Bar(n: Int, s: String)

// We have to add the supertype by hand in the companion
object Bar extends ((Int, String) => Bar) {
  // apply is still generated for us
  ...
}
```

I was constructing blueprints in an applicative comprehension `(... |@| ... |@| ...)(FooBlueprint)` so it was helpful to treat the companions as functions. In the end I decided to use ScalaCheck instead of exhaustively trying every possible blueprint because it was taking forever, so I actually don't need the function values now. But it's useful generally so I left it there.

cquiroz:
Nice to know, I seem to remember passing `Bar.apply` as a function sometimes
Anyway :+1:

tpolecat:
Rebased. Tests pass.

PR #614, opened 2015-10-26
REL-2385 update GMOS template

tpolecat:
This updates the GMOS template, which did not require any logic changes. Added a sanity check just to ensure that template expansion works in all configurations.


cquiroz:
Great :+1:

tpolecat:
Rebased. Tests pass.

PR #616, opened 2015-10-27
REL-2387: Add F2 Custom Mask "Slit Width"

fnussber:
This PR adds functionality that allows users to select a slit width in case they are using a custom mask with F2. This is relevant for smart gcal and for ITC calculations where the slit width is a mandatory parameter.

(Note that the current smartgcal design did not anticipate custom masks with arbitrary slit widths and the lookup mechanism in use today only works with the F2 FPUnits implemented as enums. In order to be able to use smartgcals for custom masks, we currently only support a limited set of custom slit widths which each correspond to one of the F2 FPUnits. Andy S. is aware of this and we plan to revise the smartgcal design at some point in the future. The same is true for GMOS btw.)


bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/flamingos2/Flamingos2.java
```diff
@@ -1259,6 +1279,9 @@ else if (v != null)
         v = Pio.getValue(paramSet, FPU_MASK_PROP);
         if (v != null) setFpuCustomMask(v);
 
+        v = Pio.getValue(paramSet, CUSTOM_SLIT_WIDTH_PROP);
+        if (v != null) setCustomSlitWidth(CustomSlitWidth.valueOf(v));
```
cquiroz: Do you need to define a default if there is none in the serialized value?

fnussber: In that case the default `Unknown` set during object initialization will be used. This covers the case of old programs that don't have this value.

cquiroz:
Looks good :+1:

PR #617, opened 2015-10-27
REL-2511: Fix error when selecting F2 pinhole masks

fnussber:
Tiny PR to fix an issue where a match error was encountered because of an incomplete match statement.


cquiroz:
:+1:

PR #618, opened 2015-10-28
OCSADV-374: ITC spectroscopy saturation warnings

fnussber:
This PR unifies and makes more generic the mechanism to create warnings for ITC calculations. This saves some code, will make it easier to add warnings in the future and also extends the current saturation and linearity limit warnings to spectroscopy calculations (they currently exist only for imaging calculations).


bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/ItcService.scala
```diff
@@ -76,13 +87,29 @@ final case class ItcSpectroscopyResult(charts: List[SpcChartData], warnings: Lis
...
+  }
+
 }
 
 object ItcSpectroscopyResult {
 
   // java compatibility
-  def apply(charts: java.util.List[SpcChartData], warnings: java.util.List[ItcWarning]) =
-    new ItcSpectroscopyResult(charts.toList, warnings.toList)
+  def apply(charts: java.util.List[SpcChartData], warnings: List[ItcWarning]) =
```
cquiroz: Funny that for a java compatible method there is both java and scala `List`

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/base/Instrument.java
```diff
@@ -258,6 +260,20 @@ public static String getSuffix() {
     public List<TransmissionElement> getComponents() {
         return new ArrayList<>(components);
     }
+
```
cquiroz: Do you think it makes sense to some day merge these Instruments with the main Instrument model in the OT?

cquiroz:
Looks fine :+1:

bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/ItcService.scala
fnussber: Right, the second parameter is created with Scala code, however the method is called from Java, so that's where the mix comes from. With ITC I am following very much a step-by-step clean-up and migrate to Scala where doable/useful strategy, which leads to quite a mix of Scala/Java code in some places. This will get better over time..

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/base/Instrument.java
fnussber: In particular this part of the Instrument code which represents a series of `TransmissionElement`s (i.e. any optical elements in the light path like filters, gratings etc) is very ITC specific and will stay with the ITC code. However the instruments still contain some code which should be moved to or replaced by code from the main model. This is an ongoing process and I would say I've already reduced the "instrument" code in ITC by about 50%.

PR #619, opened 2015-10-28
OCSADV-461: BAGS exception

sraaphorst:
Fixed an issue with the way new primary targets were chosen when a BAGS target exists and a manual primary target was deleted.


cquiroz:
:+1:

PR #623, opened 2015-10-28
REL-2368: Update QVis to support FT observations on queue nights

fnussber:
The rules for scheduling Fast Turnaround programs have changed: Originally they were meant to be scheduled on specific FT nights only but recently it has been decided that they should be scheduled like additional queue programs on any nights that are available for queue observing.

This means that the special Fast Turnaround constraints in QVis have become obsolete and can be removed unceremoniously, which is what this PR is doing.

### DIFF
--- a/bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/filter/ui/FilterElement.scala
+++ b/bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/filter/ui/FilterElement.scala
@@ -336,13 +336,14 @@ object FilterElement {
     }
 
     def filter = init match {
-      case IsActive(_) => IsActive(selection)
-      case IsCompleted(_) => IsCompleted(selection)
-      case IsRollover(_) => IsRollover(selection)
-      case HasTimingConstraints(_) => HasTimingConstraints(selection)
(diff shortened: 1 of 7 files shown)


bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/filter/ui/FilterElement.scala
```diff
@@ -336,13 +336,14 @@ object FilterElement {
...
+      case IsActive(_)                => IsActive(selection)
+      case IsCompleted(_)             => IsCompleted(selection)
+      case IsRollover(_)              => IsRollover(selection)
+      case HasTimingConstraints(_)    => HasTimingConstraints(selection)
       case HasElevationConstraints(_) => HasElevationConstraints(selection)
-      case HasPreImaging(_) => HasPreImaging(selection)
-      case IsNonSidereal(_) => IsNonSidereal(selection)
+      case HasPreImaging(_)           => HasPreImaging(selection)
+      case IsNonSidereal(_)           => IsNonSidereal(selection)
+      case HasDummyTarget(_)          => HasDummyTarget(selection)
```
fnussber: This match statement was incomplete.

tpolecat:
:+1:

PR #625, opened 2015-10-29
REL-2512: Overwrite old motion values on Simbad query

cquiroz:
Simple PR to reset PM, RV and Z if they don't come on a simbad query


fnussber:
:+1:

PR #627, opened 2015-10-30
OCSADV-376 / 377: ITC show peak e- full well filling % and ADU

fnussber:
Addition of separate columns that display the full well filling % and ADU value for ITC calculations. These values show how saturated (or un-saturated) the observations are going to be and allow to fine tune the observation exposure time.

This screenshot shows an example for GMOS imaging with its three separate CCDs:

image

And another example for F2 spectroscopy:

image


cquiroz:
Nicely documented :+1:

PR #628, opened 2015-10-30
REL-2359: Add Daytime Calibration option to QPT

fnussber:
Currently the QPT only allows to schedule night time calibrations. The aim of this PR is to allow QCs to schedule daytime calibrations as part of the nightly plans if needed.

This PR also deals with a serialization issue that stems from the use of `NonEmptyList` in `BandsList`. `NonEmptyList` is not serializable and breaks the communication between the ODB and the QPT. I am planning to add a serialization test case to cover this, but for now my main goal is to push the functionality into the test release. @cquiroz : you probably want to have a quick look at the changes I made to `BandsList`.

### DIFF
--- a/bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/BandsList.scala
+++ b/bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/BandsList.scala
@@ -1,7 +1,5 @@
 package edu.gemini.spModel.core
 
-import edu.gemini.spModel.core.SiderealTarget
-
 import scalaz._
 import Scalaz._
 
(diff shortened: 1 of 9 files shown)


bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/BandsList.scala
```diff
@@ -10,9 +8,9 @@ import Scalaz._
  * It is used, e.g. to extract a magnitude from a target
  */
 sealed trait BandsList {
-  val bands: NonEmptyList[MagnitudeBand]
-  def extract(t: SiderealTarget) = bands.map(t.magnitudeIn).list.find(_.isDefined).flatten
-  def bandSupported(b: MagnitudeBand) = bands.list.contains(b)
+  val bands: List[MagnitudeBand]
```
cquiroz: The reason this was a `NEL` is to indicate that it cannot be empty and I'd think that is a good thing. On the other hand this is the simplest solution to solve the serialization problem so this fix is ok
I have to wonder though how this class ends up being part of the model, it seems a bit removed to require serialization

cquiroz:
Looks fine, are the AGS tests running ok?

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/BandsList.scala
fnussber: Agreed that `NEL`s are a good thing conceptually, but the fact that they're not serializable seems a bit of a design flaw(?). The `Obs` objects in the `MiniModel` contain `AgsAnalysis` objects which in turn contain `BandsList`s. It seems this was added in order to allow selecting observations with certain guide star properties; @sraaphorst probably knows the details.

fnussber:
Rebased and tests work. @cquiroz : if you're ok with that I am going to merge this one.

cquiroz:
Sure

PR #629, opened 2015-10-30
REL-2514: Update Simbad query to include flux system

cquiroz:
Simbad just updated their VoTable output at our request to include the magnitude system. This handles some special cases where the magnitude system does not match the default one

screenshot 2015-10-30 10 22 44


fnussber:
:+1:

PR #630, opened 2015-10-30
Fix for weird compilation error

cquiroz:
Strange compilation error when referencing `CoordinateTypes.Epoch` from `ModelConverters`
It is fixed by adding an explicit import of the Units class, maybe it is a compiler bug?


tpolecat:
:+1:

PR #631, opened 2015-10-30
OCSADV-443: BagsResult implementation

sraaphorst:
Representation of Bags target in `GuideProbeTargets` changed from `Option<SPTarget>` to multi-valued `BagsResult` to allow for three states as required for BAGS feedback.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/BagsResult.scala
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/BagsResult.scala
@@ -12,7 +12,7 @@ sealed trait BagsResult extends Cloneable {
   override def clone: BagsResult = this
 
   override def equals(obj: scala.Any): Boolean = obj match {
-    case other: BagsResult => target.equals(other.target)
+    case other: BagsResult => target.equals(other.target) && id.equals(other.id)
     case _                 => false
   }
 
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/BagsResult.scala
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/BagsResult.scala
@@ -11,11 +11,6 @@ sealed trait BagsResult extends Cloneable {
   def targetAsJava: GOption[SPTarget] = target.asGeminiOpt
   override def clone: BagsResult = this
 
-  override def equals(obj: scala.Any): Boolean = obj match {
-    case other: BagsResult => target.equals(other.target) && id.equals(other.id)
-    case _                 => false
-  }
-
   def getParamSet(factory: PioFactory): ParamSet = {
     val paramSet = factory.createParamSet(BagsResult.BagsResultParamSetName)
     Pio.addParam(factory, paramSet, BagsResult.BagsResultParamIdName, id)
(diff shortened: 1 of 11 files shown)


bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/BagsResult.scala
```diff
@@ -0,0 +1,70 @@
...
+
+import edu.gemini.shared.util.immutable.{Option => GOption}
+import edu.gemini.spModel.obs.context.ObsContext
+import edu.gemini.spModel.pio.{Pio, PioFactory, ParamSet}
+import edu.gemini.spModel.target.SPTarget
+import edu.gemini.spModel.rich.shared.immutable._
+
+sealed trait BagsResult extends Cloneable {
+  val id: String
+  val observationHash: String = ""
```
cquiroz: What's the role of `observationHash`? It doesn't seem to be overidden on the derived classes but it is used in equals

sraaphorst: Sorry. My bad. That was supposed to have been removed from this. I'm going to do that and push another commit.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/BagsResult.scala
```diff
@@ -0,0 +1,68 @@
...
+import edu.gemini.spModel.rich.shared.immutable._
+
+sealed trait BagsResult extends Cloneable {
+  val id: String
+  val target: Option[SPTarget] = None
+  def targetAsJava: GOption[SPTarget] = target.asGeminiOpt
+  override def clone: BagsResult = this
+
+  override def equals(obj: scala.Any): Boolean = obj match {
+    case other: BagsResult => target.equals(other.target)
```
cquiroz: This means that e.g. `NoSearchPerformed == NoTargetFound` is true

sraaphorst: Hmmm... good point. Not sure what the easiest way is to fix this with case classes of a sealed trait where I want to compare data members.
I can compare `id`s or something more elaborate.

cquiroz: Could you move the equals to only `WithTarget`?

sraaphorst: Of course. Ha! Thanks!

cquiroz: Actually the default `equals` of the case class should suffice right?

sraaphorst: Hmmm... now that you mention it, my guess is that it will, but I could not quite remember how equality is implemented.

I'm also not sure if, as this progresses, we'll want to compare the SPTargets or the underlying ITargets to avoid comparing watchers, but I'm going to take it out completely for now until the need to do something fancier than the default becomes apparent.

cquiroz:
Looks fine :+1:

PR #632, opened 2015-10-30
REL-2323: Altair + GMOS-N AGS strategies

sraaphorst:
This PR adds multi-probe strategies to GMOS-N when Altair is being used to allow for flexure targets.
There are two cases in particular that are needed as per REL-2323: Altair + GMOS OI, and PWFS1 + GMOS OI.
To accomplish this, I created a somewhat rudimentary `MultiProbeStrategy` class that has some limitations as described in the comments.

Part of the PR requested some minor changes to `AltairForm` with radio button label changes and tool tip text added, and I used this as an opportunity to clean up the code somewhat.

### DIFF
--- a/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/MultiProbeStrategy.scala
+++ b/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/MultiProbeStrategy.scala
@@ -13,7 +13,16 @@ import scala.concurrent.Future
 
 import scala.concurrent.ExecutionContext.Implicits.global
 
-
+/**
+ * Represents an AGS strategy consisting of multiple substrategies.
+ *
+ * There are several inherent shortcomings with this:
+ * 1. select requires a position angle, and querying each substrategy may return a number of position angles.
+ *    In the current implementation, the position angle for the first strategy is returned.
+ * 2. probeBands must return one of a fixed number of BandsLists. We currently try to find a common one, or one
+ *    that is a subset of all of those of the substrategies. If that is not possible, we try to find a single band
+ *    that is a subset of those of the substrategies. If that fails, we just return R.
+ */
 case class MultiProbeStrategy(key: AgsStrategyKey, strategies: List[AgsStrategy]) extends AgsStrategy {
   override def magnitudes(ctx: ObsContext, mt: MagnitudeTable): List[(GuideProbe, AgsMagnitude.MagnitudeCalc)] =
     strategies.flatMap(_.magnitudes(ctx, mt))
@@ -52,8 +61,8 @@ case class MultiProbeStrategy(key: AgsStrategyKey, strategies: List[AgsStrategy]
       val bLSet = Set(bL.bands.list)
       bandsLL.forall(bL2 => bLSet.subsetOf(Set(bL2.bands.list)))
     }).getOrElse {
-      // TODO: We must have A band, at least, that is a subset of all the bands.
-      // TODO: Otherwise bork and return R bands list. Not sure what else to do here.
+      // We must have a band, at least, that is a subset of all the bands.
+      // Otherwise bork and return R bands list. Not sure what else to do here.
       val bandInt = bandsLL.map(bL => bL.bands.list.toSet).reduce((s1,s2) => s1.intersect(s2))
       bandInt.headOption.map(SingleBand).getOrElse(RBandsList)
     }
--- a/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/MultiProbeStrategy.scala
+++ b/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/MultiProbeStrategy.scala
@@ -5,7 +5,7 @@ import edu.gemini.ags.api.AgsStrategy.Estimate
 import edu.gemini.ags.api.{AgsAnalysis, AgsMagnitude, AgsStrategy}
 import edu.gemini.catalog.api.CatalogQuery
 import edu.gemini.spModel.ags.AgsStrategyKey
-import edu.gemini.spModel.core.{SingleBand, RBandsList, SiderealTarget}
+import edu.gemini.spModel.core.{RBandsList, SiderealTarget}
 import edu.gemini.spModel.guide.{ValidatableGuideProbe, GuideProbe}
 import edu.gemini.spModel.obs.context.ObsContext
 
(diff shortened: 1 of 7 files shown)


bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/MultiProbeStrategy.scala
```diff
@@ -0,0 +1,70 @@
...
+ * Represents an AGS strategy consisting of multiple substrategies.
+ *
+ * There are several inherent shortcomings with this:
+ * 1. select requires a position angle, and querying each substrategy may return a number of position angles.
+ *    In the current implementation, the position angle for the first strategy is returned.
+ * 2. probeBands must return one of a fixed number of BandsLists. We currently try to find a common one, or one
+ *    that is a subset of all of those of the substrategies. If that is not possible, we try to find a single band
+ *    that is a subset of those of the substrategies. If that fails, we just return R.
+ */
+case class MultiProbeStrategy(key: AgsStrategyKey, strategies: List[AgsStrategy]) extends AgsStrategy {
```
cquiroz: I fear this class :smile: I think it should be thoroughly tested like `ScienceProbeStrategy`

Some things that worry me:
* What does it mean to do an analyze? Is it per probe?
* catalog queries support multiple queries but it is just to accomodate GeMS and I was planning to remove it as you can do GeMS with a single query
* I think `probeBands` should be a superset of bands rather than a subset, but I'm ready to be convinced otherwise

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/MultiProbeStrategy.scala
```diff
@@ -0,0 +1,70 @@
...
+    strategies.flatMap(_.analyze(ctx, mt))
+
+  override def analyze(ctx: ObsContext, mt: MagnitudeTable, guideProbe: ValidatableGuideProbe, guideStar: SiderealTarget): Option[AgsAnalysis] =
+    AgsAnalysis.analysis(ctx, mt, guideProbe, guideStar, probeBands)
+
+  override def catalogQueries(ctx: ObsContext, mt: MagnitudeTable): List[CatalogQuery] =
+    strategies.flatMap(_.catalogQueries(ctx, mt))
+
+  override def candidates(ctx: ObsContext, mt: MagnitudeTable): Future[List[(GuideProbe, List[SiderealTarget])]] =
+    Future.sequence(strategies.map(_.candidates(ctx, mt))).map(_.flatten)
```
tpolecat: FYI `.map` and then `.sequence` can be rewritten as `.traverse`. Would be nice if stdlib had `.traverseM` which would let us get rid of the `.flatten` as well.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/MultiProbeStrategy.scala
```diff
@@ -0,0 +1,70 @@
...
+import scala.concurrent.Future
+
+import scala.concurrent.ExecutionContext.Implicits.global
+
+/**
+ * Represents an AGS strategy consisting of multiple substrategies.
+ *
+ * There are several inherent shortcomings with this:
+ * 1. select requires a position angle, and querying each substrategy may return a number of position angles.
+ *    In the current implementation, the position angle for the first strategy is returned.
```
tpolecat: I'm not clear on why this is ok. Can you say some more words about this?

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/MultiProbeStrategy.scala
```diff
@@ -0,0 +1,70 @@
...
+
+  override val probeBands = {
+    val bandsLL = strategies.map(_.probeBands)
+
+    // Find a BandsList that is a subset of the others.
+    bandsLL.find(bL => {
+      val bLSet = Set(bL.bands.list)
+      bandsLL.forall(bL2 => bLSet.subsetOf(Set(bL2.bands.list)))
+    }).getOrElse {
+      // We must have a band, at least, that is a subset of all the bands.
```
tpolecat: If you use braces instead of parens you can avoid the nesting.

```scala
bandsLL.find { bL => 
  val bLSet = Set(bL.bands.list)
  bandsLL.forall(bL2 => bLSet.subsetOf(Set(bL2.bands.list)))
 } getOrElse { ...
```

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/MultiProbeStrategy.scala
sraaphorst: I am changing this to make `MultiProbeStrategy` a sealed abstract class, and removing `probeBands` from it to require that they be specified directly in subclasses (but yes, I agree that this should have been a superset and not a subset). In all the cases where we are using this (GMOS-N OI + Altair and GMOS-N OI + PWFS1), it is always going to be the R-bands list anyway, so I don't think we should worry about doing any calculation to determine this and risk errors / failure in finding an appropriate band list.

I'm not sure what you mean about catalog queries supporting multiple queries and having this removed: will there be a problem if I perform a query for each of the AGS strategies combined in this strategy? In my testing in the OT, it seems to work right now and just combine the results, which is exactly what I want.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/MultiProbeStrategy.scala
sraaphorst: I'm changing this to be more explicit, as you'll see in my next commit in a few minutes.

Currently, the only cases we are interested in are those which involve guide probes where:
1. One of the guide probes is NOT symmetric with respect to rotations around the base / an offset position; and
2. A second guide probe IS symmetric with respect to rotations around the base / an offset position;
Thus, I *think* it makes sense to use the PA as returned by the first guide probe, unless I am possibly misunderstanding something that could come up with respect to offsets. (I am not sure - please tell me if this is the case.)

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/MultiProbeStrategy.scala
sraaphorst: I'll put together some tests today for my (pending, in a couple minutes) update to this PR.

PR #633, opened 2015-10-31
OCSADV-443: Fixed broken GPT PIO.

sraaphorst:
I had a fundamental misunderstanding about how PIO worked and thus thought I had tested it when it turns out that I hadn't.

This repairs the PIO break that was introduced by the changes to the data model for `GuideProbeTargets`, making things more robust at the same time. I have tested this manually, but it still requires a formal test case, which I will add later.
(I didn't want to leave this broken in the build.)


sraaphorst:
Actually, hold on. There may be (will be) an issue with backward compatibility. I am going to have to do one more commit. Sorry.

sraaphorst:
Alright. Backward compatibility has been restored and this is ready for review.

cquiroz:
:+1:

PR #634, opened 2015-10-31
OCSADV-355B: Allow input of numbers in exponential notation.

fnussber:
The `NumberBoxWidget` is the base for many of the UI elements that allow to enter numbers in the OT, in particular also the `NumericPropertySheet`. So far this widget didn't allow to enter numbers in scientific notation, e.g. `1e-15`.

This PR extends the `NumberBoxWidget` to allow entry of "incomplete" exponential numbers, this is needed because the numbers are validated after every single key stroke. In addition to this, the `NumberFormatter`s and `DecimalFormatter`s used for the `NumericPropertySheet` are modified so that they automatically default to exponential notation for very small and very big numbers. Using those formatters has also the advantage that they deal with corner cases like empty strings and - how convenient - incomplete exponential numbers (e.g. `1e` or `1e-`) by just ignoring the incomplete exponential part.

The way numbers are formatted (max number of fraction digits, limits for exponential notation) is somewhat arbitrary, so I expect this task to be reopened for fine adjustments after some testers have used it, but for now it solves the main issue of not being able to enter numbers in exponential notation.

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details/NumericPropertySheet.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details/NumericPropertySheet.scala
@@ -1,7 +1,7 @@
 package jsky.app.ot.gemini.editor.targetComponent.details
 
 import java.awt.{GridBagConstraints, GridBagLayout, Insets}
-import java.text.{ParseException, NumberFormat}
+import java.text.{DecimalFormat, NumberFormat, ParseException}
 import java.util.Locale
 import javax.swing.JPanel
 
@@ -16,8 +16,8 @@ import jsky.util.gui.{NumberBoxWidget, TextBoxWidget, TextBoxWidgetWatcher}
 import squants.{Quantity, UnitOfMeasure}
 
 import scala.swing.ListView.Renderer
-import scala.swing.event.SelectionChanged
 import scala.swing._
+import scala.swing.event.SelectionChanged
 import scalaz.std.list._
 import scalaz.syntax.functor._
 import scalaz.syntax.id._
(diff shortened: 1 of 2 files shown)


cquiroz:
`NumberBoxWidget` is used in many places, are we sure this change won't change the way numbers are displayed in unintended locations?
There is a Scala Version of this at `NumberField` Does it also need updating?

fnussber:
I've been wondering about side effects, too, so here is what I *think* is going to happen, but I might be overlooking something:
* `NumberBoxWidget` is used all over the place, but I don't think adding the capability to enter numbers in exponential notation should have any negative side effects (?). The formatting is dealt with by the the callers of `NumberBoxWidget`, so the way numbers are displayed in general is not impacted. As a side note: the way `NumberBoxWidget` deals with negative numbers is currently broken: enter a `-` and you can enter anything in that field; this PR fixes that behavior, too.

* The changes to the formatting will only impact the `NumericPropertySheet` class which is currently only used in the target editor. So the impact of this will be easily testable. Currently the formatting is going to change somewhat, e.g. currently "0.0001" will be turned into "1.0E-4", while the new formatting allows for more digits after the decimal place before resorting to exponential notation, but that can easily be changed, if needed. I think here we need Andy's and Bryan's feedback.
 
* Finally, `NumberField` is mostly used in the PIT and in a couple of places in the OT and comes with its own formatter. For now I would not change this, but we may need to revisit this in case we add capabilities to enter source definitions to the PIT, which (I think) is currently not the case.

cquiroz:
I did a similar change for the `redshift` field, supporting a numeric field that changes depending on the selection. could you verify that your change doesn't break it

fnussber:
The `redshift` field uses its own formatters, I did not change those, values will still be displayed as before.

cquiroz:
Ok, besides the risk of something missformatted I think this looks good, hopefully any issues about formatting will surface soon

PR #635, opened 2015-11-01
REL-2432: Update the GMOS-S Hamamatsu gain and read noise in the OT

arturog8m:
A tiny PR to update the GMOS-S gain & read noise values after work done on the Hamamatsu detectors. Added a simple test to confirm the specs for GN & GS are correct.


cquiroz:
:+1:

fnussber:
Do the tests run? This will probably break the ITC tests because the calculation results will slightly change.

arturog8m:
Ops. I didn't check the ITC test. I'll do that, thanks for pointing it out

arturog8m:
Right, ITCs tests are broken. How do we update the ITC tests baseline?

fnussber:
Execute `BaselineTest.create()` and replace the file `baseline.txt` in the test resources with the new one. From the doc in `BaselineTest`: 
`Execute create() in order to update the baseline. The file is written to the current output directory (e.g. ../ocs/app/itc/idea/out/test/edu.gemini.itc-2015001.6.1/baseline.txt when working with IntelliJ) and needs to be manually copied from there into the resources folder!`
Executing all the test (about 5000) will take 5 to 6 minutes. After replacing the `baseline.txt` file you can check the updated results by executing `BaselineAllSpec` which will execute a random subset of all those tests (about 10 per instrument if I remember correctly).

fnussber:
Here's the rough location the file can be found in assuming you're using IntelliJ and have the all_bundles project open: `file:///Users/fnussber/Projects/ocs/app/all-bundles/idea/out/test/edu.gemini.itc.web-2016001.1.1/baseline.txt`

arturog8m:
Thanks, got it. I couldn't run `BaselineTest.create()` from IntelliJ though - got an exception like `scala.collection.parallel.CompositeThrowable: Multiple exceptions thrown during a parallel computation....`, but worked fine from command line. I must have something misconfigured in Idea, but it's too late now to try to deal with it :). 

Added a new commit with the updated `baseline.txt`, now all ITC tests pass. Thanks for preventing me from breaking the build!

arturog8m:
And I need a faster computer. It took about 20m to run the tests on my computer!

fnussber:
*gosh* 20 minutes is indeed long. But hey, when I started working on ITC a few months ago executing all those tests took about an hour(!) on my machine :) In any case: :+1:  from the ITC command center.

PR #636, opened 2015-11-01
OCSADV-468 fix pluto conversion

tpolecat:
I was re-using the PIO `Params` defining Pluto's orbital elements, which you can't do ... adding the same `Param` to a second `ParamSet` removes it from wherever it was, so a program with more than one instance of Pluto ended up with the orbital elements only on the last instance.

Fixed, test updated.


bundle/edu.gemini.spModel.io/src/test/scala/edu/gemini/spModel/io/impl/migration/to2016A/PlutoDemotionTest.scala
```diff
@@ -12,7 +12,7 @@ import scala.collection.JavaConverters._
 class PlutoDemotionTest extends MigrationTest {
 
   @Test def testTemplateConversion(): Unit =
-    withTestProgram("pluto.xml", { (_,p) => validateProgram(p) })
+    withTestProgram("GN-2011B-Q-40.xml", { (_,p) => validateProgram(p) })
```
cquiroz: Excellent to have this explicitly tested

cquiroz:
:+1:

tpolecat:
I'm actually going to minimize this test; we shouldn't have a whole program checked in like this.

PR #637, opened 2015-11-01
OCSADV-468 fix pluto conversion

tpolecat:
I was re-using the PIO `Params` defining Pluto's orbital elements, which you can't do ... adding the same `Param` to a second `ParamSet` removes it from wherever it was, so a program with more than one instance of Pluto ended up with the orbital elements only on the last instance.

Fixed, test updated to include two instances of Pluto. Fails without change, works with.

Note: this is a re-do of an earlier PR, with a more minimal test setup.


tpolecat:
Rebased. Tests pass.

PR #638, opened 2015-11-02
OCSADV-473: NIRI linearity and saturation limits are switched

fnussber:
Trivial bug fix.


cquiroz:
Cannot comment too much on the numbers :smile: , so let's trust you :+1: 
Can this stuff be automatically tested?

fnussber:
The numbers *are* the tests. This is the baseline, i.e. hash values for the results/outputs for sets of configurations for each instruments. This doesn't allow to test single bits and pieces but works so far very well for reliable regression testing. Now that some parts of the code are more decoupled I am thinking about adding more tests that allow more targeted testing but I haven't spent time on that yet.

PR #639, opened 2015-11-02
OCSADV-475: BAGS disable and RA / Dec restore.

sraaphorst:
This temporarily turns off BAGS by commenting out the `watch` and `unwatch` methods for programs so that they are never registered for automatic lookups.

Furthermore, the behavior of the RA and Dec text widgets in `CoordinateEditor` have been restored to previous behavior (with changes commented out - we can decide whether or not to remove these when we have a solution to the precision and updating issues).


cquiroz:
:+1:

PR #641, opened 2015-11-02
REL-2506

sraaphorst:
Take two on this PR.

I forgot initially to also change the brightness limit with respect to the faintness limit, as for CWFS, they should always be 7.5 apart.

This is a simple change that changes the faintness limit for CWFS with respect to the REL task, and most of the work done here is in modifying the test cases to similarly reflect that change.

There is one result that seems strange to me below. @cquiroz - if you could take a quick look, as well as possibly @tpolecat or @swalker2m I would very much appreciate it, because it seems like it shouldn't be happening and may reveal a bug somewhere, or it might just be a misunderstanding on my part.

I will put a line note at the relevant area.


bundle/edu.gemini.ags/src/test/scala/edu/gemini/ags/impl/GemsStrategySpec.scala
```diff
@@ -381,9 +417,9 @@ class GemsStrategySpec extends Specification with NoTimeConversions {
       val newCtx = selection.map(applySelection(ctx, _)).getOrElse(ctx)
       val analysis = gemsStrategy.analyze(newCtx, ProbeLimitsTable.loadOrThrow())
       analysis.collect {
-        case AgsAnalysis.Usable(Canopus.Wfs.cwfs1, st, GuideSpeed.FAST, AgsGuideQuality.DeliversRequestedIq, _) if st.some == cwfs1    => Canopus.Wfs.cwfs1
+        case AgsAnalysis.Usable(Canopus.Wfs.cwfs1, st, GuideSpeed.FAST, AgsGuideQuality.DeliversRequestedIq, _) if st.some == cwfs1 => Canopus.Wfs.cwfs1
         case AgsAnalysis.Usable(Canopus.Wfs.cwfs2, st, GuideSpeed.FAST, AgsGuideQuality.DeliversRequestedIq, _) if st.some == cwfs2 => Canopus.Wfs.cwfs2
-        case AgsAnalysis.Usable(Canopus.Wfs.cwfs3, st, GuideSpeed.FAST, AgsGuideQuality.DeliversRequestedIq, _) if st.some == cwfs3 => Canopus.Wfs.cwfs3
+        case AgsAnalysis.NotReachable(Canopus.Wfs.cwfs3, st, _) if st.some == cwfs3                                               => Canopus.Wfs.cwfs3
```
sraaphorst: This is the relevant part of the code as mentioned in the PR description.

I am not sure why a Gems strategy should be returning a guide star that has an AGS analysis of `NotReachable`. This seems incorrect to me. Or am I missing something?

bundle/edu.gemini.ags/src/test/scala/edu/gemini/ags/gems/GemsResultsAnalyzerSpec.scala
```diff
@@ -65,7 +65,7 @@ class GemsResultsAnalyzerSpec extends MascotProgress with Specification with NoT
       }
 
       System.out.println("gems results: size = " + gemsGuideStars.size)
-      gemsGuideStars should have size 164
+      gemsGuideStars should have size 247
```
cquiroz: You expect this number to grow when the limits are wider, so this points in the right direction, the bad side is the mascot should be slower given we have more targets

bundle/edu.gemini.ags/src/test/scala/edu/gemini/ags/impl/GemsStrategySpec.scala
cquiroz: I think it is possible, though you could think that the strategy may just not return a star if it is unreachable

cquiroz:
I'm sorry that you had to confront this, the tests are not very good in the sense that we don't have a good way to independently find what stars would be selected. They are good at detecting regressions though
:+1:

sraaphorst:
No worries! And thanks, @cquiroz for the comments.

PR #642, opened 2015-11-02
OCSADV-354B: Change string and add tooltip for undefined user SEDs.

fnussber:
Changes to a string, addition of a tooltip and a small refactoring.

### DIFF
--- a/bundle/edu.gemini.itc/src/main/scala/edu/gemini/itc/service/ItcServiceImpl.scala
+++ b/bundle/edu.gemini.itc/src/main/scala/edu/gemini/itc/service/ItcServiceImpl.scala
@@ -9,7 +9,6 @@ import edu.gemini.itc.gnirs.GnirsRecipe
 import edu.gemini.itc.gsaoi.GsaoiRecipe
 import edu.gemini.itc.nifs.NifsRecipe
 import edu.gemini.itc.niri.NiriRecipe
-import edu.gemini.itc.operation.ImagingS2NMethodACalculation
 import edu.gemini.itc.shared._
 import edu.gemini.spModel.core.{AuxFileSpectrum, UserDefinedSpectrum, SpectralDistribution, SPProgramID}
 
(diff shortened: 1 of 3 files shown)


bundle/edu.gemini.itc/src/main/scala/edu/gemini/itc/service/ItcServiceImpl.scala
```diff
@@ -28,25 +27,33 @@ class ItcServiceImpl extends ItcService {
...
-      val spectrum      = new String(spectrumBytes)
-      UserDefinedSpectrum(name, spectrum)
+    // update parameters sent from client with stuff that needs to be done on the server
+    val updatedParams: ItcParameters = {
+
+      // Get the SED data from an aux file. For now we can assume that the ITC service is running on the same
+      // machine as the database (localhost). In case this setup changes, we need to change this here, too.
+      def readAuxFile(id: String, name: String): SpectralDistribution = {
+        val programId     = SPProgramID.toProgramID(id)
+        val spectrumBytes = new AuxFileClient("localhost", 8443).fetchToMemory(programId, name)
```
cquiroz: This looks like a future source of bugs :smile:, but I don't know how you could make it configurable in a sensible way

cquiroz:
the code looks fine :+1:

bundle/edu.gemini.itc/src/main/scala/edu/gemini/itc/service/ItcServiceImpl.scala
fnussber: You mean localhost and port number? Yeah, that's a point. But making it configurable only helps if people remember to change the configuration.. pick your poison, I guess?

cquiroz: That's right, we'll likely forget either one

PR #643, opened 2015-11-03
OCSADV-480: Include percent well filling in ITC web output

fnussber:
This PR adds some ITC calculation results that are already displayed in the OT to the web app output.

I also refactored the `GmosPrinter` a tiny bit to simplify it and replaced two manually created saturation warnings for Michelle and TRecs with generic ones that are used by all the other instruments.


cquiroz:
:+1:

PR #645, opened 2015-11-04
OCSADV-482: Make BAGS target non-editable

sraaphorst:
This PR accomplishes the following:

1. Changes to `EdCompTargetList` to make BAGS targets non-editable.
2. Fix of `OtItemEditor.updateEnabledState`, which I believe was incorrect before. (If any call was made to this method with `false` to disable a widget, I believe that the widget could then never be made enabled with another call with `true`. I have tested and it appears to work correctly now, but if I misunderstood the original intent of this very limited use method, please let me know.)
3. Cleaned up `OtItemEditor`, using Java 8 lambda expressions and `Optional`.

### DIFF
--- a/bundle/jsky.app.ot/src/main/java/jsky/app/ot/editor/OtItemEditor.java
+++ b/bundle/jsky.app.ot/src/main/java/jsky/app/ot/editor/OtItemEditor.java
@@ -24,9 +24,8 @@ import javax.swing.*;
 import javax.swing.plaf.basic.BasicArrowButton;
 import java.awt.Component;
 import java.awt.Container;
-import java.util.HashSet;
-import java.util.Hashtable;
-import java.util.Set;
+import java.util.*;
+import java.util.function.Predicate;
 
 
 /**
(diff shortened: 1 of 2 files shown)


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -60,12 +60,12 @@ final class BagsManager(executor: ScheduledThreadPoolExecutor) {
    * consideration for a BAGS lookup.
    */
   def watch(prog: ISPProgram): Unit = {
-    //    synchronized {
-    //      state += prog.getProgramID
-    //      prog.addStructureChangeListener(StructurePropertyChangeListener)
-    //      prog.addCompositeChangeListener(CompositePropertyChangeListener)
-    //    }
-    //    prog.getAllObservations.asScala.foreach(enqueue(_, 0L, initialEnqueue = true))
+        synchronized {
```
cquiroz: wasn't bags enabled in a previous PR?

sraaphorst: Yes, but it is as of yet unmerged, but the changes are the exact same between the two PRs (and I needed BAGS enabled for testing for this), so I suspect there will not be conflicts?

cquiroz: If there are conflicts they should be easy to fix

cquiroz:
:+1:

PR #646, opened 2015-11-04
REL-2517: update program container version

swalker2m:
The 2016A migration code will always run, even in the future, if we don't update the program container version.

````scala
object To2016A extends Migration {

  val Version_2016A = Version.`match`("2016A-1")

  def isPre2016A(c: Container): Boolean =
    c.getVersion.compareTo(Version_2016A) < 0

  // Entry point here
  def updateProgram(d: Document): Unit =
    d.containers.find(_.getKind == SpIOTags.PROGRAM).filter(isPre2016A).foreach { _ =>
      conversions.foreach(_.apply(d))
    }

````


arturog8m:
:+1:

sraaphorst:
:+1:

PR #647, opened 2015-11-05
REL-2143: Copied GMOS-S E2V observations have wrong CCD manufacturer

fnussber:
From 2016A onwards it is not allowed anymore to use GMOS-S with E2V detectors and PIs can not set the detector type to E2V. However it is possible to copy observations from previous semesters which have the detector set to E2V.

Adding some code to avoid this on copying seemed to be too complicated and too prone to introduce unwanted side-effects for something that is actually only a transitional problem. Therefore Andy and I decided to instead add a P2 error for GMOS-S observations in programs for semester 2016A or later that have the detector set to E2V.

I also added some test cases and hope that the new 'RuleSpec' abstract base class (modeled after `WfsRuleSpec`) and `GmosSpec` prove useful for other people implementing rule tests in the future. The current Java based tests in `GmosRuleTest` are very unwieldy to use.

### DIFF
--- a/bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/gmos/GmosRule.java
+++ b/bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/gmos/GmosRule.java
@@ -1275,7 +1275,7 @@ public final class GmosRule implements IRule {
                         // apply if E2V is selected
                         final DetectorManufacturer ccd = ((InstGmosSouth) instrument).getDetectorManufacturer();
                         if (ccd == DetectorManufacturer.E2V) {
-                            return new Problem(ERROR, PREFIX + "POST_2015B_GMOS_S_WITH_E2V", msg, SequenceRule.getInstrumentOrSequenceNode(step, elems));
+                            return new Problem(ERROR, PREFIX + "POST_2015B_GMOS_S_WITH_E2V_RULE", msg, SequenceRule.getInstrumentOrSequenceNode(step, elems));
                         }
                     }
                 }
--- a/bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/gmos/GmosRule.java
+++ b/bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/gmos/GmosRule.java
@@ -1266,7 +1266,7 @@ public final class GmosRule implements IRule {
         public Problem check(Config config, int step, ObservationElements elems, Object state) {
             // apply this rule to GMOS-S
             final SPInstObsComp instrument = elems.getInstrument();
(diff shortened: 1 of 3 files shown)


bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/gmos/GmosRule.java
```diff
@@ -1256,6 +1242,48 @@ public Problem check(Config config, int step, ObservationElements elems, Object
         }
     }
 
+    /**
+     * ERROR: If GMOS-S is used with E2V after semester 2015B.
+     * This can happen when PIs copy parts of old programs into new ones. PIs will not be able to change the detector
+     * settings themselves and will have to have a staff member help them. This is not ideal, but since this is only
```
cquiroz: Why is this the case, can't we let PIs change the setting?

bundle/edu.gemini.p2checker/src/test/scala/edu/gemini/p2checker/rules/RuleSpec.scala
```diff
@@ -0,0 +1,56 @@
...
+import edu.gemini.spModel.target.obsComp.TargetObsComp
+import org.specs2.mutable.Specification
+
+import scala.collection.JavaConverters._
+import scalaz.syntax.id._
+
+/**
+ * Generic test bed for arbitrary P2 checker rule tests.
+ */
+abstract class RuleSpec extends Specification {
```
cquiroz: This looks great, do we have any other tests for P2 rules?

cquiroz:
:+1:

bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/gmos/GmosRule.java
fnussber: I asked Andy the same question, he prefers to keep this setting disabled for PIs so that they can  not (accidentally) change it.

bundle/edu.gemini.p2checker/src/test/scala/edu/gemini/p2checker/rules/RuleSpec.scala
fnussber: Many, see `AbstractRuleTest`, `GeneralRuleTest` and `GmosRuleTest` for Java and then there are more written in Scala. I was tempted to start moving some from `GmosRuleTest`, I think with this approach we could probably have all those tests with about 1/4 of the code, but in the end there are only so many hours in a day.

cquiroz: you always have weekends :smile:

PR #648, opened 2015-11-06
OCSADV-472: Keep SED aux file selection from updating on single key strokes

fnussber:
OCSADV-472 describes some issues with BAGS updates. One of them is that the SED aux file selector for user defined source distributions updates its list of available files on every single key stroke. Needless to say that this wasn't such a brilliant idea.

Currently there is no way to get notified when aux files are changed (added/removed) and therefore the best way to deal with updating the list of available SED aux files seems to be to update it every time the user opens the combobox popup menu.

Because this update is triggered when the user clicks on the combobox to display the available values we need to produce those immediately and therefore the load operation that gets the aux file list from the server is done synchronously and will be aborted after 3 seconds.

I was also thinking about providing some `Publisher` based mechanism on top of the `AuxFileClient` and `AuxFileModel` but the advantage of the approach I followed in this PR is that it is contained in a single place and is less risky. Also, I did not want to start introducing something new; I think the whole event handling in the OT is a topic we need to have some broader discussion at some point.

Example of some user defined spectral distributions in the OT:

image

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details/SourceDetailsEditor.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details/SourceDetailsEditor.scala
@@ -1,14 +1,16 @@
 package jsky.app.ot.gemini.editor.targetComponent.details
 
 import java.awt.{Component, GridBagConstraints, Insets}
+import java.util.concurrent.TimeoutException
 import javax.swing.JPanel
+import javax.swing.event.{PopupMenuEvent, PopupMenuListener}
 
 import edu.gemini.auxfile.client.AuxFileClient
 import edu.gemini.auxfile.copier.AuxFileType
 import edu.gemini.pot.sp.ISPNode
 import edu.gemini.shared.util.immutable.{Option => GOption}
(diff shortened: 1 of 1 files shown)


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details/SourceDetailsEditor.scala
```diff
@@ -274,43 +258,106 @@ final class SourceDetailsEditor extends GridBagPanel with TelescopePosEditor wit
...
+      * when the combobox is expanded to show the selection, therefore if the result is not returned from
+      * the server in a short time, we simply ignore the result.
+      */
+    private def getAuxFiles(programId: SPProgramID): Set[AuxFileSpectrum] = {
+
+      try {
+        import scala.collection.JavaConversions._
+
+        // block UI for at most 3 seconds, if getting the aux file listing takes longer, the result will be ignored
+        Await.result(Future {
```
cquiroz: Can't this be made without blocking? Or is this necessary because of the `deafTo` and `listenTo` calls surrounding `getAuxFiles`?

fnussber: This is done when the user opens the dropdown menu of the combobox, basically the UI is waiting for the values to be displayed at this moment. If I do this non-blocking, the current values (which could be out of date) will be displayed while I am updating the values in the background. I agree this isn't great, but I couldn't come up with a better place to hook up and get notified that it is necessary to update the combobox model.

PR #649, opened 2015-11-06
OCSADV-271: Support 'embiggen' feature

cquiroz:
In the old catalog severs the queries were widened to increase cache efficiency. This meant that when searching for nearby objects there was a good chance that the data would be already available locally
This PR brings this capability into the new catalogs. Queries are widened 150% with up to 10 arcmins to avoid covering too large of a sky swath

### DIFF
--- a/bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableClient.scala
+++ b/bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableClient.scala
@@ -4,7 +4,7 @@ import java.net.{URL, UnknownHostException}
 import java.util.concurrent.atomic.AtomicInteger
 import java.util.logging.Logger
 
-import edu.gemini.catalog.api.{NameCatalogQuery, ConeSearchCatalogQuery, CatalogQuery}
+import edu.gemini.catalog.api.{RadiusConstraint, NameCatalogQuery, ConeSearchCatalogQuery, CatalogQuery}
 import edu.gemini.spModel.core.Angle
 import edu.gemini.spModel.core.SiderealTarget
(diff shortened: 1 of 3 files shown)


swalker2m:
Is this widening transparent to the caller?  Or does the caller have to filter out results that only match the wider query?

cquiroz:
It is transparent, as there is a filtering stage after we get the results and that uses the original, un-widened, range. That filter comes with the client's request and it is used to filter the results from the server or the cache

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableClient.scala
```diff
@@ -103,7 +105,9 @@ trait CachedBackend extends VoTableBackend {
       def go(pos: Int):Option[(Int, QueryResult)] =
         a.lift(pos) match {
           case None                                                            => None
-          case Some(CacheEntry(SearchKey(query:ConeSearchCatalogQuery, r), v)) => if (query.isSuperSetOf(k.query)) Some((pos, v)) else go(pos + 1)
+          case Some(CacheEntry(SearchKey(query:ConeSearchCatalogQuery, _), v)) =>
+            // Note we need to compare against the widened query
+            if (widen(query).isSuperSetOf(k.query)) Some((pos, v)) else go(pos + 1)
           case _                                                               => None // Not caching named queries so far
```
swalker2m: I suppose you could collapse the two branches that return None.  In fact, maybe you can

````scala
a.lift(pos).collect {
  case Some(CacheEntry ...) => ..
}
````

swalker2m:
:+1:

cquiroz:
Added suggested fix and tests are passing

PR #650, opened 2015-11-06
REL-2519: Update Flamingos2 filter transmission data

fnussber:
Not much to see here. This PR just updates the ITC data files that describe the transmission characteristics of a bunch of F2 filters. It also includes an update to the regression test baseline.


swalker2m:
:+1:

tpolecat:
The subject line in the email was truncated to **Update Flaming...** and it was fun to imagine what the rest of that sentence might be.

PR #652, opened 2015-11-07
REL-2359B: Add Daytime Calibration option to QPT

fnussber:
This is a follow up on an earlier PR which added UI capabilities for filtering/selecting daytime calibrations in the QPT. So far daytime calibrations didn't really make it through the filtering mechanisms of the `ObsQueryFunctor` because their step count was set to 0 in the `PlannedTimeSummary`.

This PR has two commits, the first one contains some trivial code brush up, the relevant changes are in 
https://github.com/gemini-hlsw/ocs/commit/69f44692afb0140e861f3d3090c59f584f6ff43f :

I am not sure what the reason was (is) to treat Acq and Daytime Cals the way it was in `PlannedTimeSummary` but trying to analyze the code I think that removing this special treatment for `DAY_CAL` obs types should be ok, because:

1) The planned steps are only used in the `ObsQueryFunctor` to exclude observations without planned steps to show up in the QPT, which is exactly the behavior we are trying to change (see `PlannedTimeSummaryService::getPlannedSteps` and `SPObsCache::getPlannedSteps` respectively).

2) The `PlannedTimeSummary` returned from `getTotalTime()` is used in different places, however since `PlannedTimeCalculator::toPlannedStepSumamry` only takes `ChargeClass.Program` into account and `DAY_CAL` has `ChargeClass.NONCHARGED` I think the returned time will be 0 as it was before.

Therefore, I think this change is ok, but of course I may overlook something relevant here, in particular because I don't know all the details where these times are relevant. (Tests are working, and I think there are quite a few tests that exercise the services  that deal with calculating times.)

### DIFF
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/obs/plannedtime/PlannedTimeSummaryService.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/obs/plannedtime/PlannedTimeSummaryService.java
@@ -1,12 +1,5 @@
-// Copyright 2001 Association for Universities for Research in Astronomy, Inc.,
-// Observatory Control System, Gemini Telescopes Project.
-// See the file LICENSE for complete details.
-//
-// $Id: PlannedTimeSummaryService.java 46768 2012-07-16 18:58:53Z rnorris $
-//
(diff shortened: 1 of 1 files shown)


cquiroz:
Looks good :+1:

PR #653, opened 2015-11-07
OCSADV-386: Bug fix for peak pixel for GNIRS cross dispersion

fnussber:
This PR looks much scarier than it actually is: it is basically a refactoring that unifies the result data objects for GNIRS ITC calculations with the ones of the other instruments, which allows it to share some common functionality to calculate the peak pixel signal which in turn fixes an issue with how this calculation is done now.

Folded into this PR is a bunch of dead code elimination in `SpecS2NVisitor` and `NiriRecipe`, and an update to some tooltips in `ItcTableModel`.


bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gnirs/GnirsRecipe.java
```diff
@@ -140,12 +138,12 @@ public GnirsSpectroscopyResult calculateSpectroscopy() {
                 _obsDetailParameters.getSkyApertureDiameter());
 
         if (instrument.XDisp_IsUsed()) {
-            final VisitableSampledSpectrum[] sedOrder = new VisitableSampledSpectrum[6];
+            final VisitableSampledSpectrum[] sedOrder = new VisitableSampledSpectrum[ORDERS];
```
cquiroz: That's a good change!

cquiroz:
Looks good, do you have tests related to these changes?

fnussber:
ITC is covered by a set of regression tests, that's why I often have to update the `baseline.txt` file when doing changes. At least that allows me to confirm that only those things change I expect to change. The (expected) changes then need to be verified by hand unfortunately. I hope to add some more tests in the future, the more I manage to streamline the old code (and understand it) the easier it will be to add meaningful low level tests.

cquiroz:
:+1:

PR #654, opened 2015-11-09
Support publishing p1 model to artifactory

cquiroz:
With this PR we are able to publish the bundles p1 and spModel to the internal artifactory repository directly from sbt. I couldn't enable it globally to publish all bundles but we need these 2 for ITAC


fnussber:
Nice to see artifactory being put to some use. :+1:

arturog8m:
I understand from this that we need Artifactory to be alive so we can build the ITAC software. Could you check that the existing ITAC documentation we have in the wiki actually refers to this and it is accurate?

tpolecat:
:+1:

PR #655, opened 2015-11-09
REL-2407-D: Initial Data Manager

swalker2m:
This preliminary PR introduces proposed changes for the Data Manager for 2016A.  The Data Manager was originally designed and implemented in 2005.  It had to work with the ODB over RMI and with the Gemini Science Archive one dataset at a time.  It moved and modified FITS files directly, sent emails to contact scientists and generally did a lot of work that is no longer necessary.  In 2016A we will be switching to an in-house archive service that takes over much of this work. For all of these reasons the original Data Manager has been replaced in its entirety.

The PR is preliminary because:

- The code is largely untested and lacking unit / property tests.
- Even if it were bug-free, it may not be quite what the science staff has in mind or may not work well with the new archive.

I will continue adding unit tests and bug fixes but soon we will need to begin user testing.

There are three commits.

1. Mostly removal of the original Data Manager.  Worth skipping.
2. Contains the starting point for the new Data Manager.
3. [User-oriented documentation](https://github.com/gemini-hlsw/ocs/blob/435f0589477627aba52344a30a5cec44de9973f4/bundle/edu.gemini.dataman.app/doc/Dataman.md) that nevertheless is worth reading first before looking at any of the code.  (I must warn that my editor reports that it scores “-47.98” for “Reading Ease” and has a “Fog Index” of “40.40”.  I don’t know what that means but it seems about right.)


cquiroz:
What's the editor that reports "Reading Ease" and "Fog index"?

swalker2m:
It's actually a markdown preview tool, not the editor itself.  It's called "Marked 2".  You edit the markdown file in whatever editor and it shows the styled version updated with each save.

andrewwstephens:
Nice!  The Gunning fog index estimates the years of formal education needed to understand text on a first reading:  https://en.wikipedia.org/wiki/Gunning_fog_index

cquiroz:
I have to wonder what will take you from "-47.98" Reading Ease to let's say "-48.03", seems awfully precise

tpolecat:
It's a lot of code but it looks good to me.

arturog8m:
Sorry for the conflicts... I hope they are easy to resolve.

cquiroz:
I read maybe half and it seems very good, A bit worried about the security on the GSA side

arturog8m:
The GSA security issue is a valid concern, but outside the scope of this PR.

arturog8m:
1,293 additions and 8,429 deletions. Just for this, I say :+1:

swalker2m:
Rebased, conflict resolved, and updated.  All the commits past the first three are new since this was last reviewed.  Also, here's a picture of the Observing Log GUI with a hypothetical error message for reference.

obslog

arturog8m:
Ship it. Good. :+1:

tpolecat:
New commits all look fine. :+1:

PR #656, opened 2015-11-09
OCSADV-453: BAGS feedback

sraaphorst:
This is the PR to allow for BAGS status information to be displayed in the `GuidingFeedback` area of the target environment (i.e. at the bottom where guide star quality is also displayed).

Changes here:
1. Now when a non-BAGS primary is selected, the BAGS target is cleared. When no primary is selected, BAGS runs.
2. Feedback about the status of BAGS is displayed as stated above.
3. BAGS sets the `BagsResult` properly to `NoTargetFound` when a catalog lookup turns up nothing. (This prevents BAGS from running on startup for observations where it previous failed to find results.)

I also reimplemented `GuidingFeedbackEditor` in scala (from Java) as it was a small class and this made things easier with interacting with other scala classes without the need of many `asGemini...` and `asJava...` etc. methods.


bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
```diff
@@ -110,6 +111,10 @@ public void keyPressed(KeyEvent event) {
 
         _agsPub.subscribe((obs, oldOptions, newOptions) -> updateGuiding());
 
+        BagsManager.instance().addBagsStatusListener((obs, oldStatus, newStatus) -> {
+            if (obs == getContextObservation())
```
cquiroz: Is it correct to use `==`, instead of `equals`?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -101,25 +102,46 @@ final class BagsManager(executor: ScheduledThreadPoolExecutor) {
...
-              // Otherwise construct an obs context, verify that it's bagworthy, and go
-              ObsContext.create(obs).asScalaOpt.filter(o => !initialEnqueue || o.isEligibleForBags).foreach { ctx =>
+        // If dequeue is false this means that (a) another task scheduled *after* me ended up
+        // running before me, so their result is as good as mine would have been and we're done;
+        // or (b) we don't care about that program anymore, so we're done.
+        if (dequeue(key, obs.getProgramID)) {
+          // Otherwise construct an obs context, verify that it's bagworthy, and go
+          ObsContext.create(obs).asScalaOpt.filter(obsCtxFilter).foreach { ctx =>
+            bagsStatus(BagsStatus.Pending)
+            executor.schedule(new Thread {
```
cquiroz: I thought you want to pass a `Runnable` to the executor rather than a `Thread` but I'm a bit rusty on this. What's the difference between passing directly a `Thread`?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -129,37 +151,44 @@ final class BagsManager(executor: ScheduledThreadPoolExecutor) {
...
+                          worker(TpeContext(obs), selOpt)
+                          obs.getProgram.addStructureChangeListener(StructurePropertyChangeListener)
+                          obs.getProgram.addCompositeChangeListener(CompositePropertyChangeListener)
                         }
+                        bagsClearStatus()
                       }
 
+                    // NOTE: We sleep for a brief duration BEFORE enqueue to allow the error message to be readable.
                     // We don't want to print the stack trace if the host is simply unreachable.
                     // This is reported only as a GenericError in a CatalogException, unfortunately.
```
cquiroz: Probably too late to improve this, but would you benefit by the catalog producing a more specific error if the host is unreachable?

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/GuideProbeTargets.java
```diff
@@ -366,6 +366,11 @@ public boolean primaryIsBagsTarget() {
     }
 
     /**
+     * Determine if the primary guide star exists but is not a BAGS target.
+     */
+    public boolean primaryIsManual() { return primaryTarget.exists(manualTargets::contains); }
```
swalker2m: What happens if BAGS runs but finds an already existing manual guide star?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -173,12 +202,16 @@ final class BagsManager(executor: ScheduledThreadPoolExecutor) {
                     if (Try {
                       val futDone = Await.ready(fut, 30.seconds)
```
cquiroz: I'm still unclear why are we mixing the Java executors with Scala futures, could we just use the Scala Future, doing the updates on a for each of the future? Or do you have to wait for completion for another reason?

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
sraaphorst: I think here we actually DO want the references to be the same, rather than do an in-depth comparison of the objects. The in-depth comparison would be costly, and it only applies if the references are the same.

cquiroz:
Can you put a screenshot on the visual changes to `GuidingFeedback`?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
sraaphorst: The reason I used a `Thread` instead of a `Runnable` is because a `Thread` allows for priority setting, which we wanted to ensure is used despite much discussion over whether or not it was relevant to make BAGS run at a lower priority and not interfere with the OT speed.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
sraaphorst: I think that could be useful, but at least for this first test release, I don't think we should worry about it just yet.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
cquiroz: I believe that when using an executor you can define a `Thread Factory` and that lets you select the priority or any other thread properties

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/GuideProbeTargets.java
sraaphorst: Oh, good point. That's not a test case I had thought of.

If there is a manual guide star that is already in the list and no primary, then BAGS would run and create a new BAGS entry for the same manual guide star. If there is a primary, BAGS won't run at all, so no issue.

Not sure the best way to handle this, to be honest: if we designate that star as the BAGS star, then it would no longer be a manual target, and additionally would no longer be editable, which is problematic.

The other option would be to have the manual star that BAGS would have selected be marked as the primary, but not designate it the BAGS star, which seems confusing.

Otherwise, this may require some more in-depth processing of allowing the BAGS star to be in the list of manual targets, and be editable if it is there. Seems a bit complicated to be implementing right now.

I'm frankly not quite sure the best way to handle this right off.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
sraaphorst: The problem here is that if we allow things to continue BEFORE the `Future` completes, the `Future` acts independently of the thread, and thus the thread ends and immediately a new thread is started by the executor without waiting for the previous catalog lookup to be done. Thus, things run as if we have a massive number of threads scheduled and bombard the catalog server.

I'm not sure if there is a better way to do this, unfortunately, as this is certainly not elegant.
For a for each of the future, will that suspend execution until the `Future` completes?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
sraaphorst: Nice. I didn't know about this. I'll change it back to `Runnable` and use a `ThreadFactory` instead.

sraaphorst:
Here is an example where a BAGS lookup is pending, i.e. a task has been approved for submission to the executor, but has not yet run. This is where the results will appear when the task is also running, and when it has failed.

Sorry for my hideous arrow drawing skills.
feedback

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/GuideProbeTargets.java
sraaphorst: I think the best thing to do is to leave this as is right now, i.e. have BAGS create a duplicate target, and discuss with Andy and Bryan what they think is best, unless anyone else has any ideas.

There doesn't seem to be a UI consistent, obvious way to handle it to me.

swalker2m: It seems like it there wouldn't be an issue with having a bags result that happens to be one of the manual options?  Could you flip the test?  That is,

primary is bags = primary exists and is equal to the bags result target (if any)
primary is manual = ! primary is bags

In fact, if I recall the fact that there is a bags result means it must be primary?

swalker2m:
It seems like we might be giving them conflicting messages?  It's red _and_ it's green at the same time?  If there is no bags result at all, just the "pending" message makes sense to me, say on a yellow warning background.  If there is a bags result but bags is currently running (because of an update), show the existing message but put a spinner or something in front to indicate activity?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -32,15 +32,13 @@ final class BagsManager(executor: ScheduledThreadPoolExecutor) {
...
-      }
+    // Check the primary guide group to see if there are any manual targets. If there are, do not run BAGS.
+    def hasManualPrimary: Boolean = {
+      ctx.getTargets.getGuideEnvironment.getPrimary.asScalaOpt.exists(_.getAll.asScalaList.exists(_.primaryIsManual))
+    }
+
+    def isMissingSearch: Boolean = {
+      ctx.getTargets.getGuideEnvironment.getPrimary.asScalaOpt.forall(_.getAll.asScalaList.exists(_.getBagsResult == BagsResult.NoSearchPerformed))
     }
   }
```
tpolecat: Do you see an opportunity for abstraction here?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -101,25 +102,45 @@ final class BagsManager(executor: ScheduledThreadPoolExecutor) {
...
+          val oldState = bagsStatuses.get(key)
+          bagsStatuses += ((key, status))
+          notifyBagsStatusListeners(observation, oldState, Some(status))
+        }
+        def bagsClearStatus() = bagsStatuses.synchronized {
+          val oldState = bagsStatuses.get(key)
+          bagsStatuses -= key
+          notifyBagsStatusListeners(observation, oldState, None)
+        }
+
```
tpolecat: Are you sure you want to synchronize on the immutable value referenced by a `var`? I have a hard time understanding the intent. Why isn't `bagsStatuses` part of `BagsState`? We now have two independent hunks of shared state.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -197,11 +229,38 @@ final class BagsManager(executor: ScheduledThreadPoolExecutor) {
...
+      listeners = l :: listeners
+
+  def removeBagsStatusListener(l: BagsStatusListener): Unit =
+    listeners = listeners.filterNot(_ == l)
+
+  def clearBagsStatusListeners(): Unit =
+    listeners = Nil
+
+  private def notifyBagsStatusListeners(obs: ISPObservation, oldStatus: Option[BagsStatus], newStatus: Option[BagsStatus]): Unit = {
+    listeners.foreach(_.bagsStatusChanged(obs, oldStatus.asGeminiOpt, newStatus.asGeminiOpt))
```
tpolecat: Always put a `try ... catch` around notifications like this ... if one listener throws an exception from `bagsStatusChanged` it shouldn't affect the other listeners.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/GuidingFeedbackEditor.scala
```diff
@@ -0,0 +1,38 @@
...
+    val mt = OT.getMagnitudeTable
+
+    // Construct the rows for the table. Optionally a BAGS row, and then a list of AGS analysis rows.
+    val rows = {
+      // A lot of nulls can happen here, unfortunately, so proceed with caution.
+      val bagsRow = Option(if (node == null) null else {
+        val obs = node.getContextObservation
+        if (obs == null) null else obs.getNodeKey
+      }).flatMap(BagsManager.instance.bagsStatus).map(BagsStatusRow)
+
```
tpolecat: Please write this in terms of a `for` comprehension.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
sraaphorst: I had initially tried to put `bagsStatus` as part of `BagsState`, but was receiving a deadlock, I believe. Honestly, I didn't have much time to look into it due to rushing to complete everything, but now that you mention it, I suppose it doesn't make a lot of sense to synch on the var, does it?

I'll look into it with a bit more detail.

tpolecat:
I think we should not rush this PR. State management in `BagsManager` is subtle and must be correct.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
sraaphorst: Is there any reason to do an explicit `try`...`catch` instead of just embedding the whole thing in a `Try{...}`, i.e.:

`listeners.foreach(l => Try { l.bagsStatusChanged(obs, oldStatus.asGeminiOpt, newStatus.asGeminiOpt)})

tpolecat: I prefer that we not use `Try` because it has some design issues, but I'm not going to make that argument right now. Whatever you do please be sure to log the exception.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
sraaphorst: Sure. I guess since it was just two boolean functions differing in the existential modifier that it didn't initially occur to me or seem important. I've gone ahead and refactored in my recent commit as suggested.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/GuidingFeedbackEditor.scala
sraaphorst: Thanks. Much cleaner. Done.

sraaphorst:
Alright. I agree with @tpolecat and will do the build without this PR, and we can simply do another test release once this PR is merged.

I'd also like to take into consideration in more detail the suggestions for feedback as given by @swalker2m.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
sraaphorst: I suspect that this has to do with querying the `BagsManager` for the `BagsStatus` as is done in the `GuidingFeedbackEditor`. There may be a way around this by having the `GuidingFeedbackEditor` handle the BAGS status events instead of the `EdCompTargetList`, but this seems a trifle complicated and inconvenient, and will probably require no longer having this class implement the `TelescopePosEditor` as the `edit` method will no longer be sufficient to contain enough information, I wouldn't think, to express this.

tpolecat:
:-1: I need to fork this and make some changes.

sraaphorst:
@tpolecat Should I leave the PR up?
Let me know if there's anything you need me to do.

tpolecat:
Yeah you can leave it up.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -97,29 +102,82 @@ final class BagsManager(executor: ScheduledThreadPoolExecutor) {
...
+            state = state.setStatus(key, status)
+            notifyBagsStatusListeners(observation, Some(oldStatus), Some(status))
+          case _ => // ignore
+        }
+        def bagsClearStatus(): Unit = state.statuses.get(key) match {
+          case Some(oldStatus) =>
+            state = state.clearStatus(key)
+            notifyBagsStatusListeners(observation, Some(oldStatus), None)
+          case _ => // Ignore
+        }
```
tpolecat: These methods are called from another thread that's not holding the state lock.

PR #657, opened 2015-11-09
remove OptimizableGuideProbeGroup.optimize

tpolecat:
This is incidental to the BAGS work ... I discovered that I was refactoring dead code, which is always my favorite thing.

Rebased. Tests run.


cquiroz:
:+1: less code, good...

PR #658, opened 2015-11-10
OCSADV-447: GNIRS ITC error -7847

fnussber:
Some input values caused errors in ITC and debugging the issue showed that the root cause was an error in how the `start` and `end` values in `MagnitudeBand` were calculated. This was fixed and a simple sanity test introduced.

Then I was wondering why, after fixing this, all my regression tests were still working. Turns out that there was some special code in some classes that (accidentally) deals with this case. This code has been there forever and by chance masked out the problem with the `MagnitudeBand` start and end values, which only caused another error for some specific cases.

Since I can not see why this special code would be needed I just removed it and while doing that I also removed some unused methods from some interfaces together with their implementations.

All regression tests are still running. Some manual tests show that the error is gone now.


cquiroz:
:+1:

PR #659, opened 2015-11-10
REL-2472:Fix errors in GPI unblocked modes

arturog8m:
A simple PR to update GPI unblocked modes so they don't set the parameter `useCal` to `true`, and fix the setting of `APOD_K2` when using `K2_UNBLOCKED`. Once these changes are in the OCS, we need to update the GPI SW to be able to match these configurations.

### DIFF
--- a/bundle/edu.gemini.spModel.io/src/main/scala/edu/gemini/spModel/io/impl/migration/to2016A/To2016A.scala
+++ b/bundle/edu.gemini.spModel.io/src/main/scala/edu/gemini/spModel/io/impl/migration/to2016A/To2016A.scala
@@ -67,10 +67,10 @@ object To2016A extends Migration {
     val isUnblocked = Set("UNBLOCKED_Y", "UNBLOCKED_H", "UNBLOCKED_J", "UNBLOCKED_K1", "UNBLOCKED_K2")
     for {
       gpi <- d.findContainers(SPComponentType.INSTRUMENT_GPI)
-      ps  <- Option(gpi.getParamSet("GPI")) if Option(ps.getParam("observingMode")).map(_.getValue).exists(isUnblocked)
-    } {
-      Option(ps.getParam("useCal")).foreach(_.setValue("false"))
-    }
+      ps  <- Option(gpi.getParamSet("GPI"))
+      m   <- Option(ps.getParam("observingMode")) if isUnblocked(m.getValue)
+      use <- Option(ps.getParam("useCal"))
+    } use.setValue("false")
   }
 
 
--- a/bundle/edu.gemini.spModel.io/src/main/scala/edu/gemini/spModel/io/impl/migration/to2016A/To2016A.scala
+++ b/bundle/edu.gemini.spModel.io/src/main/scala/edu/gemini/spModel/io/impl/migration/to2016A/To2016A.scala
@@ -1,10 +1,11 @@
 package edu.gemini.spModel.io.impl.migration.to2016A
 
+import edu.gemini.pot.sp.SPComponentType
 import edu.gemini.spModel.io.impl.SpIOTags
 import edu.gemini.spModel.io.impl.migration.Migration
 import edu.gemini.spModel.io.impl.migration.PioSyntax._
 import edu.gemini.spModel.pio.{Container, Document, Param, Version}
-import edu.gemini.spModel.pio.xml.{PioXmlUtil}
+import edu.gemini.spModel.pio.xml.PioXmlUtil
 
 import scala.collection.JavaConverters._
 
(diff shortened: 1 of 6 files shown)


bundle/edu.gemini.spModel.io/src/main/scala/edu/gemini/spModel/io/impl/migration/to2016A/To2016A.scala
```diff
@@ -59,6 +60,20 @@ object To2016A extends Migration {
...
+    val isUnblocked = Set("UNBLOCKED_Y", "UNBLOCKED_H", "UNBLOCKED_J", "UNBLOCKED_K1", "UNBLOCKED_K2")
+    for {
+      gpi <- d.findContainers(SPComponentType.INSTRUMENT_GPI)
+      ps  <- Option(gpi.getParamSet("GPI")) if Option(ps.getParam("observingMode")).map(_.getValue).exists(isUnblocked)
+    } {
+      Option(ps.getParam("useCal")).foreach(_.setValue("false"))
+    }
+  }
+
+
```
tpolecat: Perhaps something like:
```scala
for {
  gpi <- d.findContainers(SPComponentType.INSTRUMENT_GPI)
  ps  <- Option(gpi.getParamSet("GPI")) 
  m   <- Option(ps.getParam("observingMode")) if unblocked(m.getValue)
  use <- Option(ps.getParam("useCal"))
} use.setValue("false")
```

bundle/edu.gemini.spModel.io/src/main/scala/edu/gemini/spModel/io/impl/migration/to2016A/To2016A.scala
```diff
@@ -59,6 +60,20 @@ object To2016A extends Migration {
       plutoParams.foreach(ps.addParam)
     }
 
+
+  //All unblocked modes in GPI must set useCal = false
+  private def updateGpiUnblockedModes(d: Document): Unit = {
+
+    val isUnblocked = Set("UNBLOCKED_Y", "UNBLOCKED_H", "UNBLOCKED_J", "UNBLOCKED_K1", "UNBLOCKED_K2")
```
cquiroz: Can you get this list directly from the `Gpi1 object?

tpolecat:
:+1: modulo style comment above

bundle/edu.gemini.spModel.io/src/test/scala/edu/gemini/spModel/io/impl/migration/to2016A/GpiUnblockedModesTest.scala
```diff
@@ -0,0 +1,39 @@
...
+import edu.gemini.pot.sp.{SPComponentType, ISPObservation, ISPProgram}
+import edu.gemini.spModel.gemini.gpi.Gpi
+import edu.gemini.spModel.io.impl.migration.MigrationTest
+import org.junit.{Assert, Test}
+import scala.collection.JavaConverters._
+
+class GpiUnblockedModesTest extends MigrationTest {
+
+  @Test def testGpiConversion(): Unit =
+    withTestProgram("gpi.xml", { (_,p) => validateProgram(p) })
```
cquiroz: Where does the migration take place? During the call `withTestProgram1?

cquiroz:
Looks good to me, how are we doing the LUT updates for the seqexec?

bundle/edu.gemini.spModel.io/src/main/scala/edu/gemini/spModel/io/impl/migration/to2016A/To2016A.scala
arturog8m: Thanks! Yes, that's way more civilized and clear. I'll update the code

bundle/edu.gemini.spModel.io/src/main/scala/edu/gemini/spModel/io/impl/migration/to2016A/To2016A.scala
arturog8m: Yes, I thought about that, but the code ends up being more verbose. As there are not too many of them, I thought it was better to just copy the names I needed, carefully.  The test in `GpiUnblockedModesTest` checks that the names are correct,as it uses the real types in `Gpi`. If there were a typo in any of these, that test would fail, so I'm relatively confident with it.

bundle/edu.gemini.spModel.io/src/test/scala/edu/gemini/spModel/io/impl/migration/to2016A/GpiUnblockedModesTest.scala
arturog8m: Right. See the `MigrationTest` trait.

arturog8m:
Once I get an OCS version with these changes in place, I'll work with Fredrik to update the seqexec and GPI LUTs. Once I have those tested, I'll check them in SVN and we'll call this one good.

arturog8m:
Rebased, test pass.

PR #662, opened 2015-11-12
REL-2535: Catalog Query conditions text color not reverting properly

cquiroz:
Small PR that refreshes the coloring of the conditions text when the obs context is reloaded


fnussber:
:+1:

PR #665, opened 2015-11-13
OCSADV-345: Provide default analysis methods. 

fnussber:
Tiny PR which deals with settings some default values for the ITC analysis method depending on instrument and instrument configuration. Unfortunately we just found out that the values we set here seem not to be taken into account in the calculations as expected, but this is a bug that has been with us for a long time and needs to be addressed separately.

On the upside I also implemented a simple cache that allows to keep analysis methods for an observation until the OT is restarted which should be replaced by properly making those settings part of the model and persist them with the rest of the observation data.


cquiroz:
Looks very good :+1:

PR #666, opened 2015-11-13
REL-2253: TimeRequest dialog box Ok button needs to be pressed twice

cquiroz:
Fix for a bug on Scala swing `FormattedTextField` that emits an extra `ValueChanged` event with empty text in certain occasions. This event is used on the PIT and the new Catalog Query Tool to do validation and due to this bug we have as a side effect, that the button used to control validation needs to be clicked twice


tpolecat:
This must have been fun to figure out. Good work. :+1:

cquiroz:
it wasn't that fun :smile:

PR #667, opened 2015-11-13
REL-2407-E: improve error messages

swalker2m:
The goal of this Data Manager update is to improve error handling for problems that come up while working with the archive server.  Specifically there was a type of error response that could be generated by the server that would be handled as an "invalid response" instead of reporting the actual error message.

### DIFF
--- a/bundle/edu.gemini.dataman.app/src/main/scala/edu/gemini/dataman/query/JsonCodecs.scala
+++ b/bundle/edu.gemini.dataman.app/src/main/scala/edu/gemini/dataman/query/JsonCodecs.scala
@@ -138,9 +138,9 @@ private[query] object JsonCodecs {
 
   implicit val EncodeJsonQaResponse: EncodeJson[QaResponse] =
     EncodeJson((r: QaResponse) =>
-      ("error"  :=? r.failure)                            ->?:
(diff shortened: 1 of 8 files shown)


cquiroz:
:+1:

bundle/edu.gemini.dataman.app/src/main/scala/edu/gemini/dataman/query/JsonCodecs.scala
```diff
@@ -156,4 +156,47 @@ private[query] object JsonCodecs {
     )
 
   implicit val CodecJsonQaResponse = CodecJson.derived[QaResponse]
+
+  // ** EitherQaResponse
+
+  import GsaQaUpdateQuery.EitherQaResponse  // String \/ QaResponse
+
+  implicit val EncodeJsonEitherQaResponse: EncodeJson[EitherQaResponse] =
```
swalker2m: This is probably a bit sketchy.  It works and the test cases pass but there is a predefined scalaz either codec provided by argonaut.  It might be worth it to define a specific simple type for the error case, define a codec for that which adds the quirky `result` field and then use the provided codec.  Perhaps.

tpolecat:
Looks good. :+1:

PR #668, opened 2015-11-13
REL-2537: GMOS doesn't update for Slow Read/Low gain

arturog8m:
There was an excess of initiative removing code that resulted in listeners associated to one of the radio buttons in the GMOS OT not being connected properly, and another field in the OT doing things it shouldn't. This PR fixes those. 

First commit 8d5ee2b is the actual fix. In 3a19e78 I removed a piece of code that didn't make any sense to me, so take a look at that as well.


cquiroz:
Seems good to me :+1:

tpolecat:
:+1:

PR #671, opened 2015-11-16
REL-2465 update GPI blueprint

tpolecat:
- update GPI blueprint (again)
- update logic to exclude `{4}` when setting mode
- add test for new logic


bundle/edu.gemini.phase2.skeleton.servlet/src/test/scala/edu/gemini/phase2/skeleton/factory/TemplateSpec.scala
```diff
@@ -79,6 +79,15 @@ abstract class TemplateSpec(xmlName: String) { this: Specification =>
       .collect { case Success(n) => n }
       .toSet
 
+  /** Return the set of all integer library IDs that appear in the group. */
```
tpolecat: Oops that comment is wrong. Standby.

PR #672, opened 2015-11-16
REL-2525: Target node title in node tree not set correctly.

sraaphorst:
The target node title in the node tree was not being set correctly.

This was due to the fact that format of the title of the target node had changed from `Targets (...)` to one of `Comet:`, `Minor Planet:`, `Solar System:`, or `Target:`, and the check to determine if the title needed to be rebuilt was still based on the old title format and had not been updated.

This updates as required and solves the problem.

I also `final`ed what I could in this class.


cquiroz:
Looks fine :+1:

PR #673, opened 2015-11-16
REL-2544: Fixed editability of target details

sraaphorst:
When making the updates for BAGS, I used the incorrect boolean expression to set whether target details are editable.

This tiny PR fixes that oversight.


cquiroz:
:+1:

PR #676, opened 2015-11-17
REL-2542: A couple issues with unbagging

sraaphorst:
This short PR fixes two issues that Bryan identified in the rolling back of BAGS to the previous AGS logic:

1. The targets in the target editor are indexed by guide probe. Previously, they were indexed starting at 1, but in the BAGS unrolling, this was accidentally changed to have them indexing at 0. They are now once again indexed starting at 1 by the changes to `TelescopePosTableWidget`.

2. If AGS was run multiple times, an old target, if having the same name as the new target, was replaced by the new target and set as the primary. In the unrolling of BAGS, the new target was simply added, so the same target could appear multiple times. The old behaviour has been restored through restoring code to `AgsStrategy` and adding a method to `GuideProbeTargets` that corresponds to the old `OptionsList.setPrimary` method.

### DIFF
--- a/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsStrategy.scala
+++ b/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsStrategy.scala
@@ -85,12 +85,30 @@ object AgsStrategy {
      * Creates a new TargetEnvironment with guide stars for each assignment in
      * the Selection.
      */
-    def applyTo(env: TargetEnvironment): TargetEnvironment =
+    def applyTo(env: TargetEnvironment): TargetEnvironment = {
+      def findMatching(gpt: GuideProbeTargets, target: SPTarget): Option[SPTarget] =
+        Option(target.getTarget.getName).flatMap { n =>
+          gpt.getTargets.asScalaList.find { t =>
+            Option(t.getTarget.getName).map(_.trim).exists(_ == n.trim)
+          }
+        }
+
       (env /: assignments) { (curEnv, ass) =>
         val target = new SPTarget(HmsDegTarget.fromSkyObject(ass.guideStar.toOldModel))
         val oldGpt = curEnv.getPrimaryGuideProbeTargets(ass.guideProbe).asScalaOpt
-        val newGpt = oldGpt.getOrElse(GuideProbeTargets.create(ass.guideProbe)).withManualPrimary(target)
+
+        val newGpt = oldGpt.fold(GuideProbeTargets.create(ass.guideProbe, target).withExistingPrimary(target)) { gpt =>
+          // We already have guide probe targets for guide probe.
+          // Does one with the same target name already exist? If so, mark it as
+          // primary and replace it with the target we just made.  If not, add the
+          // target and mark it as primary.
+          findMatching(gpt, target).fold(gpt.withManualPrimary(target)) { existing =>
+            gpt.withExistingPrimary(existing).setPrimary(target)
+          }
+        }
+
         curEnv.putPrimaryGuideProbeTargets(newGpt)
       }
+    }
   }
 }
--- a/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsStrategy.scala
+++ b/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsStrategy.scala
@@ -90,7 +90,7 @@ object AgsStrategy {
         def name(t: SPTarget): Option[String] =
           Option(t.getTarget.getName).map(_.trim)
 
-        name(target).flatMap(n => gpt.getTargets.asScalaList.find(name(_).exists(_ == n.trim)))
+        name(target).flatMap(n => gpt.getTargets.asScalaList.find(name(_).exists(_ == n)))
       }
 
       (env /: assignments) { (curEnv, ass) =>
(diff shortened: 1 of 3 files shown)


bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsStrategy.scala
```diff
@@ -85,12 +85,30 @@ object AgsStrategy {
      * Creates a new TargetEnvironment with guide stars for each assignment in
      * the Selection.
      */
-    def applyTo(env: TargetEnvironment): TargetEnvironment =
+    def applyTo(env: TargetEnvironment): TargetEnvironment = {
+      def findMatching(gpt: GuideProbeTargets, target: SPTarget): Option[SPTarget] =
+        Option(target.getTarget.getName).flatMap { n =>
+          gpt.getTargets.asScalaList.find { t =>
+            Option(t.getTarget.getName).map(_.trim).exists(_ == n.trim)
```
cquiroz: I guess that can be compacted to `.exists(_.trim == n.trim)`

swalker2m: Or we might just pull out the name fetching and trimming since it is done twice.  Is name matching sufficient?  Seems like it would be but I don't remember how it used to work.

````scala
  def name(t: SPTarget): Option[String] = Option(t.getTarget.getName).map(_.trim)
````

Comment (sraaphorst) [inline bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsStrategy.scala]: Good call. I just pulled this out of the old pre-BAGS code and copied it in verbatim, but I will make the recommended change.

**Comment by cquiroz** [comment]
Can we unit test the change to indexing?

File: bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsStrategy.scala [hunk 2]
```diff
@@ -85,12 +85,30 @@ object AgsStrategy {
...
         val oldGpt = curEnv.getPrimaryGuideProbeTargets(ass.guideProbe).asScalaOpt
-        val newGpt = oldGpt.getOrElse(GuideProbeTargets.create(ass.guideProbe)).withManualPrimary(target)
+
+        val newGpt = oldGpt.fold(GuideProbeTargets.create(ass.guideProbe, target).withExistingPrimary(target)) { gpt =>
+          // We already have guide probe targets for guide probe.
+          // Does one with the same target name already exist? If so, mark it as
+          // primary and replace it with the target we just made.  If not, add the
+          // target and mark it as primary.
+          findMatching(gpt, target).fold(gpt.withManualPrimary(target)) { existing =>
+            gpt.withExistingPrimary(existing).setPrimary(target)
```
swalker2m: If you find a matching one, should it be updated at all?  Will this not mark the node as edited (and therefore require a sync)?  Did it used to work this way?  I can't remember.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/TelescopePosTableWidget.java
```diff
@@ -315,7 +315,7 @@ private static Icon bagsIcon(boolean enabled) {
                             final Option<AgsGuideQuality> quality = guideQuality(ags, guideProbe, target);
                             final boolean enabled = isPrimaryGroup && primary.exists(target::equals);
 
-                            final Row row = new GuideTargetRow(isActive, quality, enabled, guideProbe, tup._2(), target, baseCoords, when);
+                            final Row row = new GuideTargetRow(isActive, quality, enabled, guideProbe, tup._2() + 1, target, baseCoords, when);
```
swalker2m: A task for another day, I think, but the two branches of the `if` on line 284 look very similar.

sraaphorst:
There's no reason that we couldn't make a unit test for this, but I think that if we are going to make a unit test for the `TableData` in `TelescopePostTableWidget`, it should probably be more comprehensive than just checking the indexing.

But yes, I agree with Shane that this is something to do at a later point, perhaps after the Dec release.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsStrategy.scala
sraaphorst: Yeah, it did work this way previously.

I guess that the idea is that if the user edits the AGS target and then invokes AGS, then this restores it in its new and pristine state.

We might be able to get around some syncing if we compare the underlying `ITarget`s after matching on name to see if the replacement is actually necessary (i.e. `!oldITarget.equals(newITarget)`), but this is essentially verbatim pre-BAGS behaviour.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/TelescopePosTableWidget.java
sraaphorst: Oh yeah. There is definitely room for significant cleaning and refactoring here.
Since this will be changing with BAGS, I'd rather make a note to do the refactoring then instead of here in this PR.

swalker2m:
Let's make the test release with this. :+1:

PR #677, opened 2015-11-17
REL-2476: AGS for visitor instruments not working

cquiroz:
When selecting the strategy for visitor instruments it could happen that the site isn't properly selected and AGS fails. This PR will default to the selected strategy site if none is available from the program


sraaphorst:
Looks good. :+1:

PR #678, opened 2015-11-17
REL-2547: set demo science proprietary period

swalker2m:
A simple PR to set the default proprietary period for demo science programs.


sraaphorst:
:+1:

PR #679, opened 2015-11-17
REL-2429-B Modification of the GRACES template

tpolecat:
This updates the GRACES template and adds two notes to each template group. Ideally these should appear only once at the top level but the template expander doesn't allow for this yet. Opened [OCSADV-443](http://jira.gemini.edu:8080/browse/OCSINF-443) to address this in a future release.


arturog8m:
:+1:

PR #682, opened 2015-11-18
OCSADV-492: Bug fix: Added missing break statement.

fnussber:
Bug fix.


swalker2m:
Once you get used to writing Scala `match` statements it's almost impossible to remember to put `break`s in Java's switch statements. :+1:

PR #683, opened 2015-11-18
REL-2407-H: Data Manager Bug Fixes

swalker2m:
A handful of updates to archive queries and parsing:

1. Adds "progid=" and "obsid=" filters so that arbitrary program names like `Andy` will work and not result in attempts to download the entire archive.
2. Adds the "RAW" filter on the query to avoid getting processed datasets in the results.
3. Treats the dataset label as optional in archive query results.  This is necessary because sometimes datasets are created from engineering screens without an associated science program.  The failure to parse one such dataset in a list shouldn't invalidate them all.

### DIFF
--- a/bundle/edu.gemini.dataman.app/src/test/scala/edu/gemini/dataman/core/Arbitraries.scala
+++ b/bundle/edu.gemini.dataman.app/src/test/scala/edu/gemini/dataman/core/Arbitraries.scala
@@ -9,11 +9,14 @@ import org.scalacheck.Arbitrary._
 import java.time.{LocalDate, ZoneId}
 import java.time.format.DateTimeFormatter
 
+import scalaz._
+import Scalaz._
+
 trait Arbitraries extends edu.gemini.spModel.dataset.Arbitraries {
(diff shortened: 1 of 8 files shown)


bundle/edu.gemini.dataman.app/src/test/scala/edu/gemini/dataman/core/Arbitraries.scala
```diff
@@ -9,11 +9,14 @@ import org.scalacheck.Arbitrary._
...
 
+import scalaz._
+import Scalaz._
+
 trait Arbitraries extends edu.gemini.spModel.dataset.Arbitraries {
   implicit val arbGsaRecord: Arbitrary[GsaRecord] =
     Arbitrary {
       for {
-        label <- arbitrary[DatasetLabel]
+        label <- Gen.frequency((95, arbDatasetLabel.arbitrary.map(some)), (5, none[DatasetLabel]))
```
cquiroz: Nice

cquiroz:
:+1:

PR #685, opened 2015-11-19
REL-2165: copy science observing conditions during template instantiation

swalker2m:
A small requested change to the template instantiation functor.  It now copies science observing conditions for all observation classes except `DAY_CAL`.


arturog8m:
Looks good, thanks for adding a test case for this :+1:

PR #689, opened 2015-11-19
REL-2407-J: Minor Data Manager Updates

swalker2m:
Three updates:

1. Split the thread pool that polls the summit and archive in two.  One half dedicated to the summit polling and one half to the public archive.  Since the network is so sketchy at GS, the public archive link is likely to be slow and unreliable and I don't think that should impact the relatively more important summit poll.

2. Resume filtering with the `progid=` and `obsid=` feature of the archive server.  This should allow us to put all non-standard program ids back in the test database.

3. Don't eat `java.lang.InterruptedException`. `\/.fromTryCatch` catches non-fatal exceptions, including `InterruptedException`.  Interrupts are sent to threads to stop them. Catching `InterruptedException` prevents the thread from stopping normally.  Switched `\/.fromTryCatch` for the same implementation used in a later version of scalaz, in `\/.fromTryCatchNonFatal`.  Also, rethrow the `InterruptedException` caught by `Task.attemptRun`.

### DIFF
--- a/bundle/edu.gemini.dataman.app/src/main/scala/edu/gemini/dataman/core/package.scala
+++ b/bundle/edu.gemini.dataman.app/src/main/scala/edu/gemini/dataman/core/package.scala
@@ -1,13 +1,16 @@
 package edu.gemini.dataman
 
 import edu.gemini.dataman.core.DmanFailure.DmanException
+import edu.gemini.spModel.core.catchingNonFatal
 import edu.gemini.spModel.dataset.{DatasetExecRecord, DatasetQaRecord}
+
+import scalaz.concurrent._
 
 import scalaz._
 import Scalaz._
-import scalaz.concurrent._
 
 package object core {
+
   /** When an action is applied, it may update one or more QA and/or exec
     * records.
     */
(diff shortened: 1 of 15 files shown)


bundle/edu.gemini.dataman.app/src/main/scala/edu/gemini/dataman/core/package.scala
```diff
@@ -33,7 +36,10 @@ package object core {
     def fail(f: => DmanFailure): DmanAction[Nothing] = EitherT(Task.delay(f.left))
 
     def mergeFailure[A](result: Throwable \/ TryDman[A]): TryDman[A] =
-      result.fold(ex => DmanFailure.DmanException(ex).left, identity)
+      result.fold({
+        case ie: InterruptedException => throw ie
```
cquiroz: Wouldn't you rather return a left with the exception? Or you want here to kill the thread?

cquiroz:
:+1:

bundle/edu.gemini.dataman.app/src/main/scala/edu/gemini/dataman/core/package.scala
swalker2m: By handling `InterruptedException` and turning it into a left, as I was doing here before, we handle the interrupt and the thread continues to run.  I'd need to check everywhere whether the `DmanFailure` is due to an `InterruptedException`.  Admittedly though, I'm not completely sure this is the right thing to do.

swalker2m:
I'll merge this now so that we can do a test release.  I'd be very interested in additional feedback though.

PR #690, opened 2015-11-19
REL-2515 leaky listeners

tpolecat:
`TargetObsComp` listens to change events on its contained `SPTargets` but these listeners were not removed when the target environment was updated. This led to out of control leakage during target editing. Fixed that and made `WatchablePos` slightly less idiotic.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/WatchablePos.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/WatchablePos.java
@@ -1,43 +1,32 @@
 package edu.gemini.spModel.target;
 
 import java.io.Serializable;
-import java.util.Vector;
+import java.util.Set;
+import java.util.HashSet;
 
 public class WatchablePos implements Cloneable, Serializable {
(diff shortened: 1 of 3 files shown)


swalker2m:
Makes sense to me.  :+1:

sraaphorst:
Looks good, and glad this was tracked down. :+1:

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/WatchablePos.java
```diff
@@ -1,43 +1,32 @@
...
 
 import java.io.Serializable;
-import java.util.Vector;
+import java.util.Set;
+import java.util.HashSet;
 
 public class WatchablePos implements Cloneable, Serializable {
 
-    private transient Vector<TelescopePosWatcher> _watchers;
+    private transient Set<TelescopePosWatcher> _watchers;
```
cquiroz: Can it be final to avoid the null checks?

swalker2m:
Merging to include in the test release.

PR #691, opened 2015-11-20
REL-497: Support for GMOS IFU-2 mode

fnussber:
This code is an attempt to salvage some of Sabrina's work and make it available in the December release. It implements IFU-2 support for GMOS.

In short, when using IFU-2 mode the incoming signal is split and shifted onto different areas of the CCD array ("left"/"blue" and "right"/"red"). This PR contains some magical code that calculates the shifts that are applied to the input signal to know where it will end up on the CCDs and also some code to transform between wavelength space and pixel space.

This allows to create a new graph which shows the resulting signal in pixel space including the gaps between the different CCDs:

image

As always I am heavily relying on the regression tests. Many of the changes in `GmosRecipe` are due to some refactoring of the code that creates the charts. The functionality is the same, except for the additional new chart.

While integrating the code I noticed several areas that will need further attention:
* `DetectorsTransmissionVisitor` is overly complicated and should be simplified and if possible made more generic to be prepared for future instruments that have more than one CCD.
* The `CCDGapCalc` class is a conversion of some IRAF script that somehow deals with unicorns and turns them into zebras. And back. My gut feeling is that this code should/could become part of the `DetectorTransmissionVisitor`, but for now it is what it is. Sabrina promised to add test cases.
* In general the way results are modeled in the ITC is *very* awkward and for the first time I think I can claim to really fully understand what's going on(!). I used that insight to change a few things of how results are presented in the IFU cases.
* The IFU-2 case can currently not be implemented for the Hamamatsu CCDs because the "application" of the input signal on the CCDs in order to get the resulting signal is done *before* the input signal is split into the read and blue part (i.e. shifted in wavelength/pixel space) and therefore the calculated response of the input signal on the CCD is not correct for the case of the Hamamatsu CCD array where the position is relevant. Which is not the case for E2V. So for now we only support E2V.

Ok, 'nuff said. I go dream of unicorns now.


cquiroz:
:+1:

andrewwstephens:
Could you point to the unicorn <-> zebra conversion ?

fnussber:
It is part of the Grand Transmogrifier, aka [CCDGapCalc]( https://github.com/fnussber/ocs/blob/632df2defac07873e0f67b9dd8ded427a8da19b3/bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gmos/CCDGapCalc.java).

tpolecat:
image

fnussber:
Awww. I want one, can I have one? Prrretty please..?

PR #692, opened 2015-11-20
REL-2167: GNIRS P2 check for Decker / Acquisition Mirror

sraaphorst:
We want to generate a warning for GNIRS when the Decker is set to Acquisition Mirror and the acquisition mirror is Out.

Previously, I was looking at the instrument node for the settings for decker and acquisition mirror.
The decker, however, is never changed in the instrument node (why it is included here, I don't know, since this seems unused), and can only be set in GNIRS sequence items.

If not set in the sequence items, it is inferred to be "empty" by SeqExec and set to something other than Acquisition Mirror based on other parameters.

Thus, now the P2 check examines the sequence instead of the instrument configuration to properly generate the warning.

(Also fixed a spelling error in comments in `ItemKey`.)


cquiroz:
:+1:

PR #693, opened 2015-11-20
REL-2552:UI pause when changing conditions in Catalog Query Tool

cquiroz:
Refreshing the table model during a conditions change may take some time and block the UI. This PR moves the table model update and replotting to be executed on the swing thread outside the event loop


sraaphorst:
Looks fine to me. :+1:

swalker2m:
It's still going to block the UI?  The swing thread and the event loop are the same.  The difference is that the event that triggered the action finishes before the repotting takes place.

PR #694, opened 2015-11-20
REL-2556: Reset unplot state

cquiroz:
Tiny PR, resets the unplot/plot button to the "Unplot" state when a new target is reloaded


sraaphorst:
:+1:

PR #695, opened 2015-11-20
REL-2555: change texes template program

swalker2m:
In coordination with REL-2165, this update removes observing conditions from Telluric standard star observations.  Template instantiation will now use the science observing conditions when creating these observations.


cquiroz:
:+1:

PR #699, opened 2015-11-20
OCSADV-493: Bug fix: Deal properly with GS IFU-N&S cases.

fnussber:
Bug fix that adds case statements for some missing cases.


tpolecat:
:+1:

PR #700, opened 2015-11-21
REL-2143: Change cut-off semester for "no E2V rule for GS"

fnussber:
Cut off date has been changed to 2014A, i.e. GMOS-S must use Hamamatsu for 2014B and later. Added test cases to make sure that error is also given for ENG and CAL programs.


cquiroz:
Looks good, So we are kind of flagging older proposals? I guess you cannot enter new 2014B proposals, can you?

fnussber:
Yes, older proposals will be flagged now with an error. But the main reason for this rule is however to flag current and future proposals, in particular for cases where people copy/paste observations from older proposals into new ones. AFAIK only Gemini staff could create new proposals for semesters in the past, but I don't think there's a reason why anyone would do that.

PR #702, opened 2015-11-21
Bugfix: Plate scale values for E2V and Hamamatsu were mixed up

fnussber:
Resetting the value to the one for E2V (which is the only CCD array we currently support for IFU-2).


arturog8m:
This should have a REL task, so testers are aware of this change.

PR #703, opened 2015-11-22
REL-2407-K: Data Manager bug fix and update

swalker2m:
Two minor updates that fell out of Friday's testing session.

1. A bug fix for the OT Observing Log GUI.  The bug was that updates to the exec log were not always reflected in the GUI.  Here I removed a `JSplitPane` that had been used in the staff version of the "data analysis" tab, yet the code was still looking for the `JSplitPane` to determine whether it was configured for staff or PI mode.  The fix was to make the staff vs. PI mode more explicit.

2. We decided to support different polling rates for the "tonight", "this week", and "all" periods for the summit FITS server vs. the public archive. In particular, I increased the "tonight" period for the summit server to 15 seconds at Paul's suggestion.  The other two haven't changed but now could be made distinct.


cquiroz:
Looks good :+1:

PR #704, opened 2015-11-23
OCSADV-497: Old magnitudes remain after Simbad query without new magnitudes

cquiroz:
Tiny PR that clears the magnitudes if none is retrieved from Simbad


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details/SiderealNameEditor.scala
```diff
@@ -115,6 +115,8 @@ final class SiderealNameEditor(mags: MagnitudeEditor) extends TelescopePosEditor
       }
       if (i.magnitudes.nonEmpty) {
         mags.replaceMagnitudes(i.magnitudes.map(_.toOldModel).asImList)
+      } else {
+        mags.replaceMagnitudes(DefaultImList.create())
```
swalker2m: Do you need the `if` in this case?  Would it not be the same if you just always use the first branch?

`mags.replaceMagnitudes(i.magnitudes.map(_.toOldModel).asImList)`

cquiroz: :+1:

swalker2m:
:+1:

PR #705, opened 2015-11-23
REL-2557: OT memory leak?

cquiroz:
This PR fixes a bug on the way a combo box is rendered, the extra foreground call makes Swing to call the rendering method over and over leading to a CPU spike.


sraaphorst:
Good job on finding this one so quickly. :+1:

PR #706, opened 2015-11-23
REL-2500: Update wavelength for F2 K_LONG filter

arturog8m:
Small PR that updates the wavelength for F2 K Long Filter from 2.0 to 2.2 um.


cquiroz:
Hmm, this should affect the PIT and ITAC too, de we need more tickets or do you prefer to update the P1 model right away?

arturog8m:
Good point. I think I figured how to make the changes in support of the PIT, see ddcc739. However, I don't know about ITAC. Where is that done?

cquiroz:
Actually ITAC will read these changes from the p1 model. I think we are good now :+1:

arturog8m:
Excellent!

PR #707, opened 2015-11-23
REL-2551: open dialog issue

swalker2m:
In the open program dialog, we filter out local programs for which the current key doesn't provide access. For example, if you now have a lowly "user" key you might not see all programs you checked out when you had a staff key.

However, we do not check whether the peer associated with a local program matches the currently selected peer.  So if you have a "staff" key from one peer, it will show you the local programs from all peers *even though you can only work with those that match the selected key's peer*.  In these cross-site cases we show the program in the open dialog table and report an "Error" in the "Changes" column because we don't have permission to lookup the status of the program.

This PR updates the local program filtering to also remove local programs associated with a peer other than the selected key's peer.  This is more consistent with the other open dialog filters along the bottom and seems to be what is expected by Andy as discussed in REL-2551.


cquiroz:
Looks fine :+1:

PR #708, opened 2015-11-24
OCSADV-498: Display multiple IFU plots vertically in OT

fnussber:
This is a layout change that should make it easier for users to check the charts that are generated for ITC calculations with multiple IFU elements for GMOS and NIFS.

For each IFU element ITC creates a set of charts. These are currently all displayed horizontally in the OT which is cumbersome if there are many IFU elements. This PR changes this by allowing charts to be grouped (in this case per IFU element) and then display the charts for each group (IFU element) stacked vertically instead of horizontally.

Example: GMOS-N with multiple IFU elements, results per IFU element stacked vertically in the OT:

image

### DIFF
--- a/bundle/edu.gemini.itc/src/main/scala/edu/gemini/itc/base/Recipe.scala
+++ b/bundle/edu.gemini.itc/src/main/scala/edu/gemini/itc/base/Recipe.scala
@@ -84,6 +84,9 @@ object Recipe {
 
   // === Spectroscopy
 
+  /** Collects the relevant information from the internal result in a simplified data object and collects some additional
+    * information from the data series like e.g. the max signal to noise. For each individual CCD ITC basically
+    * does a full analysis and gives us a separate result. */
   def toCcdData(r: SpectroscopyResult, charts: List[SpcChartData]): ItcCcd = {
     val s2nChart: SpcChartData = charts.find(_.chartType == S2NChart).get
     val singleSNRatioVals: List[Double] = s2nChart.allSeries(SingleS2NData).map(_.yValues.max)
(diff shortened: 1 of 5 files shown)


bundle/edu.gemini.itc/src/main/scala/edu/gemini/itc/base/Recipe.scala
```diff
@@ -93,11 +96,20 @@ object Recipe {
...
+  // === Java helpers
+
+  // One result (CCD) and a simple set of charts, this covers most cases.
   def serviceResult(r: SpectroscopyResult, charts: java.util.List[SpcChartData]): ItcSpectroscopyResult =
-    ItcSpectroscopyResult(List(toCcdData(r, charts.toList)), charts.toList)
+    ItcSpectroscopyResult(List(toCcdData(r, charts.toList)), List(SpcChartGroup(charts.toList)))
+
+  // One result (CCD) and a set of groups of charts, this covers NIFS (1 CCD and separate groups for IFU cases).
+  def serviceGroupedResult(r: SpectroscopyResult, charts: java.util.List[java.util.List[SpcChartData]]): ItcSpectroscopyResult =
+    ItcSpectroscopyResult(List(toCcdData(r, charts.toList.flatten)), charts.toList.map(l => SpcChartGroup(l.toList)))
```
cquiroz: I thought you would normally use `charts.asScala`, is the `toList` some of the automatic conversions?

cquiroz:
Looks fine :+1:

bundle/edu.gemini.itc/src/main/scala/edu/gemini/itc/base/Recipe.scala
fnussber: I think you would have to write `charts.asScalaBuffer.toList` (?). Using `toList` just does an implicit conversion and seems simpler. But maybe I am missing something.

PR #709, opened 2015-11-24
OCSADV-490: Include GMOS gain and read mode in ITC web output 

fnussber:
Adding some config info on the ITC web app output.

image


cquiroz:
Why does the test file change in this case? does baseline.txt contain the expected output of the webapp?

fnussber:
This is correct. On the todo list is to add regression tests that only take the calculation results into account but not the web output. (Originally the web output and the calculation output was so intertwined that there was no way to do that. Now the things are in place but I haven't done it yet.)

PR #711, opened 2015-11-24
OCSADV-495: GNIRS OT ITC recognition of imaging mode

fnussber:
This PR adds some code that detects if GNIRS is in imaging mode (i.e. the acquisition mirror is in). ITC currently does not support imaging calculations for GNIRS, so what this PR is doing is that all imaging configs will be filtered/ignored on the "ITC Spectroscopy" tab. (If we ever added imaging for GNIRS to ITC the imaging configurations would show up on the "ITC Imaging" tab instead.)


cquiroz:
Looks fine to me

PR #712, opened 2015-11-24
REL-2539: F2 NullPointerException

fnussber:
This PR fixes a NPE that happens because the custom slit width property isn't always set properly in the config sequence.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/flamingos2/Flamingos2.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/flamingos2/Flamingos2.java
@@ -1217,7 +1217,7 @@ public final class Flamingos2 extends ParallacticAngleSupportInst
         Pio.addParam(factory, paramSet, FPU_PROP, _fpu.name());
         if (_fpu == FPUnit.CUSTOM_MASK) {
             Pio.addParam(factory, paramSet, FPU_MASK_PROP, _fpuMaskLabel);
-            Pio.addParam(factory, paramSet, CUSTOM_SLIT_WIDTH_PROP, customSlitWidth.name());
+            Pio.addParam(factory, paramSet, CUSTOM_SLIT_WIDTH_PROP.getName(), customSlitWidth.name());
         }
 
         Pio.addParam(factory, paramSet, MOS_PREIMAGING_PROP.getName(), getMosPreimaging().name());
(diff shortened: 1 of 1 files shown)


cquiroz:
:+1:

fnussber:
No, this information is not used by the seqexec, it is only here to allow the users to define a slit width which can be used by the ITC to do its calculations.

arturog8m:
good. Just for curiosity.. how come the `.getName()` thing fixes a NPE? How does this work?

fnussber:
The change to use `getName()` instead of (implicitly) `toString()` doesn't have anything to do with the NPE. However I think it makes more sense to use `getName()` as the lookup key for the parameter in theses cases. The rest of the F2 code uses either of those two, I think it should use `getName()` always, but I only changed it for custom slit width property which I added recently.

swalker2m:
The `Pio.getValue` method that takes a `PropertyDescriptor` does the call to `getName` for you.  The problem seems to be that we don't support a `getValue` that takes a `PropertyDescriptor` and a default value.  We should probably add that to `Pio`.

swalker2m:
Still the question remains, at least for me, how does this actually fix a NPE?  Was there some code looking for the wrong property key because of this?

fnussber:
Line 1327. Setting the property.

arturog8m:
Ahhh!. Did not catch that one with the rest of the changes. That makes sense.

swalker2m:
I see, so now we will set the custom slit width property regardless of whether we're using a custom mask or not.  That's a bit surprising but probably okay.  This could conceivably have an impact on the seqexec since it changes what it will receive in the non-custom mask case but I doubt it will be an issue.

fnussber:
I am open for better solutions. For now this is the best I could come up with. Sorry.

swalker2m:
If the seqexec actually doesn't use this information, maybe we don't need to include it in the sequence? @jluhrs can you confirm that the seqexec will ignore the F2 custom slit width value?

fnussber:
The problem this is trying to solve is that the values needs to be set for the smartgcal lookups. However it was not always set at the beginning of a sequence created with an iterator.

fnussber:
Do you want me to revert this one? I was hoping for another test release and was not planning to work on this one any longer. But it seems to be important.

arturog8m:
There are no references to `customSlitWidth` in the seqexec, so it shouldn't affect the seqexec, but we should test it. That is tcl so who knows what it actually does.

fnussber:
This doesn't surprise me because I just added this about 2 weeks ago.

swalker2m:
Okay, I see.  I think this is all fine.  I was just confused about what was going on.  We might add appropriate methods that take `PropertyDescriptor` where missing at some point instead of explicitly calling `getName` but no need to revert this!

arturog8m:
Agreed! Tomorrow they will be doing regression testing, so I can ask them to verify F2 explicitly.

swalker2m:
Wrt to the last comment though, I believe in the new sequence model it won't matter about property descriptor `getName` and all the rest anyway.

jluhrs:
The Seqexec does not know about this `customSlitWidth` property. If it finds it in the sequence it will just ignore it.

PR #713, opened 2015-11-24
REL-2476: AGS for visitor instruments not working

cquiroz:
This PR removes the site for visitors instruments used in the Catalog Query Tool. A side effect is that the AGS limits are selected with the site from the chosen guiders rather than the one on the context.
This fixes a bug where visitor instruments may not always select the correct strategy since a visitor can be either on the north or south


swalker2m:
Why does the Catalog Query Tool have an impact on AGS?

cquiroz:
There is really no impact on this case. The CQT ask AGS for  a strategy and it uses it to calculate the star quality. One of the params of the selection is the site. This PR improves the selection of the site for visitors

swalker2m:
:+1:

PR #714, opened 2015-11-24
REL-2560: ODB telnet settings

sraaphorst:
We ran into an issue this morning where two telnet connections were closed on the client end, but still maintained / blocked on the ODB server, and as the default settings of no timeout and only allowing two active telnet connections, we could not access the ODB through telnet.

This tiny PR increases the number of maximum allowed telnet connections and sets a timeout for blocking reads to prevent this problem in the future.

### DIFF
--- a/app/spdb/build.sbt
+++ b/app/spdb/build.sbt
@@ -155,8 +155,10 @@ def with_gogo(version: Version) = AppConfig(
 def with_remote_gogo(version: Version) = AppConfig(
   id = "with-remote-gogo",
   props = Map(
-    "gosh.args"              -> "--nointeractive",
-    "osgi.shell.telnet.port" -> "8224"
+    "gosh.args"                       -> "--nointeractive",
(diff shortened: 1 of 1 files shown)


cquiroz:
Did you test the timeout?

sraaphorst:
It doesn't seem to time out telnet connections that are still active on the client end, so unless I can install an spdb on a different machine and then drop the connection without logging out to simulate what happened in the original issue, I'm not sure how to test this. :-/

cquiroz:
I think that's ok, these are difficult to test and as far as I remember the osgi telnet service is very old, so who knows what it really works

sraaphorst:
I'm going to try to see if I can simulate the behaviour by closing sessions unexpectedly from sbfosxdev-mp1 to my local machine before merging to see if I can generate the same conditions that came up this morning, but from the Apache Felix docs, this just basically sets the `SO_TIMEOUT` option on the socket to the number supplied (measured in ms).

PR #715, opened 2015-11-25
OCSADV-496: Avoid spikes around Hamamatsu CCD gaps.

fnussber:
This PR works around an issue that is caused by how gaps are added to the signal and how some (in this case unwanted) interpolation is causing "ghost" values to be added at the chip borders that result in wrong and ugly spikes in the resulting charts.

Yes I know, this is a hack. The handling of chip gaps needs a major redo. The way the gaps are currently added not only shows some weird behavior with the spikes this PR is dealing with but is complicated and cumbersome and was (and still is) also a pain when trying to get the IFU-2 stuff to work. For now I am afraid this is the best I can do.

Before (spikes around the second gap):
image

After (no spikes):
image


bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gmos/GmosRecipe.java
```diff
@@ -466,4 +473,12 @@ private static Color ccdDarkColor(final int ccdIndex) {
             default: throw new Error();
         }
     }
+
+    // In the multi-CCD case we have to force the first and last y values to 0 to cancel out signal and s2n spikes
+    // caused by interpolation, resampling or some other effect of how data is calculated around CCD gaps.
+    // TODO: DetectorTransmissionVisitor needs a serious overhaul so that this behavior becomes more predictable.
+    private static void fixGapBorders(double[][] data) {
+        data[1][0]                  = 0.0;
```
cquiroz: I guess this is the `hack`. It doesn't look that bad, in many cases you need to fix border conditions like this. At least I don't have a big problem with this

cquiroz:
Looks good to me :+1:

PR #716, opened 2015-11-25
REL-2558: Fixes problem with F2 Skeleton

arturog8m:
This PR removes the changes made to the Phase 1 Model to update the label names for the F2 K-Long Filter. The reason to remove these is that this will make existing Phase 1 documents to generate errors when skeletons are generated in December. The changes to the OT display values are preserved. This was tested with a P1 document sent by Andy and works okay now 

These changes should be added at the same time we release a new version of the PIT in March, if we really want to keep the names of the filters updated in the PIT.


cquiroz:
:+1:

PR #717, opened 2015-11-25
REL-2563: Overwrite mode in F2 Custom MDF box

arturog8m:
This PR addresses an issue in the F2 Custom MDB entry box - the textbox is always in 'overwrite mode' so that you can't insert characters in the middle of the string without overwriting other characters. 

There must be a way to set an appropriate `Format` in the `createStringInstance` method  that would do the same (I tried that, but I couldn't make it to persist the changes in properties), hence this other alternative that is not as good but I think it is safe in this context. I can't think of a case in the OT where we do want the overwrite mode to be set to `true` - if that happens in the future, we either need to put more effort in this, or just set the flag back in the specific instance of this class. 

The actual change is in 342ac6c. The other commit is just code cleanup.

### DIFF
--- a/bundle/edu.gemini.shared.gui/src/main/java/edu/gemini/shared/gui/bean/TextFieldPropertyCtrl.java
+++ b/bundle/edu.gemini.shared.gui/src/main/java/edu/gemini/shared/gui/bean/TextFieldPropertyCtrl.java
@@ -1,7 +1,3 @@
-//
-// $
-//
-
 package edu.gemini.shared.gui.bean;
 
 import edu.gemini.shared.util.immutable.None;
@@ -13,11 +9,8 @@ import javax.swing.event.DocumentEvent;
 import javax.swing.event.DocumentListener;
 import javax.swing.text.DefaultFormatter;
 import javax.swing.text.Document;
-import java.beans.PropertyChangeEvent;
-import java.beans.PropertyChangeListener;
 import java.beans.PropertyDescriptor;
 import java.text.*;
-import java.util.logging.Logger;
 
 /**
  * A PropertyCtrl implementation for text fields.  Currently support is
@@ -26,7 +19,6 @@ import java.util.logging.Logger;
  * @param <T> property type
  */
 public final class TextFieldPropertyCtrl<B, T> extends PropertyCtrl<B, T> {
-    private static final Logger LOG = Logger.getLogger(TextFieldPropertyCtrl.class.getName());
 
     /**
      * An encapsulation of all the information required in order to format and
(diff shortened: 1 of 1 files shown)


sraaphorst:
I'm curious as to why this text field is in overwrite mode in the first place and how it gets to be in that state. An initial look through the code doesn't reveal anything to me that should do that, but perhaps I'm missing something?

cquiroz:
Is the overwrite mode used anywhere else?

swalker2m:
Art and I were looking at this earlier.  The `TextFieldPropertyCtrl` class in question states in its Java Doc that "currently support is provided for editing integers and doubles" but in two places we use it for `String` properties.  It seems to work fine for integers and doubles but the `String` usages (one in the visitor instrument name editor and one in the F2 custom mask name field) suffered this issue with the overwrite mode.

I assume the problem was never noticed because it simply doesn't get much use outside of integer and double fields.  We didn't know why the default would be overwrite, nor why integer and double fields work fine without setting the overwrite property to `false`.  Monkey testing with the two `String` usages and with a few integer/double cases indicates that all is fine when we use the `false` setting, but none of this is very satisfying.

arturog8m:
No, the overwrite mode isn't used anywhere else. The fact that in the `FomatSupport` for the `String` instance of this control the `Format` is set to `null`, whereas for Decimal and Integer fields is set to specific number formatters, seems to give a clue as that those things set the overwrite property to what you expect. 

The `DefaultFormatter` documentation says that `setOverwriteMode` behavior is:

> Configures the behavior when inserting characters. If overwriteMode is true (the default), new characters overwrite existing characters in the model.

So, the default is `overwriteMode=true` which I wasn't expecting at all. I suspect the formatters for Decimal and Integer, etc are taking care of this internally, somehow, but I really didn't have much time to look into this any further.

sraaphorst:
What a horrible design `JFormattedTextField` has, given that you can either supply a `Format` (none of whose implementations have overwrite options, it seems) or an `AbstractFormatter`, with the `DefaultFormatter` implementation having the only instance of `setOverwrite` I can find, bizarrely defaulting to `true` to make it completely inconsistent with any other formatting configuration of the text field.

sraaphorst:
Indeed, looking at the Java source, it doesn't even seem like you can enable this behaviour other than by using `DefaultFormatter` or one of its subclasses. I guess the idea is to use an `InternationalFormatter` or `MaskFormatter`, where this overwrite might make sense, and then pass them via the constructor a `Format` or mask `String`, instead of supplying the `Format` directly to the `JFormattedTextField`.

Yuck.

tpolecat:
Yeah Swing is a disaster. After a while you stopped being surprised by this kind of thing.

sraaphorst:
Anyways, I guess this makes about as much sense to do as anything else, so :+1:

PR #718, opened 2015-11-25
REL-2562: Speed improvements to Catalog Query Tool for larger sets

cquiroz:
Small updates to the Catalog Query Tool improving performance by reusing calculated objects and swing components. The most important change is avoiding multiple redraws when updating the conditions boxes. The last change breaks REL-2535, but a different fix is included in this PR

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/edu/gemini/catalog/ui/QueryResultsFrame.scala
+++ b/bundle/jsky.app.ot/src/main/scala/edu/gemini/catalog/ui/QueryResultsFrame.scala
@@ -55,12 +55,17 @@ object QueryResultsFrame extends Frame with PreferredSizeFrame {
     override val flipAction = "Plot"
   }
 
+  private object GuidingFeedbackRenderer {
+    val icons = AgsGuideQuality.All.map(q => q -> GuidingIcon(q, enabled = true)).toMap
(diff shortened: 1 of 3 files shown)


bundle/jsky.app.ot/src/main/scala/edu/gemini/catalog/ui/QueryResultsFrame.scala
```diff
@@ -55,12 +55,17 @@ object QueryResultsFrame extends Frame with PreferredSizeFrame {
     override val flipAction = "Plot"
   }
 
+  private object GuidingFeedbackRenderer {
+    val icons = AgsGuideQuality.All.map(q => q -> GuidingIcon(q, enabled = true)).toMap
```
tpolecat: Just FYI you can use `fproduct` here, which takes `F[A] => (A => B) => F[(A, B)]` ... so

```scala
val icons = AgsGuideQuality.All.fproduct(GuidingIcon(_, enabled = true)).toMap
```

cquiroz: Nice one

tpolecat:
This looks good. :+1:

PR #719, opened 2015-11-26
REL-2508: Smartgcal service startup issues.

fnussber:
The smartgcal servlet shows some erratic behavior on startup. It sometimes fails and sometimes starts without problems. Currently I don't have a good idea of what's going on.

This PR contains a minimally invasive change to the smartgcal startup that I hope will leave us with some additional information about what's going wrong when trying to startup the svn command line environment, which is the root cause of why the smartgcal servlet is (sometimes) not available.

Once we have a better idea of what's going on we can move the diagnostic command execution into the `catch` statement, for now I would like to see its output even if things go ok.

Add-On: From looking at the code I am now thinking that there might be some synchronization issues/race conditions in the code, this would explain the erratic behavior. I'll follow up with that asap.

And also: From looking at the update dates of those files (most are several years old), I wonder if the whole current mechanism isn't total overkill and those files should just become ordinary resources...
image

### DIFF
--- a/bundle/edu.gemini.smartgcal.servlet/src/main/java/edu/gemini/smartgcal/servlet/repository/CalibrationSubversionRepository.java
+++ b/bundle/edu.gemini.smartgcal.servlet/src/main/java/edu/gemini/smartgcal/servlet/repository/CalibrationSubversionRepository.java
@@ -5,11 +5,13 @@ import edu.gemini.spModel.gemini.calunit.smartgcal.CalibrationFile;
 import edu.gemini.spModel.gemini.calunit.smartgcal.CalibrationRepository;
 import edu.gemini.spModel.gemini.calunit.smartgcal.Version;
 import org.tigris.subversion.svnclientadapter.*;
-import org.tigris.subversion.svnclientadapter.commandline.CmdLineClientAdapterFactory;
+import org.tigris.subversion.svnclientadapter.commandline.*;
 
 import java.io.*;
 import java.net.MalformedURLException;
 import java.util.Date;
+import java.util.logging.Logger;
+import java.util.logging.Level;
 
 /**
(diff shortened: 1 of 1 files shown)


cquiroz:
Looks fine :+1:

arturog8m:
We should discuss the idea of making these files just resources. Maybe we can move them to github and people can simply update them there? Anyhow, we can talk about this in December. For now, let's add this to figure out what's going on. :+1:

bundle/edu.gemini.smartgcal.servlet/src/main/java/edu/gemini/smartgcal/servlet/repository/CalibrationSubversionRepository.java
```diff
@@ -40,6 +33,33 @@ public CalibrationSubversionRepository(String url, String user, String password)
...
+        // If svn is not available CmdLineClientAdapterFactory.setup() will fail with an exception.
+        // Unfortunately the original exception is swallowed and the one thrown by setup() does not have a
+        // case and is generic and therefore gives no clues as to what went wrong. This additional code tries
+        // to execute "svn --version" manually (which is what the setup() function is doing) and log the exception,
+        // hopefully giving us some additional information.
+        try {
+            LOG.info("Trying to execute svn --version to validate svn availability");
+            final CmdLineClientAdapter12 cmd = new CmdLineClientAdapter12(new CmdLineNotificationHandler());
+            cmd.getVersion();
+            LOG.info("svn --version: OK");
```
swalker2m: I think it was a good idea to move this out of a static initializer block and to explicitly do this test before trying the setup. So :+1: let's merge this.  Having gotten the version information though, we might as well log what it is just in case it somehow finds a different version of svn than we were expecting.

PR #720, opened 2015-11-26
OCSADV-241: Convert non-gems ManualGS to use VoTable

cquiroz:
This PR removes a bunch of classes from the old catalog bundle that are not in use anymore. All the code compiles, tests passes and the OT starts without troubles


swalker2m:
This is marvelous.  Is there any risk though that you've removed something still in use somewhere, say via reflection?  Could we just as well wait until after the release for this?

arturog8m:
Shane just took the words out of my mouth... I also think it would be best to do this right after the release, just in case.

cquiroz:
I tested quite a bit, but of course there could be some magic missing. e.g. I had removed the AstroCat** classes and those were called via reflection
Let's push it to the next release, I'll keep the PR open

swalker2m:
:+1:

PR #721, opened 2015-11-26
REL-2407-L: Logging Level Updates

swalker2m:
The Data Manager is currently generating copious logging output.  While it is probably a good idea to leave it that way initially, it makes it hard to spot non-data manager issues in the log.  This PR creates  shared data manager "detail" and "json" logging levels that are consulted when writing a variety of low-level log messages.  It also adds a shell command to allow changing the logging levels at runtime.

The overall logging level is configured to `INFO` in the spdb.  By default "detail" messages will be written at `INFO` and so will appear in the log.  The "json" level controls whether the JSON posted to and retrieved from the GSA is logged.  By default it will be written at `FINE` and will not appear.  To make general "detail" messages disappear and "json" appear, for example, the relevant commands would be:

````
g! dataman detail
INFO (logging on)
g! dataman detail FINE
Logging detail level set to FINE (logging off)
g! dataman json
FINE (logging off)
g! dataman json INFO
Logging json level set to INFO (logging on)
````

### DIFF
--- a/app/spdb/conf/logging.properties
+++ b/app/spdb/conf/logging.properties
@@ -10,7 +10,5 @@ java.util.logging.FileHandler.limit = 10000000
 java.util.logging.FileHandler.count = 100
 
 #edu.gemini.dataman.level = FINE
-#edu.gemini.datasetfile.level = FINE
-#edu.gemini.datasetrecord.level = FINE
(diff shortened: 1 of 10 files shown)


cquiroz:
:+1:

PR #724, opened 2015-11-27
REL-2500: Update F2 K-long filter label

fnussber:
The title says it all. Low impact last minute change.


cquiroz:
:+1:

cquiroz:
Does it have to change on the OT too?

fnussber:
It has already been changed in the OT, however I forgot to change it on the web page. A future ITC web app will maybe leverage some form of a web framework to generate those pages from the data available so that these things don't need to be done manually.

PR #725, opened 2015-11-27
REL-2567: Default sky fibers for IFU in ITC web form

fnussber:
Set default values for IFU sky fibres on web pages to 250 and add some explanatory text. Low impact late minute change.


cquiroz:
Do you need to change it in the OT too? In any case :+1:

fnussber:
The OT already calculates those values based on the current GMOS configuration. A future ITC web app will maybe leverage some form of a web framework to calculate the default values on the web page dynamically. Or just use javascript. ;)

PR #726, opened 2015-11-27
REL-2568: Add ITC support for GMOS mixed filters

fnussber:
A set of dat files for GMOS mixed filters was missing in the ITC resources. Adding the dat files allows to support those filters without any changes to the code. Low risk last minute change.


cquiroz:
:+1:

fnussber:
Ok, actually, I also need to make those filters available on the GMOS-N and GMOS-S web pages. Done.

PR #727, opened 2015-11-28
REL-2573: Fix ITC GMOS IFU-2 slit separation

fnussber:
The IFU-2 slit separation was using a wrong input value and therefore produced wrong results. The value needed is the "lines per mm" value of the GMOS dispersers/gratings, which we currently don't have.

To solve this issue I added the missing value to the different GMOS dispersers and fed it into the calculation of the split separation which now delivers the expected results.

Fiddling around with the enums in the base model is of course a relatively risky operation at this point, however I can not immediately think of any problematic side-effect this particular change could have (?), so I assume this is overall still a relatively low risk change since the added values are new and are not used anywhere else.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gmos/GmosNorthType.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gmos/GmosNorthType.java
@@ -1,16 +1,6 @@
-//
-// $Id: GmosNorthType.java 44720 2012-04-24 15:45:03Z nbarriga $
-//
-
 package edu.gemini.spModel.gemini.gmos;
 
 import edu.gemini.spModel.type.*;
-import edu.gemini.spModel.gemini.gmos.GmosCommonType.DetectorManufacturer;
-
-import java.util.Set;
-import java.util.HashSet;
-import java.util.Arrays;
-import java.util.Collections;
 
 /**
  *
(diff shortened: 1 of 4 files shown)


bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gmos/GmosNorthType.java
```diff
@@ -80,27 +80,29 @@ public StageModeNorth parse(String name, StageModeNorth defaultValue) {
...
-        B1200_G5301("B1200_G5301", "B1200"),
-        R831_G5302("R831_G5302", "R831"),
-        B600_G5303("B600_G5303", "B600") {public boolean isObsolete() {return true;}},
-        B600_G5307("B600_G5307", "B600"),
-        R600_G5304("R600_G5304", "R600"),
-        R400_G5305("R400_G5305", "R400"),
-        R150_G5306("R150_G5306", "R150"),
-        R150_G5308("R150_G5308", "R150"),
+    public enum DisperserNorth implements GmosCommonType.Disperser, ObsoletableSpType {
+        MIRROR("Mirror", "mirror", 0),
```
fnussber: You could argue that the lines per mm value should be an optional since the mirror doesn't have it. However I think the actual issue here is that the mirror isn't a disperser/grating and if anything should be an optional value then it is the disperser element. *Every* disperser element must have a "lines per mm" value. So I was a bit reluctant to model this as an optional in order to deal with a design flaw(?) somewhere else.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gmos/GmosNorthType.java
```diff
@@ -80,27 +80,29 @@ public StageModeNorth parse(String name, StageModeNorth defaultValue) {
...
         ;
 
         /** The default Disperser value **/
         public static DisperserNorth DEFAULT = MIRROR;
 
-        private String _displayValue;
-        private String _logValue;
+        private final String _displayValue;
+        private final String _logValue;
+        private final int _linesPerMm;
```
arturog8m: I think it would be best to name this what it actually represents, not its units. Looking at http://www.gemini.edu/sciops/instruments/gmos/spectroscopy-overview/gratings, looks like we are trying to model the Gmos gratings' "Ruling Density", which happens to be measured in lines per mm. So, my initial thought is that this should be called "rulingDensity" or something like that. As it is, it's like calling "metersPerSecond" something that is actually representing velocity. What do you think?

andrewwstephens:
:+1:  That sounds good to me.  I'll take the blame for propagating the poor variable name.

fnussber:
Renamed as suggested `linesPerMm` => `rulingDensity`.

arturog8m:
Looks good, :+1:  - I'll go ahead and merge this one so it goes in our (hopefully) final test release today.

PR #728, opened 2015-11-28
OCSADV-500: Display the aperture size in ITC Signal plot title

fnussber:
From the JIRA task: The 2015A infrared ITCs (F2, GNIRS, NIFS, NIRI) displayed the aperture size in the spectroscopy "Signal and sqrt(Background)" plot title. It would be useful to display this information in all of the 2016A ITC spectroscopy plot titles to remind users when multiple pixels are being summed. A suggested title is: "Signal & Sqrt(Background) summed in an aperture of N pix diameter"

I also streamlined the titles a little bit and added a line break for titles that are (too) long for a single line.

This is a low risk last minute change, low risk because it only impacts the chart titles, but no calculations etc.


arturog8m:
This looks fine to me :+1:  - I'll go ahead and merge it, so we make a new test release now.

PR #729, opened 2015-11-30
OCSADV-404: Import search target from the TPE

cquiroz:
This PR adds a button that will reload the currently selected target from the TPE back into the CQT, thus allowing the user, e.g. to drag the target and have that into the CQT.
I named the button "Sync from Obs." but I'm not that happy with the name. I'm open to suggestions. On the old Catalog Navigator this was called "From Image".

Here is a screenshot:

 
screenshot 2015-11-30 10 32 42

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/edu/gemini/catalog/ui/QueryResultsFrame.scala
+++ b/bundle/jsky.app.ot/src/main/scala/edu/gemini/catalog/ui/QueryResultsFrame.scala
@@ -15,6 +15,7 @@ import edu.gemini.catalog.ui.tpe.CatalogImageDisplay
 import edu.gemini.catalog.votable._
 import edu.gemini.pot.sp.SPComponentType
 import edu.gemini.pot.ModelConverters._
+import edu.gemini.shared.util.immutable.ScalaConverters._
 import edu.gemini.shared.gui.textComponent.{SelectOnFocus, TextRenderer, NumberField}
 import edu.gemini.shared.gui.{ButtonFlattener, GlassLabel, SortableTable}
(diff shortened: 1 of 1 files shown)


arturog8m:
What is CQT?

cquiroz:
Catalog Query Tool

arturog8m:
Something on the UI doesn't work for me. The "+" plus button floating above the new "Sync from Obs." looks strange. I don't really have a good suggestion, however.. maybe putting it next to the Query button will be better? Or improving the "+" button to make sure it's clear what that thing is going to do. I'm sorry I don't have a better suggestion, but as it is looks weird in my opinion

arturog8m:
I wonder if the option to sync from the TPE should be in the TPE itself...

tpolecat:
Code looks fine. No opinion about UI. :+1:

arturog8m:
W/respect to the button label - what about "Load from Image"? At least we call "Image" the TPE in the main toolbar in the OT.

cquiroz:
I agree on the UI, I'm not that happy either with it. I'll think ways to improve it.
About the label, now I'm thinking "Reset from Obs." because you are resetting the form

arturog8m:
I think we can discuss the label later  - I'm not fond of using abbreviations in buttons, I'd rather put just Reset and a tooltip with a full description to what the button actually does - However, I don't see reasons to keep this PR open because of that :+1:

cquiroz:
The UI has improved a bit (I think) by adding  a separator for the buttons, renaming the button to "Reset" and move it to the right

What do you think?

screenshot 2015-12-02 09 54 52

arturog8m:
:+1:

PR #730, opened 2015-11-30
REL-2165: reapply nighttime cal should not lose observing conditions

swalker2m:
This PR addresses an issue wherein the observing conditions would be reset upon template reapply. The previous PR (feature/REL-2165) updated template instantiation to copy the science observing conditions for nighttime calibrations as well as science and acquisitions.  It did not update template reapply though, so nighttime calibrations would lose observing conditions when updated.  In discussions with Bryan, it was decided that not only should conditions be kept on reapply, but also any changes to the target component or position angle.


arturog8m:
Looks good, and great you added tests! :+1:

PR #732, opened 2015-12-01
REL-2576: Hide "Peak Pixel" columns and warnings in ITC

fnussber:
This PR hides/removes peak pixel count related values in the ITC tables in the OT and on the web pages output.

We went back and forward on this a little bit but finally decided that the values and warnings as they are now are confusing and not what users want to see because of the way the peak pixel values are currently calculated.

Sorry for this being so late, but I think the risk associated is relatively low because this PR really just removes some table columns and print statements and also disables a set of warnings but has no impact on any calculations. (However the regression test baseline needs to be updated because of the changes in the web pages output.)


arturog8m:
Looks good, and low risk. If we merge this, we will make an official release with it, I'm waiting for confirmation from Andy that he saw a preview of these changes in dev, to avoid delays in the installation testing tomorrow.

andrewwstephens:
:+1:

PR #734, opened 2015-12-01
OCSADV-503: Replace legacy catalog image loading classes (aka remove reflection)

cquiroz:
This PR replaces legacy code used to load images on the TPE. On the old catalog image loading was abstracted like if it were another catalog. In my opinion the abstraction is overkill and I have a new class hierarchy for image loading separate from the targets catalog. This can be found mainly on ImageCatalog.scala. Configuration is no longer in xml, instead the image servers are scala objects
As a side effect I'm able to remove a lot of legacy code that is no longer needed


bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/image/ImageCatalog.scala
```diff
@@ -0,0 +1,62 @@
...
+  private val SKY_USER_CATALOG = "jsky.catalog.sky"
+
+  val all = List(DssGeminiNorth, DssGeminiSouth, DssESO, Dss2ESO, Dss2iESO, MassImgJ, MassImgH, MassImgK)
+
+  val defaultImageServer = DssGeminiNorth
+
+  def user = all.find(_.id == Preferences.get(SKY_USER_CATALOG, defaultImageServer.id)).getOrElse(defaultImageServer)
+
+  def user(is: ImageCatalog) = Preferences.set(SKY_USER_CATALOG, is.id)
+}
```
tpolecat: Please add return types and doc for these methods, since they're side-effecty and have an unadvertised dependency on `Preferences`.

bundle/jsky.app.ot/src/main/java/jsky/image/gui/DivaMainImageDisplay.java
```diff
@@ -692,6 +693,7 @@ protected void saveHistory(boolean merge) {
                 l = _historyList;
             Preferences.getPreferences().serialize(HISTORY_LIST_DATA_NAME, ImageHistoryItem.map(l));
         } catch (Exception e) {
+            // Ignore
         }
```
tpolecat: Ignore? Really? Nothing to see here?

tpolecat:
Brilliant. :100:

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/image/ImageCatalog.scala
cquiroz: That's a good comment, They don't look very side-effecting, but they are indeed. `def user` should at least have a `()`

bundle/jsky.app.ot/src/main/java/jsky/image/gui/DivaMainImageDisplay.java
cquiroz: Lazy me, I will add at least logging

bundle/jsky.app.ot/src/main/java/jsky/catalog/gui/CatalogNavigator.java
```diff
@@ -1,1243 +0,0 @@
...
-    /**
-     * Return the top level catalog directory to use based on the value of the
-     * <em>jsky.catalog.directory</em> system property, which may be set to the name of
-     * the class implementing the CatalogDirectory interface. The default is
-     * to use the {@link jsky.catalog.astrocat.AstroCatConfig} class (The
-     * {@link jsky.catalog.skycat.SkycatConfigFile} class is another alternative).
-     * The static method in the given class named <em>getDirectory()</em> is called to
-     * return a reference to the top level CatalogDirectory.
-     */
-    public static CatalogDirectory getCatalogDirectory() {
```
cquiroz: This method was guilty of many reflection issues in the past. It is not in use anymore

PR #735, opened 2015-12-02
REL-2576B: Show peak pixel values and warnings for spectroscopy.

fnussber:
This PR re-enables the peak pixel values and saturation and linearity warnings on the web output and in the OT ITC tables for spectroscopy.

We turned this temporarily off for the release because the peak pixel values that are calculated are wrong/misleading for spectroscopy but there was no simple way to fix this. There is a separate JIRA task that will deal with fixing this issue for good sometime in the future.


arturog8m:
I'm a bit confused. If we revert these changes, we are assuming we will do additional work on this before the next release - otherwise we will need to revert the code once again. Why not wait and revert these changes when we actually do the actual fix for the issue that made us revert this change in the first place?

fnussber:
I was assuming we were going to tackle this til March, but you're right, that might not be the case.

PR #736, opened 2015-12-02
OCSADV-511: template group sync bug

swalker2m:
This PR fixes a bug in which the template group version information is incremented whenever anything in the template group changes.  The bug arrises from a merge correction to prevent two distinct template groups created in distinct OTs from having the same `VersionToken` number.  However, the template group node version should be updated only if the `VersionToken` is in fact changed.


tpolecat:
parsing...

image

bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs2/TemplateNumberingCorrection.scala
```diff
@@ -21,8 +21,14 @@ class TemplateNumberingCorrection(lifespanId: LifespanId, nodeMap: Map[SPNodeKey
     def correct(tokens: List[(VersionToken, SPNodeKey)]): TryVcs[MergePlan] = {
       def updateToken(tgLoc: TreeLoc[MergeNode], vt: VersionToken): TryVcs[TreeLoc[MergeNode]] =
         tgLoc.getLabel match {
-          case m@Modified(_, _, tg: TemplateGroup, _, _) =>
-            TryVcs(tgLoc.modifyLabel(_ => m.copy(dob = tg.copy <| (_.setVersionToken(vt)))))
+          case m@Modified(_, nv, tg: TemplateGroup, _, _) =>
+            if (tg.getVersionToken == vt)
```
tpolecat: This might read better as two cases with a guard on the first one. But it's fine as-is.

tpolecat:
Thanks for adding a test. :+1:

arturog8m:
giphy

Ship it!

PR #738, opened 2015-12-04
REL-2581: Configure PIT to 2016B Semester

cquiroz:
This PR sets the build for the 2016B PIT and adds the initial upconversion code for older proposals

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -6,7 +6,7 @@ organization in Global := "edu.gemini.ocs"
 
 ocsVersion in ThisBuild := OcsVersion("2016A", true, 1, 1, 1)
 
-pitVersion in ThisBuild := OcsVersion("2016A", false, 1, 1, 0)
+pitVersion in ThisBuild := OcsVersion("2016B", true, 2, 1, 0)
 
 // Bundles by default use the ocsVersion; this is overridden in bundles used only by the PIT
 version in ThisBuild := ocsVersion.value.toOsgiVersion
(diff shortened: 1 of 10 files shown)


build.sbt
```diff
@@ -6,7 +6,7 @@ organization in Global := "edu.gemini.ocs"
 
 ocsVersion in ThisBuild := OcsVersion("2016A", true, 1, 1, 1)
 
-pitVersion in ThisBuild := OcsVersion("2016A", false, 1, 1, 0)
+pitVersion in ThisBuild := OcsVersion("2016B", true, 2, 1, 0)
```
sraaphorst: Just out of curiosity: why will the version numbering for 2016B start at 2.1 and not 1.1?

sraaphorst:
Looks fine to me. Annoying about having to translate back and forth between all those mutable / immutable objects! :+1:

build.sbt
cquiroz: Historically we have made it so that the PIT is [Year].[Semester].[Minor] where Semester is either 1 or 2. In this case being 2016B we make it 2016B.2.1. We may review this policy for this round

andrewwstephens: Your explanation makes sense, but it does seem strange to start with version 2...

cquiroz: Right, the original problem is that OSGi doesn't support letters on the version number and I wanted to have an easy way to tell the semesters apart. So to make OSGi happy we made the bundle version 2016.2.1 which is easy to parse by humans too

PR #739, opened 2015-12-04
REL-2565: Update F2 K-long filter label in PIT

cquiroz:
This PR fixes the K-long typo and adds upconversion logic for legacy proposals using the bogus value


tpolecat:
:+1: very happy there's a test!

PR #743, opened 2015-12-10
REL-2412A: PIT duplication check with the in-house archive

cquiroz:
To query GSA from the PIT, I'd like to reuse the query methods build for the datamanager. However, we wouldn't want to include the datamanager in the PIT, so this PR splits the `edu.gemini.dataman.app` bundle into two, one containing `dataman` itself and the other the methods to query GSA, This new bundle is called `edu.gemini.gsa.query`

It is more or less as straight package split between the bundles

Please comment if we agree with the approach in general

### DIFF
--- a/bundle/edu.gemini.dataman.app/src/main/scala/edu/gemini/dataman/app/GsaPostAction.scala
+++ b/bundle/edu.gemini.dataman.app/src/main/scala/edu/gemini/dataman/app/GsaPostAction.scala
@@ -1,9 +1,8 @@
 package edu.gemini.dataman.app
 
 import edu.gemini.dataman.core._
-import edu.gemini.gsa.core.{QaResponse, QaRequest}
 
-import edu.gemini.gsa.query.{GsaResponse, GsaAuth, GsaQaUpdateQuery, GsaHost}
+import edu.gemini.gsa.query._
 import edu.gemini.pot.spdb.IDBDatabaseService
 import edu.gemini.spModel.core.Site
 import edu.gemini.spModel.dataset.QaRequestStatus.ProcessingPost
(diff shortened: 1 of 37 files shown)


bundle/edu.gemini.dataman.app/src/main/scala/edu/gemini/dataman/app/GsaPostAction.scala
```diff
@@ -1,7 +1,9 @@
 package edu.gemini.dataman.app
 
 import edu.gemini.dataman.core._
-import edu.gemini.dataman.query.{GsaQaUpdateQuery, GsaResponse}
+import edu.gemini.gsa.core.{QaResponse, QaRequest}
+
+import edu.gemini.gsa.query.{GsaResponse, GsaAuth, GsaQaUpdateQuery, GsaHost}
```
swalker2m: In the new bundle you preserve the `core` vs `query` split that the data manager has.  We might want to rearrange it a bit thinking about which parts are considered part of the public API -- in the data manager there is no public API and nothing is exported but this bundle is being made for use from multiple bundles.  Or perhaps we just dump it all into one package?  There's not much to it after all.

I was looking for the bundle's `build.sbt` to see what is exported vs. not exported but it's not part of this PR?

`GsaHost` moves into `query` from `core` was that intentional?

cquiroz:
Added bundle's build.sbt

cquiroz:
On terms of the public API I basically want to have access to `GsaRecordQuery` and any related classes needed including `GsaHost`, `GsaRecord`, etc

swalker2m:
I think maybe just put all of it in one package (?) but otherwise :+1:

PR #744, opened 2015-12-11
Feature/ocsadv 391

jluhrs:
Implemented support for FLAMINGOS-2


jluhrs:
I'll just go ahead and merge this code, as it does not affect other applications. Then we can work on improving it.

PR #746, opened 2015-12-14
REL-2582: Replace PIT target model.

cquiroz:
This PR continues the task OCSADV-155 of replacing the PIT model with the new target model. In this particular PR the `Coordinates` class and related classes are replaced.
Though this is not strictly part of the PIT 2016B package, this change is required to implement REL-2412 and REL-2388
This PR will probably have side effects on ITAC that need to be dealt on before the next release (REL-2589)


tpolecat:
Nice. :+1:

PR #747, opened 2015-12-15
Reduction on warnings

cquiroz:
This PR removes a few hundred warnings and dead code from the older modules.
There is no attempt to fully upgrade the code to modern standards but at least we'd avoid so many warnings. There are still quite a few warnings to go :cry:


fnussber:
This is *massive*. Super awesome.

bundle/edu.gemini.shared.gui/src/main/java/edu/gemini/shared/gui/calendar/LunarDateSymbolSupplier.java
```diff
@@ -128,20 +105,20 @@ public int compareTo(Object o) {
 
         int _yearCount;
 
-        List _years;  // each entry is a List of Months
+        List<List<List<DateSymbol>>> _years;  // each entry is a List of Months
```
sraaphorst: Ugh. How horrible this is!
Having it properly typed, though, will of course make it much easier to understand what is going on.

sraaphorst:
Wow. This is incredibly impressive.

I think we need to all send YOU a cake this Friday and then be forced to watch you savour it in thanks and humility for all the work you put into this.

PR #748, opened 2015-12-15
seqexec config syntax

tpolecat:
This cleans up the syntax for extracting a value from a config, so instead of

```scala
extract[FPUnit](config, a)
```

You say

```scala
config.extract(a).as[FPUnit]
```

Debatable whether this is an actual improvement but I think it reads better.

### DIFF
--- a/bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/TCS.scala
+++ b/bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/TCS.scala
@@ -90,7 +90,8 @@ object TCS {
   }
 
   // Helper functions for incrementally building a TCS configuration
-  def buildWithConjunction[T, P, S](f: P => T => T)(p: S \/ P): T => T = p.map(f).getOrElse((t: T) => t)
+  def buildWithConjunction[T, P, S](f: P => T => T)(p: S \/ P): T => T = 
+    p.map(f).getOrElse((t: T) => t)
 
   def buildWithConjunction[T, P, Q, S](f: (P, Q) => T => T)(p: S \/ P, q: S \/ Q): T => T = ( for {
       r <- p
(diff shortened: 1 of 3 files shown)


tpolecat:
BTW @arturog8m it's a little easier if you comment on the PR rather than on the commit. Not a big deal but it saves some clicking around.

arturog8m:
Sounds good! :+1: 

Agreed on commenting on the PR instead of the commit - I meant to do that, but I didn't realized I was on the wrong place.

bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/TCS.scala
```diff
@@ -137,13 +138,13 @@ object TCS {
 
   def fromSequenceConfig(config: Config)(s0: TcsConfig): TcsConfig = {
     List(
-      build(buildPwfs1Config, new ItemKey(TELESCOPE_KEY, GUIDE_WITH_PWFS1_PROP), config),
-      build(buildPwfs2Config, new ItemKey(TELESCOPE_KEY, GUIDE_WITH_PWFS2_PROP), config),
-      build(buildOiwfsConfig, new ItemKey(TELESCOPE_KEY, GUIDE_WITH_OIWFS_PROP), config),
-      build(buildOffsetConfig, new ItemKey(TELESCOPE_KEY, P_OFFSET_PROP), new ItemKey(TELESCOPE_KEY, Q_OFFSET_PROP),
-        config)
+      build(buildPwfs1Config,  TELESCOPE_KEY / GUIDE_WITH_PWFS1_PROP, config),
```
cquiroz: Where are all the capital letter constants defined? If on the scala side we may rather use the capitalization of `TelescopeKey`, right?

tpolecat: Yes, I will look at caps in another PR.

tpolecat:
Ok I added the property descriptor thing.

cquiroz:
Looks good :+1:

PR #750, opened 2015-12-16
Replace A => A with Endo[A]

tpolecat:
It's fairly common to have a "pipeline" of functions of type `A => A` and the identity value `a => a` for a "no-op", and this is what we see in this code that manipulates a `TcsConfig`. 

It turns out that we can do this more easily by using `Endo[A]` which simply wraps `A => A` and provides a monoid where `|+|` is `andThen` and `zero` is the identity function.

### DIFF
--- a/bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/TCS.scala
+++ b/bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/TCS.scala
@@ -14,7 +14,7 @@ import scala.language.implicitConversions
 import scala.reflect.ClassTag
 import scalaz.concurrent.Task
 import scalaz._, Scalaz._
-import squants.space.{Millimeters, LengthConversions}
+import squants.space.{ Millimeters, LengthConversions } 
 
 /**
  * Created by jluhrs on 4/23/15.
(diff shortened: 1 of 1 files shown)


bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/TCS.scala
```diff
@@ -136,15 +126,13 @@ object TCS {
...
+  def fromSequenceConfig(config: Config)(s0: TcsConfig): TcsConfig =
     List(
       build(buildPwfs1Config,  TELESCOPE_KEY / GUIDE_WITH_PWFS1_PROP, config),
       build(buildPwfs2Config,  TELESCOPE_KEY / GUIDE_WITH_PWFS2_PROP, config),
       build(buildOiwfsConfig,  TELESCOPE_KEY / GUIDE_WITH_OIWFS_PROP, config),
       build(buildOffsetConfig, TELESCOPE_KEY / P_OFFSET_PROP, TELESCOPE_KEY / Q_OFFSET_PROP, config)
-    ).foldLeft(s0)((b, f) => f(b))
-
-  }
+    ).suml.apply(s0)
```
cquiroz: Does this takes all the `Endo`s and `|+|` them together, and then apply `A=>A` to `s0`?

tpolecat: Correct.

sraaphorst:
Very cool to see endomorphism monoids in scalaz. I don't understand the code in its entirety, but based on the concept and what I do grok, :+1:

cquiroz:
:+1: too

PR #751, opened 2015-12-16
OCSADV-515: Target Editor BAGS changes and cleanup

sraaphorst:
This is a major cleanup with some changes to the target editor in preparation for BAGS.
The decision was made that the use of GuideGroups will now be mandatory and more explicit, so that BAGS acquired target(s) can all be placed in an explicit guide group.

In preparation for this, I overhauled the target editor to always display guide groups whenever any guide probe targets are listed, and prevented the collapsing of guide group nodes in the target editor tree.

In the past, when a single grouping of guide probe targets was provided, while in the data model, they belonged to a guide group, the guide group was not explicitly shown in the target editor:
old - one group

Now, as you can see, the guide group is always shown:
new - one group

Additionally, here is an image showing the old target editor with multiple guide groups.
You can see that the guide groups are collapsible (with one being collapsed):
old - collapsed groups

Now, collapsing is no longer permissible, and all guide groups will always be displayed, along with icons identifying them as guide groups (the folders):
new - multiple groups

In addition to these changes, there was a general mish-mash of code for drag and dropping of targets where in a few places, the code used arrays to try to allow for drag and drop of multiple targets. Then, since we only allow (in the UI and in the actual drop logic) ONE target to be dragged and dropped at a time, I simplified the code to no longer be a hybrid managing the possibility of multiple targets while in reality only handling one target by moving everything to only handling one target.

The rest of the changes are essentially a cleanup. I am sorry for the horrible diff that you'll see between the versions of `EdCompTargetList`, but in the old code, internal classes and listener members were littered throughout the class without rhyme and reason, making the code (at least, for me) hideously unpleasant to navigate. I have tried to conglomerate all of the methods together with some refactoring and rewriting, and then placed all the listener members and classes at the end of the file. Due to the major reordering, though, this will invariably look awful here in github.

Also note that the documentation for the `JXTreeTable` that is used (an external custom provided Swing widget) is horrible, and due to forcing the guide groups to always be open required substantial modification to the existing code in `TelescopePosTableWidget`.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/TelescopePosWatcher.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/TelescopePosWatcher.java
@@ -6,9 +6,7 @@ package edu.gemini.spModel.target;
  */
 @FunctionalInterface
 public interface TelescopePosWatcher {
-
     void telescopePosUpdate(WatchablePos tp);
-
 }
 
 
--- a/bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
+++ b/bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
@@ -11,7 +11,6 @@ import edu.gemini.spModel.pio.ParamSet;
 import edu.gemini.spModel.pio.xml.PioXmlFactory;
 import edu.gemini.spModel.target.SPTarget;
 import edu.gemini.spModel.target.TelescopePosWatcher;
-import edu.gemini.spModel.target.WatchablePos;
 import edu.gemini.spModel.target.env.*;
 import edu.gemini.spModel.target.obsComp.TargetObsComp;
 import edu.gemini.spModel.target.obsComp.TargetSelection;
@@ -57,7 +56,7 @@ public final class EdCompTargetList extends OtItemEditor<ISPObsComponent, Target
         _w = new TelescopeForm(this);
 
         _w.addComponentListener(new ComponentAdapter() {
-            public void componentShown(ComponentEvent componentEvent) {
+            public void componentShown(final ComponentEvent componentEvent) {
                 toggleAgsGuiElements();
             }
         });
@@ -76,7 +75,7 @@ public final class EdCompTargetList extends OtItemEditor<ISPObsComponent, Target
         );
 
         _w.guideGroupName.addWatcher(new TextBoxWidgetWatcher() {
-            public void textBoxKeyPress(TextBoxWidget tbwe) {
+            @Override public void textBoxKeyPress(final TextBoxWidget tbwe) {
                 SwingUtilities.invokeLater(() -> {
                     final String name = _w.guideGroupName.getText();
                     final GuideGroup newGroup = _curGroup.setName(name);
@@ -93,13 +92,10 @@ public final class EdCompTargetList extends OtItemEditor<ISPObsComponent, Target
                     _w.guideGroupName.requestFocus(); // otherwise focus is lost during event handling
                 });
             }
-            public void textBoxAction(TextBoxWidget tbwe) {
-            }
         });
 
         _w.positionTable.addKeyListener(new KeyAdapter() {
-            @Override
-            public void keyPressed(KeyEvent event) {
+            @Override public void keyPressed(final KeyEvent event) {
                 switch (event.getKeyCode()) {
                     // Make the delete and backspace buttons delete selected positions.
                     case KeyEvent.VK_DELETE:
@@ -114,7 +110,7 @@ public final class EdCompTargetList extends OtItemEditor<ISPObsComponent, Target
 
     }
 
-    @Override protected void updateEnabledState(boolean enabled) {
+    @Override protected void updateEnabledState(final boolean enabled) {
         if (enabled != isEnabled()) {
             setEnabled(enabled);
             updateEnabledState(getWindow().getComponents(), enabled);
@@ -136,7 +132,7 @@ public final class EdCompTargetList extends OtItemEditor<ISPObsComponent, Target
         _w.newMenu.setEnabled(enabled && inst != null);
     }
 
-    protected void updateDetailEditorEnabledState(boolean enabled) {
+    protected void updateDetailEditorEnabledState(final boolean enabled) {
         // Update enabled state for all detail widgets.
         _w.detailEditor.allEditorsJava().stream().forEach(ed -> updateEnabledState(new Component[]{ed}, enabled));
     }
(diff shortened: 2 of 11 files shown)


bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/TelescopePosWatcher.java
```diff
@@ -11,10 +11,9 @@
  * An interface supported by clients of TelescopePos who want to
  * be notified when the positions changes in some way.
  */
+@FunctionalInterface
```
cquiroz: :+1:

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
```diff
@@ -146,163 +142,64 @@ protected void updateDetailEditorEnabledState(boolean enabled) {
...
-                }
-                return new Some<>(new TargetClipboard(group));
-            }
-            return new Some<>(new TargetClipboard(target));
-        }
+    // Common code to manage the position watcher on the current position around an action that modifies it.
+    // We also check for the presence of a specified target in the target environment.
+    private void manageCurPosIfEnvContainsTarget(final SPTarget target, final Runnable action) {
+        final TargetObsComp obsComp = getDataObject();
+        if (obsComp == null) return;
```
cquiroz: So many null checks, I wish this were Scala

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
```diff
@@ -317,29 +214,22 @@ public void init() {
...
+        _w.positionTable.reinit(obsComp);
         _w.guidingControls.manualGuideStarButton().peer().setVisible(GuideStarSupport.supportsManualGuideStarSelection(getNode()));
         updateGuiding();
         _agsPub.watch(getContextObservation());
     }
 
     // OtItemEditor
-    protected void cleanup() {
+    @Override protected void cleanup() {
         _agsPub.watch(null);
```
cquiroz: This looks weird, what does it mean to watch `null`?

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
```diff
@@ -372,68 +258,84 @@ private void updateGuiding(final TargetEnvironment env) {
...
-                }
-            } else {
-                final GuideGroup grp = TargetSelection.getGuideGroup(env, node);
-                if (grp != null) if (getDataObject() != null) {
-                    final TargetEnvironment env1 = getDataObject().getTargetEnvironment();
-                    if (env1.getGroups().contains(grp)) {
 
-                        if (_curPos != null) _curPos.deleteWatcher(posWatcher);
+    private static boolean enablePrimary(final SPTarget target, final TargetEnvironment env) {
+        return env.getBase() != target && !env.getUserTargets().contains(target);
```
cquiroz: `!equals`?

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
```diff
@@ -520,129 +418,133 @@ public void actionPerformed(ActionEvent actionEvent) {
...
+            return lab;
+        }
+    };
 
-        // Get the set of guiders that are referenced but not legal in this context, if any.  Any
-        // "available" guider is legal, anything left over is referenced but not really available.
-        final Set<GuideProbe> illegalSet = env.getOrCreatePrimaryGuideGroup().getReferencedGuiders();
-        illegalSet.removeAll(avail);
+    private final TelescopePosWatcher posWatcher = tp -> {
+        if (tp != _curPos) {
```
cquiroz: `!equals`?

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
```diff
@@ -764,6 +659,60 @@ public void actionPerformed(ActionEvent e) {
...
+                spTarget.setTarget(target.clone());
+                spTarget.setMagnitudes(mag);
+            } else if (group != null && this.group != null) {
+                final GuideGroup newGroup = group.setAll(this.group.cloneTargets().getAll());
+                // XXX TODO: add a helper method in the model to replace a guide group
+                final TargetEnvironment env = dataObject.getTargetEnvironment();
+                final GuideEnvironment ge = dataObject.getTargetEnvironment().getGuideEnvironment();
+                final ImList<GuideGroup> options = ge.getOptions();
+                final ArrayList<GuideGroup> list = new ArrayList<>(options.size());
+                options.foreach(gg -> list.add(gg == group ? newGroup : gg));
```
cquiroz: `equals`?

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/TelescopePosTableWidget.java
```diff
@@ -227,36 +189,39 @@ private static Icon bagsIcon(boolean enabled) {
...
+        private static Option<AgsGuideQuality> guideQuality(final Option<Tuple2<ObsContext, AgsMagnitude.MagnitudeTable>> ags,
+                                                            final GuideProbe guideProbe, final SPTarget guideStar) {
             return ags.flatMap(tup -> {
                 if (guideProbe instanceof ValidatableGuideProbe) {
+                    final ObsContext ctx = tup._1();
+                    final AgsMagnitude.MagnitudeTable magTable = tup._2();
                     final ValidatableGuideProbe vgp = (ValidatableGuideProbe) guideProbe;
-                    return AgsRegistrar.instance().currentStrategyForJava(tup._1()).map(strategy -> {
-                        final Option<AgsAnalysis> agsAnalysis = strategy.analyzeForJava(tup._1(), tup._2(), vgp, ModelConverters.toSideralTarget(guideStar));
+                    return AgsRegistrar.instance().currentStrategyForJava(ctx).map(strategy -> {
```
cquiroz: Would this ever be called for GEMS? I ask because we still have 2 ways to call Gems AGS, the one you are using should be preferred but in many places we are still using the legacy one

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/TelescopePosTableWidget.java
```diff
@@ -349,16 +319,13 @@ public int indexOf(final SPTarget target) {
...
                 ++index;
-                // don't count collapsed tree nodes!
-                if (treeTable.isExpanded(path)) {
-                    for (Row row2 : row.children()) {
-                        if (row2.target().getOrNull() == target) return index;
-                        ++index;
-                    }
+
+                for (final Row row2 : row.children()) {
+                    if (row2.target().getOrNull() == target) return index;
```
cquiroz: `equals`?

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/TelescopePosTableWidget.java
```diff
@@ -368,33 +335,29 @@ public int indexOf(final GuideGroup group) {
             if (group == null) return -1;
 
             int index = 0;
-            for (Row row : rows) {
+            for (final Row row : rows) {
                 if (row.group().getOrNull() == group) return index;
```
cquiroz: `equals?`

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/TelescopePosTableWidget.java
```diff
@@ -472,40 +432,34 @@ public Object getValueAt(Object node, int columnIndex) {
     }
 
     // Return the world coordinates for the given target
-    private static Option<WorldCoords> getWorldCoords(SPTarget tp, Option<Long> when) {
+    private static Option<WorldCoords> getWorldCoords(final SPTarget tp, final Option<Long> when) {
```
cquiroz: Couldn't we migrate to the new `Coordinates` class?

cquiroz:
It is difficult to parse the whole code but I guess the results speak for themselves. It still seems a bit strange that you can't collapse a guide group, but I assume this was discussed with science, right?

andrewwstephens:
We discussed this in Hilo, and it was explained that guide groups are needed to separate automatic and manual guide stars.  In practice, pre-BAGS observations will have only manual guide stars, and >99% of post-BAGS observations will have only automatic guide stars.

swalker2m:
We discussed this yesterday briefly but for the record, I think we should get rid of the `JXTreeTable` altogether and the folder icons.  Then we can make the names of the targets/groups flush with the left and add a bit of a gap before the guide star target names to indicate nesting.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
sraaphorst: You're telling me :-).

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
sraaphorst: Looking at `AgsContextPublisher`, this seems to be a really horrible trick that basically says to:
1. Unwatch any currently watched observations; and
2. Clear out the `AgsContext` by setting it to null through a bit of ridiculousness using `ImOption` via a call to an `AgsContext.create` method with `null`, instead of simply doing a `null` check and if null, setting the `AgsContext` to `null` directly.

Seems completely straightforward, no? :laughing: 
This should probably be simplified to not be so bizarre, perhaps by the addition of an `unwatch` method or probably better by adding a `watch(Option[ISPObservation])` method, as calling with `None` seems like it makes a lot more sense and wouldn't introduce any issues with regards to the `AgsContext`.

There is also an unused method in `AgsContextPublisher`. I think I will look at simplifying this, and thanks for pointing it out.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
sraaphorst: In all of these cases, we actually DO want `==` or `!=`, as we are working directly with references to the same objects. If we used `equals` instead, we would end up having to compare watcher lists on the `SPTarget`s and the underlying `ITarget`s, which is computationally unnecessary, and indeed, it is even possible it could result in false positives for two freshly created `SPTargets` which would have identical freshly created `ITarget`s with regards to `equals`.

Absolutely contrary to programmer training and intuition :smile:

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/TelescopePosTableWidget.java
sraaphorst: Yes, this is definitely used for GeMS.
We do need to get rid of the legacy GeMS stuff soon, as having it is unnecessarily cluttering up things and adds to confusion.
(I know I was confused by this when doing some of the `BagsManager` coding.)

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
cquiroz: Right, that would be good, passing null to a method is certainly code smell

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
cquiroz: Ok, I make the mistake of using `==` in Java after so much scala, but in this case seems ok

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/TelescopePosTableWidget.java
sraaphorst: I don't see why not, but I would have to ask a couple of questions first. For example, the distance as displayed in the target editor represents what? The **angular distance** between the base position and the target?

I'm also not entirely sure how to use the `when` parameter with regards to `Coordinates`; right now, it is completely ignored in the `ITarget` anyways, so I suspect we could just ignore it entirely without issue and use `Coordinates` without worrying about it, but I'm not sure if that is right. Advice, anyone? @andrewwstephens or @tpolecat ?

sraaphorst:
As per the discussion with @swalker2m and @arturog8m I will aim to entirely eliminate the `JXTreeTable`. Since this is working, though, as it currently stands, as per my follow up with @arturog8m  I would like to merge these changes and then create a second PR attempting to drop the use of the hideous `JXTreeTable`. I'm just not certain on the surface how easy this will be.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/TelescopePosTableWidget.java
cquiroz: BTW, there is a method `Coordinates.angularDistance(that: Coordinates)`

andrewwstephens: @sraaphorst - Correct, the "Distance" in the Target Environment table is the angular distance in arcminutes between the base position and the target.

sraaphorst:
All changes made as per the PR comments, with substantial cleanup of code.
Merging this and will now continue work under this feature branch to eliminate the `JXTreeTable`.

PR #752, opened 2015-12-16
Immutable library cleanup and functional interfaces

sraaphorst:
I was feeling inspired by @cquiroz and his herculean efforts, so last night I tried to do some cleaning up of the edu.gemini.shared.util.immutable classes to get rid of template parameter warnings, exploit backing lists better for our `ImList`s, remove unnecessary comments, and add `@FunctionalInterface` annotations where it made sense to do so.

All test cases still pass and I don't think this will break anything (e.g. serializability), but please let me know if you see any issues.


cquiroz:
Very good :+1: You went beyond the types and set final to params, this is great

PR #753, opened 2015-12-17
REL-2412B: PIT duplication check with the in-house archive

cquiroz:
This PR replaces the legacy page-scrapping GSA client (That was really heroic) with a new one for the new archive using JSON, The API remains largely the same requiring minimal changes to the PIT
This probably needs some testing by the end-users as there are many options on the archive that this PR doesn't use


tpolecat:
This looks great. :+1:

PR #754, opened 2015-12-18
REL-2582B: Replace PIT target model

cquiroz:
This PR fixes a couple of bugs outputting `Declination` values, given that the underlying `Angle` doesn't support negatives. This requires new formatters on the output

### DIFF
--- a/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/editor/EphemerisElementEditor.scala
+++ b/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/editor/EphemerisElementEditor.scala
@@ -42,7 +42,7 @@ class EphemerisElementEditor(e: EphemerisElement) extends StdModalEditor[Ephemer
     })
     setValue(new Date(e.validAt))
   }
-  object RA extends RATextField(e.coords.ra.toAngle.toDegrees)
+  object RA extends RATextField(e.coords.ra)
   object Dec extends DecTextField(e.coords.dec)
   object Mag extends NumberField(e.magnitude, allowEmpty = false)
 
--- a/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/editor/EphemerisElementEditor.scala
+++ b/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/editor/EphemerisElementEditor.scala
@@ -1,16 +1,18 @@
 package edu.gemini.pit.ui.editor
 
-
 import edu.gemini.model.p1.immutable.EphemerisElement
 
 import edu.gemini.pit.ui.util._
 import edu.gemini.shared.gui.textComponent.NumberField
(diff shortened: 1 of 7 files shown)


bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/editor/EphemerisElementEditor.scala
```diff
@@ -41,22 +43,14 @@ class EphemerisElementEditor(e: EphemerisElement) extends StdModalEditor[Ephemer
     setValue(new Date(e.validAt))
   }
   object RA extends RATextField(e.coords.ra.toAngle.toDegrees)
-  object Dec extends DecTextField(e.coords.dec.toDegrees)
+  object Dec extends DecTextField(e.coords.dec)
```
sraaphorst: Just out of curiosity, why does the `DecTextField` take `e.coords.dec`, whereas the `RATextField` needs the `toAngle.toDegrees` conversion done?

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/util/RaDecFields.scala
```diff
@@ -23,9 +24,23 @@ class DegreeFormatter(caption: String, inRange: Option[Double => Boolean] = None
...
+
+  def fromString(s: String):Option[Declination] = try {
+    s.toDouble match {
+      case d if inRange.forall(f => f(d)) => Declination.fromAngle(Angle.fromDegrees(d))
+      case _                              => None
+    }
+  } catch {
+    case e:Exception             => None
+  }
+}
```
sraaphorst: I think all of this can be simplified using `Try` instead:
`def fromString(s: String): Option[Declination] = Try {
  s.toDouble match {
    case d if inRange.forall(f => f(d)) => Declination.fromAngle(Angle.fromDegrees(d))
}}.toOption`

should do the trick, I think.

sraaphorst: Ugh... sorry for the horrible formatting. I apparently didn't do that quite correctly.

cquiroz: Thanks for the suggestion, this is a copy of similar code, but I could use the format you suggest instead

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/editor/EphemerisElementEditor.scala
cquiroz: In the previous model both RA and Dec were represented as Doubles (Angle), I just translated this straight to the new model but this breaks on Dec given that it doesn't display correctly for negative angles. I should reconsider this and make `RATextField` operate on `RightAscencion` as you suggest

cquiroz:
Incorporated the suggestions from the PR, rebased and all test pass

PR #755, opened 2015-12-19

cquiroz:
This PR changes the phase 1 model of  Texes, Dssi and Phoenix to request the user to indicate which site the proposal belongs to. Included are model changes, UI updates on the decision tree, upconversion logic and multiple tests.
Most of the added lines are test files


tpolecat:
:+1: good.

PR #757, opened 2015-12-21
REL-2522: Implement backend monitors for hbfphase1-lv1

cquiroz:
This PR changes the p1-monitor configuration to run on the new phase1 host and adds support for FT and partner proposals
Note that the configuration will change once we move out of the testing phase


tpolecat:
:+1:

PR #758, opened 2015-12-21
seqexec companion objects

tpolecat:
This cleans up capitalization issues reported earlier and replaces singleton implementations with `object`s.


cquiroz:
:+1: Very good, very scalish

PR #763, opened 2015-12-28

cquiroz:
This PR adds several checks for PIT proposals mostly around conditions, LGS and band 3


tpolecat:
:+1: man the problem checking in P1 is so much nicer than P2 checking.

cquiroz:
Certainly it is, you can essentially read the checks

PR #765, opened 2015-12-29
Port to scala 2.11

cquiroz:
This PR is a proposal to port the OCS to scala 2.11. It is not fully finalized as there are some open issues but I wanted to share it to keep discussing tho pros/cons

The port moves scala from `2.10.4` to `2.11.7`. In general, this does not break a lot of code but it breaks several scala dependencies. They need to be changed to use `2.11.X` as well. As a general rule I used the minimal version that supports `2.11.7` trying not to introduce code changes

The following is a list of the dependencies changed:

1 **specs** Specs was changed to `2.3.12` which is the oldest one supporting scala 2.11.7. This change produces many code changes to the tests but it is workable.
2 **scalaz** This goes from `7.0.5` to `7.0.6` which supports scala `2.11.7`. This change produced no cascading code changes
3 **Argonaut** The same version is kept `6.0.4` but we use the `2.11.7` variant
4 **scala-xml** This is no longer part of the scala library, it must be imported as a regular lib at version `1.0.5`
5 **scala-parser-combinator** This is no longer part of the scala library, it must be imported as a regular lib at version `1.0.4`
6 **squants** Moved to version `0.6.1` including some of florian's patches and with OSGi headers
7 **scala-swing** Moved to use the official version for scala `2.11`. version `2.0.0-M2`

Libraries that require additional work
1 **slick** I haven't yet changed this as the versions supporting 2.11 bring a large API change, so we'd need to rework the modules depending on it, at least `edu.gemini.sp.vcs.log` and `edu.gemini.util.security`
~~2 **breeze** Updating to 2.11 breaks binary compatibility with breeze, I have updated this bit by bit and managed to get breeze `0.6.1` running correctly but we should update it up to version `0.8` to be fully `2.11` compatible~~
3 **scala-actors** Actors are deprecated and go to its own separate library. in a few modules, the support was removed and replaced by `Future` but in more extensive cases (like on the PIT) I kept the actors using the library version `2.11.7`

At this stage of the port all tests pass and I can build the PIT, p1pdfmaker and p1_monitor and run them successfully, however neither QPT, the OT or the SPDB are still running due to the dependencies on **slick** and **breeze**

**NOTE** As much as possible all dependencies are expressed as managed ones. Only the ones without original OSGi headers are kept as unmanaged


tpolecat:
We can probably upgrade to Slick 2 without a lot of work. This will be temporary; I want to move to doobie once we're on scalaz 7.1.

swalker2m:
I'd be in favor of going right to scalaz 7.1.  That would give us serialization of `\/` and `Zipper` and make the bags model refactor work less horrible.

cquiroz:
I'd approach the big upgrades piece-wise, one for scala 2.11, another for scalaz, etc. Seems less involved and risky. At least in this PR I tried to minimize code changes, most of the changes are due to API changes on specs2

swalker2m:
I discussed this with Rob last week.  The two H2 databases we use are manipulated with Slick 1 which doesn't work with Scala 2.11.  He'd like to move these to doobie, but that will require Scalaz 7.1.  Sebastian and I need Scalaz 7.1 for the new BAGS model as well since more of the data structures are serializable in 7.1.  Scalaz 7.1 works with Scala 2.10 as well as 2.11.  So we think the the plan should be to do the Scalaz 7.1 update first, then doobie and finally Scala 2.11.

arturog8m:
Sounds like the right thing to do. BAGS needs to be finished asap, and since that's dependent on Scalaz, I'd like to get that done first, and then upgrade the rest.

cquiroz:
I want scala 2.11 ASAP given that most of scala.js is compatible with 2.11 rather than 2.10.

cquiroz:
Rebased onto develop and ported the code to support breeze on its latest version 0.12 on scala 2.11

cquiroz:
Rebase onto develop and it now uses scalaz 7.1.6

cquiroz:
This should be complete now, all tests are running and I can build and start all the applications
I had some doubts about the library breeze that depends on a different shapeless version than doobie but I can launch the OT and run an AGS queries without big troubles
As much as possible this PR uses managed dependencies

project/src/main/scala/edu/gemini/osgi/tools/idea/IdeaProject.scala
```diff
@@ -157,11 +157,7 @@ class IdeaProject(idea: Idea, scalaInstance: ScalaInstance, imls: List[File]) {
       </component>
       {rootComponent(javaVersion)}
       <component name="ProjectRunConfigurationManager"/>
-      <component name="ScalaCompilerConfiguration">
-        <plugins>
-          <plugin path="$PROJECT_DIR$/../../../lib/compile/paradise_2.10.5-2.0.1.jar" />
-        </plugins>
-      </component>
+      <component name="ScalaCompilerConfiguration"/>
```
tpolecat: I am a little bit irritated that we are able to delete this so soon after I figured it out.

tpolecat:
This looks great. Thanks for all your work on this! :+1:

PR #766, opened 2015-12-30
Idea project generation update.

swalker2m:
A couple of minor updates to Idea project generation.

1) Update project and module XML files so that a conversion and backup is not performed on startup. (At least for Idea 14.  Further updates might be needed for Idea 15.)

2) Always replace module files when the project is generated regardless of whether they have been edited.  In practice not touching existing updated modules just meant having to delete all `.iml` files before regenerating the project file anyway.

### DIFF
--- a/project/src/main/scala/edu/gemini/osgi/tools/idea/IdeaProject.scala
+++ b/project/src/main/scala/edu/gemini/osgi/tools/idea/IdeaProject.scala
@@ -164,6 +164,7 @@ class IdeaProject(idea: Idea, scalaInstance: ScalaInstance, imls: List[File]) {
       </component>
       <component name="VcsDirectoryMappings"/>
       {libraryTable}
+      <component name="ScalaCompilerConfiguration" />
     </project>
 
   private def initialProject(javaVersion: String): xml.Elem =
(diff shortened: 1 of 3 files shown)


project/src/main/scala/edu/gemini/osgi/tools/idea/IdeaProject.scala
```diff
@@ -318,19 +319,20 @@ def project: xml.Elem =
...
-          {scalaInstance.allJars.map(jarUrl)}
-        </CLASSES>
-      </library>
-      <library name="scala-library">
-        <CLASSES>
-          {scalaInstance.allJars.map(jarUrl)}
-        </CLASSES>
+      <library name="scala-sdk" type="Scala">
+        <properties>
+          <option name="languageLevel" value="Scala_2_11" />
```
tpolecat: Hm what does this option do?

swalker2m: Unclear.  I think the corresponding Java setting controls which language features are supported by the editor.  At any rate, it should probably match the Scala version we're actually using so I updated it to read from `scala.util.Properties.versionNumberString`.

tpolecat:
I'm still on IDEA 13 ... guess I should upgrade. Anyway :+1:

PR #767, opened 2016-01-06
OCSADV-515B: Change Target Editor table to be a JTable instead of a JXTreeTable

sraaphorst:
(Sorry for this somewhat large and muddled PR. I did some cleanup and removal of unused code that I found as I made these changes, which probably should have been included in a separate PR.)

This extends the changes made in OCSADV-515 and deals with the target editor table `TelescopePosTableWidget` and related classes.

The idea here is that previously, due to the use of `GuideGroup`s, we needed the rows of the target editor table to act like a tree, with targets being assigned to a `GuideGroup`, which could be expanded or collapsed as desired. Additionally, if a single `GuideGroup` existed, it was "hidden" in the sense that the targets simply appeared in the table instead of explicitly showing them as falling under the `GuideGroup`. (Note that this change was made in the PR for OCSADV-515 already.)

With respect to discussions moving forward with BAGS, we have decided to always show `GuideGroup`s expanded, and BAGS targets will always appear in their own uneditable `GuideGroup`. Thus, the tree structure of the table is no longer necessary. Previously, `TelescopePosTableWidget` extended `JXTreeTable` (a third party class), but as this is no longer required, the main goal of this PR is to instead extend the standard Swing `JTable`.

Here is a before image, with the changes made to always show `GuideGroup`s, based on OCSADV-515 while still using the `JXTreeTable`:
old

And here is an example of what things look like after this PR, using the `JTable` instead. Note that the removal of the icons is deliberate for simplicity:
new


arturog8m:
I really like the new UI, looks much more clean, and I think this is a
major improvement. Well done!

bundle/edu.gemini.shared.gui/src/main/java/edu/gemini/shared/gui/TableUtil.java
```diff
@@ -1,311 +0,0 @@
...
-            public void actionPerformed(ActionEvent ae) {
-                jd.dispose();
-            }
-        });
-        jd.setLocationRelativeTo(null);   // Center the dialog on the window
-        return jd;
-    }
-
-}
-
```
swalker2m: RIP `TableUtil`.

sraaphorst:
Thanks, @arturog8m ! I appreciate it :smile:

bundle/edu.gemini.shared.gui/src/main/java/edu/gemini/shared/gui/TableUtil.java
sraaphorst: According to Find Usages, it won't be missed :grinning:

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/TelescopePosTableDropTarget.java
```diff
@@ -201,30 +204,31 @@ private boolean dropNodes(final int action, final Transferable transferable, fin
...
+        _selection = idx == -1 ? None.instance() : new Some<>(idx);
+        treeIgnoreSelectionAction(table::clearSelection);
     }
 
-    /** Restore the tree selection: we only allow one node to be selected. */
-    private void restoreTreeSelection() {
-        treeIgnoreSelectionAction(() -> _selection.map(_tree::getRowForPath).ifPresent(_tree::selectRowAt));
+    /** Restore the table selection: we only allow one row to be selected. */
+    private void restoreTableSelection() {
+        treeIgnoreSelectionAction(() -> _selection.foreach(table::selectRowAt));
```
swalker2m: Must have been awful getting drag and drop to work.  What happens when you try to drag a guide star out of a group or a "user" target into a group?

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/TelescopePosTableWidget.java
```diff
@@ -202,16 +201,21 @@ public String getValue(final Row row) {
...
             @Override public Option<GuideGroup> group() { return group; }
             @Override public List<Row> children()       { return children; }
-            @Override public Icon getIcon()             { return groupIcon; }
         }
 
+        // Collection of rows, which may include "subrows".
         private final ImList<Row> rows;
+
+        // Total number of rows, including subrows. Precalculated as swing uses this value frequently.
+        private final int numRows;
```
swalker2m: It is a bit confusing (just as it was before this PR too) that "row" sometimes means a normal table row and sometimes means a group of rows.  I think it's just a terminology thing.  I shouldn't have called the `Target \/ Group` concept a "Row" all those years ago.  Not sure what else to suggest though.

bundle/jsky.util.gui/src/main/java/jsky/util/gui/ToggleButtonWidget.java
```diff
@@ -1,94 +0,0 @@
...
-import javax.swing.border.BevelBorder;
-import java.awt.Font;
-import java.awt.event.ItemEvent;
-import java.awt.event.ItemListener;
-import java.util.Collections;
-import java.util.List;
-import java.util.Vector;
-
-/**
- * A toggle button widget originally based on Bongo, but now ported to Swing.
```
swalker2m: Bongo, good times.

swalker2m:
I don't see any problems with it and all the code cleanup is appreciated.  It isn't much code but I can see that it must have been tricky to get right.  I agree with Art, the table definitely looks much nicer now. :+1:

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/TelescopePosTableDropTarget.java
sraaphorst: It wasn't as bad as you'd think.

Guide stars can only be dragged from one group to another (otherwise the drag op fails), and user targets cannot be dragged into anything at all, which is what I think is the expected behaviour.

andrewwstephens:
Should one group be labeled "Automatic" and the others "Manual"?

sraaphorst:
@andrewwstephens That will come later, when the data model changes are in place. :accept:

cquiroz:
Just a nitpick, perhaps the "WIP" commits should be merged and given a meaningful comment. Too late for this as it was already merged but we may consider it for the future

PR #768, opened 2016-01-07
OCSADV-516: GuideProbeTargets changes

sraaphorst:
In the data model changes that will be implemented for BAGS, BAGS targets are no longer associated with `GuideProbeTarget`s but belong in their own `GuideGroup`s.

This PR, then, rolls back the `GuideProbeTargets` from their adjustments for BAGS to once again not contain any BAGS targets and instead be an `OptionsList` as it once was, and covers all the changes in the rest of the code base that this revert requires.

I also did some general code cleanup as I went along.


bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/GuideGroup.java
```diff
@@ -369,7 +364,7 @@ public boolean containsTarget(SPTarget target) {
 
     @Override
     public GuideGroup removeTarget(final SPTarget target) {
-        final ImList<GuideProbeTargets> updated = guideTargets.map(gpt -> gpt.removeTargetSelectPrimary(target));
```
swalker2m: This version of `GuideGroup` is going away anyway in favor of the implementation based on the new model, but is this what you meant?  It looks like the semantics change there?

sraaphorst: They actually do the exact same thing. It's just that when we divorced `GuideProbeTargets` from `OptionsList`, I decided to rename the method to make it more explicit that if it was the primary target being removed, then another primary was selected. You can see this in `OptionsList.UpdateOps.remove`, which is what is called by `GuideProbeTargets.remove`.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -1,244 +0,0 @@
...
-        val newGG = gg.setAll(DefaultImList.create(bagslessGpts.asJavaCollection))
-        val idx = ge.getOptions.indexOf(gg)
-        ge.setOptions(ge.getOptions.updated(idx, newGG))
-      } else {
-        ge.update(OptionsList.UpdateOps.remove(gg))
-      }
-    }
-    oldEnv.setGuideEnvironment(newGuideEnv)
-  }
-}
```
swalker2m: We'll still need something like the `BagsManager`.  I guess the idea is to bring it back when the model changes are done?

swalker2m:
I think this looks good. :+1: :pray:  I found myself wanting to separate out the meat of the change from the cleanups though.  Taking the initiative to cleanup code is laudable for sure but if there are 100s of lines of unrelated updates, let's strive to put them in a different commit if not a different PR altogether.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
sraaphorst: Yes... I was vacillating between commenting it out or just removing it, but I figured we can bring it back easily enough when the time comes, and there is no need to have it kicking around until then, especially since it will require fairly substantial changes.

sraaphorst:
Sure. I will do my best to resist the urge. :smile:

PR #769, opened 2016-01-07
jsky.util.gui cleanup

sraaphorst:
This is going to be pretty awful, I think.

I was bored earlier in the week in the evening, so I started looking at the code in `jsky.util.gui` and saw that there was a huge amount of room for cleanup and code elimination.

That led me down a few bread crumb trails into other packages as well, and this is the result.
There were many things like unused status loggers and HTML Query result handlers that were always unused / set to null, so I removed these, along with a number of unnecessary files.

Not sure if this will be considered too drastic of a cleanup effort, but I thought I would release it to the unwashed masses of the OCS programmers all the same.


bundle/edu.gemini.catalog/src/main/java/jsky/catalog/RangeSearchCondition.java
```diff
@@ -113,12 +85,12 @@ public boolean isTrueFor(Comparable val) {
...
+        if (_minInclusive) {
             return ((_minVal == null || _minVal.compareTo(val) <= 0)
                     && (_maxVal == null || _maxVal.compareTo(val) > 0));
         }
 
-        if (!_minInclusive && _maxInclusive) {
+        if (_maxInclusive) {
             return ((_minVal == null || _minVal.compareTo(val) < 0)
                     && (_maxVal == null || _maxVal.compareTo(val) >= 0));
         }
```
tpolecat: Is this logic change valid? What if it's both `_minInclusive` and `_maxInclusive`?

bundle/edu.gemini.catalog/src/main/java/jsky/catalog/TablePlotSymbol.java
```diff
@@ -569,6 +533,7 @@ public String getRatio() {
     }
 
     /** Return the symbol's ratio, evaluating the expression for the given row. */
+    @SuppressWarnings("EmptyCatchBlock")
```
tpolecat: I think we should fix the empty catch blocks rather than disabling the warnings. Swallowing `Throwable` is not ok.

bundle/jsky.app.ot/src/main/resources/jsky/app/ot/gemini/altair/AltairForm.jfd
```diff
@@ -169,7 +169,7 @@
       </void> 
       <void method="add"> 
        <object class="com.jformdesigner.model.FormComponent"> 
-        <string>jsky.util.gui.OptionWidget</string> 
+        <string>javax.swing.JRadioButton</string>
```
fnussber: I think the jfd files can go away entirely. They're a relict of the JFormDesigner times.

bundle/edu.gemini.catalog/src/main/java/jsky/catalog/RangeSearchCondition.java
sraaphorst: That case is addressed on L83 just above.
(Yeah, I admit this was confusing as it didn't appear directly in the diff.)
The only reason I caught it was because IDEA was squawking about boolean conditions always being true.

bundle/edu.gemini.catalog/src/main/java/jsky/catalog/TablePlotSymbol.java
sraaphorst: I'm happy to go through and make whatever changes people think should be done.
Logging the errors?

bundle/jsky.app.ot/src/main/resources/jsky/app/ot/gemini/altair/AltairForm.jfd
sraaphorst: I've been wondering for awhile why we still have these and if they are ever used, since JFormDesigner is no longer used, and I believe requires a license that we don't even presently have, if memory serves.

arturog8m: I suggest to remove them all and check that things still build - they should. If you find some showstopper there, we regroup. 

We aren't planning on using JFD anymore, so I think getting rid of all references to it would be a good thing to do.

sraaphorst:
Alright. All JFD files have been removed and everything still unsurprisingly builds as it should!
I have updated this PR with a new commit.

sraaphorst:
Hmmm... that commit didn't work its way into this PR. Weird.
I'll look at that.

sraaphorst:
Oh, there we go. No idea what happened. Sorry for the stream of comments.

sraaphorst:
Since this PR is a bit big (especially now with the JFD file removal), I think I'm going to close it and re-PR it in small, more manageable pieces over the course of the week, maybe with 1-2 small PRs per day.

PR #771, opened 2016-01-11
OCSINF-444: Removal of GmosGUI, PhoenixGUI, and OptionWidget.

sraaphorst:
This PR entails the following clean up:

= Removal of the no-longer-used `GmosGUI` and `PhoenixGUI`.
= The `OptionWidget` class (extends `JRadioButton`) was used in several places but none of the minimal added functionality of this class over `JRadioButton` was ever used, so I eliminated this class (and the accompanying `OptionWidgetWatcher`) and directly replaced all uses with `JRadioButton`.


swalker2m:
Good.  :+1:

PR #772, opened 2016-01-11
OCSINF-444C: Removal of CheckBoxWidget and cleanup of table widgets

sraaphorst:
Yet another PR in my ongoing cleanup. This one addresses the following:

= Removal of the `CheckBoxWidget` class, since this simply extends `JCheckBox`, is only used in `AltairForm`, and is unnecessary.
= `ArrayTableCellRenderer` constructor modified since parameter was never used.
= Cleaned up `TableDisplay` thoroughly, which is the only class that uses `ArrayTableCellRenderer`.
= Removed minor redundancy from `SaveableAsHTML` interface.


swalker2m:
M'kay.  :+1:

PR #773, opened 2016-01-12
Horizons Service for Ephemeris Lookup

tpolecat:
This PR adds yet another version of the Horizons service, which provides support for automated ephemeris lookup.

- `search` fetches a list of `HorizonsDesignation` (with a text name for display, as needed) given a partial name. There are three cases for the user to consider (comet, asteroid, major body) and eight cases internally, for differing output formats.
- `lookupEphemeris` takes a `HorizonsDesignation` and a few other details and looks up an `Ephemeris`. This lookup should never fail or be ambiguous (it's an error condition if this happens).

The API will likely change, and will probably move from server to client. For now it's just basic library support proving that it works (or at least seems to).

### DIFF
--- a/bundle/edu.gemini.horizons.server/src/main/scala/edu/gemini/horizons/server/backend/HS2.scala
+++ b/bundle/edu.gemini.horizons.server/src/main/scala/edu/gemini/horizons/server/backend/HS2.scala
@@ -29,8 +29,9 @@ object HorizonsService2 {
 
   /** The type of failures that might arise when talking to HORIZONS. */
   sealed trait HS2Error
-  case class HorizonsError(e: HorizonsException) extends HS2Error
-  case class ParseError(input: List[String], message: String) extends HS2Error
+  case class  HorizonsError(e: HorizonsException) extends HS2Error
+  case class  ParseError(input: List[String], message: String) extends HS2Error
+  case object EphemerisEmpty extends HS2Error
 
   /** A value annotated with a name. Used when multiple results are returned. */
   case class Row[A](a: A, name: String)
@@ -60,7 +61,7 @@ object HorizonsService2 {
       HS2.fromDisjunction(parseResponse(search, lines).leftMap(ParseError(lines, _)))
 
     // And finally
-    horizonsRequestLines2(queryParams) >>= parseLines
+    horizonsRequestLines(queryParams) >>= parseLines
 
   }
 
@@ -123,7 +124,7 @@ object HorizonsService2 {
       IO(CgiReplyBuilder.buildResponse(m.getResponseBodyAsStream, m.getRequestCharSet)).map(toEphemeris)
 
     // And finally      
-    horizonsRequest2(queryParams)(buildEphemeris)
+    horizonsRequest(queryParams)(buildEphemeris).ensure(EphemerisEmpty)(_.size > 0)
 
   }
 
@@ -259,8 +260,8 @@ object HorizonsService2 {
   }
 
   // Construct a program thet performs a HORIZONS request and yields the response as lines of text.
-  private def horizonsRequestLines2(params: Map[String, String]): HS2[List[String]] =
-    horizonsRequest2(params) { method =>
+  private def horizonsRequestLines(params: Map[String, String]): HS2[List[String]] =
+    horizonsRequest(params) { method =>
       IO {
         val s = Source.fromInputStream(method.getResponseBodyAsStream, method.getRequestCharSet)
         try     s.getLines.toList
@@ -276,7 +277,7 @@ object HorizonsService2 {
 
   // Construct a program thet performs a HORIZONS request and transforms the respons with the
   // provided function.
-  private def horizonsRequest2[A](params: Map[String, String])(f: GetMethod => IO[A]): HS2[A] =
+  private def horizonsRequest[A](params: Map[String, String])(f: GetMethod => IO[A]): HS2[A] =
     EitherT {
       IO(new GetMethod(CgiHorizonsConstants.HORIZONS_URL)).using { (method: GetMethod) =>
         IO {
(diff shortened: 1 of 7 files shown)


bundle/edu.gemini.horizons.server/src/main/scala/edu/gemini/horizons/server/backend/HS2.scala
```diff
@@ -0,0 +1,299 @@
...
+import java.io.IOException
+import CgiHorizonsConstants._
+import edu.gemini.horizons.api.HorizonsReply
+import java.util.Date
+import java.text.SimpleDateFormat
+
+import scala.collection.JavaConverters._
+import scala.util.matching.Regex
+import scala.io.Source
+import scalaz._, Scalaz._, scalaz.effect.{ IO, Resource }
```
swalker2m: I thought `Task` was the new `IO`?

tpolecat: This particular bit of work can't be done in parallel so I wanted to preclude that option by using `IO`, but it's mostly laziness ... the exception and resource handling support in `IO` is missing from `Task` and I wanted to use both here.

bundle/edu.gemini.horizons.server/src/main/scala/edu/gemini/horizons/server/backend/HS2.scala
```diff
@@ -0,0 +1,299 @@
...
+   * total elements spread uniformly over the over the given time period. The computed ephemeris may
+   * contain slightly more or fewer entries than requested due to rounding, or many fewer for very
+   * short timespans (no more than one entry will be returned per minute).
+   */
+  def lookupEphemeris(target: HorizonsDesignation, site: Site, start: Date, stop: Date, elems: Int): HS2[Ephemeris] = {
+
+    def formatDate(date: Date): String = {
+      val df = new SimpleDateFormat("yyyy-MMM-dd HH:mm:ss")
+      s"'${df.format(date)}'"
+    }
```
swalker2m: Channeling @cquiroz here: perhaps we should use `Instant` / `DateTimeFormatter` etc.?  Or is it being API compatible with the old horizons service or something?

tpolecat: Yeah we should probably switch and propagate the change back into `Semester` which is where the instances will usually come from. I think maybe I'll open another ticket for it.

bundle/edu.gemini.horizons.server/src/main/scala/edu/gemini/horizons/server/backend/HS2.scala
```diff
@@ -0,0 +1,299 @@
...
+        IO(m.releaseConnection)
+    }
+
+  // Construct a program thet performs a HORIZONS request and transforms the respons with the
+  // provided function.
+  private def horizonsRequest[A](params: Map[String, String])(f: GetMethod => IO[A]): HS2[A] =
+    EitherT {
+      IO(new GetMethod(CgiHorizonsConstants.HORIZONS_URL)).using { (method: GetMethod) =>
+        IO {
+          HorizonsService2.synchronized {
```
swalker2m: There's nothing mutable about this (?) but the issue is that it limits the number of queries from the same IP?  Might be worth a comment.

tpolecat: Yep, I will add a comment. Strictly speaking I'm not quite sure this is adequate because the `GetMethod` remains open and results could in principle still be streaming back after the monitor is released. However I have not been able to make this happen so I'm going to keep an eye on it. I guess an explicit mutex would be safer.

bundle/edu.gemini.horizons.server/src/main/scala/edu/gemini/horizons/server/backend/HS2.scala
```diff
@@ -0,0 +1,299 @@
...
+          // Single result with form: JPL/HORIZONS         1P/Halley           2015-Dec-31 11:40:21
+          lazy val case2 =
+            """  +([^/]+)/(.+?)  """.r.findFirstMatchIn(header).map { m =>
+              List(Row(HorizonsDesignation.Comet(m.group(1)), m.group(2)))
+            } \/> "Could not match '1P/Halley' header pattern."
+
+          // First one that works!
+          case0 orElse
+          case1 orElse
+          case2 orElse "Could not parse the header line.".left
```
swalker2m: Is there any benefit from clarifying "Could not parse the header line as a comet."?

swalker2m:
Looks very clean, very compact. Huge improvement. :+1:

bundle/edu.gemini.horizons.server/src/main/scala/edu/gemini/horizons/server/backend/HS2.scala
tpolecat: Might as well.

bundle/edu.gemini.horizons.server/src/main/scala/edu/gemini/horizons/server/backend/HS2.scala
fnussber: This code looks indeed nice and I like how `IO` provides a `using` statement. So if `Task` should generally be preferred over `IO` (why?) then will these `IO` specific functionalities being added to `Task` in future `scalaz` versions?

tpolecat: These things are [straightforward to build](https://github.com/tpolecat/doobie/blob/master/core/src/main/scala/doobie/util/catchable.scala) in terms of `Catchable` (which is the exception-trapping effect that `Task` implements) but nobody has done this for scalaz because the plan was to replace both `Task` and `IO` with something better by now. I might do it for 7.2 if nothing happens in the interim; the library support for `IO` is much better, but it doesn't support concurrency (which is why everyone is using `Task` instead).

PR #774, opened 2016-01-18
REL- 2571: Add help page links to PIT tabs

cquiroz:
Small PR that adds a tooltip to the help button on the PIT sub windows

### DIFF
--- a/bundle/edu.gemini.ui.workspace/src/main/java/edu/gemini/ui/workspace/util/SimpleInternalFrame.java
+++ b/bundle/edu.gemini.ui.workspace/src/main/java/edu/gemini/ui/workspace/util/SimpleInternalFrame.java
@@ -93,31 +93,7 @@ public class SimpleInternalFrame extends JPanel {
     public SimpleInternalFrame(String title) {
         this(null, title, null, null, null, null, null);
     }
-    
-    
-    /**
-     * Constructs a SimpleInternalFrame with the specified 
-     * icon, and title.
-     * 
-     * @param icon        the initial icon
-     * @param title       the initial title
-     */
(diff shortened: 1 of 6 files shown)


bundle/edu.gemini.ui.workspace/src/main/java/edu/gemini/ui/workspace/util/SimpleInternalFrame.java
```diff
@@ -320,7 +274,7 @@ public void setSelected(boolean newValue) {
      * @param bar     the panel's tool bar
      * @return the panel's built header area
      */
-    private JPanel buildHeader(JLabel label, JToolBar bar, Action helpAction, Icon helpIcon) {
+    private JPanel buildHeader(JLabel label, JToolBar bar, Action helpAction, Icon helpIcon, String hint) {
```
sraaphorst: Apart from `null` distaste and wondering if we should instead use an `Option<String>` here, with a `hintOption.foreach(help::setToolTipText)`,  :+1:

cquiroz: Should we keep using the old Java `Option` class? Or do we prefer `Optional` for Java 1.8

sraaphorst: That's something I've also been wondering about. In most new code I write, I use Java 8 `Optional`, and in old code, I keep Gemini `Option`, and nobody seems to have complained thus far :smile:

cquiroz:
Went using `Optional` and removed the use of nulls. A small case class encapsulates the parameters for the help button

PR #776, opened 2016-01-18

cquiroz:
The PDF output needs some tweaks given that certain instrument can specify either side
Some lovely xsl programming included


sraaphorst:
I can't claim to know exactly what is going on in all steps here, but it looks straightforward enough, so I'll tentatively :+1:

PR #777, opened 2016-01-18

cquiroz:
With this PR double clicking on a problem entry related to conditions will open the conditions editor, it used to just focus on the wrong condition


sraaphorst:
Looks good to me. :+1:

PR #778, opened 2016-01-19

cquiroz:
Missing telescope selection for NOAO template used to generate phase 1 proposal pdfs. Sorry it broke the build


swalker2m:
:+1:

PR #779, opened 2016-01-19
OCSADV-139: update to scalaz 7.1.6

swalker2m:
This PR updates the codebase to scalaz 7.1.6.  A few notable changes:

* The scalaz update required updating argonaut from 6.0.4 to 6.1.  Unfortunately that also drags in monocle and quasiquotes.  Monocle lenses are pretty cool though.

* Most everything is invariant in 7.1.  That necessitated a number of changes. For example in this code, even though `B >: A`, because it is invariant `Slick[B]` is no longer a supertype of `Slick[A]`.  The "solution" I used was to map just to explicitly up-cast the type from `A` to `B`.  I'm guessing there is a better way to do this.

````scala
//type Slick[+A] = Kleisli[IO, Session, A]
type Slick[A] = Kleisli[IO, Session, A]

implicit class SlickOps[A](a: Slick[A]) {
  def except[B >: A](b: Throwable => Slick[B]): Slick[B] =
    Slick.session >>= { s =>
//      a(s).except { 
      a(s).map(x => x: B).except { 
        case t: Throwable => b(t)(s)
      } .liftIO[Slick]
    }
````

* `flatmap` was removed from `Validation`, since `Validation` isn't a monad.  Since we flatmapped `Validation` in a couple of places, the solution was to just flip back and forth to/from disjunction.

* There is some weirdness with imports that I never figured out.  This appeared with getting a `Monoid[Double]` instance, which comes up for example in `OptionOps.orZero`.  I just switched those to `| 0.0` etc.

* Disjunction adds the scala `merge` method which works when both the left and right sides are of a "commonly assignable" type.  That's the theory anyway but when pressed by incomprehensible compile errors I resorted to `fold(identity, identity)`.

* The order of `BooleanOps.either` was [reversed](https://github.com/scalaz/scalaz/commit/03cd235225e04e6731708f9b84f4f570327502de)

### CODE REVIEW

File: bundle/edu.gemini.dataman.app/build.sbt [hunk 1]
```diff
@@ -7,9 +7,13 @@ name := "edu.gemini.dataman.app"
...
 unmanagedJars in Compile ++= Seq(
-  new File(baseDirectory.value, "../../lib/bundle/argonaut_2.10-6.0.4.jar"),
+  new File(baseDirectory.value, "../../lib/bundle/argonaut_2.10-6.1.jar"),
+  new File(baseDirectory.value, "../../lib/bundle/monocle-core_2.10-1.1.0.jar"),
+  new File(baseDirectory.value, "../../lib/bundle/monocle-macro_2.10-1.1.0.jar"),
+  new File(baseDirectory.value, "../../lib/bundle/quasiquotes_2.10-2.0.1.jar"),
+  new File(baseDirectory.value, "../../lib/bundle/org.scala-lang.scala-reflect_2.10.1.v20130302-092018-VFINAL-33e32179fd.jar"),
   new File(baseDirectory.value, "../../lib/bundle/osgi.cmpn-4.3.1.jar"),
-  new File(baseDirectory.value, "../../lib/bundle/scalaz-concurrent_2.10-7.0.5.jar")
+  new File(baseDirectory.value, "../../lib/bundle/scalaz-concurrent_2.10-7.1.6.jar")
```
cquiroz: Could we use managed jars instead of having the jar on lib/bundle?

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/view/partner/PartnerView.scala
```diff
@@ -94,7 +94,7 @@ class PartnerView extends BorderPanel with BoundView[Proposal] {view =>
       BorderFactory.createEmptyBorder(2, 4, 2, 4))
     override def refresh(m:Option[Proposal]) {
 
-      val (reqTime, minTime) = ~m.map(_.proposalClass).map(r => (r.requestedTime.hours, r.minRequestedTime.hours))
+      val (reqTime, minTime) = m.map(_.proposalClass).map(r => (r.requestedTime.hours, r.minRequestedTime.hours)) | ((0.0, 0.0))
```
cquiroz: Does this mean there is no `zero` defined for tuples?

cquiroz:
Great work!!!

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/view/partner/PartnerView.scala
swalker2m: I think it is there if you can figure out how to use the right imports.  It seems that 

````scala
import scalaz._, Scalaz._
````

is no longer always sufficient or correct.  I just punted.

File: bundle/edu.gemini.dataman.app/build.sbt (hunk 1 above)
Comment (swalker2m) [inline bundle/edu.gemini.dataman.app/build.sbt]: I used `sed` so it was trivial to do from the command line.  We're only using managed dependencies for tests at this point. A change like that would probably be more appropriate in a separate PR.  At any rate I'm not familiar enough with the build to say what if any impact it would have on building bundles and apps.

File: bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/view/partner/PartnerView.scala (hunk 2 above)
Comment (tpolecat) [inline bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/view/partner/PartnerView.scala]: It looks like it was using `Monoid[Double]` which shouldn't exist ... not sure how this ever worked.

**Comment by tpolecat** [comment]
> The "solution" I used was to map just to explicitly up-cast the type from A to B. I'm guessing there is a better way to do this.

Yeah I added a `.widen` operation for 7.2 that lets you say `fa.widen[B]` but it didn't make it into 7.1, so yeah `fa.map(x => x : B)` or `fa.map[B](identity)` or something like that is what you need to do.

> Since we flatmapped Validation in a couple of places, the solution was to just flip back and forth to/from disjunction.

Yep.

> There is some weirdness with imports that I never figured out. This appeared with getting a Monoid[Double] instance, which comes up for example in OptionOps.orZero.

There is no `Semigroup[Double]` (and therefore no `Monoid`) because floating-point ops are not associative. I'm surprised this worked in 7.0.

> Disjunction adds the scala merge method which works when both the left and right sides are of a "commonly assignable" type. That's the theory anyway but when pressed by incomprehensible compile errors I resorted to fold(identity, identity).

Hm yeah you have to provide a type argument; it won't find the LUB for you. So `.merge[Woozle]`.

File: bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/Target.scala [hunk 3]
```diff
@@ -24,32 +24,38 @@ object Target extends TargetLenses
...
+
   val name: Target @> String =
     Lens(_.fold(
-      TooTarget.name.run,
-      SiderealTarget.name.run,
-      NonSiderealTarget.name.run
+      runTarget(TooTarget.name),
+      runTarget(SiderealTarget.name),
+      runTarget(NonSiderealTarget.name)
     ))
```
tpolecat: You had asked about this on iChat. If it compiles here we're fine, so let's leave it. Might make sense to switch to monocle lenses anyway.

tpolecat:
This looks great! I'm glad there weren't any big problems. :+1:

PR #781, opened 2016-01-19
REL-1232: Add ToDo for Band 3 observations

cquiroz:
Small problem checker for band 3 proposals without observations


tpolecat:
:+1:

PR #782, opened 2016-01-21
REL-1692: B semester 11hr incorrect error

cquiroz:
Tiny PR


arturog8m:
Looks pretty straightforward :+1:  Are there test cases for these things?

sraaphorst:
I think I see how this fixes the issue of the RA range, but does it address generating an error when it should be generating a warning?

If so, tiny :+1: for a tiny PR :smile:

cquiroz:
@arturog8m Thanks for reminding me about it, I pushed a fix for a broken test case
@sraaphorst The nature of the issue has changed, so the actual description on REL-1692 is not correct after I discussed it with Andy, I think these ranges/warnings are ok

PR #783, opened 2016-01-21
Enable sbt cached resolution

cquiroz:
I found this little gem, With sbt 0.13.7 you can enable cached resolution, now when I change the project and reload instead of 30 mins it takes 30seconds

http://www.scala-sbt.org/0.13/docs/Cached-Resolution.html

I don't know about side effects of moving to 0.13.7 but I'm able to e.g. build the PIT without any problems with this patch

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -13,6 +13,8 @@ version in ThisBuild := ocsVersion.value.toOsgiVersion
 
 scalaVersion in ThisBuild := "2.10.4"
 
+updateOptions := updateOptions.value.withCachedResolution(true)
+
 // Note that this is not a standard setting; it's used for building IDEA modules.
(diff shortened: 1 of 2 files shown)


sraaphorst:
I think if this works, we will have to sing your praises forever and send you your very own cake.

tpolecat:
It should never take 30 minutes (`reload plugins` and then `clean` and then restart should fix it) but this sounds great. :+1:

cquiroz:
Well, I have measured even more than 30m (even with reload plugins), though I think my system has some weird setting, since nobody else gets it so slow
This setting avoids that ivy checks all the dependencies upon reload, which apparently is very slow, especially for multi-module builds

PR #784, opened 2016-01-22
Remove warnings on obslog bundle

cquiroz:
Another cleanup PR, this one removes all warnings produced on the bundle edu.gemini.obslog, it is mostly mechanical work


swalker2m:
All looks reasonable to me.  Having more type information will make it easier to maintain. :+1:

PR #785, opened 2016-01-22
Revert "Enable sbt cached resolution"

tpolecat:
This reverts commit b4bd9f51e0fe9ed74e9a4130a34c4ab4b553c6ca.

@cquiroz I was getting compiler crashes during a compile phase that sbt introduces, so I had to revert this commit locally to get my world to compile again. Can you see if a clean full sbt build works for you?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -13,8 +13,6 @@ version in ThisBuild := ocsVersion.value.toOsgiVersion
 
 scalaVersion in ThisBuild := "2.10.4"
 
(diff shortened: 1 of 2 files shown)


cquiroz:
Right, it breaks on scala 2.10.4 though it works on 2.11.7, I assume it is a bad interaction with sbt
We should re-enable when we move to 2.11

cquiroz:
:+1:

PR #788, opened 2016-01-26
Removed warnings on shared.util p2checker

cquiroz:
Another PR removing warnings, nothing too special
Some methods were removed, given that they are marked unused by IDEA


tpolecat:
:+1:

PR #789, opened 2016-01-27
Fix for package export bug

cquiroz:
PR #788 introduced a bug with the exported packages as it contains a class on package `edu.gemini.shared` which likely shouldn't be exported
This PR moves the class on `package.scala` to its own file


tpolecat:
What was the effect of this bug? What is the harm of including that class in exports?

Not arguing with the fix, just trying to understand what the bug was.

cquiroz:
IntegerIsIntegral ended up in the package `edu.gemini.shared` and that broke OSGi when building an app. You could have exported that package but given that we have 5 bundles under `edu.gemini.shared.*` makes it a bit risky

tpolecat:
Ah ok. Yeah the overlapping packages are a mess. :+1:

PR #790, opened 2016-01-27

cquiroz:
This PR fixes a bug when clicking on the problems table with a conditions-related issue. The PIT would open the conditions editor but apply the updated conditions to the first condition on the plan. This is visible if you have many condition groups on your proposal


swalker2m:
:+1:

PR #791, opened 2016-01-27
REL-2412: PIT duplication check with the in-house archive

cquiroz:
The PIT breaks when doing GSA queries because monocle jars were not included after the scalaz migration.
I include a small bug fix removing an exception that appears when the PIT is closed from the felix console


tpolecat:
:+1:

PR #792, opened 2016-01-28
REL-1692: Update PIT RA & Dec limits for 2016B

cquiroz:
Small change to DEC limits for GS on Gems and B semester


swalker2m:
:+1:

PR #793, opened 2016-01-28
Compilation error fixes

cquiroz:
Compilation breaks in a few places as `IntegerIsIntegral` was moved to another package. This was tested with a full project compilation

### DIFF
--- a/bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs2/MergePlan.scala
+++ b/bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs2/MergePlan.scala
@@ -3,6 +3,7 @@ package edu.gemini.sp.vcs2
 import edu.gemini.pot.sp._
 import edu.gemini.pot.sp.version._
 import edu.gemini.shared.util._
+import edu.gemini.shared.util.IntegerIsIntegral._
 import edu.gemini.sp.vcs2.NodeDetail.Obs
 import edu.gemini.sp.vcs2.VcsFailure._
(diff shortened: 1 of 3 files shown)


bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs2/MergePlan.scala
```diff
@@ -3,6 +3,7 @@ package edu.gemini.sp.vcs2
 import edu.gemini.pot.sp._
 import edu.gemini.pot.sp.version._
 import edu.gemini.shared.util._
+import edu.gemini.shared.util.IntegerIsIntegral._
```
swalker2m: I'm not opposed to this but I could use help understanding what is going on.  With

````scala
import edu.gemini.shared.util._
````
you get `IntegerIsIntegral` which provides the `Integral` instance needed by `VersionVector`, I think.  The actual methods in `Integral` aren't directly referenced here?

What was the compile error?

cquiroz: It all starts from a warning given by scalac, that you shouldn't have top-level constructs like `IntegerIsIntegral` in a package object. The fix seems simple, just move it out but then you loose the implicit that exposes `IntegerIsIntegral`. My solution (which maybe far from perfect) was to move it out of the package object and add an object `IntegerIsIntegral` with the implicit conversion

swalker2m: Thanks, that makes sense. :+1:

PR #794, opened 2016-01-28
Merge bundles

cquiroz:
This PR merges the bundle `edu.gemini.model.p1.visibility` into `edu.gemini.model.p1`. The visibility bundle only has 3 files and it is only used by the PIT, IMHO it is just overhead to have it as a separate bundle


sraaphorst:
:+1:

swalker2m:
Apart from relying on P1 model, the low resolution estimate it makes seems fairly PIT-specific. :+1:

PR #795, opened 2016-01-29
Support travis CI

cquiroz:
This PR integrate Travis CI into the OCS. After some experimentation I got it to build on travis on my ocs fork and we can try now in the main OCS branch. It may take some experimentation to get right but this is a start

Here is a link to the last successful build
https://travis-ci.org/cquiroz/ocs/builds/105728173

It should generate HipChat notifications too
screenshot 2016-01-29 15 43 49

Some files had to be renamed as they work on in OSX but break tests in Linux


tpolecat:
Hm looks like the credentials stuff for svn didn't work.

cquiroz:
They should be setup on travis so they don't remain on the repo, I'll check that

cquiroz:
I've read quite a bit about this and there is no practical way to provide the encrypted values to tarvis-ci for pull requests. I can see why, it would be very easy for somebody to make a pull request and export the credentials.
@fnussber suggested that we could not get the credentials and instead use a template as he does for ITC, maybe he can expand this and we could change the script to use that method for PRs and the svn with real credentials for builds after code has been merged
This assumes the tests don't use the actual credentials

tpolecat:
I think Jenkins can do this so we might look into that. It runs as a normal user so it could have access to svn. Last time I set it up it was a single jarfile and all you have to do is `java -jar ...` to run it.

cquiroz:
I setup jenkins on the build machine. However for PR support we need to open it up to the outside so GitHub can call it to start the build. That's not something I'm looking forward to do

tpolecat:
Just the prospect of dealing with ISG?

cquiroz:
Mostly that part :+1:

cquiroz:
The build is now working for PRs, if we were going to merge it would use the actual credentials from svn. I'm happy with this setup, it should tell us if the build is broken at least and requires no maintenance (as far as I know)

https://travis-ci.org/gemini-hlsw/ocs/builds/107249915

Thanks to @fnussber to point out how to make the build without the credentials

What do you think?

swalker2m:
It seems fine to me if PRs build with template credentials.  For that matter, I would have been happy if just the final merged develop branch built with CI.  Having it check your PR before you merge is gravy.   It doesn't check every intermediate commit though does it?  It seems like that would be cumbersome if it did.

:+1:

tpolecat:
Awesome!!! :100:

cquiroz:
As far as I can tell, it only checks the full PR, otherwise I imagine is too much load on their servers to build each commit. Our build takes around 24 mins to run :D
Once we merge I need to verify if it will build the merged branch correctly

PR #796, opened 2016-02-01
REL-2202: Apple developer signature for PIT Mac app

cquiroz:
This PR updates the build to support signing Mac Applications thus becoming acceptable to GateKeeper. So far it has been tested only for the PIT
The code will check if the certificate exists on the system and only then do the signing, Several steps of signing must be done before the file is acceptable.
The signature can be verified running the command below on the app:

    spctl -v -a <path.to>\Gemini...app

The certificate itself is not included as it cannot be made public but it must be installed in each system where the application is built. Note that in the code the SHA hash of the certificate and certificate id are included. That information is of no use to a potential attacker


swalker2m:
:+1:

PR #797, opened 2016-02-03
New Target Editors

tpolecat:
Ok this adds the initial target editors for the new model. I apologize for the volume of code. Best way to review might be to fetch my branch so you can run the test UI.

- I added a transient new-model `Target` member to `SPTarget`. This is unused in mainline OCS code but provides a way to build and test the new editors.
- The editors themselves are copy/paste of existing ones, updated to use lenses and the new immutable model. A bunch of stuff went away because the model is so much simpler now. There is still some refactoring that could be done but it's ok for now.
- The non-sidereal target editor uses the new Horizons support and it seems solid. Target lookup happens in two stages: name search to determine the permanent id, and then a search using this id to retrieve the ephemeris. The robot can repeat this search for fine-grained data.
- Horizons back-end functionality was moved to the client bundle because Horizons doesn't allow concurrent queries from the same machine. So **you need to rebuild your IDEA project** after this is merged. Existing server functionality is still there but will go away. 

To be done:

- TPE integration (will need to happen after editor is integrated).
- Custom ephemeris upload (need to pull some code out of PIT).
- UI tweaks, as needed.

You can test this from sbt:

`> bundle_jsky_app_ot/runMain jsky.app.ot.gemini.editor.targetComponent.details2.TestFrame`

As you edit the target you will see the updated versions on the console. To save space the `toString` for ephimerides is just the size `«41»`.

Builds clean, tests pass. :chicken:


cquiroz:
Good, I really have to give thanks that the code is so extensively documented. Before delving into the code, I have a couple of general questions:

* Some of the code uses the same names and looks on the surface a bit like the old UI classes, used on the current target editor. Will those old classes disappear in time?
* We had a short discussion about reusing this editor on the PIT once we integrate ITC, Do you see any big problems doing that in the future?
* We should check together the code that you need from the PIT, given that we are so close to the next PIT release. Or perhaps I should just do a release branch
* We have some RA/Dec text fields in the OT, should we reuse those?
* Would you publish some screenshots to check the UI

swalker2m:
I just skimmed the code really but it looks fine. :+1:  I played around with the editor a bit and everything seems to work well.

ceres

What is the "Query 1;"?  I was surprised there was no way to actually see the ephemeris elements probably it doesn't matter.

swalker2m:
@cquiroz The idea, I believe, is that these editors will replace the existing ones when we totally switch over to the new target model.  Right now they just edit a transient `_newTarget` member of `SPTarget` that has no impact on anything else in the OT.  In the future though, `SPTarget` will just have listener support and a mutable reference to the new target.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/SPTarget.java
```diff
@@ -41,8 +42,17 @@ public void setTarget(final ITarget target) {
      * if the contained target is of the specified type.
      */
     public void setTargetType(final ITarget.Tag tag) {
-        if (tag != _target.getTag())
-            setTarget(ITarget.forTag(tag));
+        if (tag != _target.getTag()) {
```
cquiroz: Should it be `!equals`?

bundle/jsky.app.ot/build.sbt
```diff
@@ -142,4 +142,4 @@ OsgiKeys.exportPackage := Seq(
   "jsky.science.util",
   "jsky.science.util.gui")
 
-
+fork := true
```
cquiroz: Why do we need to fork the VM here?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details2/Horizons.scala
```diff
@@ -0,0 +1,279 @@
...
+
+      // Use the cached result if possible
+      getCachedResult(service, hObjId, hObjType, date).map(_.right).filter(_ => useCache).getOrElse {
+
+        // New query
+        service.setInitialDate(date)
+        service.setObjectId(hObjId)
+        service.setObjectType(hObjType)
+
+        // Here is the blocking call. If we get null back then it means the user cancelled or there
```
cquiroz: What is shown to the user while the horizon query is progressing?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details2/MagnitudeEditor.scala
```diff
@@ -0,0 +1,374 @@
+package jsky.app.ot.gemini.editor.targetComponent.details2
+
+import java.awt.{GridBagConstraints, Insets, GridBagLayout, Component, Dimension}
+import java.awt.event.{ActionEvent, ActionListener}
+import java.beans.{PropertyChangeEvent, PropertyChangeListener}
+import java.text.{NumberFormat, DecimalFormat}
+
+import javax.swing.{SwingUtilities, JScrollPane, BorderFactory, JPanel, DefaultComboBoxModel, JFormattedTextField, JLabel, JList, ListCellRenderer, JButton, JComboBox, JTextField}
```
cquiroz: Should we use the scala swing API?

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/SPTarget.java
swalker2m: The tags are enum elements so it seems okay.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details2/NonSiderealNameEditor.scala
```diff
@@ -0,0 +1,152 @@
...
+      }
+
+    def show(msg: String): HS2[Unit] =
+      HS2.delay(Swing.onEDT(GlassLabel.show(SwingUtilities.getRootPane(name), msg)))
+
+    def hide: HS2[Unit] =
+      HS2.delay(Swing.onEDT(GlassLabel.hide(SwingUtilities.getRootPane(name))))
+
+    def lookup(d: HorizonsDesignation, site: Site): HS2[Ephemeris] =
+      show("Fetching Ephemeris...") *> HorizonsService2.lookupEphemeris(d, site, 1000) // arbitrary
```
cquiroz: I guess here we are showing the glass pane while doing the lookup, nice code :+1:

swalker2m:
One thing I noticed is that when you re-query, it still asks for the target type.  I guess the point though is that there really isn't any need to query again since you get ephemeris for the entire semester?

It might be useful to show the target type in the editor, but on the other hand maybe it's always obvious to an astronomer anyway.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details2/NonSiderealNameEditor.scala
```diff
@@ -0,0 +1,152 @@
...
+            null)
+        } .map(_.asInstanceOf[Wrapper].unwrap)
+      } >>= {
+        case Some(r) => oneResult(r)
+        case None    => ().point[HS2]
+      }
+
+    def oneResult(r: Row[_ <: HorizonsDesignation]): HS2[Unit] =
+      updateDesignation(r.a, r.name).as(site) >>= {
+        case Some(site) => lookup(r.a, site) <* hide >>= updateEphem
```
cquiroz: Could you gives a short explanation about the `<* hide >>=` piece? It looks like, lookup, if fails hide and the update but I'm not sure, there is a similar one below `*> askSearch >>=`

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details2/package.scala
```diff
@@ -0,0 +1,84 @@
...
+    def objectType = ObjectType.MAJOR_BODY
+  }
+
+  implicit class CoordinateParamOps(p: CoordinateParam) {
+    def setOrZero(d: java.lang.Double): Unit =
+      p.setValue(if (d == null) 0.0 else d.doubleValue)
+  }
+
+  def forkSwingWorker[A <: AnyRef](constructImpl: => A)(finishedImpl: Throwable \/ A => Unit): Unit =
+    new SwingWorker {
```
cquiroz: Seems `SwingWorker` is deprecated, Perhaps we can avoid its use in new code?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details2/Horizons.scala
tpolecat: Sorry, that file is unused now. I need to remove it.

bundle/jsky.app.ot/build.sbt
tpolecat: It's just so you can run the test GUI and exit with **⌘+Q** without killing sbt.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details2/MagnitudeEditor.scala
tpolecat: This is a direct port of the Java code, which is pretty complicated. So I didn't want to rewrite the GUI logic.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details2/NonSiderealNameEditor.scala
tpolecat: It means lookup and then hide the glass pane, but we don't care about the value of `hide`. These operators are defined for all applicative functors:

- `a *> b` is the same as `(a |@| b)((aa, bb) => bb)`
- `a <* b` is the same as  `(a |@| b)((aa, bb) => aa)`

So the expression `lookup(r.a, site) <* hide >>= updateEphem` is just shorthand for

```scala
for {
  x <- lookup(r.a, site)
  _ <- hide
  e <- updateEphem(x)
} yield e
```

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details2/package.scala
tpolecat: I'm using it once but it may be used in some stuff I didn't touch. If that's the only place it's used I'll get rid of it.

tpolecat:
> Some of the code uses the same names and looks on the surface a bit like the old UI classes, used on the current target editor. Will those old classes disappear in time?

Yes, this is a copy/paste of the old editors, which were written anticipating the change to immutable targets. The old package will disappear.

> We had a short discussion about reusing this editor on the PIT once we integrate ITC, Do you see any big problems doing that in the future?

This all relies on the change notifications provided by `SPTarget` so it would take some refactoring. Depending on what happens with the "new" OCS this may or may not be worth the trouble.

> We should check together the code that you need from the PIT, given that we are so close to the next PIT release. Or perhaps I should just do a release branch

Ok, or I could just duplicate the code in a common place and then you can delete the code from the PIT next semester.

> We have some RA/Dec text fields in the OT, should we reuse those?

Possibly. They're awful. The PIT versions may be better.

tpolecat:
@swalker2m 

> What is the "Query 1;"? I was surprised there was no way to actually see the ephemeris elements probably it doesn't matter.

That is the actual Horizons query that uniquely identifies the target. You just happened to pick the most boring asteroid in the solar system. The query for Zhejiangdaxue is `DES=2007 JU34`. Maybe we shouldn't show it but it seems like it might be helpful; you can `telnet horizons.jpl.nasa.gov 6775` and type that in and see exactly what OCS is looking at.

I'll go ahead and add the current RA/Dec to the UI just as a sanity check.

tpolecat:
Ok got rid of `SwingWorker` and the unused Horizons code, thanks @cquiroz.

Any more comments?

cquiroz:
Nice, we'll get rid of `SwingWorker` altogether some day. Any reason to prefer `Task` over, e.g. `SwingUtilities.invokeLater`? Either way is fine with me

tpolecat:
@cquiroz I think the point of `SwingWorker` (and `Task` here) is that you want your stuff to *not* be on the EDT to avoid making the UI unresponsive, and when you *do* want things to be on it (like showing and hiding the glass-pane) you use `SwingUtilities.invokeLater` or `Swing.onEDT` to do it.

swalker2m:
I figured the "1" identified Ceres somehow but the semicolon looked suspicious.  Indeed typing `1;` into the telnet address gives you Ceres though.

tpolecat:
@swalker2m indeed, `1` without the semicolon is Mercury. `DES=1;CAP` is Halley's comet. It's fantastic.

PR #798, opened 2016-02-04
Don't import p1 mutable classes

cquiroz:
Mini improvement to the IDEAproject generation logic. With this PR the generated IDEA project won't attempt to import the classes on the mutable p1 model by default. They are still available and the project compiles fine but avoids a common mistake I make when letting IDEA do the imports


tpolecat:
:+1:

PR #800, opened 2016-02-04
REL-2617: Remove H_STAR and H_LIWA observing modes from PIT

cquiroz:
This PR hides the GPI modes H_Star and H_Liwa on the UI and adds upconversion code to convert existing proposal with those modes and convert them to H Direct


swalker2m:
:+1:

PR #805, opened 2016-02-08
REL-2622: Data Manager Discovered Dataset Bug

swalker2m:
This grim little PR fixes a fault report (FR-36116) that was triggered by the Data Manager update for 2016A.  Usually new datasets are recorded in the ODB via events from the seqexec.  Sometimes however the seqexec event is never generated and yet the dataset is created by the instrument and added to the archive.  When the Data Manager discovers a new dataset via archive poll results, it adds a new record to the corresponding Observing Log in the ODB.

The behavior change in 2016A was to no longer attempt to compute the sequence step configuration corresponding to the dataset.  Unfortunately though this information is needed by the Observing Log webapp, particularly for exporting the observing log to a text file.  When expected sequence step information is not available, it blows up.

This PR restores the pre-2016A behavior of adding a sequence step configuration for newly discovered datasets.  It also changes the dataset filename to strip off the ".fits" extension to be consistent with the WDBA and allow the Observing Log webapp to parse out the file number.

### DIFF
--- a/bundle/edu.gemini.dataman.app/src/main/scala/edu/gemini/dataman/app/ObsLogActions.scala
+++ b/bundle/edu.gemini.dataman.app/src/main/scala/edu/gemini/dataman/app/ObsLogActions.scala
@@ -4,6 +4,8 @@ import edu.gemini.dataman.app.ModelActions._
 import edu.gemini.dataman.core._
 import edu.gemini.pot.sp.SPObservationID
 import edu.gemini.pot.spdb.IDBDatabaseService
+import edu.gemini.spModel.config.map.ConfigValMapInstances
+import edu.gemini.spModel.config.{DatasetConfigService, ConfigBridge}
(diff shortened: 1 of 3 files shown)


tpolecat:
I always read `DmanAction` as `DamnAction`. Thanks for writing a test. :+1:

cquiroz:
Note that this didn't build against the rel branch on travis

bundle/edu.gemini.dataman.app/src/main/scala/edu/gemini/dataman/app/ObsLogActions.scala
```diff
@@ -256,6 +276,8 @@ final class ObsLogActions(odb: IDBDatabaseService) {
 object ObsLogActions {
   val Log = Logger.getLogger(ObsLogActions.getClass.getName)
 
+  val FitsFile = """(\V*?)(?:\.fits)?$""".r
```
cquiroz: Should this be case-insensitive?

cquiroz:
:+1:

PR #806, opened 2016-02-08
key database slick → doobie

tpolecat:
This adds a `KeyDatabase` implementation based on [doobie](https://github.com/tpolecat/doobie).
- This uses the same schema and logic as the existing Slick-based implementation. We need to update to hashed passwords (REL-2624).
- The Slick implementation is still there but the doobie one is used by default now.


bundle/edu.gemini.util.security/src/test/resources/testdb.sql
```diff
@@ -0,0 +1,22 @@
...
+    NAME VARCHAR NOT NULL,
+    HASH VARCHAR NOT NULL,
+    VERSION INTEGER DEFAULT 1 NOT NULL
+);   
+-- 5 +/- SELECT COUNT(*) FROM PUBLIC.KEYS;     
+INSERT INTO PUBLIC.KEYS(CLASS, NAME, HASH, VERSION) VALUES
+('User', 'bob@dobbs.com', 'foodle', 1),
+('Program', 'GS-2010A-Q-3', 'blippy', 1),
+('Staff', 'Gemini', 'hoox', 2),
+('Affiliate', 'Chile', 'huevon', 123),
```
cquiroz: :+1:

swalker2m:
:+1:

bundle/edu.gemini.util.security/build.sbt
```diff
@@ -9,9 +9,18 @@ name := "edu.gemini.util.security"
...
   new File(baseDirectory.value, "../../lib/bundle/scala-slick_2.10-1.0.0.jar"),
+  new File(baseDirectory.value, "../../lib/bundle/scala-reflect-2.10.5.jar"),
   new File(baseDirectory.value, "../../lib/bundle/scala-swing_2.10-2.0.0-SNAPSHOT.jar"),
   new File(baseDirectory.value, "../../lib/bundle/scalaz-core_2.10-7.1.6.jar"),
-  new File(baseDirectory.value, "../../lib/bundle/scalaz-effect_2.10-7.1.6.jar"))
+  new File(baseDirectory.value, "../../lib/bundle/scalaz-concurrent_2.10-7.1.6.jar"),
+  new File(baseDirectory.value, "../../lib/bundle/scalaz-effect_2.10-7.1.6.jar"),
+  new File(baseDirectory.value, "../../lib/bundle/scalaz-stream_2.10-0.7.2a.jar"),
+  new File(baseDirectory.value, "../../lib/bundle/doobie-core_2.10-0.2.4-SNAPSHOT.jar"),
+  new File(baseDirectory.value, "../../lib/bundle/shapeless_2.10-2.2.5.jar"),
```
cquiroz: On the scala 2.11 branch we use the latest version of breeze which depends on shapeless 2.0, We'll have to check at integration time if there is conflict here

bundle/edu.gemini.util.security/src/main/scala/edu/gemini/util/security/auth/keychain/KeySchema2.scala
```diff
@@ -0,0 +1,157 @@
...
+import java.security.Principal
+import java.util.logging.Logger
+
+import scalaz._, Scalaz._, effect.IO
+
+object KeySchema2 extends KeyMappers2 {
+  lazy val Log: Logger = java.util.logging.Logger.getLogger(KeySchema.getClass.getName)
+
+  // H2-specific SqlState
+  val TABLE_OR_VIEW_NOT_FOUND = SqlState("42S02")
```
cquiroz: Is this an error constant?

bundle/edu.gemini.util.security/src/main/scala/edu/gemini/util/security/auth/keychain/KeySchema2.scala
```diff
@@ -0,0 +1,157 @@
...
+
+  ////// DATABASE OPERATIONS
+
+  val createSchema: ConnectionIO[Unit] = 
+    sql"""
+      create table VERSION (
+        VALUE INTEGER NOT NULL
+      );
+      create table KEYS (
+        CLASS   VARCHAR NOT NULL,
```
cquiroz: Do we have to write the class name on the db?

bundle/edu.gemini.util.security/src/main/scala/edu/gemini/util/security/auth/keychain/KeySchema2.scala
```diff
@@ -0,0 +1,157 @@
...
+    val Affiliate = "Affiliate"
+    val Program   = "Program"
+    val User      = "User"
+    val Staff     = "Staff"
+    val Visitor   = "Visitor"
+  }
+
+  // Clazz extractor
+  implicit class EnhanceGP(p: GeminiPrincipal) {
+    def clazz: String = p match {
```
cquiroz: I guess this will act as a marker column to tell which type of object each row represent, right?

cquiroz:
I skimmed the doobie documentation, which was very useful to understand this PR.
The code seems very good :100: , with tests and self creation of the schema at startup, I'd like that we get rid of the slick dependency altogether before merging into scala 2.11
A general question, we are using H2 right now, can you imagine a scenario that would require us to move to e.g. postgres?

bundle/edu.gemini.util.security/src/main/scala/edu/gemini/util/security/auth/keychain/KeySchema2.scala
tpolecat: Yes, it's the error code from H2 when you reference an unknown table.

bundle/edu.gemini.util.security/src/main/scala/edu/gemini/util/security/auth/keychain/KeySchema2.scala
tpolecat: We write a tag that tells you what kind of principal we're talking about. It's not the literal class name,

bundle/edu.gemini.util.security/src/main/scala/edu/gemini/util/security/auth/keychain/KeySchema2.scala
tpolecat: Right.

tpolecat:
@cquiroz yes, I'm adding a vcs log implementation with doobie and then once both are merged I'll remove Slick in another PR.

And yes, I think for the OCS re-design we're thinking about I would like to use PostgreSQL for the back end (assuming we decide to use a relational database).

PR #809, opened 2016-02-09
OCSADV-508, REL-2611: Make ITC "Signal" graph for 1 pixel

fnussber:
This PR is a collection of refactorings, cleanups and F2 bug fixes (REL-2611). The main focus was a bunch of refactorings that allow to calculate the signal and background strength in  a different way, i.e. for a single pixel instead of for all pixels in the slit/aperture. This is necessary to calculate the peak pixel flux for a single pixel appropriately, which in turn allows to give saturation warnings for spectroscopy observations.

This is work in progress, while the signal and background are calculated properly in most cases now it doesn't work yet for IFU calculations (GMOS and NIFS), unfortunately this turns out to be again relatively complicated. However Andy and I agreed it would be nice to have this set of changes out there for testing asap, so we are hopeful this PR can make the next test release.


swalker2m:
It looks like the build failure was from the recent Data Manager update that I made, which has now been fixed (or so I claim).

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/base/Disperser.java
```diff
@@ -0,0 +1,48 @@
...
+    double dispersion();
+
+    /** Spectral resolution in nm for a 0.5-arcsec slit.
+     *  This value can be used to extrapolate the resolution for any given slit which is the default
+     *  behavior for most instruments and dispersers. */
+    double resolutionHalfArcsecSlit();
+
+    /** Calculates the size of a spectral resolution element in [nm] for the given slit.
+     *  By default this is done by extrapolating from the 0.5-arcsec resolution. */
+    default double resolution(final Slit slit) {
```
cquiroz: default methods :+1:

cquiroz:
I can't really comment on the internals of the code as they are very ITC specific. However the code seems good, and tests are passing so :+1:

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/base/Disperser.java
fnussber: AKA poor man's trait ;)

fnussber:
I am still mostly moving around code and in particular getting rid of copy-paste things, like e.g. whole blocks of code that are not used at all etc. Also adding comments. Overall I think we are getting closer to the point were future refactorings can make things a lot better and easier by introducing more concepts like the ones new here (`Slit` and `Disperser`). Baby steps. And overall probably still faster than a rewrite, in particular since we need to keep it working AND add new functionality. Just saying.

PR #810, opened 2016-02-10
update doobie jar w/dynamicimport-package header

tpolecat:
This fixes a classloading issue in the vcs bundle initializer.


cquiroz:
Could you have used it as a managed dependency?

tpolecat:
Not at the moment. OSGi header for shapeless is currently [broken](https://github.com/milessabin/shapeless/issues/510) but it's fixed for the next version so we can try again.

PR #811, opened 2016-02-10
Turn off PollServiceSpec tests.

swalker2m:
Skips `PollServiceSpec` tests since they cannot run reliably in travis.

### DIFF
--- a/bundle/edu.gemini.dataman.app/src/main/scala/edu/gemini/dataman/app/PollService.scala
+++ b/bundle/edu.gemini.dataman.app/src/main/scala/edu/gemini/dataman/app/PollService.scala
@@ -40,6 +40,17 @@ sealed trait PollService {
 
   /** Stops the poll service and cleans up resources. */
   def shutdown(): Unit
+
+  /** Returns true if the PollService is not currently polling and has no
+    * pending ids to poll at the moment this method is called.
+    */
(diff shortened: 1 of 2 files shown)


cquiroz:
I'd argue that if a test cannot run reliably because the hardware is perhaps too slow, the test should be changed to run in that environment rather than just be ignored
We probably have a bunch of tests that have that kind of timeouts built it

swalker2m:
The problem is that there is no good answer for how long an operation should "usually" take.  If I increase the timeouts to ridiculous values then the test phase of the build will, corresponding, take ridiculously longer to finish.

cquiroz:
Yes, it is difficult to tell how long to wait, is it talking to a remote service?
Short of rethinking the whole test I think is ok to have a longer timeout considering the (presummably) slower hardware

swalker2m:
@cquiroz I looked closely at the tests and found that really only one couldn't be fixed with a ridiculously long timeout.  The others should still terminate long before the timeout, presumably.  So I increased the timeout considerably and was able, I think, to change the problematic test to make it work in this environment.  Thanks for the pushback.  In the end the `PollServiceSpec` is not only still viable but also better than it was.  That's my hope anyway.  We'll see what verdict Travis renders.

cquiroz:
Travis agrees with you :+1:

PR #814, opened 2016-02-12
FIX: Added missing dependencies to ITC standalone app.

fnussber:
Dependencies that were introduced with `doobie` were missing in the ITC standalone server app.

(This is a stripped down server app that only runs the ITC and is currently used by science for development and is also what is running on itc.gemini.edu to provide the ITC calculation web pages to the outside world.)


cquiroz:
:+1:

PR #815, opened 2016-02-12
vcs log → doobie

tpolecat:
This adds a VCS log implementation that uses doobie rather than Slick.
- This uses the same schema and logic as the existing Slick-based implementation.
- The Slick implementation is still there but the doobie one is used by default now.
- IDEA project generation had to be changed to add the macro paradise plugin, which is required by Shapeless for Scala 2.10; this can go away when we're on 2.11


cquiroz:
The code looks good to me, though there is a failing test. :+1: 
About the database evolution, I see you have the changes in the code with version number. Is the version checked when starting? Perhaps, we should document how to update the code when the database schema changes
I'm looking forward to integrate these changes into the scala 2.11 branch

tpolecat:
Ugh, the failure is a compiler bug. I thought I had finally worked around it but I guess not.

tpolecat:
Confirmed that the failure is a thread safety issue with object initialization in 2.10 … shouldn't happen in production but it can be provoked by running a bunch of tests in parallel. Turned off parallel test execution as a workaround.

arturog8m:
Will this go away in 2.11?

cquiroz:
Could we leave  it as it is and enable it on the scala 2.11 port?

tpolecat:
Yes, it's fixed in 2.11 so we can turin parallel execution back on.

tpolecat:
Damn it. Still failing.

tpolecat:
Ok the times are all offset by 8h so it's a timezone thing. Working on it.

cquiroz:
I guess you can pass the timezone for the jvm with `-Duser.timezone=America/Chicago`, though that may not be the most correct way to handle this

tpolecat:
Sigh. Timezone offsetting doesn't work correctly. It rolls the time but not the date. Working on it...

PR #818, opened 2016-02-15
Cleaned sequence commands

jluhrs:
The main goal of this task was to clean the sequence commands. In the initial prototype, commands could only be triggered from the console. The code that implemented the console commands had details that should have been in the sequence command themselves, so I fixed that. Also, I modified some commands to provide an immediate response in case of an error. For example, the method to run a sequence would have returned a Task, that has to be run to know if the sequence could actually be run. But if there was no problem, running that Task would actually run the sequence. So now the command is more like a "start sequence", that return a value that tells if the sequence started or there was a problem.

A secondary issue that I tried to fix is that a previous modification that changed some singletons to be objects exposed an initialization problem.

### DIFF
--- a/bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Executor.scala
+++ b/bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Executor.scala
@@ -32,20 +32,20 @@ object Step {
       q <- observe.leftMap(NonEmptyList(_))
     } yield StepResult(p, q)
 
-  def step(config: Config): (Set[System], Step) = {
+  def step(config: Config): SeqexecFailure \/ (Set[System], Step) = {
     val instName = config.getItemValue(new ItemKey(INSTRUMENT_KEY, INSTRUMENT_NAME_PROP))
     val instrument = instName match {
       case GmosSouth.name => Some(GmosSouth)
-      case Flamingos2.name => Some(Flamingos2(Flamingos2ControllerEpics))
+      case Flamingos2.name => Some(Flamingos2(Flamingos2ControllerSim))
       case _ => None
     }
 
     instrument map { a => {
-        val systems = List(Tcs(TcsControllerEpics), a)
-  //      val systems = List(Tcs(TcsControllerSim), a)
-        (systems.toSet, step(systems.map(_.configure(config)), a.observe(config)))
+//        val systems = List(Tcs(TcsControllerEpics), a)
+        val systems = List(Tcs(TcsControllerSim), a)
+        (systems.toSet, step(systems.map(_.configure(config)), a.observe(config))).right
       }
-    } getOrElse((Set[System](), EitherT(Task(NonEmptyList[SeqexecFailure](UnrecognizedInstrument(instName.toString)).left[StepResult]))))
+    } getOrElse(UnrecognizedInstrument(instName.toString).left[(Set[System], Step)])
   }
 
 }
--- a/bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/ExecutorImpl.scala
+++ b/bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/ExecutorImpl.scala
@@ -1,5 +1,8 @@
 package edu.gemini.seqexec.server
 
+import edu.gemini.seqexec.shared.{SeqFailure, SeqExecService}
+import edu.gemini.spModel.core.Peer
+
 import scalaz._
 import Scalaz._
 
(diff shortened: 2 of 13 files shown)


bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Executor.scala
```diff
@@ -32,20 +32,20 @@ object Step {
       q <- observe.leftMap(NonEmptyList(_))
     } yield StepResult(p, q)
 
-  def step(config: Config): (Set[System], Step) = {
+  def step(config: Config): SeqexecFailure \/ (Set[System], Step) = {
     val instName = config.getItemValue(new ItemKey(INSTRUMENT_KEY, INSTRUMENT_NAME_PROP))
     val instrument = instName match {
       case GmosSouth.name => Some(GmosSouth)
-      case Flamingos2.name => Some(Flamingos2(Flamingos2ControllerEpics))
+      case Flamingos2.name => Some(Flamingos2(Flamingos2ControllerSim))
```
cquiroz: I guess these controls if we talk to the real systems or are in simulation, right?

bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/ExecutorImpl.scala
```diff
@@ -13,39 +16,56 @@ import scalaz.concurrent.Task
...
+    }}
+  }
 
   // todo: it is an error to run a sequence with an existing ExecState != s
   private def runSeq(id: SPObservationID, s: ExecState): Task[(ExecState, NonEmptyList[SeqexecFailure] \/ Unit)] =
     cancelRef.modify(_ - id) *> recordState(id)(s) *> run(go(id), recordState(id))(s)
 
-  def start(id: SPObservationID, sequenceConfig: ConfigSequence): Task[(ExecState, NonEmptyList[SeqexecFailure] \/ Unit)] = {
-    sequence(sequenceConfig) match {
+  def start(id: SPObservationID): SeqexecFailure \/ Unit = for {
```
cquiroz: The return type tells me that I either get an error or I get nothing. I'd expect, in the non-case error, to get something I can use to follow or interrupt the sequence, or do we rely on the id only?

bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/ExecutorImpl.scala
```diff
@@ -13,39 +16,56 @@ import scalaz.concurrent.Task
...
-  def continue(id: SPObservationID): Task[(Option[ExecState], NonEmptyList[SeqexecFailure] \/ Unit)] =
-    getState(id) >>= {
-      case -\/(e) => Task.now((None, NonEmptyList(e).left))
-      case \/-(s) => runSeq(id, s).map(_.leftMap(_.some))
-    }
+  def continue(id: SPObservationID): SeqexecFailure \/ Unit = continueTask(id) map (_.runAsync(???))
+  private def continueTask(id: SPObservationID): SeqexecFailure \/ Task[(Option[ExecState], NonEmptyList[SeqexecFailure] \/ Unit)] =
+    getState(id).run map {runSeq(id, _).map(_.leftMap(_.some))}
+
+  def state(id: SPObservationID): SeqexecFailure \/ ExecState = getState(id).run
```
cquiroz: I guess you can poll this method to follow the sequence, but I'd like to have a method that pushes me changes, it could be like a callback or perhaps other nicer construct

bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Flamingos2Epics.scala
```diff
@@ -99,26 +100,27 @@ object Flamingos2Epics {
...
         .buildAll()
 
-      val f2State = CaService.getInstance().getStatusAcceptor("flamingos2::status")
+      instance = new Flamingos2Epics(service)
 
       TrySeq(())
     } catch {
       case c: Throwable =>
-        Log.warning("Flamingos2Epics: Problem initializing EPICS service: " + c.getMessage)
+        Log.warning("Flamingos2Epics: Problem initializing EPICS service: " + c.getMessage + "\n"
```
cquiroz: Can you pass the full exception to `Log.warning` rather than the stack trace?

bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Flamingos2Epics.scala
```diff
@@ -99,26 +100,27 @@ object Flamingos2Epics {
...
 }
 
-object Flamingos2EpicsInitializer {
+object Flamingos2Epics extends EpicsSystem {
+  
   val Log = Logger.getLogger(getClass.getName)
   val CA_CONFIG_FILE = "/Flamingos2.xml"
+  
+  // Ugly use of null. But if it is used before initialization, then the application deserves to crash.
+  var instance: Flamingos2Epics = null
```
cquiroz: I think in these cases it maybe better to make `Flamingos2Epics` into a concrete class and pass the `CaService` as a param

bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Executor.scala
jluhrs: That is a pending issue. At some point I have to code that selection, based in some property.

bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/ExecutorImpl.scala
jluhrs: At that level, it's just the id.

bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Flamingos2Epics.scala
jluhrs: It is like that. I use the "instance" var only to access the only instance of the class. Maybe I should use a method to access the only instance, that checks if it exists and creates it otherwise.

cquiroz: You may be able to achieve the same making your instance a `lazy val` and pulling the `CaService` from the singleton class

bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Flamingos2Epics.scala
```diff
@@ -99,26 +100,27 @@ object Flamingos2Epics {
...
 
-  def init: TrySeq[Unit] = {
+  override def init(service: CaService): TrySeq[Unit] = {
     try {
       (new XMLBuilder).fromStream(this.getClass.getResourceAsStream(CA_CONFIG_FILE))
```
cquiroz: What's the use of this line?

jluhrs: XMLBuilder builds command senders and status acceptors from a XML configuration file.

cquiroz: Looks like a no-op from here, does it set some state in `Flamingos2Epics`?

jluhrs: No, it creates a series of command senders and status acceptors in CaService, which then Flamingos2Epics retrieves by name.

The XML configuration files are created from the old CA files in Seqexec with a little tool I wrote. They have the list of all EPICS channels that are read or written by the application. I certain improvement would be to have the tool generate code to create all the command senders and status acceptors, allowing for direct references and compile checks. Hopefully in the (near) future...

bundle/edu.gemini.seqexec.server/src/main/scala/edu/gemini/seqexec/server/Flamingos2Epics.scala
```diff
@@ -99,26 +100,34 @@ object Flamingos2Epics {
...
+  
+  // Still using a var and null, but at least now it's hidden. Attempts to access the single instance will
+  // now result in an Exception with a meaningful message, instead of a NullPointerExecption
+  private var instanceInternal: Flamingos2Epics = null
+  lazy val instance: Flamingos2Epics = {
+    if (instanceInternal == null) {
+      throw new Exception("Attempt to reference Flamingos2Epics single instance before initialization.")
+    }
+    instanceInternal
+  }
```
swalker2m: I haven't been following this closely but this seems a bit unusual at least.  Is there a reason to not just make a `Flamingos2Epics` when you need it?

If you are going to do this though, you could hide the `null` with (assuming scalaz syntax is imported):

````scala
private var instanceInternal = Option.empty[Flamingos2Epics]

def instance: Flamingos2Epics =
  instanceInternal | throw new Exception("...")
````

jluhrs: Yes, I should have used an Option. I'll fix it.
The reason I wanted to treat Flamingos2Epics as a singleton is that it is not allowed to have more than one instance, because that means more than one object will try to control the instrument. But I could create and destroy the single instance on demand.

PR #819, opened 2016-02-15
REL-2632: Small TableUtil bug

sraaphorst:
The former `TableUtil` was not calculating column widths correctly as it was using the wrong renderers, so I made an attempt to fix this properly.

However, I made a small error in doing so, which used the wrong renderer to calculate header widths if a header renderer was not explicitly defined, and resulted in an exception in terms of the data contained in the data.

This can be seen by creating a new GSAOI observation and attempting to do a GeMS manual guide star search: a table will pop open and a complaint will be issued about String vs. Boolean types.

This PR simply changes the header renderer to be the correct one, i.e. a default, instead of using the cell renderer if the header renderer is not explicitly defined.


cquiroz:
Looks fine to me :+1:

PR #821, opened 2016-02-16
Feature/REL-491: GMOS ITC should not allow spatial binning for IFU mode

spakzad:
Edited Gmos.java so that if the user selects an IFU and a spatial binning that is not 1, the web form will produce an error message.  Changed two tests in BaselineGmos.scala that allowed an IFU and a spatial binning of 2.  Changed the baseline.txt.


bundle/edu.gemini.itc/src/test/java/edu/gemini/itc/gmos/CCDGapCalcTest.java
```diff
@@ -0,0 +1,24 @@
...
+public class CCDGapCalcTest {
+
+    @Test
+    public void thisIsATest() {
+        final double actualResult0  = CCDGapCalc.calcIfu2Shift(750.0, 831.0);
+        final double expectedResult0 = 2736.24773782;           // Taken from gfoneshift in PYRAF
+        final double actualResult1  = CCDGapCalc.calcIfu2Shift(750.0, 150.0);
+        final double expectedResult1 = 3645.34948931;           // Taken from gfoneshift in PYRAF
+
+        Assert.assertTrue("this is true", true);
```
cquiroz: This line could be removed

cquiroz:
This PR looks good to me, we could just merge it :+1:

bundle/edu.gemini.itc/src/test/java/edu/gemini/itc/gmos/CCDGapCalcTest.java
sabrinapak: Hi,

I don't know what line you are referring to, but I assume it is the one that says "[Add more details!!!]".

I've tried to add more details, but I'm having trouble doing a pull request.  The line above should be replaced with the following text:
* CCDGapCalc.java is based on the IRAF script gfoneshift.cl, which takes the central wavelength
* and the lpmm and outputs the shift. This is used in the GMOS IFU-2 calculations.
* This test was created to compare the output of CCDGapCalc.java to gfoneshift.cl.

spakzad: Hi,

I don't know what line you are referring to, but I assume it is the one that says "[Add more details!!!]".

I've tried to add more details, but I'm having trouble doing a pull request. The line above should be replaced with the following text:

CCDGapCalc.java is based on the IRAF script gfoneshift.cl, which takes the central wavelength
and the lpmm and outputs the shift. This is used in the GMOS IFU-2 calculations.
This test was created to compare the output of CCDGapCalc.java to gfoneshift.cl.

spakzad:
This pull request has been closed after discussions with Andy Stephens and Kathy Roth (turns out the ITC should allow spatial binning, but strongly discourage it, like in the OT).

bundle/edu.gemini.itc/src/test/java/edu/gemini/itc/gmos/CCDGapCalcTest.java
cquiroz: I was referring to
```
 Assert.assertTrue("this is true", true);
```

But that's ok anyway, just small detail

PR #822, opened 2016-02-17
Minimize build logging to make Travis output more useful.

tpolecat:
This PR removes the goddamn AGS logging, as you might surmise. With some luck this will bring the log size down far enough that we don't need to use the raw one. Let's see.

### DIFF
--- a/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsResultsAnalyzer.scala
+++ b/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsResultsAnalyzer.scala
@@ -131,10 +131,10 @@ object GemsResultsAnalyzer {
   }
 
   private def printResults(result: List[GemsGuideStars]) {
-    Log.info("Results:")
-    result.zipWithIndex.foreach { case (s, i) =>
-      Log.info(s"result #$i : $s")
-    }
+    // Log.info("Results:")
+    // result.zipWithIndex.foreach { case (s, i) =>
+    //   Log.info(s"result #$i : $s")
+    // }
   }
 
   private def sortResultsByRanking(list: List[GemsGuideStars]): List[GemsGuideStars] = {
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,4 +29,4 @@ before_script:
   - if [ $TRAVIS_SECURE_ENV_VARS == false ]; then
       cp $TRAVIS_BUILD_DIR/project/OcsCredentials.scala.template $TRAVIS_BUILD_DIR/project/OcsCredentials.scala;
     fi
-script: sbt $TRAVIS_SCALA_VERSION test:compile test
+script: sbt $TRAVIS_SCALA_VERSION --warn test:compile test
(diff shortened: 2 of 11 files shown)


cquiroz:
It looks like it wasn't enough to get to the raw one. Anyway, I'm ok with this change, I doubt anybody uses the log output.

andrewwstephens:
Does this affect logging to the console?

tpolecat:
@cquiroz hm still not enough. i'll see if i can turn down some other things.
@andrewwstephens yes, this affects console output

andrewwstephens:
The console output is useful.  :disappointed:

tpolecat:
Console output is not really intended to be observable …

andrewwstephens:
How about an option to enable verbose console output?

tpolecat:
@andrewwstephens we could do that .. it's a config setting. Most of the irritating output was `println`s rather than proper log messages but we could change those.

cquiroz:
Looks like a big portion of the extra output is complaints about warnings, perhaps you can pass a param to sbt to hide those? I wouldn't want to blank suppress them though

swalker2m:
The data manager tests are a bit verbose with the logging.  I'll see about calming that a bit.  Perhaps for running the test cases the log level should be upped to severe?  If something fails you can always rerun it with a lower log level if you want.

I'm not sure but I think at least in the long run, I think I'd rather get the build and test log under control than turn off the colors in order to make the raw log easier to read.

tpolecat:
@swalker2m yeah maybe cranking up the log level in tests would work. I don't know how to do that but how hard could it be? heh-heh

But honestly I think almost all of the logging is useless so I'd like to go through and get rid of a bunch.

tpolecat:
@swalker2m (and anyone else) you can make a PR to this branch if you want to get in on the action.

cquiroz:
It seems all tests are green now :+1:

swalker2m:
Looks like we're still almost 8,000 lines too long.

cquiroz:
Can't we pass a java.util.logging properties file that only shows SEVERE?

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsResultsAnalyzer.scala
```diff
@@ -130,10 +130,10 @@ object GemsResultsAnalyzer {
...
 
   private def printResults(result: List[GemsGuideStars]) {
-    Log.info("Results:")
-    result.zipWithIndex.foreach { case (s, i) =>
-      Log.info(s"result #$i : $s")
-    }
+    // Log.info("Results:")
+    // result.zipWithIndex.foreach { case (s, i) =>
+    //   Log.info(s"result #$i : $s")
+    // }
```
swalker2m: One idea for this would be to make them `Log.fine("...")` instead of commenting them out.  Then @andrewwstephens could lower the log level in his OT and see all the minute details.

tpolecat:
I'm going to play around with sbt settings ... you can turn off all the `[info] Resolving ...` messages and so on.

tpolecat:
Rebased to 2.11

tpolecat:
Ok with `sbt --warn` we don't get any test output but presumably we would if a test failed ... ?

In any case it now fits in the standard log.

cquiroz:
Nice, scala 2.11 brings its own set of extra warnings that may increase the logs

swalker2m:
There's a lot more that can be done to clean up logging but is there any reason not to merge this now? :+1:

tpolecat:
I'm going to try a few more things this morning but yeah I'll merge it today.

tpolecat:
Sorry, this got way behind `develop`. Closing now, will revisit.

PR #823, opened 2016-02-17
remove slick

tpolecat:
This removes Slick from OCS.


cquiroz:
Is it in any of the apps definitions?

tpolecat:
I did a search and `slick` doesn't appear in any `.sbt`, `.scala`, or `.java` files so I think we're ok.

tpolecat:
Hm where is Travis?

cquiroz:
Hmm, good question, maybe it is waiting for #818 which is process?

tpolecat:
@cquiroz ok that fixed it. I did a bunch of head resets and forced pushes so it managed to confuse travis/

tpolecat:
Ok all green. Comments?

cquiroz:
:+1:

PR #824, opened 2016-02-17
REL-2644: fix confirm group change bug

swalker2m:
In an offset position list, you can configure custom guiding options for specific guide probes.  For example, you can specify that PWFS2 be parked at a particular offset position.  Over in the target component, you can specify multiple guide groups, some of which may have a PWFS2 guide star while others do not.  When you have custom guide configuration for a guide probe and then switch guide groups to one that doesn't include the probe for which there is custom configuration, the OT should display a confirmation dialog informing you of the consequences and allowing you to reconsider.

This is a small bug fix that takes longer to explain than fix.  The upshot is that we were calling `OffsetUtil.allOffsetPosLists` with a target component node instead of with its "context" observation.  That results in an empty list of offset position lists instead of all the offset position lists in the observation.


cquiroz:
Looks fine to me, let's see what travis says about it

PR #825, opened 2016-02-17
REL-2576B: Activate peak pixel infromation and warnings for spectroscopy

fnussber:
After we noticed that the peak pixel values were not calculated properly for the December release (peak pixel was based on all pixels in the aperture instead of just a single pixel), we decided to turn peak pixel information and warnings off until we can fix this in order not to confuse the users.

PR https://github.com/gemini-hlsw/ocs/pull/809 addressed this issue (with the notable exception of IFUs),  so that we can turn on those warnings again. IFU support will follow in a later PR.


cquiroz:
:+1:

PR #827, opened 2016-02-18
REL-552 Added imaging option to GNIRS ITC, and Altair option for imaging.

olesjasmirnova:


bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gnirs/Gnirs.java
```diff
@@ -110,46 +150,38 @@ public Gnirs(GnirsParameters gp, ObservationDetails odp) {
...
+        if (_mode instanceof Spectroscopy) {
+            _gratingOptics = new GnirsGratingOptics(
+                    getDirectory() + "/" + getPrefix(), _grating.get(),
+                    _centralWavelength,
+                    _detector.getDetectorPixels(),
+                    1,
+                    isLongCamera() ? LONG_CAMERA_SCALE_FACTOR : 1,
+                    1);
+
+            if (_grating.equals(Disperser.D_10) && !isLongCamera())
```
cquiroz: I think this will always fail, `_grating` is an instance of `Option<Disperser>`

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gnirs/Gnirs.java
```diff
@@ -110,46 +150,38 @@ public Gnirs(GnirsParameters gp, ObservationDetails odp) {
...
+        }
         addComponent(_detector);
-
-
     }
 
     public edu.gemini.itc.base.Disperser disperser(final int order) {
         return new GnirsGratingOptics(
-                getDirectory() + "/" + getPrefix(), _grating,
+                getDirectory() + "/" + getPrefix(), _grating.get(),
```
cquiroz: Calls to `get()` may fail at runtime, `getOrElse` is recommended or we should return `Option<Disperser>`

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gnirs/Gnirs.java
```diff
@@ -250,11 +293,14 @@ public double getCentralWavelength() {
...
     public TransmissionElement getGratingOrderNTransmission(int order) {
-        return GnirsGratingsTransmission.getOrderNTransmission(_grating, order);
+        return GnirsGratingsTransmission.getOrderNTransmission(_grating.get(), order);
     }
 
     private double correctedCentralWavelength() {
-        if (!isXDispUsed()) {
+        if (_mode instanceof Imaging) {
+            _Filter = Filter.fromFile(getPrefix(), params.filter().get().name(), getDirectory() + "/");
+            return (int) _Filter.getEffectiveWavelength();
```
cquiroz: Wouldn't the conversion to `(int)` make it loose precision? The method is returning `double`

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gnirs/Gnirs.java
swalker2m: Perhaps it could be something like:

````scala
if (_grating.exists(_ == Disperser.D_10) && !isLongCamera())
````

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gnirs/Gnirs.java
olesjasmirnova: True. It was a copy-paste from some other place. I'll just remove the casting.

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gnirs/Gnirs.java
olesjasmirnova: `disperser` is only called from `GnirsRecipe.calculateSpectroscopy`, which is only called when we the user is asking for spectroscopy (i.e. explicitly defines a disperser), and in that case we're guaranteed to have a value defined for `_grating`.

If you still think the changes need to be made, as the use is restricted to `GnirsRecipe` I could make the method to return `Option<Disperser>` - not too many changes to the code.

fnussber: I should add that this is a more general problem than just in this one place; the reason being that (as @olesjasmirnova points out) for spectroscopy and imaging calculations different preconditions are true (e.g. disperser *must* be present for spectroscopy).

These are sometimes checked (using `Exception`s to signal the problem to the user) and sometimes not. Originally all of these optional objects (filters, disperser) were handled as present or `null`, and to make this more explicit I started to use `Option`s but never took the time to really clean this up through all levels of the call hierarchy, that's why there are some places with `get`s.

Not trying to say this is good, just trying to point out the history. I was always planning to clean this up and introduce a clearer separation between spectroscopy and imaging cases which would allow to do the checking of preconditions (disperser present? ..etc) more explicit and turn the optional objects into the actual ones in the same step and then pass those into the calculations. But I guess someone else will have to follow up with that. So please be nice with @olesjasmirnova when she uses `get` in some places ;)

cquiroz: Sorry if I was too blunt, in no case I want to discourage any of our collaborators. @olesjasmirnova If you have logical reasons why an `Option` should be valid (like in this case), it is ok IMHO to use `get()` directly

fnussber: No worries. It can certainly be improved and it was correct to point it out of course. I was just trying to give a bit more context.

fnussber:
@olesjasmirnova: You will have to merge the latest changes into your branch, otherwise it can not be merged.

PR #830, opened 2016-02-19
REL-2649: Hide TEXES in the 2016B PIT

cquiroz:
Small PR that hides Texes on the PIT and remove some deprecation warnings


sraaphorst:
Looks completely straightforward. :+1:

PR #831, opened 2016-02-19
REL-2522: Implement backend monitors for hbfphase1-lv1

cquiroz:
This PR contains updates to the phase1 monitor including:

- Remove the use of the legacy actors library
- Support that PDF templates can be assigned on a per monitor basis
- Setup final configuration of the PIT and P1-monitor for the final release

### DIFF
--- a/bundle/edu.gemini.p1monitor/src/main/scala/edu/gemini/p1monitor/P1MonitorDirMonitor.scala
+++ b/bundle/edu.gemini.p1monitor/src/main/scala/edu/gemini/p1monitor/P1MonitorDirMonitor.scala
@@ -34,22 +34,24 @@ class P1MonitorDirMonitor(cfg: P1MonitorConfig) extends DirListener {
       originalsDir.mkdirs()
     }
 
-    def writeUpdated(attachment: File, proposal: Proposal, f: File) {
+    def writeUpdated(attachment: File, proposal: Proposal, f: File):Proposal = {
       // Use just the name to avoid hardcoding the file
       val attachedFileName = attachment.getName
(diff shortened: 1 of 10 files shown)


bundle/edu.gemini.p1monitor/src/main/scala/edu/gemini/p1monitor/P1MonitorDirMonitor.scala
```diff
@@ -36,22 +34,24 @@ class P1MonitorDirMonitor(cfg: P1MonitorConfig) extends DirListener {
...
-      newPDFile.foreach {
+      newPDFile.flatMap {
         case p =>
           // This sounds like a good idea, let Proposal to do the IO, but this will need some help to avoid writing
           // the absolute path of the file
-          for {
+          (for {
             conversion <- ProposalIo.readAndConvert(f)
-          } yield writeUpdated(p, conversion.proposal, f)
+          } yield writeUpdated(p, conversion.proposal, f)).toOption
```
swalker2m: Not that I haven't done it anyway, but I always feel a bit sheepish about doing side-effects in a map.  Can the actual writing to a file be deferred so that the map is pure?

cquiroz: Well, the p1-monitor was done at our initial introduction to Scala, you can see many, non very good decisions in here. Given how small it is and the little time I have, I never bothered to improve it substantially.
I hope to do a rewrite when we do the real p1-backend

swalker2m:
:+1:

PR #832, opened 2016-02-19
Removal/Reduction of warnings on the qpt bundles

cquiroz:
Almost all warnings gone on the QPT bundles, In some places I resorted to blanket '@SupressWarnings` a few classes, as there were just too many to handle, sorry


tpolecat:
:+1:

PR #833, opened 2016-02-19
Phoenix and DSSI should display a single site where it is installed

cquiroz:
Last minute request to change the PIT UI, displaying only the sites where a visitor instrument is really installed


tpolecat:
:+1:

PR #835, opened 2016-02-20
prep for target model swap

tpolecat:
This PR is in preparation for swapping the target model; the intent is to minimize the number of references to `ITarget` since this type is going away. Here we are removing references in the form `spt.getTarget.foo` by putting a method `foo` on `SPTarget` that delegates to the underlying `ITarget`. These will be replaced with equivalent operations on the underlying `Target` and then inlined back into the current call sites, where appropriate.

The changes here are largely mechanical. More will follow.

### DIFF
--- a/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/Strategy.scala
+++ b/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/Strategy.scala
@@ -13,7 +13,7 @@ import edu.gemini.spModel.guide.{GuideProbe, GuideProbeUtil}
 import edu.gemini.spModel.obs.context.ObsContext
 import edu.gemini.spModel.rich.shared.immutable._
 import edu.gemini.spModel.target.obsComp.PwfsGuideProbe
-import edu.gemini.spModel.target.system.NonSiderealTarget
+import edu.gemini.spModel.target.system.{ITarget, NonSiderealTarget}
 
 import scala.Function.const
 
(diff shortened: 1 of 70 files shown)


bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/Strategy.scala
```diff
@@ -72,7 +72,7 @@ object Strategy {
   }
 
   def isSidereal(ctx: ObsContext): Boolean =
-    !ctx.getTargets.getBase.getTarget.isInstanceOf[NonSiderealTarget]
+    ctx.getTargets.getBase.isSidereal
```
cquiroz: Just this line is worth the change :smile:

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/system/TransitionalSPTarget.java
```diff
@@ -2,9 +2,11 @@
 
 import edu.gemini.shared.skyobject.Magnitude;
 import edu.gemini.shared.util.immutable.ImList;
+import edu.gemini.shared.util.immutable.Option;
 import edu.gemini.spModel.core.SpatialProfile;
 import edu.gemini.spModel.core.SpectralDistribution;
 import edu.gemini.spModel.target.WatchablePos;
+import scala.None$;
```
cquiroz: Is this import really being used?

cquiroz:
Looks very good :+1:

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/system/TransitionalSPTarget.java
tpolecat: Ah, nope. Thanks.

PR #836, opened 2016-02-23
OCSADV-519-C: Initial AgsHash function

swalker2m:
This PR introduces a simple hash function for AGS.  In particular, given an `ObsContext` it will compute a hash value based on variables that play a role in AGS calculations.  The plan is to store this information in an LRU cache in the OT to avoid performing AGS lookups unless there is a chance that the algorithm will come up with a different result.

I think I did a somewhat reasonable job taking into account the various factors in play but nevertheless I expect I've overlooked some details.  This function will likely grow a bit as missing elements are discovered but should serve as a starting point.

### DIFF
--- a/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsHash.scala
+++ b/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsHash.scala
@@ -61,7 +61,7 @@ object AgsHash {
       val base = t.getBase.getTarget
 
       def toData(coord: GemOption[java.lang.Double]): Int =
-        coord.asScalaOpt.map(_.doubleValue.scaled(8)).##
+        coord.asScalaOpt.map(_.doubleValue).##
 
       val ra  = toData(base.getRaDegrees(time))
       val dec = toData(base.getDecDegrees(time))
@@ -72,12 +72,12 @@ object AgsHash {
     // Offset Positions, which are returned in a Set.  Order is not important
     // for the purpose of AGS calculations.
     M3.unorderedHash(ctx.getSciencePositions.asScala.map { o =>
-      (o.p.arcsec.scaled(3), o.q.arcsec.scaled(3))
+      (o.p.arcsec, o.q.arcsec)
     }) +=: buf
 
     // Position Angle
     Option(ctx.getPositionAngle).foreach { a =>
-      a.degrees.scaled(3).## +=: buf
+      a.degrees.## +=: buf
     }
 
     // Position Angle Constraint
@@ -139,10 +139,5 @@ object AgsHash {
 
     def arcsec: Double =
       a.toArcsecs.getMagnitude
-  }
-
-  private implicit class DoubleOps(d: Double) {
-    def scaled(scale: Int): Double =
-      BigDecimal(d).setScale(scale, BigDecimal.RoundingMode.HALF_UP).doubleValue()
   }
 }
--- a/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsHash.scala
+++ b/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsHash.scala
@@ -1,0 +1,124 @@
+package edu.gemini.ags.api
+
+import edu.gemini.shared.util.immutable.ScalaConverters._
+import edu.gemini.shared.util.immutable.{Option => GemOption}
+import edu.gemini.spModel.ags.AgsStrategyKey
+import edu.gemini.spModel.gemini.flamingos2.Flamingos2
+import edu.gemini.spModel.gemini.gmos.{InstGmosNorth, InstGmosSouth}
+import edu.gemini.spModel.obs.context.ObsContext
+
+import java.time.Instant
+
+import scala.collection.mutable.ListBuffer
+import scala.util.hashing.{MurmurHash3 => M3}
+
+import scala.collection.JavaConverters._
+
+/** Computes an `Int` hash value that corresponds to the set of inputs to the
+  * AGS lookup.  If an observation is changed in some way and yet the hash
+  * algorithm computes the same value, there is no need to re-do the AGS search.
+  */
+object AgsHash {
+
+  /** Calculates the AGS hash for this context, using the current base position
+    * location.
+    *
+    * <em>Warning.</em> This method is not referentially transparent as it may
+    * provide distinct answers for the same input at different times
+    * (particularly for non-sidereal targets).
+    */
+  def hashNow(ctx: ObsContext): Int =
+    hash(ctx, Instant.now())
+
+  /** Calculates the AGS hash for this context, calculating the base position
+    * corresponding to the given time.
+    */
+  def hash(ctx: ObsContext, when: Instant): Int =
+    hash(ctx, when.toEpochMilli)
+
+
+  /** Calculates the AGS hash for this context, calculating the base position
+    * corresponding to the given time expressed in milliseconds since 1970.
+    */
+  def hash(ctx: ObsContext, when: Long): Int = {
+
+    val buf = ListBuffer.empty[Int]
+
+    // AGS Strategy
+    val strategyKey = AgsRegistrar.currentStrategy(ctx).map(_.key)
+    strategyKey.foreach { s =>
+      M3.stringHash(s.id) +=: buf
+    }
+
+    // Conditions
+    Option(ctx.getConditions).foreach { c =>
+      c.cc.## +=: c.iq.## +=: c.sb.## +=: buf
+    }
+
+    // Base Position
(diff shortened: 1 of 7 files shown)


cquiroz:
AGS (With the big exception of GEMS) is fairly quick, keeping in mind that the catalog queries are cached. Why is it needed to cache them in the OT?
Does it play a role in the sync with SPDB?

swalker2m:
@cquiroz I debated this myself, especially considering that BAGS runs in a lower priority thread.  This feature is at best an optimization and not something that we absolutely need to make it work.  To be clear, we aren't caching the AGS or catalog results.  The hash value will differentiate updates that require recalculating AGS vs. those that are innocuous from the point of view of AGS.  However fast the AGS calculation may be, the claim is that the hash computation is faster, does not require I/O, and will generate less memory churn.

The alternative to using the hash computation is that every change to any aspect of an observation will trigger a hit on the catalog server, downloading and evaluating candidates and computing the "best" alternative.  For example, each time the repeat count is updated in observe node despite the fact that this has no impact on the results.  Also, each time the user syncs her program AGS would otherwise need to recalculate results for all observations since we may be fetching updates that haven't yet been verified by AGS.

cquiroz:
Just a clarification, catalog queries are cached on the client, as long as the base doesn't move much, there shouldn't be a hit to the catalog server

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsHash.scala
```diff
@@ -0,0 +1,148 @@
...
+object AgsHash {
+
+  /** Calculates the AGS hash for this context, using the current base position
+    * location.
+    *
+    * <em>Warning.</em> This method is not referentially transparent as it may
+    * provide distinct answers for the same input at different times
+    * (particularly for non-sidereal targets).
+    */
+  def hashNow(ctx: ObsContext): Int =
```
cquiroz: Why is time a parameter on the hash? You'd think the hash is only dependent on the observation configuration

swalker2m:
Thanks, I had forgotten that catalog results are cached in the client too.  That definitely should help. I don't think it means that the hash calculation is not useful though.  AGS calculations for vignetting are not trivial and GeMS is insanely complex.  Without the hash we'd also needed to take extreme measures to avoid AGS computations while typing in text fields like one for the observation name.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsHash.scala
```diff
@@ -0,0 +1,148 @@
...
+        Option(ctx.getInstrument).foreach {
+          _.pwfs1VignettingClearance.getMagnitude.## +=: buf
+        }
+
+      case Pwfs2NorthKey | Pwfs2SouthKey =>
+        Option(ctx.getInstrument).foreach {
+          _.pwfs2VignettingClearance.getMagnitude.## +=: buf
+        }
+
+      case GemsKey                       =>
```
cquiroz: I think for GEMS you may also need NIR band?

cquiroz:
Thanks, for the clarifications, I guess we didn't need this before because AGS was manually triggered.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsHash.scala
swalker2m: The problem is that the base position is an extremely important factor, and the where the base position is depends on the time.

cquiroz: What time is that?, having `Instant.now()` gives the impression that is the current local time. If so, you'd think the hash would change every millisecond. I'm probably missing something

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsHash.scala
swalker2m: GeMS is a little confusing because the code for the manual search offers many more options.  When an AGS computation is done, it fixes most of the options.  The things we need to include are those that differ based on the `ObsContext` alone.  For example, whether it is using GSAOI or F2.

It looked to me like the NIR band was being fixed at `None` when `search` is called in the AGS `select` method.  Since it is `None` the instrument in use will determine how the filtering is done, for example in the [`GemsMagnitudeTable`](https://github.com/gemini-hlsw/ocs/blob/develop/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsMagnitudeTable.scala#L92).

I could easily be wrong though.  If you see where the NIR band is otherwise extracted from something in the context, let me know and I'll have to add that.

cquiroz: As you say GEMS is complex, checking the [search criteria](https://github.com/gemini-hlsw/ocs/blob/develop/bundle/edu.gemini.ags/src/main/java/edu/gemini/ags/gems/GemsGuideStarSearchOptions.java#L147-L189) It seems NIR and tip tilt mode are involved
On the other hand I don't know how/where those are set on the obscontext

sraaphorst: @swalker2m's comment seems right to me: the only AGS code I can see always sets the NIR band to `None`, so it should not depend on the `ObsContext` and thus can be ignored in computing the hash value?

`GemsStrategy.select` and `GemsStrategy.candidates`, which are what ultimately result in the search, both assign `None` to this value, which is ultimately passed all the way through the `GemsGuideStarSearchOptions` code @cquiroz highlighted to `GemsMagnitudeTable.LimitsCalculator`.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsHash.scala
swalker2m: I expect we won't be calling `hashNow` but rather the other `hash` methods with an explicit time that will be taken from the observing block.  This is the same time that will be used for parallactic angle calculations and (in particular) non-sidereal target coordinates.

Even if you do call `hashNow` though, the hash computation will only differ if the base position changes significantly.  The coordinates are rounded.  For sidereal targets a long time will usually pass before that happens even for the fastest moving targets -- and in fact we aren't yet even considering proper motion etc.  Even a non-sidereal target would have to be moving fast to have a different hash value every millisecond.

At any rate, the hash value is only as valid as our determination of where the target is in the sky so it *should* give us a different answer if the target is in a different location.

This raises a good point though, and I don't know the answer: how much movement should be sufficient to invalidate a hash value?  I wanted milli-arcseconds to count but I'm not sure if that is correct or even if I achieved it with a scale of 8 digits.  I picked 8 because 1 mas is 1/3,600,000 degrees or about 0.00000027778 degrees.  This may well be stupid though.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsHash.scala
cquiroz: I'm pretty sure the tiptilt mode has influence, that's not to say that it may be fixed on the higher level. Does this mean that we have a bunch or redundant code in GEMS AGS?

swalker2m: I believe `GemsStrategy.select` is used by AGS, like all the other AGS strategy implementations.  It in turn calls `GemsStrategy.search` passing in a fixed value of `GemsTipTiltMode.canopus`.  I wouldn't be surprised to be proven wrong though. I have difficulty following all the twists and turns of GeMS mascot algorithms.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsHash.scala
swalker2m: I guess technically any real change in base position should be enough to trigger a change to the hash calculation.  Double values are complicated. I was worried about small rounding-error differences being counted but now I think that doesn't matter.  We aren't going to be constantly coming up with slightly different coordinates for the same time.  I think I'll remove rounding of double values altogether.

cquiroz: I thought that was the case, a small change on time doesn't affect the base position, For AGS and sidereal targets I'd image you need a large movement to invalidate it

cquiroz:
The PR looks good, it is a pity that we don't understand GeMS AGS well enough to have clear answers. Great work with the Murmur3 hash

tpolecat:
This looks good. It worries me a bit that this will have to be updated as AGS evolves but I guess that's unavoidable. :+1:

PR #837, opened 2016-02-23
P1 monitor bug fix

cquiroz:
This is a tiny PR for a border case bug when the p1 monitor gets  a proposal with multiple visitors in each site

### DIFF
--- a/bundle/edu.gemini.p1monitor/src/main/scala/edu/gemini/p1monitor/P1MonitorMailer.scala
+++ b/bundle/edu.gemini.p1monitor/src/main/scala/edu/gemini/p1monitor/P1MonitorMailer.scala
@@ -134,7 +134,7 @@ class P1MonitorMailer(cfg: P1MonitorConfig) {
 
   private def getInstrumentsString(prop: Proposal): String = prop.observations.map {
       obs => obs.blueprint match {
-        case Some(bp: GeminiBlueprintBase) => s"${bp.instrument.site.name} (${bp.instrument.id})"
+        case Some(bp: GeminiBlueprintBase) => s"${bp.site.name} (${bp.instrument.id})"
         case _                             => ""
       }
     }.distinct.mkString(", ")


bundle/edu.gemini.p1monitor/src/main/scala/edu/gemini/p1monitor/P1MonitorMailer.scala
```diff
@@ -134,7 +134,7 @@ class P1MonitorMailer(cfg: P1MonitorConfig) {
 
   private def getInstrumentsString(prop: Proposal): String = prop.observations.map {
       obs => obs.blueprint match {
-        case Some(bp: GeminiBlueprintBase) => s"${bp.instrument.site.name} (${bp.instrument.id})"
```
swalker2m: What is the difference between a blueprint site and an instrument site?

swalker2m:
Is there an associated JIRA task?

bundle/edu.gemini.p1monitor/src/main/scala/edu/gemini/p1monitor/P1MonitorMailer.scala
cquiroz: This is something we added on the 2016B to differentiate visitor instruments that can be in either site. In all honesty, we should likely remove the site in instrument, but that maybe involved, especially in terms of upconverting old proposals

cquiroz:
It should be part of REL-2463

bundle/edu.gemini.p1monitor/src/main/scala/edu/gemini/p1monitor/P1MonitorMailer.scala
swalker2m: I'm guessing the instrument site is used elsewhere?  Is it not a worry that the two could be inconsistent?  If we need to do this for now to avoid making waves, I understand but we should probably at least add an issue to evaluate this because it seems like we're not modeling the situation correctly.  Perhaps for visitor instruments the site needs to be supplied when an instance is created or something.

cquiroz:
That's a valid point, the site was added as we have some named visitor instruments (Phoenix, Texes, DSSI) that can eventually be made available to either site (even during the same semester I suppose). Most likely we should get rid of the instrument site, which is hardcoded and use the blueprint site instead with sensible defaults.
I've made a task to track this REL-2657

swalker2m:
:+1:

PR #838, opened 2016-02-23
REL-2593: Create Links to ASCII Data for GMOS IFU-2 Pixel Signal and …

spakzad:
…Background

I hope most of these changes are obvious.  This was the last JIRA task that Florian helped me with before he left.


cquiroz:
Looks very good, I'll merge right away

PR #839, opened 2016-02-23
REL-2656: ITC fails with K0III star SED

spakzad:
Minor change made so the ITC works when KOIII star is selected.


cquiroz:
:+1:

PR #840, opened 2016-02-25
REL-2661: Ignore negative parallaxes from Simbad

cquiroz:
Small PR that makes negative parallax values equal to 0, to avoid confusing the TCC. Some warnings were removed too.


swalker2m:
Hmm, strange that SIMBAD would just flat out give erroneous results.  I wonder how we know that the non-negative values are okay.  Anyway, :+1:

PR #841, opened 2016-02-25
OCSADV-514-B: Add Auto Disabled

swalker2m:
This is an addendum to the model refactoring for the `GuideEnvironment` and friends.  In particular it introduces a third kind of `AutomaticGroup`, `Disabled`, joining `Initial` and `Active`.  `Disabled` is used when importing old programs from XML that do not have an automatic group.  This addresses a concern voiced by Andy when the automatic group concept was proposed in Hawaii in December.  Previously old observations would have been set to `Initial` and available for BAGS searches.  Now BAGS will be made to ignore `Disabled` automatic groups.

In the UI, these will show up with a distinctive name like "Auto (Disabled)".  It isn't done yet, but I propose that for unobserved observations you should be able to select the automatic group as primary and have it switch to `Initial`.  Beyond that I hope no other controls are needed to switch back and forth to/from `Disabled` though it wouldn't be too difficult to implement.

autodisabled

I also moved the migration code that adds an automatic group on import from pre-2016B programs into the common migration area.  The caveat being some really old pre-2010B migration that I left in place.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/env/GuideEnvironment.scala
+++ b/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/env/GuideEnvironment.scala
@@ -205,7 +205,8 @@ final case class GuideEnvironment(guideEnv: GuideEnv) extends TargetContainer {
 
 object GuideEnvironment {
   val ParamSetName = "guideEnv"
-  val Initial: GuideEnvironment = GuideEnvironment(GuideEnv.initial)
+  val Initial: GuideEnvironment  = GuideEnvironment(GuideEnv.Initial)
+  val Disabled: GuideEnvironment = GuideEnvironment(GuideEnv.Disabled)
 
   val env: GuideEnvironment @> GuideEnv =
     Lens.lensu((a,b) => a.copy(b), _.guideEnv)
@@ -216,7 +217,9 @@ object GuideEnvironment {
   val manual: GuideEnvironment @> Option[OptsList[ManualGroup]] =
     env >=> GuideEnv.manual
 
-  private def initializeOptions(opts0: List[GuideGroup], primary0: Int): GuideEnvironment = {
+  def create(guideGroups: OptionsList[GuideGroup]): GuideEnvironment = {
+    val opts0    = guideGroups.getOptions.asScalaList
+    val primary0 = guideGroups.getPrimaryIndex.getOrElse(0).intValue
     val (opts, primary) =
       if (opts0.headOption.exists(_.grp.isAutomatic)) (opts0, primary0)
       else (GuideGroup.AutomaticInitial :: opts0, primary0 + 1)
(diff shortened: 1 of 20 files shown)


bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/env/GuideEnvironment.scala
```diff
@@ -216,7 +217,9 @@ object GuideEnvironment {
   val manual: GuideEnvironment @> Option[OptsList[ManualGroup]] =
     env >=> GuideEnv.manual
 
-  private def initializeOptions(opts0: List[GuideGroup], primary0: Int): GuideEnvironment = {
+  def create(guideGroups: OptionsList[GuideGroup]): GuideEnvironment = {
+    val opts0    = guideGroups.getOptions.asScalaList
+    val primary0 = guideGroups.getPrimaryIndex.getOrElse(0).intValue
```
sraaphorst: Just out of curiosity, why is the `.intValue` required here for primary indices?

sraaphorst:
I didn't look over the XML file in any detail, but otherwise, looks good to me.
:+1:

andrewwstephens:
This sounds good.  How do you define an "old" program?  What will happen when you copy an executed observation which has BAGS disabled?

swalker2m:
@andrewwstephens "old" is any program exported from an OT or the ODB before the 2016B upgrade.  Since it didn't occur to me before you mentioned it, at the moment copying will maintain the "Disabled" automatic group. I'm guessing though that ideally copying would reset the automatic group to "Initial" so that BAGS starts working with it?

andrewwstephens:
Yes please reset the automatic group on copy.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/env/GuideGroup.scala
```diff
@@ -276,12 +281,13 @@ object GuideGroup extends ((GuideGrp) => GuideGroup) {
 
   val ParamSetName = "guideGroup"
 
-  private sealed trait TypeTag
-  private case object AutoInitialTag extends TypeTag
-  private case object AutoActiveTag extends TypeTag
-  private case object ManualTag extends TypeTag
+  sealed trait TypeTag
```
cquiroz: Maybe the name `TypeTag` could collide with the one in the reflection api?

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/TelescopePosTableWidget.java
```diff
@@ -204,7 +203,18 @@ default Border border(int col)     {
 
         static final class GroupRow extends AbstractRow {
             private static String extractName(final GuideGroup group) {
-                return group.getName().getOrElse("Auto");
+                final GuideGrp grp = group.grp();
+                final String name;
+                if (grp.isAutomatic()) {
+                    if (grp instanceof AutomaticGroup.Disabled$) {
+                        name = "Auto (Disabled)";
```
cquiroz: How does it know it is grayed-out on the UI?

cquiroz:
Looks very good to me :+1:

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/TelescopePosTableWidget.java
sraaphorst: This is currently done in `EdCompTargetList.setSelectionFromNode`.

The code for this should probably be moved directly into `EdCompTargetList.updateUIForGroup`, but that is a change for another PR.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/env/GuideGroup.scala
swalker2m: I didn't think about that.  It collides in the sense that that they have the same name which could be confusing, but they are in quite different scopes.  It works but I'll change it to `GroupTypeTag` or something.

swalker2m:
In the [2nd commit](https://github.com/swalker2m/ocs/commit/4fa921d489190eaef3c9e4bae7786219aceb7a27) I added the requested feature that "old" observations with `Disabled` automatic groups produce copies that have useable `Initial` automatic groups.  I renamed `GuideGroup.TypeTag` in the last commit.

Let me know if there are any further comments.  Otherwise I'll go ahead and merge this today.

PR #842, opened 2016-02-26
more prep for target model swap

tpolecat:
This PR removes some more references to `ITarget`, trivially other than calls to `getTarget().getSkycalcCoordinates()` which now need a time value.


tpolecat:
Yeah that probably makes sense. I didn't anticipate this may call sites. I'll do that here.

swalker2m:
:+1:

PR #843, opened 2016-02-26
persistent new target

tpolecat:
This makes the new `Target` carried by `SPTarget` persistent and hooks it up to PIO serialization. This is not yet visible in the UI. It will be renamed, etc., in a later PR.


cquiroz:
Looks good :+1:

PR #844, opened 2016-02-29
OCSADV-514: add position angle to active guide group

swalker2m:
This PR tweaks the `AutomaticGroup.Active` group to add a position angle.  Depending on the position angle constraint in the observation context, BAGS will sometimes find a guide star at a position angle other than the current one.  For example, if the constraint is `FIXED_180` BAGS is allowed to search at the current position angle and the current position angle plus 180º.  Previously we were not keeping track of the position angle required to reach the guide stars found by BAGS.  This means that when switching to a manual group and then later back to the automatic guide group, the position angle might no longer correspond.  With this update, we will be able to restore the position angle associated with the BAGS guide star.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/env/GuideGroup.scala
+++ b/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/env/GuideGroup.scala
@@ -1,5 +1,6 @@
 package edu.gemini.spModel.target.env
 
+import edu.gemini.spModel.core.Angle
 import edu.gemini.spModel.target.env.AutomaticGroup.{Active, Disabled, Initial}
 import edu.gemini.spModel.target.env.GuideGroup.AutoDisabledTag
 import edu.gemini.spModel.target.env.TargetCollection._
@@ -48,7 +49,7 @@ case class GuideGroup(grp: GuideGrp) extends java.lang.Iterable[GuideProbeTarget
   def contains(gp: GuideProbe): Boolean =
     grp match {
       case Initial | Disabled => false
-      case Active(ts)         => ts.member(gp)
+      case Active(ts, _)      => ts.member(gp)
       case ManualGroup(_, ts) => ts.member(gp)
     }
 
@@ -57,7 +58,7 @@ case class GuideGroup(grp: GuideGrp) extends java.lang.Iterable[GuideProbeTarget
       case ManualGroup(_, opts) =>
         opts.lookup(gp).map { o => (o.focus, o.toList) }
 
-      case Active(ts)           =>
+      case Active(ts, _)        =>
         ts.lookup(gp).map { t => (some(t), List(t)) }
 
       case Initial | Disabled   =>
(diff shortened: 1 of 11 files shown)


swalker2m:
@andrewwstephens Note there is still a bit of awkwardness here.  For example say the user lets BAGS find a guide star at position angle 0º with constraint `FIXED_180`.  Then she creates a manual group, makes it primary, changes the position angle to 180º and selects a suitable guide star.  Now she switches back to the automatic group as primary.  We can now restore the position angle to 0º so that it corresponds to the auto guide star that was found.  However if she switches back to the manual group, we aren't tracking the fact that those guide stars correspond to a position angle of 180º.

One idea we briefly considered was to keep the position angle in the manual guide groups as well.  That would require moving it from its current home in the instrument, which would be a significant effort.

swalker2m:
I think the test case failure revealed by Travis is unrelated. It compiles and tests clean locally.

````
CoordinatesSpec
be consistent with Dec offsetting (includes polar discontinuity
A counter-example is [Coordinates(RA(Angle(359.9Â°)),Dec(Angle(36.69999999999982Â°))), Angle(180.0Â°)] (after 84 tries)
````

**Comment by tpolecat** [comment]
Yeah sorry about that. I have an open issue … rounding error, as always.

**Comment by tpolecat** [comment]
I restarted the build (i.e., rolled the dice again).

File: bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/env/GuideGroup.scala [hunk 1]
```diff
@@ -236,19 +242,23 @@ case class GuideGroup(grp: GuideGrp) extends java.lang.Iterable[GuideProbeTarget
     getTargets.toList.iterator
 
   def getParamSet(f: PioFactory): ParamSet = {
-    import GuideGroup.{ParamSetName, AutoInitialTag, AutoActiveTag, ManualTag }
+    import GuideGroup.{grp => _, _}
```
cquiroz: This is new to me, what does it mean?

tpolecat: It means don't import `grp` but import everything else. Kind of weird.

swalker2m: In this case, I did it because the `GuideGroup.grp` lens would otherwise clash with the `grp` field of class `GuideGroup`.

cquiroz: Nice to know, I can imagine it being useful is some situations

cquiroz:
Looks good :+1:

tpolecat:
> One idea we briefly considered was to keep the position angle in the manual guide groups as well. That would require moving it from its current home in the instrument, which would be a significant effort.

ugh.

:+1:

andrewwstephens:
@swalker:  Thanks for pointing this out.  This sounds reasonable.  I'm having trouble reproducing your scenario because AGS forcefully updates the PA even when a Manual guide group is selected.  :confused:

swalker2m:
@andrewwstephens Thanks for pointing out that bug.  I've updated the code (in this PR) so that it will update the position angle tracked in the automatic guide group always but only update the instrument position angle if the automatic group is primary.  Unfortunately though we'll have to wait until @sraaphorst is back to see this change in action in the UI.

andrewwstephens:
@swalker2m @sraaphorst : Is it possible to use focus to know when the user is done editing a field?

swalker2m:
@andrewwstephens It is but it gets tricky. I believe there are some cases when a focus lost event is simply not generated.  For example, when you are typing in the pos angle field and then click on another OT node.  Sebastian went through this before though with the RA/Dec fields.

PR #845, opened 2016-02-29
OCSADV-139-B: cleanup

swalker2m:
Just a bit of cleanup and deprecation warning removal related to the scalaz update from a few weeks ago.  Most significantly, I switched `fromTryCatch` to `fromTryCatchNonFatal` which is what I believe we almost always want.


cquiroz:
:+1:

PR #846, opened 2016-02-29
trying out sbt -q

tpolecat:
This may or may not confuse Travis. We shall see.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,4 +29,4 @@ before_script:
   - if [ $TRAVIS_SECURE_ENV_VARS == false ]; then
       cp $TRAVIS_BUILD_DIR/project/OcsCredentials.scala.template $TRAVIS_BUILD_DIR/project/OcsCredentials.scala;
     fi
-script: sbt -q $TRAVIS_SCALA_VERSION test:compile test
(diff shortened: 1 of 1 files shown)


cquiroz:
What does the `-q` do? I couldn't find sbt command line options documentation

tpolecat:
@cquiroz I recommend using [sbt-extras](https://github.com/paulp/sbt-extras) which has a useful `-help` menu and other things. This is what Travis provides.

swalker2m:
D'oh:

Timeout (20 minutes) reached. Terminating "sbt -q test:compile test"
The command sbt -q test:compile test exited with 137.

cquiroz:
I think that is because travis expects some output
if it doesn't get anything in some timeout it thinks the process is dead. Maybe a warning level would be better

tpolecat:
Trying again with a 40 minute timeout.

swalker2m:
Interesting though that when it did fail with the timeout, it seemed to dump the entire log anyway?  If that happens on a test case failure then `-q` won't be that helpful.

tpolecat:
Yeah that's puzzling. They may not work together.

PR #847, opened 2016-03-08
REL-2670: Allow selecting FT PI Affiliation without PI address

cquiroz:
Minor PR that removes a restriction where to set the PI Partner Affiliation you need to set the institution. With this change the affiliation can be set independently


swalker2m:
The deletion certainly looks fine.  I'm not sure I understand fully why not having an institution would lead to removing the affiliation selection as it did before, but :+1:

PR #848, opened 2016-03-14
REL-2667: Add Japan to the PIT FT PI Affiliation menu

cquiroz:
This PR sets the stage for a new release of the PIT. Subaru users can now apply to FT proposal and thus the model and UI need to support "Japan" as a PI Affiliation country.
This PR contains model updates, UI changes, upconversion logic and fixes for some warnings

### DIFF
--- a/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Partners.scala
+++ b/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Partners.scala
@@ -24,17 +24,21 @@ object Partners {
   // REL-2248 Contains a list of partners that are not allowed on joint proposals
   val jointProposalNotAllowed = List[NgoPartner](NgoPartner.KR, NgoPartner.AU)
 
-  // REL-2670 A partner for FT can be either Ngo or Exchange
+  // REL-2670 A partner affiliation for an FT proposal can be either Ngo or Exchange (Subaru)
   type FtPartner = Option[NgoPartner \/ ExchangePartner]
 
   val NoPartnerAffiliation = "None"
 
   private val SubaruAffiliation = "Japan"
 
+  // Possible FT Partners: None, Ngos, Subaru
   val ftPartners:Seq[(FtPartner, String)] = {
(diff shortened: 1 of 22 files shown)


bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Partners.scala
```diff
@@ -22,12 +24,30 @@ object Partners {
...
+    case Some(-\/(p))                      => Partners.name.get(p)
+    case Some(\/-(ExchangePartner.SUBARU)) => Some(SubaruAffiliation)
+    case _                                 => None
   }
 
-  def toPartner(name: String): Option[NgoPartner] = Partners.name.find(_._2 == name).collect {
-    case (p: NgoPartner, _) => p
+  def toPartner(name: String): FtPartner = ftPartners.find(_._2 == name).collect {
+    case (Some(-\/(p)), _) => -\/(p)
+    case (Some(\/-(e)), _) => \/-(e)
```
swalker2m: It looks as if you could collapse the two cases?

````scala
case (Some(ftPartner), _) => ftPartner
````

File: bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Partners.scala [hunk 2]
```diff
@@ -22,12 +24,30 @@ object Partners {
   // REL-2248 Contains a list of partners that are not allowed on joint proposals
   val jointProposalNotAllowed = List[NgoPartner](NgoPartner.KR, NgoPartner.AU)
 
-  val ftPartners:Seq[(Option[NgoPartner], String)] = {
-    (None -> "None") :: NgoPartner.values.toList.map(p => Some(p) -> Partners.name.getOrElse(p, ""))
+  // REL-2670 A partner affiliation for an FT proposal can be either Ngo or Exchange (Subaru)
+  type FtPartner = Option[NgoPartner \/ ExchangePartner]
```
swalker2m: Keck and CFHT are also an exchange partners and yet not an FT partner.  Perhaps we should make a new `FtExchangePartner` type that includes only Subaru?

cquiroz:
Updated the code with suggestions form comments

swalker2m:
:+1:

PR #849, opened 2016-03-15
16b Target Model Converters

tpolecat:
This adds code to the `To2016B` program converter to generate new-model targets from old ones and add a zero-length scheduling block to nonsidereal observations, corresponding with the single-element ephemeris generated from the existing fixed coordinates and `validAt` time. Right now the deserialized values contain *both* old and new targets; the old target will be removed once it is unused.

### DIFF
--- a/bundle/edu.gemini.spModel.io/src/main/scala/edu/gemini/spModel/io/impl/migration/to2016B/To2016B.scala
+++ b/bundle/edu.gemini.spModel.io/src/main/scala/edu/gemini/spModel/io/impl/migration/to2016B/To2016B.scala
@@ -2,14 +2,11 @@ package edu.gemini.spModel.io.impl.migration
 package to2016B
 
 import edu.gemini.pot.sp.SPComponentType
-import edu.gemini.spModel.core.Semester
-import edu.gemini.spModel.gemini.obscomp.SPProgram
-import edu.gemini.spModel.io.impl.migration.Migration
-import edu.gemini.spModel.io.impl.migration.PioSyntax._
-import edu.gemini.spModel.obs.SPObservation
+import edu.gemini.spModel.core._
+
 import edu.gemini.spModel.pio.xml.PioXmlFactory
 import edu.gemini.spModel.pio.{Pio, ParamSet, Document, Version}
-import edu.gemini.spModel.target.SPTargetPio
+import edu.gemini.spModel.target.{TargetParamSetCodecs, SPTargetPio, SourcePio}
 import edu.gemini.spModel.target.env.GuideGroup
 
 import scala.collection.JavaConverters._
@@ -19,7 +16,15 @@ import scalaz._
 import Scalaz._
 
 object To2016B extends Migration {
+
   val version = Version.`match`("2016B-1")
+  val R0 = RightAscension.zero
+  val D0 = Declination.zero
+
+  implicit class MoreLensOps[A, B](lens: A @> B) {
+    def ?:=(op: Option[B]): State[A, B] =
+      op.fold(lens.mods(identity))(lens.assign(_))
+  }
 
   val conversions: List[Document => Unit] = List(
     updateGuideEnvironment, updateSchedulingBlocks, updateTargets // order matters!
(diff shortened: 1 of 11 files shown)


cquiroz:
The error on travis is the same we observed on my previous PR #848. I marked the broken tests as pending

bundle/edu.gemini.spModel.io/src/main/scala/edu/gemini/spModel/io/impl/migration/to2016B/To2016B.scala
```diff
@@ -68,4 +81,116 @@ object To2016B extends Migration {
...
+      // Drop it into the paramset.
+      // TODO: remove everything else!
+      t.addParamSet(TargetParamSetCodecs.TargetParamSetCodec.encode("target", newTarget))
+
+    }
+
+  // Properties common to all target types
+  def common(ps: ParamSet): State[Target, Unit] =
+    for {
+      _ <- Target.name                ?:= ps.value("name")
```
cquiroz: Can you explain what the `?:=` and `:=` operators do?

bundle/edu.gemini.spModel.io/src/test/scala/edu/gemini/spModel/io/impl/migration/MigrationTest.scala
```diff
@@ -21,11 +20,46 @@ trait MigrationTest {
...
+      val parser = new PioSpXmlParser(db.getFactory)
+      val files  = Option(dir.listFiles).getOrElse(Array.empty[File]).filter(_.getName.endsWith(".xml")).sortBy(_.getName)
+      val total  = files.length
+      files.zipWithIndex.foreach { case (f, n) =>
+        print(s"[$n/$total] ${f.getName} ... ")
+        try {
+          parser.parseDocument(new FileReader(f))
+          println(" [ok]")
+        } catch {
+          case e: Exception => println(s" ${Console.RED}[err] ${e.getMessage}${Console.RESET}")
```
cquiroz: Is the `Console.RED` used to change color in the output? ANSI codes?

tpolecat: Yeah that's all it does. It's just ANSI escapes.

bundle/edu.gemini.spModel.io/src/main/scala/edu/gemini/spModel/io/impl/migration/to2016B/To2016B.scala
tpolecat: Yeah `:=` turns a lens into a state action:  `Lens[A,B] := B` gives you a `State[A, B]`. So this lets you use imperative-looking code to write a pure functional series of lens transformations. So I compose all the "assignments" into a single `State` program and then use `exec` to apply it to an `A`.

`?:=` is defined up at the top of the file. It does the same thing but for `Lens[A,B] := Option[B]` that is a no-op if the value is empty.

bundle/edu.gemini.spModel.io/src/main/scala/edu/gemini/spModel/io/impl/migration/to2016B/To2016B.scala
```diff
@@ -68,4 +81,116 @@ object To2016B extends Migration {
...
+        for {
+          system <- t.value("system")
+          ra     <- t.value("c1").flatMap(s => (Angle.parseHMS(s) orElse Angle.parseDegrees(s)).toOption).map(RightAscension.fromAngle)
+          dec    <- t.value("c2").flatMap(s => (Angle.parseDMS(s) orElse Angle.parseDegrees(s)).toOption).flatMap(Declination.fromAngle)
+        } yield (system, Coordinates(ra, dec))
+
+      // Construct a new target
+      val newTarget: Target =
+        data.map {
+          case ("J2000", Coordinates.zero) => TooTarget.empty
```
swalker2m: By default a new target in a non-ToO observation is also Sidereal (0,0).  To be sure, you would need to also check the ToO type of the program.

tpolecat: Ah that's right. Thanks.

cquiroz:
If you rebase on develop it should pass travis checks

tpolecat:
@cquiroz yep, will do.

tpolecat:
@swalker2m ok I fixed that TOO issue and added a test for it.

swalker2m:
Looks good.  :+1:

PR #850, opened 2016-03-18
REL-2667-B: Add Japan to the PIT FT PI Affiliation menu

cquiroz:
This PR creates an extra version of p1-monitor to support Subaru as an FT partner


swalker2m:
:+1:

PR #851, opened 2016-03-21
switch to new target model, in progress

tpolecat:
This is a work in progress but is already too big to view because GitHub insists on printing every line of a deleted file. The actual new code is minimal. To see all changes you can go to this [diff](https://github.com/gemini-hlsw/ocs/pull/851.diff) and open it in a text editor that knows how to do highlighting. Sorry, there's not much else we can do.

This PR removes the old target model. `SPTarget.getTarget` now returns a new model `Target`.
- `ITarget` and all of its implementations are gone.
- `SkyObject` and `skyobject.Magnitude` are gone.
- Most functionality in `ModelConverters` is gone.

This will require (sadly) close review and a lot of testing. I will update this PR with more info when it's ready for review.

Tests to fix:

- [x] bundle_edu_gemini_spModel_io/test
- [x] bundle_edu_gemini_wdba_xmlrpc_server/test
- [x] bundle_edu_gemini_oodb_too_url/test
- [x] bundle_edu_gemini_sp_vcs_log/test
- [x] bundle_edu_gemini_phase2_skeleton_servlet/test
- [x] bundle_edu_gemini_pot/test
- [x] bundle_edu_gemini_p2checker/test

In any case :-1: this is in progress, don't merge.


bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/ConfigExtractor.scala
```diff
@@ -236,12 +234,12 @@ object ConfigExtractor {
       targets.getPrimary.asScalaOpt.fold("No guide star selected".left[SPTarget])(_.right)
 
     def extractMagnitude(guideStar: SPTarget) = {
-      val r  = guideStar.getMagnitude(Band.r)
-      val R  = guideStar.getMagnitude(Band.R)
-      val UC = guideStar.getMagnitude(Band.UC)
-      if      (r.isDefined)  r.getValue.getBrightness.right
-      else if (R.isDefined)  R.getValue.getBrightness.right
-      else if (UC.isDefined) UC.getValue.getBrightness.right
+      val r  = guideStar.getNewMagnitude(MagnitudeBand._r)
```
cquiroz: I believe there is already code doing this. class `RBandList` perhaps

bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/general/GeneralRule.java
```diff
@@ -153,162 +150,162 @@ public IP2Problems check(ObservationElements elements)  {
...
-                    }
-
-                }
-                for (final String error : errorSet) {
-                    final String id = error.equals(WFS_EMPTY_NAME_TEMPLATE) ? "WFS_EMPTY_NAME_TEMPLATE" :
-                              (error.equals(NAME_CLASH_MESSAGE) ? "NAME_CLASH_MESSAGE" : "COORD_CLASH_MESSAGE");
-                    problems.addError(PREFIX+id, error, elements.getTargetObsComponentNode().getValue());
-                }
-            }
+//            final boolean hasAltairComp = elements.hasAltair();
```
cquiroz: Will this commented code be ported too? Or isn't it needed anymore?

bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/gpi/GpiRule.java
```diff
@@ -165,14 +166,14 @@ public IP2Problems check(ObservationElements elements)  {
                 P2Problems problems = new P2Problems();
                 TargetEnvironment env = obsComp.getTargetEnvironment();
                 SPTarget base = env.getBase();
-                Option<Magnitude> imag = base.getMagnitude(Magnitude.Band.I);
-                Option<Magnitude> hmag = base.getMagnitude(Magnitude.Band.H);
+                scala.Option<Magnitude> imag = base.getNewMagnitude(MagnitudeBand.I$.MODULE$);
```
cquiroz: Great improvement over the magic `UNDEFINED_MAG` constant

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/SPTarget.scala
```diff
@@ -0,0 +1,205 @@
...
+  def getDecString(time: GOLong): GOption[String] =
+    gcoords(time)(_.dec.toAngle.formatDMS)
+
+  def getSkycalcCoordinates(time: GOLong): GOption[SCoordinates] =
+    gcoords(time)(cs => new SCoordinates(cs.ra.toDegrees, cs.dec.toDegrees))
+
+  def getSiderealTarget: Option[SiderealTarget] =
+    target.fold(_ => None, Some(_), _ => None)
+
+  def putNewMagnitude(mag: Magnitude): Unit =
```
cquiroz: Why are these methods `NewMagnitude`? Where are the old ones?

cquiroz:
As a note Sublime 3 can open the diff and show coloring

cquiroz:
I skimmed the code, This looks like a huge change with so many deleted classes. I think in general looks great. One question is why there is so much deleted UI code. Don't we need to port those?

I guess you noted the compilation error in travis

bundle/edu.gemini.ags.servlet/src/main/java/edu/gemini/ags/servlet/ToContext.java
```diff
@@ -151,9 +151,10 @@ private TargetEnvironment targetEnv(HttpServletRequest req) throws RequestExcept
             // Here we just need an SPTarget with a contained NonSiderealTarget. In the end, this
             // is just used to determine whether it is a non-sidereal target or not in order to
             // pick the right guider.
+            // TODO: track down where this happens!
```
swalker2m: I think it is probably just AGS.  It wants to know which strategy should be the default.  If it is a non-sidereal target usually a PWFS guide probe is used whereas sidereal targets use an OIWFS by default.

See [this code](https://github.com/gemini-hlsw/ocs/blob/develop/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/Strategy.scala#L65-L75) for example.

swalker2m:
> I skimmed the code, This looks like a huge change with so many deleted classes. I think in general looks great. One question is why there is so much deleted UI code. Don't we need to port those?

@cquiroz This could be just the editors for the old target model.  Rob merged in the new editors in a previous PR but they weren't being used yet.  Probably here we will make the switch over so the old ones are no longer needed.

bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/general/GeneralRule.java
tpolecat: Yes, it will be ported, it's not done yet.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/SPTarget.scala
tpolecat: They will be renamed now; for a while there were methods using  both new and old magnitudes.

tpolecat:
Right, the old target editors were deleted.

tpolecat:
Thanks for taking a look; as you no doubt noticed, it's not quite done.

tpolecat:
Ok, tests pass (at least locally). Odds and ends:

- [ ] Rename `getNewXXX` methods on `SPTarget`

And then I will review the whole diff closely.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/SingleProbeStrategy.scala
```diff
@@ -255,11 +254,10 @@ object SingleProbeStrategy {
...
-    val ra1    = st.coordinates.ra.toAngle.toRadians
-    val dec1   = st.coordinates.dec.toAngle.toRadians
-    val target = HmsDegTarget.fromSkyObject(st.toOldModel)
-    val ra2    = Angle.fromDegrees(target.getRa.getAs(Units.DEGREES)).toRadians
-    val dec2   = Angle.fromDegrees(target.getDec.getAs(Units.DEGREES)).toRadians
+    val ra1    = base.ra.toAngle.toRadians
+    val dec1   = base.dec.toAngle.toRadians
+    val ra2    = st.coordinates.ra.toAngle.toRadians
+    val dec2   = st.coordinates.dec.toAngle.toRadians
     val raDiff = ra2 - ra1
```
tpolecat: I fixed an apparent bug here (introduced in c7268d2) ... the `base` coordinates were never used and the method always returned zero. This was not caught by any test and I don't know what the impact actually is. Any ideas?

swalker2m: @tpolecat this is currently unused.  It applies when the `PosAngleConstraint` is `UNBOUNDED`, which means AGS is allowed to select any position angle required to reach the highest quality guide star.  Unbounded searches proved problematic, though I can't remember what the issue was.  Maybe someone can refresh my memory?  At any rate, it is a potential task for December to enable this feature so it is good that you caught it.  We should definitely add test cases as part of turning it on.

cquiroz: Sorry about the translation error. I vaguely remember changing this as part of the process of using the new target model

bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/SpTargetFactory.scala
```diff
@@ -1,92 +1,49 @@
...
-    sp.setName(too.name)
-    sp
+  val dra  = core.ProperMotion.deltaRA  >=> core.RightAscensionAngularVelocity.velocity >=> core.AngularVelocity.masPerYear
+  val ddec = core.ProperMotion.deltaDec >=> core.DeclinationAngularVelocity.velocity    >=> core.AngularVelocity.masPerYear
+
+  def create(t: Target, time: Long): Either[String, SPTarget] = {
+    val ct: core.Target =
+      t match {
+        case too: TooTarget         => core.TooTarget(too.name)
+        case nsd: NonSiderealTarget => initNonSidereal(nsd, time).exec(core.NonSiderealTarget.empty)
```
cquiroz: What does `exec` do? Does it execute the `State`?

bundle/edu.gemini.p2checker/src/main/scala/edu/gemini/p2checker/rules/general/GeneralRuleHelper.scala
```diff
@@ -0,0 +1,29 @@
...
+import edu.gemini.spModel.core.{NonSiderealTarget, TooTarget, SiderealTarget, Target}
+
+import scalaz._, Scalaz._
+
+class GeneralRuleHelper {
+
+  def samePosition(a: Target, b: Target): Boolean =
+    (a, b) match {
+
+      // Sidereal targets must have identical coordinates, proper motion, redshift, and parallax
```
cquiroz: Should this comparison be tolerant to precision errors?

tpolecat: Unclear, probably. The old implementation compared strings values for coordinates but exact valued for everything else. I'll change it to do the same thing. :chicken:

bundle/edu.gemini.ags.servlet/src/main/java/edu/gemini/ags/servlet/ToContext.java
tpolecat: Ok, I added a scheduling window and 1-element ephemeris to the non-sidereal target to be sure the coordinates are available. I think it's ok.

tpolecat:
Ok. There are a lot of odds and ends to take care of but I'd like to go ahead and merge this before it gets any more out of sync with `develop`. Any final words?

cquiroz:
Looks fine to me :+1:

PR #854, opened 2016-03-29
ephemeris display improvements

tpolecat:
Some minor improvements in the display of ephemeris information in the target editor.

- shows ephemeris resolution in minutes
- shows time + position a bteginning of scheduling block (if any)
- shows current position, and updates in real time

image

There is still work to do but this is an improvement.

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details2/EphemerisEditor.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details2/EphemerisEditor.scala
@@ -1,47 +1,164 @@
 package jsky.app.ot.gemini.editor.targetComponent.details2
 
-import java.awt.Color
+import java.awt.{Insets, GridBagConstraints, GridBagLayout, Color}
 import java.text.SimpleDateFormat
-import java.util.{ Date, TimeZone }
-import javax.swing.JLabel
+import java.util.{ TimerTask, Timer, Date, TimeZone }
+import javax.swing.{SwingConstants, BorderFactory, JPanel, JLabel}
 
 import edu.gemini.pot.sp.ISPNode
 import edu.gemini.shared.util.immutable.{ Option => GOption }
-import edu.gemini.spModel.core.{Declination, Angle, RightAscension, SiderealTarget, Coordinates, Target}
+import edu.gemini.spModel.core.{Ephemeris, NonSiderealTarget, Target}
 import edu.gemini.spModel.obs.context.ObsContext
 import edu.gemini.spModel.target.SPTarget
 import jsky.app.ot.gemini.editor.targetComponent.TelescopePosEditor
+import edu.gemini.shared.util.immutable.ScalaConverters._
 
 import scalaz._, Scalaz._
 
 class EphemerisEditor extends TelescopePosEditor with ReentrancyHack {
 
+  private[this] var target: Option[NonSiderealTarget] = None
+
+  // TODO: stop the timer when the window closes, or something?
+  private val timer = new Timer
+  timer.scheduleAtFixedRate(new TimerTask {
+    def run(): Unit =
+      target.flatMap(_.coords(System.currentTimeMillis)) match {
+        case Some(cs) => now.setText(cs.ra.toAngle.formatHMS + " " + cs.dec.formatDMS)
+        case None     => now.setText("--")
+      }
+    }, 1000L, 50L) // unjustifiably fast but it looks cool
+
+  val start, end, size, now, sched, schedCoords  = new JLabel <| { l =>
+    l.setForeground(Color.DARK_GRAY)
+    l.setMinimumSize(l.getMinimumSize <| (_.width = 200))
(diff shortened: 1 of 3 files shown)


tpolecat:
Hm where is Travis?

tpolecat:
Ok close and then re-open fixed it. Confidence high!

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details2/EphemerisEditor.scala
```diff
@@ -1,47 +1,164 @@
...
 
 class EphemerisEditor extends TelescopePosEditor with ReentrancyHack {
 
-  val df = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss z") <| { df =>
-    df.setTimeZone(TimeZone.getTimeZone("UTC"))
-  }
+  private[this] var target: Option[NonSiderealTarget] = None
 
-  val start, end, size = new JLabel <| { l =>
+  // TODO: stop the timer when the window closes, or something?
```
cquiroz: Could this stop the JVM from terminating? If the timer is still running

cquiroz:
I wonder if this should be replicated on the PIT to some extent? Or even better, reused

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details2/EphemerisEditor.scala
tpolecat: We call `System.exit()` so no, but I should at least create it as a daemon. I was trying to figure how to stop it when the panel is hidden but I was never able to see the relevant events so I gave up (`ComponentListener` never seemed to do anything). Any ideas?

tpolecat:
Yeah, and I need to pull the custom ephemeris file thing from the PIT. We're very close to the models being identical so I think we should be able to make the editors shareable soon.

tpolecat:
Here is a [movie](https://www.dropbox.com/s/vjzcp5mesr06u8k/moon.mov?dl=0). Kind of cool.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details2/EphemerisEditor.scala
swalker2m: Also, the `target` variable will be accessed from multiple threads.  Should it be an `AtomicReference`?

tpolecat: What is the current thinking on that? Back in the olden days I always assumed reference updates were atomic. Did the 64-bit JVM change that?

swalker2m: I think there shouldn't be an issue with a partial write like you would see with updating a Double or Long, but my understanding was that to guarantee that changes made by one thread are seen by another you have to mark the reference `@volatile` or else synchronize it.

cquiroz:
Cool, Can you put the movie directly into GH or do you need to go the dropbox route?

cquiroz:
Cool, Can you put the movie directly into GH or do you need to go the dropbox route?

tpolecat:
No, it won't let you drop movies into comments.

cquiroz:
Looks good :+1:

arturog8m:
Now they will want TPE to show an animation with the trajectory of the object....

andrewwstephens:
I'm not ruling out an animation :smile: but it seems like now we should be able to display the current and future positions without having to click a "Plot" button.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details2/EphemerisEditor.scala
```diff
@@ -1,47 +1,164 @@
...
+  private val timer = new Timer(true)
+  timer.scheduleAtFixedRate(new TimerTask {
+    def run(): Unit =
+      target.flatMap(_.coords(System.currentTimeMillis)) match {
+        case Some(cs) => now.setText(cs.ra.toAngle.formatHMS + " " + cs.dec.formatDMS)
+        case None     => now.setText("--")
```
swalker2m: We should put the `now.setText` in the event loop with `Swing.onEDT` too.

tpolecat: Good catch. I really need to stop trying to write UI code.

tpolecat:
Ok I made the `target` field volatile, moved the text field update on to the EDT, and turned down the refresh speed from 50ms to 1000ms … it looks cool when it goes that fast but there's no reason to burn CPU on it.

tpolecat:
Also made some of the flatmappy stuff easier to follow.

swalker2m:
Looks great. :+1:

PR #857, opened 2016-03-30
REL-2695: change default proprietary period

swalker2m:
A trivial update to the default proprietary period for C, DD, LP, and Q programs.


tpolecat:
:+1:

PR #859, opened 2016-04-04
OCSADV-519: BAGS GUI changes and BagsManager reintroduction

sraaphorst:
This PR finalizes the bulk of the tasks remaining to move BAGS back into the OT and additionally comprises UI tasks / fixes the remaining UI issues that were arising in the way that BAGS worked.

Note that due to myriad issues trying to rebase / merge these changes with the repo, a new branch was created and my commits were conglomerated and remade in this new branch; thus, there is no accurate commit history available for this PR.

With the inclusion of BAGS, changes to things such as the base position would trigger BAGS lookups, which could make changes to the target environment if the automatic group changed, which would cause the table model to rebuild and the target editors to be reset. Pre-BAGS, when AGS needed to be triggered manually via the Auto GS button(s), this was not problematic, since this happened at predictable times. Now, with the inclusion of BAGS, this can happen behind the scenes at any time and causes several UI issues.

This behind-the-scenes AGS lookup was problematic in particular because if a new auto group was found, a new target environment would be set for the observation, which contained copies of all targets, and thus the active editor needed to be reset to use the new target copy. This would cause the issues that I have brought up multiple times where the RA and Dec fields were reformatted / repopulated, which made editing them difficult as the contents would be reset and the caret pushed to the end of the text fields.

In order to overcome this, several changes were made:

1. If the auto group is selected, the position angle is now set to the one specified for the auto group. Additionally, we now want the guiding quality of the auto group to be evaluated with regards to its PA, even if the PA is different from the currently set PA. (This can be seen in `TelescopePosTableWidget.TelescopePosTableModel`, where guide quality of targets in a group are calculated with respect to the group, i.e. using the PA of the auto group for auto targets, and using the currently set PA for manual targets.)

This may be slightly confusing / counterintuitive, and ultimately, it seems that each guide group should have an associated PA, but since the PA is currently associated with the instrument instead of the guide groups, the UI and data model changes required to do this were deemed to be too complicated for the current release of BAGS.

2. `OtItemEditor` now has an additional `init` method that takes the old node and data object being managed, if any existed. This is to allow the `EdCompTargetList` to receive the old `ISPObsComponent` and `TargetObsComponent`, which can be used to determine if the only change made is an auto group update.

3. `EdCompTargetList` now behaves differently if the only changes made are to the auto group. When this is the case, instead of simply accepting the new `TargetObsComp` blindly, we attempt to keep as much of the old `TargetEnvironment` as possible (i.e. everything but the new auto group) to avoid using cloned targets and thus allowing us to refrain from having to reset the target editors.

4. `TelescopePosTableWidget`: I renamed to table model to be more clear. In the past, when any change was made to the target environment (including even selecting a new guide group or guide star), the entire table model was rebuilt. This caused the table selection to be reset, which triggered a cascade of events that led to the editors being reset. As previously mentioned, this only happened in a predictable fashion, so it was not an issue; however, with BAGS, it caused selection events to be trigger in `EdCompTargetList`.

Now, instead, the table model has been made (blah) mutable, and when BAGS updates occur, instead of rebuilding the entire model, we simply make changes to the model itself and then fire the appropriate table model insertion, deletion, and update events to tell listeners what has changed. This allows us to get away with not having to reset the table row selection and avoid that chain of events previously mentioned that was the issue.

Note that since BAGS can also set the position angle to the PA required by the auto group if it is active, this table modification also requires a re-evaluation of all manual guide groups to ascertain if the guide target AGS quality evaluations have changed (e.g. they may have gone from being in range of the patrol field to out of range of the patrol field).

5. `BagsManager` has been reintroduced, and hooked up to `SPViewer` to properly register / deregister observations for BAGS processing.

6. The "Auto GS" buttons and logic have been removed as they will no longer be used. (Part of this is removing the `AgsClient`.)

7. I changed a couple variable and class names for added clarity, and IDEA was very insistent on fixing some indentation, so some of the changes you see here will trivially comprise those.

The remaining task to be done is for BAGS to provide feedback regarding the lookups pending or in progress for an observation. This will not be included in the initial test release and is forthcoming.


sraaphorst:
Sorry about the weird list numbering. I suppose I don't know how to do multi-paragraph enumerated list entries with Markdown!

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
```diff
@@ -276,20 +274,93 @@ private void manageCurPosIfEnvContainsTarget(final SPTarget target, final Runnab
...
+        }
+        if (!oldNodeKey.equals(newNodeKey)) {
+            return false;
+        }
+
+        // Extract the target environments and check if the auto groups have changed.
+        final TargetEnvironment oldEnv    = oldTOC.getTargetEnvironment();
+        final TargetEnvironment newEnv    = newTOC.getTargetEnvironment();
+        final Option<GuideGroup> oldGpOpt = oldEnv.getGroups().headOption().filter(GuideGroup::isAutomatic);
+        final Option<GuideGroup> newGpOpt = newEnv.getGroups().headOption().filter(GuideGroup::isAutomatic);
```
swalker2m: We should probably add a `getAutomaticGroup` to `GuideEnvironment` so that you don't have to do this.  `GuideEnv` keeps a reference to the automatic group that can just be returned.  Something like:

````scala
def getAutomaticGroup: GuideGroup = {
   GuideGroup(guideEnv.auto)
}
````

Comment (sraaphorst) [inline bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java]: That would be useful. I'll do it so that we can avoid the filter.

File: bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java [hunk 2]
```diff
@@ -276,20 +274,93 @@ private void manageCurPosIfEnvContainsTarget(final SPTarget target, final Runnab
...
+
+        // We only want to return false if the two groups exist and contain the same targets.
+        return oldGpOpt.forall(oldGp ->
+                newGpOpt.forall(newGp -> {
+                    // Extract the target names.
+                    final List<String> oldNames = oldGp.getTargets().map(SPTarget::getName).toList();
+                    final List<String> newNames = newGp.getTargets().map(SPTarget::getName).toList();
+                    return !(oldNames.size() == newNames.size() && oldNames.containsAll(newNames));
```
swalker2m: The underlying `Target` behind an `SPTarget` is now implemented by immutable `case class`es so `equals` and `hashCode` should work correctly.  We could map to lists of `Target` and then convert them to sets and check for equality, I think?

We should probably check that the associated guide probes haven't changed as well?  In theory the targets could be the same but associated with different guide probes.

sraaphorst: I guess we could convert, for an auto group, to `Map[GuideProbe,Set[Target]]` and compare that?
Although since this is only necessary for auto groups, I think the `Set[Target]` would only ever have size 1 precisely, so we could probably just do `Map[GuideProbe,Target]` and compare that directly for equality.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/TelescopePosTableWidget.java
```diff
@@ -56,11 +58,11 @@
 
     // Note -- synchronized though I guess this should always be called from
     // gui thread...
-    public static synchronized String formatBrightness(Magnitude mag) {
+    private static synchronized String formatBrightness(Magnitude mag) {
```
cquiroz: Why should it be called from the the GUI thread? It looks like no swing calls are done here, are they?

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
sraaphorst: That's Scala, though, and this is Java, so I guess we may need to check that the maps are the same size and then iterate over the entries for one and check that it is in the other.

swalker2m: Yes, either way sounds good but I'm guessing the `Map[GuideProbe, Set[Target]]` is more natural since the `GuideProbeTargets` API you'll have to work with assumes multiple targets are possible.  As you say, in this case the set will always be a singleton set but that's fine.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
```diff
@@ -276,20 +274,93 @@ private void manageCurPosIfEnvContainsTarget(final SPTarget target, final Runnab
...
+
+            final int primaryIdx               = oldGuideEnv.getPrimaryIndex();
+            final Option<GuideGroup> autoGroup = newTOC.getTargetEnvironment().getGroups().find(GuideGroup::isAutomatic);
+
+            final TargetEnvironment newEnv = autoGroup.map(a -> {
+                final ImList<GuideGroup> newGroups = oldGuideEnv.getOptions().filter(GuideGroup::isManual).cons(a);
+                final GuideEnvironment newGuideEnv = oldGuideEnv.setOptions(newGroups).setPrimaryIndex(primaryIdx);
+                return oldEnv.setGuideEnvironment(newGuideEnv);
+            }).getOrElse(oldEnv);
+            newTOC.setTargetEnvironment(newEnv);
```
swalker2m: A fundamental assumption is that there is always an automatic group and that it always be the first group.  This is enforced in the underlying Scala model, though the old Java API doesn't enforce it.  At any rate, I think it'd be okay to just use `GuideEnvironment.setGroup(0, autoGroup)`.  In fact, maybe we could add a `setAutoGroup` to make it more explicit.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/TelescopePosTableWidget.java
sraaphorst: Not to my knowledge, but I didn't change this... I think Rob may have written it this way, so I didn't question it 😆 . I can look into it further.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
swalker2m: You should be able to compare two `java.util.Map` instances for equality directly with `equals`?

sraaphorst: I'm thinking that maybe it makes sense to add a `toMap` and `toJavaMap` to `GuideGroup` to compare, or even more, a method directly in `GuideGroup` that compares the group to another.

I don't think it's as simple as doing an equals op and that we will actually have to do some extraction since `==>>` is a sealed abstract class.

sraaphorst: Really? Huh. I didn't know that `java.util.Map` would correctly compare the contents of two maps.
(I have learned to generally expect little from Java 😄 .)

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/TelescopePosTableWidget.java
swalker2m: The `TelescopePosTableWidget` is exclusively a GUI concept, built and updated by the UI in the swing thread.  This method is synchronized because the shared `static` `DecimalFormat` is not thread safe.  It shouldn't matter if it is only used by the event loop thread but, why not synchronize it anyway I suppose.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/tpe/AgsClient.scala
```diff
@@ -1,249 +0,0 @@
-package jsky.app.ot.tpe
```
cquiroz: :+1: :+1:

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/TelescopePosTableWidget.java
```diff
@@ -350,9 +311,172 @@ private static String extractName(final GuideGroup group) {
...
+            for (final Tuple2<Integer, Row> tup2 : indexedRows.collect(Collectors.toList())) {
+                final int rowIdx = tup2._1();
+                final Row row    = tup2._2();
+
+                if (changedRowIndices.contains(rowIdx)) {
+                    // We only care about the child rows (hence + 1), and all rows have been offset by autoGroupSizeDiff.
+                    final int lowerIdx = tableRowIdx + 1 + autoGroupSizeDiff;
+                    final int upperIdx = lowerIdx + row.children().size() - 1;
+                    fireTableRowsUpdated(lowerIdx, upperIdx);
+                }
```
swalker2m: The only thing that is changing in this case is the auto group, correct?  The indices of other rows might be adjusted, but does that require a row update event?  I'm likely overlooking the obvious but the auto group insertion/deletions should be sufficient I believe.  Also I would have guessed that firing update events for the other rows that haven't changed would defeat the purpose of trying to limit the event to the minimum possible change.

Did you test this with GeMS by the way?  It is maybe the only current AGS strategy where multiple auto group rows might be added at once.

sraaphorst: I've tested it with GeMS quite a bit. (I hope thoroughly! I can't think of anything I might have missed.) Everything seems to work.

The reason why other rows may have changed is through the setting of the PA from the auto group. If the auto group is the primary and a BAGS lookup is triggered where another auto group with, say, a PA+180 is used, then the PA is set to 180. The guiding quality of the manual guide stars might (likely will) change as a result, as they will no longer be in the guider patrol field.

Thus, in order to set their guiding feedback correctly, it is necessary to recalculate each group and determine which, if any, have changed.

Does that make sense?

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/TelescopePosTableWidget.java
```diff
@@ -350,9 +311,172 @@ private static String extractName(final GuideGroup group) {
...
+                    });
+
+                    if (groupChanged) {
+                        newRows.add(gRowNew);
+                        groupRowsChanged.add(rowIdx);
+                    } else {
+                        newRows.add(row);
+                    }
+                } else {
+                    newRows.add(row);
```
swalker2m: A minor thing but if the `row` is always added we should just do so after the `if` only line 411.

sraaphorst: I think maybe I'm a bit confused as to what you mean here.
In one case, `gRowNew` is added, and not `row`.

swalker2m: Ok, yes right, I missed that.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/viewer/SPViewer.java
```diff
@@ -887,6 +892,9 @@ public void closeProgram(ISPProgram node) {
 
     /** Close the current viewer window, but don't exit the application, even if it is the last window. */
     private void closeViewer() {
+        if (getRoot() != null)
+            BagsManager.instance().unwatch(getRoot());
```
swalker2m: You can open more than a single program in a viewer.  When the viewer is closed, the cleanup should include all the programs in that viewer.  Can you confirm that this is happening as required?  It may be working fine but from this it looks like it could just be removing the active program in the view.

swalker2m:
This was a complicated task and I don't envy you for it.  We struggled to figure out ways to make the OT able to cope with changes happening outside of the event loop and didn't come up with any great ideas.  This is along the lines of the strategies we discussed though and it is very impressive that you got it all to work. :+1: pending a couple of comments above.

cquiroz:
I agree, great work, I can imagine the hurdles on breaking the loops between UI updates and model updates :+1:

sraaphorst:
Thanks to @swalker2m and @cquiroz for your comments.
All changes as recommended have been made, so if there are no further comments, I'm going to merge this so we can do a test release for Andy and Bryan.

PR #860, opened 2016-04-04
Remove ParallacticAngleDuration

tpolecat:
This removes `ParallacticAngleDuration`, which was only used in the editor and duplicated information stored in the `SchedulingBlock`.


bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/obs/SchedulingBlock.scala
```diff
@@ -1,6 +1,8 @@
 package edu.gemini.spModel.obs
 
-case class SchedulingBlock(start: Long, duration: Long)
+case class SchedulingBlock(start: Long, duration: Long) {
+  def useRemainingTime: Boolean = duration == 0L
+}
```
tpolecat: I will fix this in another PR ... `duration` needs to be an `Option`.

swalker2m:
Looks good to me, but could you comment @sraaphorst on whether there is any reason why removing what appears to be the duplicate copy of the duration in the instrument would be problematic?  I couldn't think of any or remember why we put it in both places (the scheduling block and the instrument).

sraaphorst:
I just looked at this, and it looks fine to me from what I can tell.

I can't remember for the life of me why we had the duplicate copy, either. It seems completely unnecessary. 👍

PR #861, opened 2016-04-04
REL-2648: QPT icon update

sraaphorst:
This simple PR is just an update to the QPT icons for Mac and Windows for 2016B.


sraaphorst:
Since there is nothing to see here, I am just going to merge it.

cquiroz:
Why the OT icon wasn't updated at the same time?

sraaphorst:
They were listed as separate issues in JIRA, so I did them separately, but apart from that, there's really no good reason why they shouldn't have been done together.

PR #863, opened 2016-04-04
REL-2702 Workaround for mystery o'clock

tpolecat:
Ok this works around the weirdness of the production ODB adding `o'clock` in the date formats for `validAt` ... I don't know why this is happening, but this works around it. To be continued...

### DIFF
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/SPTargetPio.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/SPTargetPio.java
@@ -42,8 +42,16 @@ public class SPTargetPio {
         try {
             return format.parse(dateStr);
         } catch (final ParseException e) {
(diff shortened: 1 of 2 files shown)


bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/SPTargetPio.java
```diff
@@ -42,8 +42,16 @@ public static synchronized Date parseDate(final String dateStr) {
...
-            return null;
+            try {
+
+                // At some point during 16A the server started writing out times with o'clock in
+                // them ... unclear why this happened. Workaround...
+                return format.parse(dateStr.replace("o'clock ", ""));
+
+            } catch (final ParseException ee) {
+                LOGGER.log(Level.WARNING, " Invalid date found " + dateStr);
+                return null;
```
cquiroz: Is it ok to return null? no problems upstream?

cquiroz:
This is :+1: please add a jira ticket to follow this up

PR #864, opened 2016-04-04
Make SchedulingBlock.duration optional

tpolecat:
This finishes a simplification started with #860. A scheduling block with duration zero had a special meaning ("remaining time"); we now use `None` for that, and any `Some(duration)` will always be positive.


cquiroz:
Looks good :+1:

PR #866, opened 2016-04-06
InstantiationFunctorTest bugfix

swalker2m:
This update was part of [PR 822](https://github.com/gemini-hlsw/ocs/pull/822), which was recently closed without merging.  I'm adding it here as a standalone PR since it took a while to diagnose and fix and is an improvement regardless of the fate of the other changes in #822.

This change fixes a bug in test case setup that caused some tests to be skipped and also avoids printing several long stack traces to the log while running the `InstantiationFunctorTest`.  There isn't a JIRA issue for this since it only impacts test case output and didn't seem worth raising to the attention of the science staff.


cquiroz:
:+1:

tpolecat:
Yeah sorry about 822 ... it was falling behind `develop` and I wasn't happy with the approach.

In any case :+1: for this.

PR #867, opened 2016-04-06
REL-2710: magnitude band rendering

tpolecat:
After spending an hour trying to change the list cell renderer I gave up and just changed the `toString`. Sorry.


cquiroz:
:+1:

PR #868, opened 2016-04-06
REL-2707 fix dms format

tpolecat:
`SPTarget.getDecString` was returning the positive angle rather than ± declination.


cquiroz:
:+1:

PR #869, opened 2016-04-08
REL-2698: HMS/DMS/Dec formatting updates

swalker2m:
This PR addresses several issues related to formatting angles.

1. Fixes a bug wherein you could end up with 60 minutes or arcminutes in a formatted RA or Dec.  For example, if you parse "0:59:59.999999" and then format it with just three digits to the right of the decimal you would end up with "0:60:00.000" instead of "1:00:00.000".

2. Makes it possible to specify precision when rounding fractional seconds (was fixed to 3 digits to the right of the decimal).

3. Changes declination / DMS to just two digits of precision to match the standard convention.

4. Makes it possible to use a separator other than `:` for the fields of a formatted RA/Dec. This is required for formatting ephemeris elements in the format expected by the TCS, which expects a single space to separate the fields.

For the commonly used syntax methods, there are no changes.  For example,

```scala
  someAngle.formatHMS
````

will select a `:` as the separator character and format to 3 decimal places as before.  The same for DMS and Declination except that it will default to 2 decimal places.  To change the separator or precision you have to use the more cumbersome `Angle.formatXXX` methods directly.

This code is a bit annoying and clumsy.  If someone sees a better way to do these things, please comment.

### DIFF
--- a/bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/Angle.scala
+++ b/bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/Angle.scala
@@ -409,18 +409,19 @@ object Angle {
 
   // Abstract over HMS/DMS
   private[core] def format3(a: Int, b: Int, c: Double, max: Int, sep: String, fractionalDigits: Int): String = {
-    val zs = "0" * fractionalDigits
-    val df = new DecimalFormat(s"00.$zs")
+    val df =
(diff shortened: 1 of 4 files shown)


cquiroz:
Looks good :+1:
Was the carry-on problem detected by user feedback?

sraaphorst:
This looks fine to me. 👍

swalker2m:
@cquiroz i noticed the issue when looking at the formatting code trying to figure out how to use it for the ephemeris robot

PR #871, opened 2016-04-08
REL-2722: BAGS now ignores instruments with no AGS strategies

sraaphorst:
Formerly, BAGS deemed GPI observations as unworthy of eligibility specifically as per an old JIRA task.
@cquiroz pointed out that this was not necessarily the best way to do things, and I agreed:
It is better if BAGS checks with `AgsRegistrar` if the current `ObsContext` allows any `AgsStrategies`, and if not, then do not register an observation with BAGS.

This allows us to add future instruments that have no strategies without having to explicitly remember to change `BagsManager`.


swalker2m:
Looks good. :+1:

PR #874, opened 2016-04-11
REL-2712: BAGS PA not being set correctly

sraaphorst:
This is a small bug where if the auto group was active, the position angle was not always being set correctly when it changed.

This would happen previously, say, if the PA was changed, and triggered a BAGS lookup that resulted in the same auto group. In that case `newEnv` and `oldEnv` were considered the same, so the subsequent PA changing code was not run.

Now, instead, if the primary guide group is the auto guide group, we simply set the associated PA.


cquiroz:
Looks good :+1:

swalker2m:
Makes sense.  👍

PR #876, opened 2016-04-11
REL-2724: Observations not eligible for BAGS must have empty auto groups

sraaphorst:
If an observation is not eligible for execution by BAGS, BAGS should perform no lookup and simply instead clear out the auto group for the observation.

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
@@ -257,7 +257,7 @@ object BagsManager {
     // If AGS results were found, apply them to the target env; otherwise, clear out any existing auto group.
     val newEnv = selOpt.map(_.applyTo(oldEnv)).getOrElse {
       val oldGuideEnv = oldEnv.getGuideEnvironment.guideEnv
-      if (oldGuideEnv.auto === AutomaticGroup.Initial) oldEnv
+      if (oldGuideEnv.auto === AutomaticGroup.Initial || oldGuideEnv.auto === AutomaticGroup.Disabled) oldEnv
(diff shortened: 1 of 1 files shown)


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -136,54 +147,51 @@ final class BagsManager(executorService: ExecutorService) {
...
-              //      running, so discard this result and do nothing,
-              //      otherwise update the model
-              //   }
-              //   on failure enqueue again, maybe with a delay depending on the failure
-              val bagsIdMsg = s"BAGS lookup on thread=${Thread.currentThread.getId} for observation=${obs.getObservationID}"
-              LOG.info(s"Performing $bagsIdMsg.")
+            ObsContext.create(obs).asScalaOpt.foreach { ctx =>
+              if (isEligibleForBags(ctx)
+                && hasBeenUpdated(obs, ctx)
+                && notObserved(obs)) {
```
swalker2m: If I read this correctly this means that observed observations will have the auto guide star they used wiped out?  That may not have been intended?

Also, won't this wipe out the auto guide star for otherwise eligible observations that have been updated but not in a way that impacts the AGS selection?

sraaphorst: Ugh. Good point. I think this needs only happen if it is not eligible for bags.
I'll take another look and see.

sraaphorst: Okay, I looked at this and changed it so that only BAGS eligibility is covered. This does some unnecessary checks (e.g. instrument has guiders), but should work otherwise.

I also fixed an issue that would result in disabled auto groups becoming initial.

swalker2m:
Seems like it should work. 👍

PR #878, opened 2016-04-11
REL-2726 - P2 Checks for Nonsidereal 

tpolecat:
This adds four P2 checks:
- `NoSchedulingBlock` - no scheduling block for non-sidereal observation
- `SchedulingBlockOutsideSemester` - if scheduling block doesn't fall within program semester
- `NoEphemerisForSchedulingBlock` - if ephemeris is undefined at scheduling block start
- `NoHorizonsDesignation` - if a nonsidereal target has no HORIZONS designation (this means the high-res lookup will fail)

### DIFF
--- a/bundle/edu.gemini.p2checker/src/main/scala/edu/gemini/p2checker/target/NonSiderealTargetRules.scala
+++ b/bundle/edu.gemini.p2checker/src/main/scala/edu/gemini/p2checker/target/NonSiderealTargetRules.scala
@@ -1,0 +1,86 @@
+package edu.gemini.p2checker.target
+
+import edu.gemini.p2checker.api.{P2Problems, IP2Problems, ObservationElements, IRule}
+import edu.gemini.shared.util.immutable.ScalaConverters._
+import edu.gemini.spModel.core._
+import scalaz._, Scalaz._
+
+final class NonSiderealTargetRules extends IRule {
+  import NonSiderealTargetRules._
+
+  implicit class SemesterOps(s: Semester) {
+    def contains(time: Long, site: Site): Boolean =
+      time >= s.getStartDate(site).getTime &&
+      time <=   s.getEndDate(site).getTime
+  }
+
+  override def check(es: ObservationElements): IP2Problems = {
+    val p2p = new P2Problems
+    p2p.append(checkNoHorizonsDesignation(es))
+    p2p.append(checkNoSchedulingBlock(es))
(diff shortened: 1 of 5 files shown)


cquiroz:
Looks fine :+1:

bundle/edu.gemini.p2checker/src/main/scala/edu/gemini/p2checker/target/NonSiderealTargetRules.scala
```diff
@@ -0,0 +1,86 @@
...
+import scalaz._, Scalaz._
+
+final class NonSiderealTargetRules extends IRule {
+  import NonSiderealTargetRules._
+
+  implicit class SemesterOps(s: Semester) {
+    def contains(time: Long, site: Site): Boolean =
+      time >= s.getStartDate(site).getTime &&
+      time <=   s.getEndDate(site).getTime
+  }
```
swalker2m: Seems like maybe something that is useful enough to go into `Semester` itself.  It's surprising we've never needed it before.

bundle/edu.gemini.p2checker/src/main/scala/edu/gemini/p2checker/target/NonSiderealTargetRules.scala
```diff
@@ -0,0 +1,86 @@
...
+          es.getObservationNode)
+    }
+
+  def checkNoEphemerisForSchedulingBlock(es: ObservationElements): IP2Problems =
+    new P2Problems <| { p2p =>
+      for {
+        ocn <- es.getTargetObsComponentNode.asScalaOpt.toList
+        toc <- es.getTargetObsComp.asScalaOpt.toList
+        tar <- toc.getTargetEnvironment.getTargets.asScalaList
+        nst <- tar.getNonSiderealTarget.toList if nst.horizonsDesignation.isEmpty
```
swalker2m: Couldn't a target have a horizons designation and yet not have ephemeris including the scheduling block start time?  That is, maybe we don't need the `if nst.horizonsDesignation.isEmpty` filter there or maybe we should but the condition should be reversed?

tpolecat: Yep, thanks. This was a copy/paste error. The condition shouldn't be there.

swalker2m:
👍

PR #879, opened 2016-04-12
OCSADV-383-B: rename ephemeris files

swalker2m:
This PR changes the names given to ephemeris files written for the TCS.  Currently the TCC (or is it the TCS?) takes the target name and appends `.eph` and then looks for a matching filename. Since there is no guarantee that human readable names entered by the user map to unique non-sidereal targets, this is a recipe for fault reports.  Instead, we’ll use the unique `HorizonsDesignation` as the filename and tell the TCC (or TCS via the TCC) what filename to look for when it loads a configuration from the WDBA.  This requires a corresponding update in the TCC/TCS.

### DIFF
--- a/bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/HorizonsDesignation.scala
+++ b/bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/HorizonsDesignation.scala
@@ -1,8 +1,9 @@
 package edu.gemini.spModel.core
 
+
 import scalaz._, Scalaz._
 
-/** 
+/**
  * Unique Horizons designation, which should allow for reproduceable ephemeris queries <b>if</b> the
  * values passed to the constructors are extracted correctly from search results. See `horizons.md`
  * in the bundle source for more information.
@@ -10,11 +11,24 @@ import scalaz._, Scalaz._
 sealed abstract class HorizonsDesignation(val queryString: String)
   extends Product with Serializable {
     def des: String // designation, human readable
+
+    import edu.gemini.spModel.core.HorizonsDesignation._
+
+    /** Exports an HorizonsDesignation to a String in a format that can be read
+      * by the `HorizonsDesignation.read` method.
+      */
+    def show: String =
+      this match {
+        case Comet(des)            => s"Comet_$des"
+        case AsteroidNewStyle(des) => s"AsteroidNew_$des"
+        case AsteroidOldStyle(num) => s"AsteroidOld_$num"
+        case MajorBody(num)        => s"MajorBody_$num"
+      }
   }
 
 object HorizonsDesignation {
 
-  /** 
+  /**
    * Designation for a comet, in the current apparition. Example: `C/1973 E1` for Kohoutek, yielding
    * the query string `DES=C/1973 E1;CAP`.
    */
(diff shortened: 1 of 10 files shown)


bundle/edu.gemini.spdb.reports.collection/src/main/scala/edu/gemini/dbTools/ephemeris/TcsEphemerisExport.scala
```diff
@@ -166,44 +168,42 @@ class TcsEphemerisExport(dir: File, night: Night, site: Site, odb: IDBDatabaseSe
...
 
-  def writeEphemeris(hid: HorizonsDesignation, name: String, em: EphemerisMap): TryExport[File] = {
-    // TODO: taking a random string and assuming it is suitable for a filename
-    // TODO: seems like asking for trouble.  any suggestions?  URL encode?
-    // TODO: still not sure what to do about file names anyway.
-    val fileName = s"${name}_${hid.toString}.eph".replaceAll("/", "-")
-
+  def writeEphemeris(hid: HorizonsDesignation, em: EphemerisMap): TryExport[File] = {
+    // See TcsConfig.ephemerisFile
+    val fileName = URLEncoder.encode(s"${hid.toString}.eph", UTF_8.name)
```
swalker2m: I would like to do the translation `HorizonsDesignation => filename` in just one place, yet it didn’t seem like an appropriate thing to add to `NonSiderealTarget` or anywhere in `core` really.  It also doesn’t seem right to add a dependency from the cronjob bundle to the WDBA or vice versa.  Suggestions?

tpolecat: The `HorizonsDesignation` query can contain semicolons and spaces as well (at least .. there could be more) which will probably screw up the TCC, so we may need to do some additional encoding.

swalker2m:
Since this has been open for a week without comment, I'll go ahead and merge it at some point today if there are no objections.

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/HorizonsDesignation.scala
```diff
@@ -1,20 +1,34 @@
...
 sealed abstract class HorizonsDesignation(val queryString: String)
   extends Product with Serializable {
     def des: String // designation, human readable
+
+    import edu.gemini.spModel.core.HorizonsDesignation._
+
+    /** Exports an HorizonsDesignation to a String in a format that can be read
+      * by the `HorizonsDesignation.read` method.
+      */
+    def show: String =
```
cquiroz: Why do we have a pair `show`/`read` using strings? Are they going to a file and written/read by different processes?

cquiroz:
Looks good :+1:

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/HorizonsDesignation.scala
swalker2m: The ephemeris robot will use `show` to come up with a string representation of the horizons id to use for the ephemeris filename.   Later when it runs again it will use `read` to convert the filename back to a horizons id so it can do housekeeping on the ephemeris files.

tpolecat:
Yeah sorry, we chatted about this and it seems fine. 👍

PR #880, opened 2016-04-13
REL-2723: Allow enabling of disabled auto groups

sraaphorst:
In the current model, when a pre-2016 program is imported, the auto guide group is set to disabled.
It must be possible to enable these auto groups, and the mechanism that we have decided to allow this to be done is to have the disabled auto group selected as the primary or double clicked.

This change simply sets a disabled auto group to an enabled one, updates the table model, and then fires an event indicating that the table has been updated. BAGS handles the rest, and performs a lookup.

### DIFF
--- a/bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/TelescopePosTableWidget.java
+++ b/bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/TelescopePosTableWidget.java
@@ -314,7 +314,7 @@ final class TelescopePosTableWidget extends JTable implements TelescopePosWatche
         // Update the auto row of a table.
         // This is called when the auto row should be changed from disabled to initial.
         void enableAutoRow(final TargetEnvironment env) {
-            final GroupRow autoGroupRow = new GroupRow(true, true, 0, env.getGuideEnvironment().automaticGroup(),
+            final GroupRow autoGroupRow = new GroupRow(true, false, 0, env.getGuideEnvironment().automaticGroup(),
                     Collections.emptyList());
             rows.set(1, autoGroupRow);
             fireTableRowsUpdated(1, 1);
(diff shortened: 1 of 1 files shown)


bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/TelescopePosTableWidget.java
```diff
@@ -311,6 +311,15 @@ private static String extractName(final GuideGroup group) {
             return tmpRows;
         }
 
+        // Update the auto row of a table.
+        // This is called when the auto row should be changed from disabled to initial.
+        void enableAutoRow(final TargetEnvironment env) {
+            final GroupRow autoGroupRow = new GroupRow(true, true, 0, env.getGuideEnvironment().automaticGroup(),
+                    Collections.emptyList());
+            rows.set(1, autoGroupRow);
+            fireTableRowsUpdated(1, 1);
```
cquiroz: Is this running on the swing thread?

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/TelescopePosTableWidget.java
```diff
@@ -918,17 +927,22 @@ void updatePrimaryStar() {
...
-                        // If we are switching to an automatic group, we also
-                        // possibly need to update the position angle.
-                        final GuideGrp grp = igg.group().grp();
-                        if (grp instanceof AutomaticGroup.Active) {
-                            updatePosAngle(((AutomaticGroup.Active) grp).posAngle());
-                        }
+
+                // If the auto group is disabled and set to primary, then make it an initial auto group.
+                final GuideGrp grp = igg.group().grp();
+                final GuideEnvironment ge = env.getGuideEnvironment();
```
cquiroz: Can we wrap `ge` in an `Option/Optional` and use a foreach to avoid the null check?

sraaphorst: Sure, no reason why not. I'll do that.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/TelescopePosTableWidget.java
sraaphorst: I believe that it will always run on the Swing thread, yes, as this code is only ever invoked by a mouse listener and an action listener. Does that sound right?

cquiroz: That sounds right

PR #881, opened 2016-04-13
REL-2709 tpe feature for nonsidereal targets

tpolecat:
This adds a TPE feature to display the track of non-sidereal objects (and removes the old implementation).

- The entire ephemeris is drawn; you can zoom out and see more.
- Labels are drawn such that they are at least 100px apart, to avoid crowding (they may be farther apart for very fast targets and/or if you're zoomed way in).
- Labels and tick marks are rotated based on the local curvature, which improves readability for curvy paths.

image

Note that right now testing will require a bit of patience due to an unrelated UI deadlock in the target editor. Closing the TPE before doing an ephemeris lookup seems to help.


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/tpe/feat/TpeEphemerisFeature.scala
```diff
@@ -0,0 +1,132 @@
...
+    as match {
+      case Nil => Nil
+      case a :: as => go(as, a, Nil)
+    }
+  }
+
+  val formatDate: Long => String = {
+    val df = new SimpleDateFormat("dd-MMM HH:mm")
+    df.setTimeZone(TimeZone.getTimeZone("UTC"))
+    t => df.synchronized(df.format(new Date(t)))
```
cquiroz: Why is it synchronized if it is a local var?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/tpe/feat/TpeEphemerisFeature.scala
```diff
@@ -0,0 +1,132 @@
...
+  def ang(p1: Point, p2: Point): Double =
+    if ((p1.x - p2.x).abs < 0.1) 0.0 // ~inf = 0
+    else {
+      val slope = (p2.y - p1.y) / (p2.x - p1.x)
+      val angle = math.atan(slope) - Math.PI / 2.0
+      if (slope < 0.0) angle - Math.PI else angle
+    }
+
+  // discard adjacent elements that meet some predicate
+  def sparsify[A](as: List[A])(f: (A, A) => Boolean): List[A] = {
```
cquiroz: Cool :+1:

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/tpe/feat/TpeEphemerisFeature.scala
tpolecat: The returned function closes over it, and can be called from multiple threads.

cquiroz:
Love it :+1:

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/tpe/feat/TpeEphemerisFeature.scala
cquiroz: Do have newer, safer formatters in the new java date api?

tpolecat: I don't know, possibly. I copy/pasted this from the old `TpeHorizonsFeature`.

PR #882, opened 2016-04-13
Remove HORIZONS server bundle.

tpolecat:
This removes the now-unused HORIZONS server.


bundle/edu.gemini.horizons.api/build.sbt
```diff
@@ -18,13 +18,10 @@ osgiSettings
 
 ocsBundleSettings
 
-OsgiKeys.bundleActivator := Some("edu.gemini.horizons.api.osgi.Activator")
```
cquiroz: Maybe this should be `None`, otherwise it could pick a default? I don't remember exactly though

tpolecat: Good question. I'll change it just to be sure.

cquiroz:
:+1:

PR #883, opened 2016-04-13
Remove remaining bits of old target model.

tpolecat:
This removes the remaining code that was in `edu.gemini.spModel.target.system`.


cquiroz:
The PR is breaking the build, some deleted code perhaps?

tpolecat:
Yeah piece-o-crap IDEA didn't catch it. Standby.

tpolecat:
This is building now, sorry.

PR #884, opened 2016-04-14
REL-2715: Making addition of manual guide groups easier

sraaphorst:
This is a weird bit of hackery that was requested by the astronomers.

In the target editor, previously, when an observation was created, the only things in the target table were the base and the auto guide group. If someone wanted a manual guide star, this required a two-step process, first adding the manual guide group and then adding the manual target to it.

The astronomers instead wanted manual targets to be able to be added in a one-step process, i.e. if there is no manual guide group, still allow users to add a manual target by automatically creating a manual guide group to hold the manual target.

This implements that bit of hackery.

### DIFF
--- a/bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
+++ b/bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
@@ -245,9 +245,14 @@ public final class EdCompTargetList extends OtItemEditor<ISPObsComponent, Target
     /**
      * Update the menu items that allow the addition of guide stars. These must be disabled for the automatic group
      * or if we are currently on the base or a user target.
+     *
+     * REL-2715 similarly demands that if there is only the auto guide group that they be enabled, with a new guide
+     * group created automatically.
      */
     private void updateGuideStarAdders() {
-        final boolean enabled = selectionIsManualGroup() || selectionIsManualTarget();
+        // If there is only one group, then it must be the auto group.
+        final boolean onlyAutoGroup = getContextTargetEnv().getGroups().size() == 1;
+        final boolean enabled       = onlyAutoGroup || selectionIsManualGroup() || selectionIsManualTarget();
         _guideStarAdders.forEach((gp, comp) -> comp.setEnabled(enabled));
     }
 
@@ -532,13 +537,17 @@ public final class EdCompTargetList extends OtItemEditor<ISPObsComponent, Target
         }
 
         @Override public void actionPerformed(final ActionEvent actionEvent) {
-            TargetEnvironment env = obsComp.getTargetEnvironment();
-            GuideEnvironment ge = env.getGuideEnvironment();
+            final TargetEnvironment env = obsComp.getTargetEnvironment();
+            final GuideEnvironment ge = env.getGuideEnvironment();
 
-            // Make sure that a group is selected and that it is not the auto group.
-            final Option<IndexedGuideGroup> iggOpt = positionTable.getSelectedGroupOrParentGroup(env);
-            if (!iggOpt.exists(igg -> !igg.group().isAutomatic()))
-                return;
+            // REL-2715 demands that we allow manual guide groups to be automatically created if only the auto
+            // group is present. This requires some ugly hacking.
+            final Option<IndexedGuideGroup> iggOpt = positionTable.getSelectedGroupOrParentGroup(env)
+                    .filter(igg -> !igg.group().isAutomatic())
+                    .orElse(() -> { return appendNewGroup(obsComp, positionTable); });
+
+            // Env may have changed at this point if we appended a new group, so we must re-retrieve env.
+            final TargetEnvironment newEnv = obsComp.getTargetEnvironment();
 
             Option<GuideProbeTargets> opt = iggOpt.flatMap(igg -> igg.group().get(probe));
             final Integer groupIndex = iggOpt.map(IndexedGuideGroup::index).getOrElse(ge.getPrimaryIndex());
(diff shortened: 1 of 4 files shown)


bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
```diff
@@ -532,13 +537,17 @@ private void showTargetTag() {
...
-            TargetEnvironment env = obsComp.getTargetEnvironment();
-            GuideEnvironment ge = env.getGuideEnvironment();
+            final TargetEnvironment env = obsComp.getTargetEnvironment();
+            final GuideEnvironment ge = env.getGuideEnvironment();
+
+            // REL-2715 demands that we allow manual guide groups to be automatically created if only the auto
+            // group is present. This requires some ugly hacking.
+            final Option<IndexedGuideGroup> iggOpt = positionTable.getSelectedGroupOrParentGroup(env)
+                    .filter(igg -> !igg.group().isAutomatic())
+                    .orElse(() -> { return appendNewGroup(obsComp, positionTable); });
```
cquiroz: Why do you need the block and return here? Would it be enough to do `.orElse(() -> appendNewGroup(obsComp, positionTable));`?

sraaphorst: Oh, you're right. It isn't needed. (I accidentally carried it across from when I was doing multiple steps here and before I refactored out appending a new guide group.)
Thanks for pointing this out. I'll push another commit.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
```diff
@@ -578,25 +587,35 @@ private void showTargetTag() {
...
-                    .setPrimaryIndex(newGroupIdx)));
+        final ImList<GuideGroup> oldGroups = ge.getOptions();
+        final int newGroupIdx              = oldGroups.size();
+        final GuideGroup newGroup          = GuideGroup.create("Manual Group");
+        final IndexedGuideGroup igg        = new IndexedGuideGroup(newGroupIdx, newGroup);
+        final ImList<GuideGroup> newGroups = oldGroups.append(newGroup);
 
-            positionTable.selectGroup(igg);
+        // OT-34: make new group primary and select it
+        if (!positionTable.confirmGroupChange(primaryGroup, newGroup)) {
```
cquiroz: This is just nitpicking, often the `if` is set like `if (condTrue) x else y`

sraaphorst: Sure, I'll change it. The `if` statement was already implemented before I came along, but no reason why it shouldn't be cleaned up.

cquiroz:
Looks good :+1:

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
```diff
@@ -245,9 +245,14 @@ private void updateUIForGroup() {
...
      * Update the menu items that allow the addition of guide stars. These must be disabled for the automatic group
      * or if we are currently on the base or a user target.
+     *
+     * REL-2715 similarly demands that if there is only the auto guide group that they be enabled, with a new guide
+     * group created automatically.
      */
     private void updateGuideStarAdders() {
-        final boolean enabled = selectionIsManualGroup() || selectionIsManualTarget();
+        // If there is only one group, then it must be the auto group.
+        final boolean onlyAutoGroup = getContextTargetEnv().getGroups().size() == 1;
```
swalker2m: This is certainly true but a bit indirect maybe.  We already have a `getAutomaticGroup` in `GuideEnvironment`, perhaps we should add a `getManualGroups`.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
```diff
@@ -532,13 +537,17 @@ private void showTargetTag() {
...
+            final Option<IndexedGuideGroup> iggOpt = positionTable.getSelectedGroupOrParentGroup(env)
+                    .filter(igg -> !igg.group().isAutomatic())
+                    .orElse(() -> appendNewGroup(obsComp, positionTable));
 
-            // Make sure that a group is selected and that it is not the auto group.
-            final Option<IndexedGuideGroup> iggOpt = positionTable.getSelectedGroupOrParentGroup(env);
-            if (!iggOpt.exists(igg -> !igg.group().isAutomatic()))
-                return;
+            // Env may have changed at this point if we appended a new group, so we must re-retrieve env.
+            final TargetEnvironment newEnv = obsComp.getTargetEnvironment();
```
swalker2m: This isn't a big deal but I was a bit confused by this at first.  What if the `appendNewGroup` didn't update the data object (which is clearly a side-effect)  but just returned a new target environment and the group it created?  It might help to leave the side effect for the end when totally finished with calculating the new target environment.  In other words, keep it pure until the end at least.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
sraaphorst: Sure, I'll do that and change the condition here.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
sraaphorst: Sounds good.

PR #887, opened 2016-04-15
REL-2706 fix deadlock

tpolecat:
This seems to fix the deadlock in the non-sidereal target editor when doing a Horizons lookup.

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details2/NonSiderealNameEditor.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details2/NonSiderealNameEditor.scala
@@ -58,12 +58,18 @@ final class NonSiderealNameEditor extends TelescopePosEditor with ReentrancyHack
 
     def updateDesignation(hd: HorizonsDesignation, name: String): HS2[Unit] =
       HS2.delay {
-        val t0 = Target.name.set(spt.getTarget, name)
-        Target.horizonsDesignation.set(t0, Some(hd)).foreach(spt.setTarget)
+        Swing.onEDT {
+          val t0 = Target.name.set(spt.getTarget, name)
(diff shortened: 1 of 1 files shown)


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/details2/NonSiderealNameEditor.scala
```diff
@@ -58,12 +58,18 @@ final class NonSiderealNameEditor extends TelescopePosEditor with ReentrancyHack
 
     def updateDesignation(hd: HorizonsDesignation, name: String): HS2[Unit] =
       HS2.delay {
-        val t0 = Target.name.set(spt.getTarget, name)
-        Target.horizonsDesignation.set(t0, Some(hd)).foreach(spt.setTarget)
+        Swing.onEDT {
```
cquiroz: Are there swing calls in here? It would seem you wouldn't need to run on the EDT if they are non UI calls

tpolecat: It triggers an event that a lot of code seems to assume always comes on the EDT ... most model changes happen due to user input so this is an unusual case. It's not ideal but I think it's probably the safest approach.

cquiroz: Right, the event is used in the EDT so it should originate in the same thread. Is that because of the target listeners?

tpolecat: It's because of stuff everywhere listening for the data object to be updated. This manifests as calls to that go all over the place in the OT.

cquiroz:
:+1:

andrewwstephens:
I've queried 20-30 nonsidereal objects with no problems ￼:+1:

PR #888, opened 2016-04-18
Scalaz 7.2

cquiroz:
This PR updates scalaz to 7.2.2. It contains changes due to removed and deprecated methods in scalaz. The apps have been tested and seem to be working fine.

One little issue is that in scala 7.2.x `unsafeTap`, also known as `<|` has been [deprecated](https://github.com/scalaz/scalaz/blob/series/7.2.x/core/src/main/scala/scalaz/syntax/IdOps.scala#L27) and marked for removal in 7.3. We use that method extensively and thus we get a bunch of new warnings. I'd suggest to leave the warnings and when we move to 7.3 we implement unsafeTap ourselves

Due to the scalaz update, we had to also update doobie, specs2, argonaut, monocle and scalacheck

A relevant factor for this upgrade: scalaz 7.2.1 is the first scalaz release with offcial scala.js support. This change would let use use the same version in both sides


bundle/edu.gemini.ags/src/test/scala/edu/gemini/ags/impl/SingleProbeStrategySpec.scala
```diff
@@ -39,7 +37,7 @@ import scala.collection.JavaConverters._
 import scalaz._
 import Scalaz._
 
-class SingleProbeStrategySpec extends Specification with NoTimeConversions {
```
swalker2m: Out of curiosity, what's the deal with `NoTimeConversions`?

cquiroz: `NoTimeConversions` was a specs2 trait that avoided conflicts among implicit methods in specs2 and `scala.concurrent.duration._`. This PR upgrades specs2 to 3.7.1 and in that version there is no need for `NoTimeConversions`

swalker2m:
This looks good. 👍 I assigned you [REL-2674](http://jira.gemini.edu:8080/browse/REL-2674) which was created to do this task.

Maybe @tpolecat or someone could explain the logic of some of the scalaz data structures switching to `IList` and `ISet`.  We have to do some extra conversions now which seems unfortunate.  It might, maybe, be worth changing a few things in a subsequent PR to add direct support for `IList` etc instead of having to convert.  Or maybe we just live with it until everything is switched to cats anyway.

tpolecat:
`IList` is a better list and `ISet` is a better set so scalaz switched ... there is also `Maybe` which is a better `Option`. It does require some conversion at the edges.

Cats doesn't have these structures but the related project Dogs will have them at some point, although it's lagging behind a bit.

tpolecat:
Great effort on this. It does require us to use a pre-release version of doobie but I think this is ok.

PR #889, opened 2016-04-18
REL-2733: Making auto disabled guide groups enableable

sraaphorst:
This morning's PR for REL-2716 accidentally introduced a bug that ignored double clicks / primary setting on the already primary guide group.

This is problematic since if the auto guide group is the only guide group and disabled, we rely on the user double-clicking or clicking the set primary button to enable the auto group and thus make BAGS run.

This fixes that by removing the check that was done to see if the currently selected guide group is the primary, and not processing the guide group if that is the case.

### DIFF
--- a/bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/TelescopePosTableWidget.java
+++ b/bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/TelescopePosTableWidget.java
@@ -928,11 +928,11 @@ final class TelescopePosTableWidget extends JTable implements TelescopePosWatche
             // If the auto group is disabled and set to primary, then make it an initial auto group.
             ImOption.apply(env.getGuideEnvironment()).foreach(ge -> {
                 final GuideGrp grp = igg.group().grp();
-                if (ge != null && grp instanceof AutomaticGroup.Disabled$) {
+                if (grp instanceof AutomaticGroup.Disabled$) {
                     final TargetEnvironment envNew = env.setGuideEnvironment(ge.setAutomaticGroup(GuideGroup.AutomaticInitial()).setPrimaryIndex(0));
                     _dataObject.setTargetEnvironment(envNew);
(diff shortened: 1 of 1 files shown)


bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/TelescopePosTableWidget.java
```diff
@@ -922,31 +922,27 @@ void updatePrimaryStar() {
...
-                        final TargetEnvironment envNew = env.setGuideEnvironment(ge.setAutomaticGroup(GuideGroup.AutomaticInitial()).setPrimaryIndex(0));
-                        _dataObject.setTargetEnvironment(envNew);
-                        _model.enableAutoRow(envNew);
+        iggOpt.foreach(igg -> {
+            final GuideGroup primary = env.getPrimaryGuideGroup();
+
+            // If the auto group is disabled and set to primary, then make it an initial auto group.
+            ImOption.apply(env.getGuideEnvironment()).foreach(ge -> {
+                final GuideGrp grp = igg.group().grp();
+                if (ge != null && grp instanceof AutomaticGroup.Disabled$) {
```
cquiroz: I guess ge can't be null if it comes from the ImOption?

sraaphorst: You are right. If `env.getGuideEnvironment` was null, the `ImOption.apply` call would turn it into `None`, effectively guaranteeing that `ge` can never be null.

I'll take this part of the `if` clause out.

swalker2m:
I didn't get a chance to review this one before it was merged but I have a question just in case.  This won't generate an edit (and a need to sync) if the group you're clicking is already primary?

sraaphorst:
Yes, that is correct. I check to make sure that an actual change would occur in the model before mucking with anything, but I'll verify to make absolutely certain.

sraaphorst:
Tested and looks fine.
Are you in? Not on HipChat?

PR #890, opened 2016-04-18
REL-2732 Fix Coordinate Enablement

tpolecat:
This simplifies the editor initialization, simply ignoring targets that don't have fixed coordinates rather than disabling the controls. They are never visible in this case so it doesn't matter ... we're doing this in a lot of places.


cquiroz:
Looks good :+1:

PR #891, opened 2016-04-18
REL-492: GMOS ITC warning when 4x spectral binning selected with IFU …

spakzad:
…mode.

Changes were only made to Gmos.java.

### DIFF
--- a/bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gmos/Gmos.java
+++ b/bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gmos/Gmos.java
@@ -335,6 +335,10 @@ public abstract class Gmos extends Instrument implements BinningProvider, Spectr
             if ((gp.fpMask().isIFU() || isIfu2()) && gp.spatialBinning() != 1) {
                 add (new ItcWarning("Spatial binning is strongly discouraged with IFU observations."));
             }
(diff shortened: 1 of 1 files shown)


cquiroz:
Looks fine, just a general question, why are there tabs `\t` in the warning, are they displayed to the user?

spakzad:
Hi Carlos,
I was trying to add tabs to make it easier to read, but I see this is not appearing in the output.  I can take them out but they don't seem to do anything.

cquiroz:
Right, that's what I thought. Would they appear on the web page? In html, tab really has no meaning, if you just mean spaces, you could put 4 spaces (or &nbsp; which is the representation for space)

spakzad:
OK, I'll fix it and re-submit it.  Thanks!

cquiroz:
the blank space was eaten in my comment,`&nbsp;` is normally the character used for a space in html. You could put for of those `&nbsp;&nbsp;&nbsp;&nbsp;` and should give you the same effect.

If you have fixes you can add them as new commits and update the Pull Request instead of closing, but closing is fine too

spakzad:
I'm attaching a sample of what it looks like now.  Is it OK or should I take the extra spaces out?  If I do take them out, should I put everything on one line like the warning above this one?
rel492_sampleoutput

andrewwstephens:
The current warning looks fine in the OT, although for consistency I would suggest using normal (not upper) case for the warning, and ending with a period since other warnings may come after this one.

warning

PR #894, opened 2016-04-19
REL-2708 fix failing major body lookup

tpolecat:
Non-sidereal name lookup was failing for some unique results for moons due to trailing whitespace differences in HORIZONS results (╯°□°）╯︵ ┻━┻ and was also renaming moons in an irritating way. Both fixed.


swalker2m:
👍

PR #895, opened 2016-04-20
REL-2720: Position angle text field updating

sraaphorst:
Changing the position angle for an instrument can trigger a BAGS lookup, and if the position angle constraint is set to allow 180 degree flips, then when BAGS selects an auto group, if the auto group is at +180, the position angle text field in the instrument editors will be reflected to show this.

This is problematic when the user is editing the position angle text field, which triggers BAGS lookups, which can set the PA and thus contents of the field currently being edited.

This change simply ignores overwriting the contents of the PA text field if the field has the focus, i.e. is currently being edited. Thus, while BAGS may store PA in the model, the text field is not updated while edited.

Once the focus is gained / lost, if necessary, the text field is updated to reflect the model as well.

### DIFF
--- a/bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/EdCompInstBase.java
+++ b/bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/EdCompInstBase.java
@@ -1,9 +1,3 @@
-// Copyright 1997 Association for Universities for Research in Astronomy, Inc.,
-// Observatory Control System, Gemini Telescopes Project.
-// See the file LICENSE for complete details.
-//
-// $Id: EdCompInstBase.java 47001 2012-07-26 19:40:02Z swalker $
-//
 package jsky.app.ot.gemini.editor;
 
 import edu.gemini.pot.sp.ISPObsComponent;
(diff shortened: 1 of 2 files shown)


bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/EdCompInstBase.java
```diff
@@ -60,26 +67,21 @@ protected void cleanup() {
...
                 if (tbwe == getPosAngleTextBox()) {
-                    /* The default position angle */
                     final double defaultPositionAngle = 0.0;
                     getDataObject().setPosAngleDegrees(tbwe.getDoubleValue(defaultPositionAngle));
                 } else if (tbwe == getExposureTimeTextBox()) {
-                    double expTime = tbwe.getDoubleValue(getDefaultExposureTime());
-                    if (isForceIntegerExposureTime()) expTime = Math.floor(expTime);
-                    getDataObject().setExposureTime(expTime);
+                    final double expTime    = tbwe.getDoubleValue(getDefaultExposureTime());
+                    final double expTimeAdj = isForceIntegerExposureTime() ? Math.floor(expTime) : expTime;
```
cquiroz: When do we force exp time to be an integer? For some instrument in particular?

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/EdCompInstBase.java
```diff
@@ -89,31 +91,39 @@ public void textBoxKeyPress(TextBoxWidget tbwe) {
...
             }
         }
-        tbwe = getExposureTimeTextBox();
-        // Added this check because TReCS doesn't have an exposure time box!
-        if (tbwe != null && getDataObject() != null) {
+    }
+
+    // Copy the data model exposure time value to the exposure time text field.
+    private void updateExpTime() {
+        if (_ignoreChanges) return;
```
cquiroz: `if (!_ignoreChanges) doX`

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/EdCompInstBase.java
sraaphorst: It looks like the functionality is in place to allow for this (i.e. there is a protected method `EdCompInstBase.isForceIntegerExposureTime` that returns `false`), but it is never overridden and thus never used.

I'm guessing we can probably just remove all of this.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/EdCompInstBase.java
sraaphorst: Sure 😄 
Changed (in both places), committed, and pushed.

cquiroz:
👍

PR #896, opened 2016-04-20
REL-492: GMOS ITC warning when 4x spectral binning selected with IFU …

spakzad:
…mode.

After consulting with Andy and Kathy, this is the warning that is generated when spectral binning equals 4.  Should look good in the OT too.


cquiroz:
Looks good to me 👍

PR #897, opened 2016-04-20
Remove seqexec related bundles

cquiroz:
This PR removes the seqexec bundles from OCS since development is taking place in ocs3. Though they don't bother much, they increase compilation time and you need to keep up with updates like for the update to scalaz 7.2


sraaphorst:
Nice. Looks good to me! 👍

PR #899, opened 2016-04-20
REL-2735 add new header format

tpolecat:
This adds a handler for HORIZONS unique asteroid result headers of the form

    JPL/HORIZONS                 (2016 GB222)                 2016-Apr-20 15:22:36

which had not been seen before.


arturog8m:
:+1:  

Ps: Wouldn't it be lovely if horizons had a normal computer interface so we wouldn't need to parse these things?

cquiroz:
👍

PR #900, opened 2016-04-21
Revert "Remove seqexec related bundles"

cquiroz:
Rats, this is a bad PR. The edu.gemini.seqexec.shared bundle should exist in spdb, I miss-understood the logic beneath.

That module should stay in ocs and be removed from ocs3, sorry.

Reverts gemini-hlsw/ocs#897


cquiroz:
Given this is just a revert of code that's never modified I'll merge right away to avoid build problems by other team members

arturog8m:
For education purposes, why this one needs to be in SPDB? I wouldn't have figured that from its name.

cquiroz:
the shared bundle contains a functor that uses spdb and lets the seqexec server read sequence information over trpc. I had the impression that it only contained classes but it contains the functor. Without it the seqexec server cannot talk to spdb
I may need to split it into 2 bundles containing just the functor and an "API" bundle, but let's see

PR #901, opened 2016-04-21
Remove seqexec server bundle

cquiroz:
Second attempt to remove the seqexec server bundle. This time only `edu.gemini.seqexec.server` is removed and `edu.gemini.seqexec.shared` has been renamed to `edu.gemini.seqexec.odb`. I prefer that name to make it more explicit and it will make it simpler to use on the ocs3 side

The following was tested:
* Full compilation
* Creation of idea project
* Launch of spdb
* Import of `edu.gemini.seqexec.odb` from ocs3


swalker2m:
This sounds fine. 👍

cquiroz:
I hope this time it works 😄

PR #903, opened 2016-04-22
OCSADV-453B: Bags feedback

sraaphorst:
This is (fingers crossed) the final BAGS component that needed implementation for June, namely feedback, which is displayed in the bottom of the target editor as per the attached image.

If someone like @tpolecat could review the synchronization of `BagsStatus`, that would be greatly appreciated.

untitled


swalker2m:
Just in case, there should be no need to add a new spinner gif unless this one is special somehow. Maybe it's smaller?  The OT has `spinner.gif` and `spinner16.gif` already.

cquiroz:
Can you do animated gifs in Swing?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -141,17 +142,37 @@ final class BagsManager(executorService: ExecutorService) {
...
-            // Otherwise construct an obs context, verify that it's bags-worthy, and go.
-            ObsContext.create(obs).asScalaOpt.foreach { ctx =>
-              val eligibleForBags = isEligibleForBags(ctx)
-              if (eligibleForBags
-                && hasBeenUpdated(obs, ctx)
-                && notObserved(obs)) {
+
+        def status(newStatus: BagsStatus): Unit = {
+          val oldStatusOpt = state.statuses.get(key)
+          if (!oldStatusOpt.exists(_.equals(newStatus))) {
```
swalker2m: I think this could be simplified to `!oldStatusOpt.contains(newStatus)`?

bundle/jsky.app.ot/src/main/java/jsky/app/ot/ags/BagsStatusListener.java
```diff
@@ -0,0 +1,9 @@
+package jsky.app.ot.ags;
+
+import edu.gemini.pot.sp.ISPObservation;
+import edu.gemini.shared.util.immutable.Option;
+
+@FunctionalInterface
```
cquiroz: 👍  I like the `@FunctionalInterface` tag

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
```diff
@@ -103,6 +103,11 @@ public void textBoxKeyPress(final TextBoxWidget tbwe) {
         });
 
         _agsPub.subscribe((obs, oldOptions, newOptions) -> updateGuiding());
+
+        BagsManager.instance().addBagsStatusListener((obs, oldStatus, newStatus) -> {
+            if (obs == getContextObservation())
```
cquiroz: `equal`?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -141,17 +142,37 @@ final class BagsManager(executorService: ExecutorService) {
...
+            state = state.setStatus(key, newStatus)
+            notifyBagsStatusListeners(observation, oldStatusOpt, Some(newStatus))
```
swalker2m: It appears to me that this is sometimes called with a lock held (in the call `status(BagsStatus.Pending)` below), which I think should be examined closely because listeners are notified.  It is also sometimes called without the lock (for example in `status(BagsStatus.Running)`) which is problematic since the state is being updated.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -171,29 +193,39 @@ final class BagsManager(executorService: ExecutorService) {
...
 
+                    // Note that we have to sleep after reporting a bags failure, as otherwise, there is no time for
+                    // the failure status message to be read before we change back to pending.
+
                     // We don't want to print the stack trace if the host is simply unreachable.
                     // This is reported only as a GenericError in a CatalogException, unfortunately.
                     case Failure(CatalogException((e: GenericError) :: _)) =>
                       LOG.warning(s"$bagsIdMsg failed: ${e.msg}")
+                      status(BagsStatus.Failed("Catalog lookup failed."))
+                      Thread.sleep(3000)
```
cquiroz: Would it block the UI? When I see this, it worries me, it seems to be system dependent to sleep for x seconds waiting for something

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -216,6 +248,26 @@ final class BagsManager(executorService: ExecutorService) {
         case _                             => // Ignore
       }
   }
+
+  /** BAGS status changes. **/
+  def bagsStatus(key: SPNodeKey): Option[BagsStatus] =
+    synchronized(state.statuses.get(key))
+
+  /** Listeners for BAGS status changes. **/
+  private var listeners: List[BagsStatusListener] = Nil
```
cquiroz: If I remember correctly you can use `CopyOnWriteArrayList` to hold listeners without synchronization. Maybe @swalker2m has some experience using it?

sraaphorst:
Oh! I must have missed spinner16, as ajax-loader looks almost identical. (spinner was way too large.)
I'll try switching to spinner16.gif and drop ajax-loader.gif if all is well.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
sraaphorst: Ugh... again, in this case, we want to compare references and not actual content.

cquiroz: 👍  I comment these often as I always makes this mistake when going back to java

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
sraaphorst: Yup. Absolutely!

sraaphorst:
Never mind. Now I see why I used ajax-loader.gif instead: spinner16.gif is not a transparent gif, so it looks really quite ugly when rendered in a label with a background colour.

Unless there are serious objections, I think I'll stick with ajax-loader.gif for this reason.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -141,17 +142,37 @@ final class BagsManager(executorService: ExecutorService) {
...
+          })
+        }
+
+
+        // If dequeue is false this means that (a) another task scheduled *after* me ended up
+        // running before me, so their result is as good as mine would have been and we're done;
+        // or (b) we don't care about that program anymore, so we're done.
+        clearStatus()
+
+        if (dequeue(key, obs.getProgramID)) {
```
swalker2m: After moving `Future` down into this `if` block we're essentially enqueueing and dequeueing immediately in the same thread.  What motivated the change?

sraaphorst:
(The other option, of course, would be to replace spinner16.gif with ajax-loader.gif.)

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
sraaphorst: I wanted to set the `BagsStatus` to `Pending` only if the observation could be dequeued, and before the `Future` was executed. Does that make sense? I was a bit wary of this, but I wasn't sure a better way to do it offhand.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
sraaphorst: It doesn't block the UI, as it isn't happening on the Swing thread.
I'm not sure a better way to allow `BagsStatus.Failed` error messages to persist for a period of readability without a sleep call somewhere: otherwise, they will immediately be flipped to `BagsStatus.Pending` in anticipation of the next lookup.

swalker2m: Instead of sleeping perhaps the pending and running status could include the results of the last attempt.  Like  "Catalog lookup failed. Will retry."  "Catalog lookup failed.  Retrying ..." or something.

sraaphorst:
@cquiroz Yes, Swing allows animated gifs, fortunately!

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -171,29 +193,39 @@ final class BagsManager(executorService: ExecutorService) {
...
                       enqueue(obs, 5000L)
                   }
                 }
-              } else if (!eligibleForBags) {
-                LOG.info(s"${obs.getObservationID} not eligible for BAGS. Clearing auto group.")
-                applySwingResults(None)
               }
+            } else if (!eligibleForBags) {
+              LOG.info(s"${obs.getObservationID} not eligible for BAGS. Clearing auto group.")
+              applySwingResults(None)
```
swalker2m: I wonder if we need a status to report this.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
swalker2m: Since the `List` itself is immutable if we can get a reference to it we don't have to worry about it changing out from under us at least.  My concern was more with holding a lock while calling listeners.

It does seem like the add/remove/clear methods below should be synchronized though.  Do we need to support arbitrary lists of listeners? If there is only one interested party perhaps the `BagsManager` could just be created with a callback function to receive status updates.

swalker2m:
I vote to replace the `spinner16.gif` file with the `ajax-loader.gif` but keep the "spinner" name assuming the transparent background option looks okay (doesn't flash etc.).  I think the so-called transparent ones I tried in the past looked bad in general. If it looks poor on the standard grey background maybe just rename `ajax-loader.gif` to `spinner16-transparent.gif`?

swalker2m:
To summarize, a few ideas are:

* Perhaps we could keep the statuses grouped by program id and not worry about clearing them until we stop watching the program.

* I think I might pull the `status` method out of the `enqueue` method, rename it to `setStatus(key: SPNodeKey, status: BagsStatus)`, and synchronize it.  `clearStatus` might not be needed if un-watching a program just removes all the status values associated with it.

* If we can get away with just one "listener" that is never added and removed (but instead set in the constructor), it becomes a bit simpler.  

* Since this is all GUI code what if you just notify the listener(s) in the swing event loop?  Otherwise you have to worry about handling the event and immediately placing the handling in the event loop anyway.  

* Instead of switching from some failure state to a clean pending state as though nothing happened, if we remember what happened before we can report it appropriately and avoid the `sleep` and clearing of statuses.  It seems like we can also rearrange the `Future` forking back the way it was then.

sraaphorst:
Excellent. Thank you both @swalker2m and @cquiroz for the fantastic feedback: much appreciated.
I'll get on these changes and let you know when they're made.

sraaphorst:
Oh, and @swalker2m, yes, there is only one listener, I believe, for status changes, namely the one created in the `EdCompTargetList` constructor, so a `List[BagsStatusListener]` is probably overkill if my understanding is correct, unless it is possible for us to have more than one `EdCompTargetList` instantiated at a given time?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
cquiroz: I think you'd need a lock to change the `var listeners` but if there is a single listener then it gets much easier

swalker2m:
You're right, there can be multiple editors open. I was going to suggest that you use `scala.swing.Publisher` to keep up with listeners and publishing but that kind of assumes the client is Scala.  I'm not sure how hard it would be to add a reaction from Java but I'm guessing not fun.

We could assume that the registration / cleanup is all done from the EDT but I guess the thing to do is keep the list maintenance synchronized and when you publish, use `Swing.onEDT { immutableListenerListReference.foreach(...) }`.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -258,17 +258,29 @@ final class BagsManager(executorService: ExecutorService) {
   private var listeners: List[BagsStatusListener] = Nil
 
   def addBagsStatusListener(l: BagsStatusListener): Unit =
-    if (!listeners.contains(l))
-      listeners = l :: listeners
+    listeners.synchronized {
```
swalker2m: We shouldn't synchronize on the object that `listeners` references since it is immutable and the reference is going to change anyway.  Probably the lock associated with `BagsManager` itself is what you want.

````scala
synchronized {
....
}
````

File: bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala [hunk 10]
```diff
@@ -258,17 +258,29 @@ final class BagsManager(executorService: ExecutorService) {
...
   def clearBagsStatusListeners(): Unit =
-    listeners = Nil
+    listeners.synchronized {
+      listeners = Nil
+    }
 
-  private def notifyBagsStatusListeners(obs: ISPObservation, oldStatus: Option[BagsStatus], newStatus: Option[BagsStatus]): Unit =
-    listeners.foreach(l => Try { l.bagsStatusChanged(obs, oldStatus.asGeminiOpt, newStatus.asGeminiOpt) } )
+  private def notifyBagsStatusListeners(obs: ISPObservation, oldStatus: Option[BagsStatus], newStatus: Option[BagsStatus]): Unit = {
+    val immutableListenerList = listeners
```
swalker2m: Technically I believe you're supposed to grab a lock to get the reference to ensure that this thread's version is up-to-date.

````scala
val ls = synchronized { listeners }
````

Comment (sraaphorst) [inline bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala]: Thanks for all the feedback here, and of course, you are right regarding the immutability and thus this making no sense.

**Comment by sraaphorst** [comment]
I'm going to wait a few more minutes before I merge this as I know this code is a bit complicated, but I'd like to get it in for a build that I'll start this evening.

Again, much thanks to both @swalker2m and @cquiroz for the mountains of help.

File: bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala [hunk 11]
```diff
@@ -141,60 +177,87 @@ final class BagsManager(executorService: ExecutorService) {
 
         val key = obs.getNodeKey
         state += key
+        if (bagsStatus(key).isEmpty)
+          setStatus(key, BagsStatus.NewPending.some)
+
         Future {
+          // Wait the delay before running.
+          Thread.sleep(delay)
```
swalker2m: Could you explain a bit why the sleep is needed?  Usually `sleep` should not be required to ensure that something is working.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -141,60 +177,87 @@ final class BagsManager(executorService: ExecutorService) {
...
+                  //      running, so discard this result and do nothing,
+                  //      otherwise update the model
+                  //   }
+                  //   on failure enqueue again, maybe with a delay depending on the failure
+                  val bagsIdMsg = s"BAGS lookup on thread=${Thread.currentThread.getId} for observation=${obs.getObservationID}"
+                  LOG.info(s"Performing $bagsIdMsg.")
+
+                  val newStatus = bagsStatus(key).fold(BagsStatus.NewRunning)(_.toRunning)
+                  setStatus(key, newStatus.some)
+                  Thread.sleep(1000)
```
swalker2m: Here too.

sraaphorst:
We're going to get the same thing with the bags status object as we did with the listeners, I believe, so the synchronization will need to be fixed there.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
sraaphorst: This may not be the right place to do it, but I suspect that it should be in the `Future`?
We pass a `delay` to `enqueue` that was never being used to give, in some cases, the system a chance to recover. Do you have any thoughts about this?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
sraaphorst: This one can be removed, I think.
It was to prevent running to not be readable before the status was changed again, but I don't think that that's important. (It was more for debugging to ensure that all was well.)

sraaphorst:
So, to answer questions, from what I can tell, it seems like Swing can oddly handle animated PNGs, but animated GIFs? I don't think so.

We may just want to switch all spinners open to the new animated PNG that I was able to create with an online tool and that works with the feedback labels.

The other option would be to create animated GIFs with the appropriate background colours for the different labels (SALMON and BANANA or whatever they are) and use those.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
tpolecat: I thought the delay here was for retrying in case of failure (catalog server unreachable and so on).

sraaphorst:
Ugh... also forgot to take out some debug `println`s. Par for the course for me.
You'd think I'd know better by now.
I'll clean that up for the next commit, too.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
sraaphorst: Yes, but where should the delay be executed? That is the question.
It was being ignored, I think, prior to this point.

swalker2m:
> We're going to get the same thing with the bags status object as we did with the listeners, I believe, so the synchronization will need to be fixed there.

Also I'm thinking we should probably avoid holding multiple locks unless necessary.  Let's just use the one lock for updating anything in the `BagsManager` state?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -54,7 +53,40 @@ final class BagsManager(executorService: ExecutorService) {
...
+          }
+          notifyBagsStatusListeners(key, oldStatusOpt, statusOpt)
+        }
+      }
+    println(s"+++ setStatus done for $key")
+  }
+
+  private def clearStatusForProgram(prog: ISPProgram): Unit = {
+    bagsStatuses.synchronized {
+        prog.getObservations.asScala.foreach { obs =>
```
swalker2m: A small leak will occur here if an observation is deleted before `clearStatusForProgram`, since `prog.getObservations` won't include that observation anymore.  We could get around it in a couple of ways.  A `Map` of `Map`s:  `Map[SPProgramID, Map[SPNodeKey, BagsStatus]]` or maybe just make the key a tuple and filter on keys that contain the program id.  On the other hand, it probably doesn't matter that much.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -141,60 +177,87 @@ final class BagsManager(executorService: ExecutorService) {
...
+                        // was running, so discard the result.
+                        if (!state.keys(key)) {
+                          LOG.info(s"$bagsIdMsg successful. Results=${opt ? "Yes" | "No"}.")
+                          applySwingResults(opt)
+                          setStatus(key, None)
+                          updateHash(obs, hash)
+                        }
+
+                      // Note that we have to sleep after reporting a bags failure, as otherwise, there is no time for
+                      // the failure status message to be read before we change back to pending.
```
swalker2m: It seems like maybe the need to sleep in order to read the status message has been taken care of by not clearing the status on an error?  So perhaps the comment simply doesn't apply now?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
swalker2m: Thanks, understood.  This seems like the right place wait to me too.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -141,60 +177,87 @@ final class BagsManager(executorService: ExecutorService) {
...
+                    val fut = strategy.select(ctx, OT.getMagnitudeTable)
+                    fut onComplete {
+                      case Success(opt) =>
+                        // If this observation is once again in the queue, then something changed while this task
+                        // was running, so discard the result.
+                        if (!state.keys(key)) {
+                          LOG.info(s"$bagsIdMsg successful. Results=${opt ? "Yes" | "No"}.")
+                          applySwingResults(opt)
+                          setStatus(key, None)
+                          updateHash(obs, hash)
```
swalker2m: Since the map update for the hash values is happening on different threads, it'll need to be synchronized as well.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
cquiroz: Sorry, why do we need the `delay`? You mention a chance to recover, recover from what?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -141,60 +177,87 @@ final class BagsManager(executorService: ExecutorService) {
...
+                        }
+
+                      // Note that we have to sleep after reporting a bags failure, as otherwise, there is no time for
+                      // the failure status message to be read before we change back to pending.
+
+                      // We don't want to print the stack trace if the host is simply unreachable.
+                      // This is reported only as a GenericError in a CatalogException, unfortunately.
+                      case Failure(CatalogException((e: GenericError) :: _)) =>
```
cquiroz: We could enhance the catalog's error exception with a specific `HostUnreachableException` if  needed

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -141,60 +177,87 @@ final class BagsManager(executorService: ExecutorService) {
...
+          setStatus(key, BagsStatus.NewPending.some)
+
         Future {
+          // Wait the delay before running.
+          Thread.sleep(delay)
+
           // If dequeue is false this means that (a) another task scheduled *after* me ended up
           // running before me, so their result is as good as mine would have been and we're done;
           // or (b) we don't care about that program anymore, so we're done.
           if (dequeue(key, obs.getProgramID)) {
```
cquiroz: I can't find where `enqueue/dequeue` are defined. where are they?

sraaphorst: Huh... that is strange. They are both methods in `BagsManager`.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
sraaphorst: I think it would be a good idea in the future (no pun intended) to enhance catalog exceptions, but I don't think it is something pressing that needs to be done sooner rather than later.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
sraaphorst: Yes, you are right 😄

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
sraaphorst: If, say, a catalog lookup fails due to lack of internet connectivity or for some other reason, we don't want the request to be immediately re-enqueued and then continue to hammer away and fail, so we want a delay before we re-execute and allow the net connection to come back up, the catalog server to recover, etc.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
cquiroz: Found them, they were hidden in the UI

sraaphorst: I was trying to find a way to highlight them and flag you, but you beat me to it 😄

sraaphorst:
BTW @cquiroz - don't worry too much about this PR right now. @swalker2m and I are going to make some changes this afternoon on it.

swalker2m:
I'm going to go ahead and close this one out.  Given the `BagsManager` refactoring this will need to change quite a bit, but should be simplified.  The new states in `BagsManager` basically correspond to the feedback we need to give so it's mostly a matter of exposing them to the UI at this point.

PR #904, opened 2016-04-22
OCSADV-383: bug in horizons start/stop time formatting

swalker2m:
This tiny PR fixes a bug in `HorizonsService2`.  When it formats the start/stop time arguments it was using the local time zone instead of UTC.  I fixed that and also switched to `DateTimeFormatter` since it is immutable and thread-safe and only one instance is needed.


cquiroz:
👏

tpolecat:
Oops. Also 👍

PR #907, opened 2016-04-26
REL-2745 add scheduling block editor at observation level

tpolecat:
This PR hacks a copy of the parametric angle time doodad into the top-level observation editor. It's not great but it provides needed functionality for today's test release.

image

Also fixed a P2 warning about missing scheduling blocks that didn't actually check to see if there was a scheduling block or not.

### DIFF
--- a/bundle/edu.gemini.p2checker/src/main/scala/edu/gemini/p2checker/target/NonSiderealTargetRules.scala
+++ b/bundle/edu.gemini.p2checker/src/main/scala/edu/gemini/p2checker/target/NonSiderealTargetRules.scala
@@ -38,12 +38,14 @@ final class NonSiderealTargetRules extends IRule {
 
   def checkNoSchedulingBlock(es: ObservationElements): IP2Problems =
     new P2Problems <| { p2p =>
-      for {
-        toc <- es.getTargetObsComp.asScalaOpt
-               if toc.getTargetEnvironment.getTargets.asScalaList.exists(_.isNonSidereal)
-      } p2p.addWarning(ERR_NO_SCHEDULING_BLOCK,
-          s"Observation ${es.getObservationNode.getObservationID} has nonsidereal targets but no scheduling block.",
+      es.getTargetObsComp.asScalaOpt.map { toc =>
+        if (toc.getTargetEnvironment.getTargets.asScalaList.exists(_.isNonSidereal) &&
+            es.getSchedulingBlock.isEmpty) {
+        p2p.addWarning(ERR_NO_SCHEDULING_BLOCK,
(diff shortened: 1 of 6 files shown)


bundle/edu.gemini.p2checker/src/main/scala/edu/gemini/p2checker/target/NonSiderealTargetRules.scala
```diff
@@ -38,12 +38,14 @@ final class NonSiderealTargetRules extends IRule {
 
   def checkNoSchedulingBlock(es: ObservationElements): IP2Problems =
     new P2Problems <| { p2p =>
-      for {
-        toc <- es.getTargetObsComp.asScalaOpt
-               if toc.getTargetEnvironment.getTargets.asScalaList.exists(_.isNonSidereal)
-      } p2p.addWarning(ERR_NO_SCHEDULING_BLOCK,
-          s"Observation ${es.getObservationNode.getObservationID} has nonsidereal targets but no scheduling block.",
+      es.getTargetObsComp.asScalaOpt.map { toc =>
```
swalker2m: It seems like `foreach` is wanted here?

tpolecat: Yep, thanks.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/parallacticangle/ParallacticAngleControls.scala
```diff
@@ -194,7 +197,6 @@ class ParallacticAngleControls extends GridBagPanel with Publisher {
     for {
       e              <- editor
       ispObservation <- Option(e.getContextObservation)
-      if ObsTargetCalculatorService.targetCalculation(ispObservation).isDefined
```
swalker2m: From what I can tell, this filter would keep anything from happening if we could not determine a `Site`, scheduling block, or target component.  Scheduling block is included below but now it looks like we'll continue anyway even if there is no known `Site` or target component?  It looks like that might be okay since the `parallacticAngle` will be `None` in that case and it will display "Target not visible" which seems reasonable.  Is that the reasoning behind this?

swalker2m:
Look good to me. 👍

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/parallacticangle/ParallacticAngleControls.scala
tpolecat: In the cases where we are doing the PA calculation the site is known, but in the general case where we're just editing the scheduling block it may not be. So this filter was preventing things from working. This is all very bad and I'm not helping matters much, but I don't have the spiritual energy to untangle this.

PR #909, opened 2016-04-26
fix infinite loop

tpolecat:
Just spotted this. Search/replace error at some point.


swalker2m:
👍

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/obs/SPObservation.java
```diff
@@ -323,7 +323,10 @@ public void setPhase2Status(ObsPhase2Status newValue) {
      * Get the scheduling block's starting time.
      */
     public Option<Long> getSchedulingBlockStart() {
-        return getSchedulingBlockStart();
+        return _schedulingBlock.flatMap(b -> {
+            final scala.Option<Object> osb = b.duration(); // boxing
+            return (osb.isDefined()) ? new Some((Long) osb.get()) : None.instance();
```
sraaphorst: `ImOption.fromScalaOpt(SchedulingBlock::duration).map(Long.box)`?

PR #911, opened 2016-04-28
REL-2750 force refresh tpe on scheduling block update

tpolecat:
This forces the TPE to update in response to scheduling block changes, which live on the observation.


swalker2m:
👍

PR #912, opened 2016-04-28
REL-2746-A add semester to obs context

tpolecat:
REL-2746 requires the target editor to know the program semester (if any) so it needs to be in the `ObsContext`. This is sufficiently noisy to warrant its own PR.

I could have done this with an overload and default arg but `ObsContext` doesn't do this elsewhere and I didn't want to introduce a new pattern. So this PR is mostly busywork.


tpolecat:
Turns out we don't need to do this.

PR #914, opened 2016-04-28
REL-2746 pad ephemeris by 1 month

tpolecat:
This changes target editor ephemeris lookup in three ways:
- It always fetches the ephemeris for the semester in which the scheduling block falls.
- It fetches an extra month before and after the semester, as requested.
- It fetches 1200 points instead of 1000, to somewhat make up for the longer interval.


swalker2m:
Looks good. 👍

PR #916, opened 2016-04-29
REL-2754 Add ephemeris refresh button

tpolecat:
This adds a button to refresh the ephemeris; it's enabled when the Horizons designation and site are known.

image


tpolecat:
You may want to add `?w=1` to the diff because most of the delta is just a change in indentation.

swalker2m:
👍

PR #917, opened 2016-05-02
REL-2754 fix "unused" bug

tpolecat:
This fixes a problem where the target would get renamed to `unused` on ephemeris refresh. Also replaces an instance variable with a method.


cquiroz:
Looks good to me :+1:

PR #919, opened 2016-05-02
REL-2599: Fixed F2 long slit geometry offsetting

sraaphorst:
Previously, the old F2 geometry feature was offsetting the geometry of long slits by `LongSlitFOVNorthPos` in a sort of hacky way that was missed when we extracted the geometry into its own class. This accounts for the offsetting properly so that the geometry of F2 long slits is now the same as it was in 2015B.

See the JIRA task for graphics showing how things looked in 2015B and how things incorrectly looked in 2016A:
http://jira.gemini.edu:8080/browse/REL-2599


cquiroz:
Looks good to me :+1:

PR #920, opened 2016-05-02
REL-2747 REL-2757 fix scheduling dialog

tpolecat:
This fixes several problems with the parallactic angle / scheduling controls and dialog:
- The relative time menu now correctly sets the start time, not the duration.
- When used in "scheduling" mode in the observation editor only the start date and time are shown; duration is not mentioned.
- It was formerly not possible to set the duration back to "remaining time", nor was it possible to set a duration of zero.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/obs/SchedulingBlock.scala
+++ b/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/obs/SchedulingBlock.scala
@@ -30,7 +30,7 @@ object SchedulingBlock {
     apply(start, Some(duration))
 
   def apply(start: Long, duration: Option[Long]): SchedulingBlock =
-    new Impl(start, duration.filter(_ > 0))
+    new Impl(start, duration.filter(_ >= 0))
 
   def unsafeFromStrings(startString: String, durationString: String): SchedulingBlock =
     apply(startString.toLong, durationString.toLong)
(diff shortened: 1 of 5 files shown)


bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/obs/SchedulingBlock.scala
```diff
@@ -30,9 +30,12 @@ object SchedulingBlock {
     apply(start, Some(duration))
 
   def apply(start: Long, duration: Option[Long]): SchedulingBlock =
-    new Impl(start, duration.filter(_ > 0))
+    new Impl(start, duration.filter(_ >= 0))
 
   def unsafeFromStrings(startString: String, durationString: String): SchedulingBlock =
     apply(startString.toLong, durationString.toLong)
 
+  def unsafeFromStrings(startString: String): SchedulingBlock =
```
cquiroz: Why are these called unsafe? Because of the `.toLong` call?

cquiroz:
Looks good to me :+1:

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/obs/SchedulingBlock.scala
swalker2m: There's no guarantee that the `String` contains a value that can be parsed as a `Long`.  I think these are used from PIO (XML) serialization though so it will all work out unless someone edits the XML.

tpolecat: Yeah it's because they throw; a proper `.fromString` would return an `Option` which might be a better approach, but it's coming from PIO as Shane mentioned and would just end up throwing on `None` anyway.

PR #922, opened 2016-05-03
REL-2662: Add support for new Altair dichroic to OT

sraaphorst:
A new Altair dichroic (wavelength) has been added, and support is required for the OT / WDBA.

This PR adds this new wavelength, and additionally, changes the names of the enum members of the old `AltairParams.Wavelength` to more accurately represent the reality of the current situation, since `WAVELENGTH_A` and `WAVELENGTH_B` no longer have the same meaning.

Additionally, `To2016B` was updated to account for these enum name changes.
I have tested it with programs with Altair components exported from 2015B and 2016A, and the import appears to be working correctly.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/altair/AltairConstants.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/altair/AltairConstants.java
@@ -1,9 +1,3 @@
-// Copyright 1997-2000
-// Association for Universities for Research in Astronomy, Inc.,
-// Observatory Control System, Gemini Telescopes Project.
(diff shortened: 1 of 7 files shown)


cquiroz:
I think it would be good to have a test for the conversion of legacy proposals if time permits

sraaphorst:
@cquiroz You mean in the general case and not just for this, correct?

cquiroz:
We have a bunch of migration tests on the package `edu.gemini.spModel.io.impl.migration`. This migration should probably be there. Again if time permits, if we need this in a hurry it seems fine to me to merge and do the tests later on

sraaphorst:
I'm going to merge it for the build, then, but I will add tests after I've got the build started and create another PR for them.

PR #923, opened 2016-05-04
REL-2731 add notes for conic target migration

tpolecat:
This updates the `To2016B` migration logic to add notes for imported conic targets:

```
Starting with 2016B, the Observing Tool supports all nonsidereal targets with ephemerides
automatically downloaded from JPL HORIZONS.  This observation used the following orbital
elements which have been migrated to a single-point ephemeris.

     Name: Pluto
    Epoch: 2457217.5
       IN: 17.36609399010031
       OM: 110.2100007229519
        W: 114.2248220688449
        A: 39.74409337717218
       EC: 0.2543036816945946
       MA: 0.0

Re-resolve the target to set the unique target identifier and to update to the most recent
ephemeris.
```

I'm ignoring targets in template groups because they targets coming from the PIT won't have orbital elements. I'm not sure if this is the right approach.

### DIFF
--- a/bundle/edu.gemini.spModel.io/src/main/scala/edu/gemini/spModel/io/impl/migration/to2016B/To2016B.scala
+++ b/bundle/edu.gemini.spModel.io/src/main/scala/edu/gemini/spModel/io/impl/migration/to2016B/To2016B.scala
@@ -219,7 +219,7 @@ object To2016B extends Migration {
       s"""
          |Starting with 2016B, the Observing Tool supports all nonsidereal targets with ephemerides
          |automatically downloaded from JPL HORIZONS.  This observation used the following orbital
-         |elements for which have been migrated to a single-point ephemeris.
+         |elements which have been migrated to a single-point ephemeris.
          |
          |     Name: $name
          |    Epoch: $epoch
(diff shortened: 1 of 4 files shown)


bundle/edu.gemini.spModel.io/src/main/scala/edu/gemini/spModel/io/impl/migration/to2016B/To2016B.scala
```diff
@@ -193,6 +202,39 @@ object To2016B extends Migration {
       case _         => sys.error("Unknown named target: " + name)
     })
 
+  // Construct a note describing the old conic target
+  def conicTargetNote(ps: ParamSet): Option[(String, String)] =
+    for {
```
cquiroz: It seems the note will only be produced if all the params are present. Is that always the case? Could/should support the case when some of the params are missing?

tpolecat: These will all be present for old conic targets.

cquiroz:
Looks good to me :+1:

PR #924, opened 2016-05-05
REL-2707: Set locale to en_US

cquiroz:
This PR adds explicit flags to all the applications setting the locale to `en_US`. I tested this by changing my locale to something else and after the launch spdb started with `en_US`

We could also set the locale using `Locale.setDefault` on the launcher but this option seems less intrusive


sraaphorst:
Thank you. About time this gets  done by someone and great on you for taking the initiative! 👍

tpolecat:
Great, thanks! 👍

PR #925, opened 2016-05-05
REL-2323B: Altar additional strategies

sraaphorst:
Partial (minimal) migration from previous REL-2323 PR to allow for WDBA changes as per REL-2518.

Now that the coding is complete, I am updating this description:

The main detail added by this PR is the inclusion of the `MultiProbeStrategy` class, which allows combining different AGS strategies together to allow them to select in concert, which is required by REL-2323 (which adds Altair + GMOS OI, and GMOS OI + PWFS1).

The challenge in this lies in the position angle constraint, which makes this non-trivial and not straightforward. For example, if the position angle constraint is FIXED + 180, if we naively run `select` on each substrategy, then it is possible that one strategy will select at the PA, and the other strategy will select at the PA + 180, which is clearly invalid.

Thus, in order to avoid having to touch the other AGS strategies, what I did was hack the `ObsContext` when the position angle constraint is not FIXED: I create a list of `ObsContext` with all possible position angles allowed by the position angle constraint, and make sure that the position angle constraint for each one is FIXED so that selecting on the `ObsContext` forces the substrategies to use the same PA.

We then mess around with the `Futures` to get the appropriate format, and rank each valid selection using a score that corresponds to the way targets are chosen for `SingleProbeStrategy` and pick the best.

Note that the position angle constraint is used in very few places in the code, so I did not have to use this technique in methods like `estimate`, where it is never examined.

### DIFF
--- a/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/MultiProbeStrategy.scala
+++ b/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/MultiProbeStrategy.scala
@@ -50,7 +50,7 @@ sealed case class MultiProbeStrategy(key: AgsStrategyKey, strategies: List[AgsSt
     def ctxToFixedList: List[ObsContext] = {
       val ctxFixed = {
         val instFixed = ctx.getInstrument match {
-          case pac : PosAngleConstraintAware =>
+          case pac : SPInstObsComp with PosAngleConstraintAware =>
             val newPac = pac.clone.asInstanceOf[SPInstObsComp with PosAngleConstraintAware]
             newPac.setPosAngleConstraint(PosAngleConstraint.FIXED)
             newPac
--- a/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/MultiProbeStrategy.scala
+++ b/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/MultiProbeStrategy.scala
@@ -95,14 +95,12 @@ sealed case class MultiProbeStrategy(key: AgsStrategyKey, strategies: List[AgsSt
         case AgsStrategy.Assignment(gp: SingleProbeStrategy.VProbe, t) => gp.calculator(cCtx).calc(t.coordinates)
       }.sum
 
-      // We can use default value of 0.0 as the guide probes for which targets with magnitudes exist should be
-      // consistent across all selections.
       val brightestMag = sel.assignments.collect {
         case AgsStrategy.Assignment(_, t) if qualitiesByTarget.get(t).exists(_ === bestQuality) =>
           probeBands.extract(t).map(_.value)
       }.collect {
         case Some(v) => v
-      }.minimum.getOrElse(0.0)
+      }.minimum.getOrElse(Double.MaxValue)
 
       (qualityIdx, vignettingIdx, bestQuality, brightestMag)
     }
--- a/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/MultiProbeStrategy.scala
+++ b/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/MultiProbeStrategy.scala
@@ -17,13 +17,10 @@ import scala.concurrent.Future
 import scalaz._
 import Scalaz._
 
-
 /**
- * Represents an AGS strategy consisting of multiple substrategies.
- *
- * Currently, this is only used to represent combinations of SingleProbeStrategies.
+ * Represents an AGS strategy consisting of multiple substrategies acting together.
  */
-class MultiProbeStrategy(val key: AgsStrategyKey, val strategies: List[AgsStrategy]) extends AgsStrategy {
+sealed case class MultiProbeStrategy(key: AgsStrategyKey, strategies: List[AgsStrategy]) extends AgsStrategy {
 
   override def magnitudes(ctx: ObsContext, mt: MagnitudeTable): List[(GuideProbe, AgsMagnitude.MagnitudeCalc)] =
     strategies.flatMap(_.magnitudes(ctx, mt))
@@ -71,13 +68,12 @@ class MultiProbeStrategy(val key: AgsStrategyKey, val strategies: List[AgsStrate
     // We have a List[ObsContext], and we want to get a Future[Option[AgsStrategy.Selection]]
     // We can take the List[ObsContext] to a Future[List[Option[AgsStrategy.Selection]]].
     // Then we simply filter out the empty Options, and of the remaining, fold to get the best selection.
-    // We MAY want to make a comparator for AgsStrategy.Selection. I'm just not sure the best way to do this.
-    // Assign values to GuideQualities, add together, pick the maximum, and then break ties by vignetting?
     val noStrategy: Option[AgsStrategy.Selection] = None
     def ctxSelect(cCtx: ObsContext): Future[(ObsContext, Option[AgsStrategy.Selection])] = {
-      val lOfF = strategies.map(_.select(cCtx, mt).map(opt => (ctx, opt)))
-      Future.fold(lOfF)(ctx -> noStrategy) { case ((_, resOpt), (_, curOpt)) =>
-        cCtx -> curOpt.map(cur => AgsStrategy.Selection(cur.posAngle, resOpt.fold(cur.assignments)(cur.assignments ++ _.assignments))).orElse(resOpt)
+      val lOfF = strategies.map(_.select(cCtx, mt))
+      Future.fold(lOfF)(cCtx -> noStrategy) {
+        case ((_, resOpt), curOpt) =>
+          cCtx -> curOpt.map(cur => AgsStrategy.Selection(cur.posAngle, resOpt.fold(cur.assignments)(cur.assignments ++ _.assignments))).orElse(resOpt)
       }
     }
 
@@ -132,5 +128,7 @@ class MultiProbeStrategy(val key: AgsStrategyKey, val strategies: List[AgsStrate
     strategies.flatMap(_.guideProbes).distinct
 }
 
-case object GmosNorthOiwfsAltair extends MultiProbeStrategy(AgsStrategyKey.GmosNorthAltairOiwfsKey, List(Strategy.GmosNorthOiwfs, Strategy.AltairAowfs))
-case object GmosNorthOiwfsPwfs1 extends MultiProbeStrategy(AgsStrategyKey.GmosNorthPwfs1OiwfsKey, List(Strategy.GmosNorthOiwfs, Strategy.Pwfs1North))
+object MultiProbeStrategy {
+  val GmosNorthOiwfsAltair = MultiProbeStrategy(AgsStrategyKey.GmosNorthAltairOiwfsKey, List(Strategy.GmosNorthOiwfs, Strategy.AltairAowfs))
+  val GmosNorthOiwfsPwfs1  = MultiProbeStrategy(AgsStrategyKey.GmosNorthPwfs1OiwfsKey, List(Strategy.GmosNorthOiwfs, Strategy.Pwfs1North))
+}
--- a/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/MultiProbeStrategy.scala
+++ b/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/MultiProbeStrategy.scala
@@ -55,7 +55,7 @@ class MultiProbeStrategy(val key: AgsStrategyKey, paStrategy: AgsStrategy, paSym
   val probeBands = RBandsList
 
   override val guideProbes: List[GuideProbe] =
-    strategies.flatMap(_.guideProbes)
+    strategies.flatMap(_.guideProbes).distinct
 }
 
 case object GmosNorthOiwfsAltair extends MultiProbeStrategy(AgsStrategyKey.GmosNorthAltairOiwfsKey, Strategy.GmosNorthOiwfs, List(Strategy.AltairAowfs))
--- a/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/MultiProbeStrategy.scala
+++ b/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/MultiProbeStrategy.scala
@@ -70,7 +70,7 @@ sealed case class MultiProbeStrategy(key: AgsStrategyKey, strategies: List[AgsSt
     def ctxSelect(cCtx: ObsContext): Future[(ObsContext, Option[AgsStrategy.Selection])] =
       Future.fold(strategies.map(_.select(cCtx, mt)))(cCtx -> noStrategy) {
         case ((_, resOpt), curOpt) =>
-          cCtx -> curOpt.map(cur => AgsStrategy.Selection(cur.posAngle, resOpt.fold(cur.assignments)(cur.assignments ++ _.assignments))).orElse(resOpt)
+          cCtx -> curOpt.map(cur => AgsStrategy.Selection(cur.posAngle, resOpt.fold(cur.assignments)(cur.assignments ++ _.assignments)))
       }
 
     // We need a way to compare AGS selections to determine which is the best across the list of generated
--- a/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/MultiProbeStrategy.scala
+++ b/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/MultiProbeStrategy.scala
@@ -6,57 +6,57 @@ import edu.gemini.ags.api.{AgsAnalysis, AgsMagnitude, AgsStrategy}
 import edu.gemini.catalog.api.CatalogQuery
 import edu.gemini.spModel.ags.AgsStrategyKey
 import edu.gemini.spModel.core.{RBandsList, SiderealTarget}
-import edu.gemini.spModel.guide.{ValidatableGuideProbe, GuideProbe}
+import edu.gemini.spModel.guide.{GuideProbe, ValidatableGuideProbe}
 import edu.gemini.spModel.obs.context.ObsContext
 
 import scala.concurrent.Future
 
-import scala.concurrent.ExecutionContext.Implicits.global
+import scalaz._
+import Scalaz._
+
 
 /**
  * Represents an AGS strategy consisting of multiple substrategies.
  *
  * Currently, this is only used to represent combinations of SingleProbeStrategies.
- *
- * Since select returns a position angle, this allows us to use the position angle of the paStrategy and ignore
- * the position angles of the (fully or approximately) rotation-symmetric (around the base pos) strategies returned
- * by select.
  */
-class MultiProbeStrategy(val key: AgsStrategyKey, paStrategy: AgsStrategy, paSymmetricStrategies: List[AgsStrategy]) extends AgsStrategy {
-  val strategies = paStrategy :: paSymmetricStrategies
+case class MultiProbeStrategy(key: AgsStrategyKey, strategies: NonEmptyList[AgsStrategy]) extends AgsStrategy {
 
   override def magnitudes(ctx: ObsContext, mt: MagnitudeTable): List[(GuideProbe, AgsMagnitude.MagnitudeCalc)] =
-    strategies.flatMap(_.magnitudes(ctx, mt))
+    strategies.toList.flatMap(_.magnitudes(ctx, mt))
 
   override def analyze(ctx: ObsContext, mt: MagnitudeTable): List[AgsAnalysis] =
-    strategies.flatMap(_.analyze(ctx, mt))
+    strategies.toList.flatMap(_.analyze(ctx, mt))
 
   override def analyze(ctx: ObsContext, mt: MagnitudeTable, guideProbe: ValidatableGuideProbe, guideStar: SiderealTarget): Option[AgsAnalysis] =
     AgsAnalysis.analysis(ctx, mt, guideProbe, guideStar, probeBands)
 
   override def catalogQueries(ctx: ObsContext, mt: MagnitudeTable): List[CatalogQuery] =
-    strategies.flatMap(_.catalogQueries(ctx, mt))
+    strategies.toList.flatMap(_.catalogQueries(ctx, mt))
 
   override def candidates(ctx: ObsContext, mt: MagnitudeTable): Future[List[(GuideProbe, List[SiderealTarget])]] =
-    Future.traverse(strategies)(_.candidates(ctx, mt)).map(_.flatten)
+    Future.traverse(strategies.toList)(_.candidates(ctx, mt)).map(_.flatten)
 
   override def estimate(ctx: ObsContext, mt: MagnitudeTable): Future[AgsStrategy.Estimate] =
-    Future.fold(strategies.map(_.estimate(ctx, mt)))(1.0)((p,est) => p * est.probability).map(Estimate(_))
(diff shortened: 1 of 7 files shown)


bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/general/GeneralRule.java
```diff
@@ -124,25 +124,39 @@ public IP2Problems check(ObservationElements elements)  {
...
         //private static final String WFS_OUT_PATROL_FIELD_TEMPLATE = "%s outside of the patrol field";
 
+        private String formatGuiderString(final Iterator<GuideProbe> iter) {
+            if (iter == null || !iter.hasNext())
+                return "";
+
+            final StringBuilder buf = new StringBuilder();
+            buf.append(iter.next().getKey());
+            while (iter.hasNext()) {
+                buf.append(", ").append(iter.next().getKey());
```
cquiroz: I really miss `mkString` 😢

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/altair/AltairParams.java
```diff
@@ -244,24 +246,24 @@ static public GuideStarType getGuideStarType(String name, GuideStarType value) {
      */
     public enum Mode implements DisplayableSpType, SequenceableSpType {
 
-        NGS     ("NGS",     GuideStarType.NGS, FieldLens.OUT, AltairAowfsGuider.instance),
-        NGS_FL  ("NGS+FL",  GuideStarType.NGS, FieldLens.IN,  AltairAowfsGuider.instance),
-        LGS     ("LGS",     GuideStarType.LGS, FieldLens.IN,  AltairAowfsGuider.instance),
-        LGS_P1  ("LGS+P1",  GuideStarType.LGS, FieldLens.IN,  PwfsGuideProbe.pwfs1),
-        LGS_OI  ("LGS+OI",  GuideStarType.LGS, FieldLens.IN,  GmosOiwfsGuideProbe.instance)
+        NGS     ("NGS",     GuideStarType.NGS, FieldLens.OUT, DefaultImList.create(AltairAowfsGuider.instance)),
+        NGS_FL  ("NGS+FL",  GuideStarType.NGS, FieldLens.IN,  DefaultImList.create(AltairAowfsGuider.instance)),
```
cquiroz: Should'n be there an NGS+OI mode according to the task?

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/altair/AltairParams.java
```diff
@@ -244,24 +246,24 @@ static public GuideStarType getGuideStarType(String name, GuideStarType value) {
...
+        LGS     ("LGS",     GuideStarType.LGS, FieldLens.IN,  DefaultImList.create(AltairAowfsGuider.instance, GmosOiwfsGuideProbe.instance)),
+        LGS_P1  ("LGS+P1",  GuideStarType.LGS, FieldLens.IN,  DefaultImList.create(PwfsGuideProbe.pwfs1, GmosOiwfsGuideProbe.instance)),
+        LGS_OI  ("LGS+OI",  GuideStarType.LGS, FieldLens.IN,  DefaultImList.create(GmosOiwfsGuideProbe.instance))
```
cquiroz: And `LGS+P1+OI`?

sraaphorst:
Okay, I have added the AGS elements. Right now, I believe, the only issue that needs to be done (which is fairly complex) is `MultiProbeStrategy`.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/MultiProbeStrategy.scala
```diff
@@ -0,0 +1,62 @@
...
+    Future.fold(strategies.map(_.select(ctx, mt)))(None: Option[AgsStrategy.Selection])((resOpt,curOpt) =>
+      curOpt.map(cur => AgsStrategy.Selection(cur.posAngle, resOpt.fold(cur.assignments)(cur.assignments ++ _.assignments))).orElse(resOpt)
+    )
+  }
+
+  // Default is R bands.
+  val probeBands = RBandsList
+
+  override val guideProbes: List[GuideProbe] =
+    strategies.flatMap(_.guideProbes)
```
cquiroz: Maybe this require a `distinct` in case you have the same probe in multiple strategies?

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/MultiProbeStrategy.scala
```diff
@@ -0,0 +1,62 @@
...
+  // As described in the comment for the class, we pick the position angle for the FIRST strategy, since it is the
+  // only one that can have a patrol field that is not (at least approximately) symmetric under rotation around the base.
+  override def select(ctx: ObsContext, mt: MagnitudeTable): Future[Option[AgsStrategy.Selection]] = {
+    Future.fold(strategies.map(_.select(ctx, mt)))(None: Option[AgsStrategy.Selection])((resOpt,curOpt) =>
+      curOpt.map(cur => AgsStrategy.Selection(cur.posAngle, resOpt.fold(cur.assignments)(cur.assignments ++ _.assignments))).orElse(resOpt)
+    )
+  }
+
+  // Default is R bands.
+  val probeBands = RBandsList
```
cquiroz: I'd think the probeBands should be taken from the strategies. It maybe far fetched but if you had, gmos strategy and niri strategy, you should include `RList` and `K`

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/MultiProbeStrategy.scala
```diff
@@ -0,0 +1,62 @@
...
+    Future.traverse(strategies)(_.candidates(ctx, mt)).map(_.flatten)
+
+  override def estimate(ctx: ObsContext, mt: MagnitudeTable): Future[AgsStrategy.Estimate] =
+    Future.fold(strategies.map(_.estimate(ctx, mt)))(1.0)((p,est) => p * est.probability).map(Estimate(_))
+
+  // As described in the comment for the class, we pick the position angle for the FIRST strategy, since it is the
+  // only one that can have a patrol field that is not (at least approximately) symmetric under rotation around the base.
+  override def select(ctx: ObsContext, mt: MagnitudeTable): Future[Option[AgsStrategy.Selection]] = {
+    Future.fold(strategies.map(_.select(ctx, mt)))(None: Option[AgsStrategy.Selection])((resOpt,curOpt) =>
+      curOpt.map(cur => AgsStrategy.Selection(cur.posAngle, resOpt.fold(cur.assignments)(cur.assignments ++ _.assignments))).orElse(resOpt)
```
cquiroz: Would `cur.posAngle` be called for each `select` result and you'd endup using the last one?

sraaphorst: Right now, this isn't going to work. This class needs a total rewrite, essentially.
I want to use the same ideas, but done in a fashion that isn't horribly broken :-).

cquiroz:
As a general comment I think this change requires a good amount of tests. Are those in the pipeline?

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/MultiProbeStrategy.scala
sraaphorst: I wasn't sure how to best handle this, with the probe bands list being a sealed trait...

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/MultiProbeStrategy.scala
sraaphorst: Yes, this is a very good idea.

sraaphorst:
I think that once we have an implementation of `MultiProbeStrategy` that we expect works, we will definitely need some thorough testing. Since this is the last of the June tasks for me, I would imagine that we would be able to fit it into the schedule.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/altair/AltairParams.java
```diff
@@ -244,24 +246,24 @@ static public GuideStarType getGuideStarType(String name, GuideStarType value) {
      */
     public enum Mode implements DisplayableSpType, SequenceableSpType {
```
cquiroz: My take on REL-2323 is that there are new modes, I think this list should be expanded. Maybe we don't need to let the user select them on the UI and they are purely to be made available via AGS

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/Strategy.scala
```diff
@@ -84,9 +86,9 @@ object Strategy {
       val ao = ctx.getAOComponent.asScalaOpt
       if (ao.exists(_.isInstanceOf[InstAltair])) {
         ao.get.asInstanceOf[InstAltair].getMode match {
-          case AltairParams.Mode.LGS_P1 => List(Pwfs1North)
+          case AltairParams.Mode.LGS_P1 => List(Pwfs1North, GmosNorthOiwfsPwfs1)
```
cquiroz: So is not possible to have LGS+P1 it has to always be LGS+P1+OI?

andrewwstephens: It *is* possible to use LGS+P1.  The LGS + PWFS1 options are:
GMOS + LGS using PWFS1 for TTF without flexure compensation (AGS selects PWFS1 only)
GMOS + LGS using PWFS1 for TTF and OIWFS for flexure compensation (AGS selects PWFS1 + OIWFS)

cquiroz:
Given the amount of code and so many commits with `WIP` in the title, should we rewrite the history to make the commits text more descriptive?

sraaphorst:
Alright, `MultiProbeStrategy` has been implemented and tested manually and looks to be working properly, so if someone could take a look at this class and offer any feedback, that would be great.

I will look at creating actual test cases today.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/Strategy.scala
sraaphorst: The inclusion of `Pwfs1North` in the list here should handle that case, no?

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/altair/AltairParams.java
sraaphorst: As per our discussion, these aren't exactly new modes in the same sense as these modes, and changing this enum would cause issues, so the list of `AgsStrategy` has been updated instead.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/altair/AltairParams.java
sraaphorst: =grins=
I think this has been addressed through our HipChat talks, about how these values are actually a subtractive process and not as you would intuitively think.

cquiroz: Is this the anti-guiders thing? The subtractive nature of this definition seems very unnatural to me. If so, at least the `_guiders` field should be renamed to `_antiGuiders` or something

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/MultiProbeStrategy.scala
```diff
@@ -0,0 +1,128 @@
...
+    // Turn into a Future[List[(ObsContext, Option[AgsStrategy.Selection])]], collect the results that have
+    // a defined selection while calculating their value, and then return the one (if one exists) of minimum value.
+    Future.sequence(ctxToFixedList.map(ctxSelect))
+      .map(_.collect { case (cCtx, Some(sel)) => (cCtx, sel, selValue(cCtx, sel)) }.minimumBy(_._3).map(_._2))
+  }
+
+
+  // From the left, take the band that is the superset of the most bands.
+  // This should but may not contain all bands, as BandsList is a sealed trait.
+  override lazy val probeBands = strategies.map(_.probeBands).reduceLeft((sup,cur) => {
```
cquiroz: I don't think this covers the (highly hypothetical) case where you have RBands + K. If necessary we can expand BandsList

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/MultiProbeStrategy.scala
```diff
@@ -0,0 +1,128 @@
...
+      }
+
+    // We need a way to compare AGS selections to determine which is the best across the list of generated
+    // ObsContexts. We assign a 4-tuple to each selection in such a way that the minimum of the 4-tuples represents
+    // the best selection.
+    // 1. Quality index is the quality converted to an int, and then summed over all targets.
+    // 2. Vignetting index is vignetting for each vignetting guide probe, summed across all.
+    // 3. Best quality across all targets. Used because for magnitude, we only want best quality class in comparison.
+    // 4. Magnitude is magnitude of the brightest star in the assignment in the best quality class.
+    def selValue(cCtx: ObsContext, sel: AgsStrategy.Selection): (Int, Double, Int, Double) = {
```
cquiroz: Perhaps instead of a 4-tuple you can make a small case class that would be more explicative

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/MultiProbeStrategy.scala
```diff
@@ -0,0 +1,128 @@
...
+  override def analyze(ctx: ObsContext, mt: MagnitudeTable): List[AgsAnalysis] =
+    strategies.flatMap(_.analyze(ctx, mt))
+
+  override def analyze(ctx: ObsContext, mt: MagnitudeTable, guideProbe: ValidatableGuideProbe, guideStar: SiderealTarget): Option[AgsAnalysis] =
+    AgsAnalysis.analysis(ctx, mt, guideProbe, guideStar, probeBands)
+
+  override def catalogQueries(ctx: ObsContext, mt: MagnitudeTable): List[CatalogQuery] =
+    strategies.flatMap(_.catalogQueries(ctx, mt))
+
+  override def candidates(ctx: ObsContext, mt: MagnitudeTable): Future[List[(GuideProbe, List[SiderealTarget])]] =
```
cquiroz: I think this can return the same guide probe multiple times. Can the upstream handle that case, or should we add the targets for the same guide probe?

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/MultiProbeStrategy.scala
```diff
@@ -0,0 +1,128 @@
...
+/**
+ * Represents an AGS strategy consisting of multiple substrategies acting together.
+ */
+sealed case class MultiProbeStrategy(key: AgsStrategyKey, strategies: List[AgsStrategy]) extends AgsStrategy {
+
+  override def magnitudes(ctx: ObsContext, mt: MagnitudeTable): List[(GuideProbe, AgsMagnitude.MagnitudeCalc)] =
+    strategies.flatMap(_.magnitudes(ctx, mt))
+
+  override def analyze(ctx: ObsContext, mt: MagnitudeTable): List[AgsAnalysis] =
+    strategies.flatMap(_.analyze(ctx, mt))
```
swalker2m: In the analysis case, the position angle is considered to be fixed so we don't have the same issue as with selection?

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/MultiProbeStrategy.scala
swalker2m: Totally not suggesting that we make this change right before the release, but maybe we should have made this return a `Future[Map[GuideProbe, List[SiderealTarget]]]` instead.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/MultiProbeStrategy.scala
```diff
@@ -0,0 +1,128 @@
...
+    AgsAnalysis.analysis(ctx, mt, guideProbe, guideStar, probeBands)
+
+  override def catalogQueries(ctx: ObsContext, mt: MagnitudeTable): List[CatalogQuery] =
+    strategies.flatMap(_.catalogQueries(ctx, mt))
+
+  override def candidates(ctx: ObsContext, mt: MagnitudeTable): Future[List[(GuideProbe, List[SiderealTarget])]] =
+    Future.traverse(strategies)(_.candidates(ctx, mt)).map(_.flatten)
+
+  override def estimate(ctx: ObsContext, mt: MagnitudeTable): Future[AgsStrategy.Estimate] =
+    Future.fold(strategies.map(_.estimate(ctx, mt)))(1.0)((p,est) => p * est.probability).map(Estimate(_))
```
swalker2m: Position angle constraint is a factor in estimation so this implementation is not kosher I think.  For example, in `SingleProbeStrategy.estimate`:

````scala
    val pac   = ctx.getPosAngleConstraint(UNBOUNDED)
    val cv    = CandidateValidator(params, mt, candidates)
    val steps = pac.steps(ctx.getPositionAngle, params.stepSize.toOldModel).toList.asScala
    val anglesWithResults  = steps.filter { angle => cv.exists(ctx.withPositionAngle(angle)) }
    val successProbability = anglesWithResults.size.toDouble / steps.size.toDouble
    AgsStrategy.Estimate.toEstimate(successProbability)
````

So it seems like you _do_ have the issue you identified already where one strategy can only find a guide star at one particular angle and another only at a different position angle.  The solution you came up with for the select case seems reasonable.

File: bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/MultiProbeStrategy.scala [hunk 14]
```diff
@@ -0,0 +1,128 @@
...
+            newPac
+          case oth : SPInstObsComp => oth
+        }
+        ctx.withInstrument(instFixed)
+      }
+      (ctx.getPosAngleConstraint match {
+        case PosAngleConstraint.FIXED                                            => List(0)
+        case PosAngleConstraint.FIXED_180 | PosAngleConstraint.PARALLACTIC_ANGLE => List(0, 180)
+        case PosAngleConstraint.UNBOUNDED                                        => Range(0, 360, 15).toList
+      }).map(a => ctxFixed.withPositionAngle(ctx.getPositionAngle.add(a, Angle.Unit.DEGREES)))
```
swalker2m: How did you come up with 15º increments for the `UNBOUNDED` case?  I notice that `SingleProbeStrategy` has a default and minimum of 10º (see `SingleProbeStrategyParams.stepSize`) but it can vary.  I don't think we have any way in general to go from `AgsStrategy` to an appropriate step size (in order to, say, pick the minimum of all the strategies) so maybe for now we should make the increment 10º and add a task to generalize this correctly in the future.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/MultiProbeStrategy.scala
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    // For each ObsContext, perform the selection and return it as a future.
+    val noStrategy: Option[AgsStrategy.Selection] = None
+    def ctxSelect(cCtx: ObsContext): Future[(ObsContext, Option[AgsStrategy.Selection])] =
+      Future.fold(strategies.map(_.select(cCtx, mt)))(cCtx -> noStrategy) {
+        case ((_, resOpt), curOpt) =>
+          cCtx -> curOpt.map(cur => AgsStrategy.Selection(cur.posAngle, resOpt.fold(cur.assignments)(cur.assignments ++ _.assignments)))
+      }
```
swalker2m: Hmm, are we sure that two or more probes can use the same guide star?  Nothing in this algorithm prohibits that?  I would guess that using a star for GMOS OIWFS would keep it from being usable as a PWFS or Altair star.  And even if not the same star, being in the vignetted area of one probe arm would keep it from being usable by another.  I worry that all of this is too complicated to introduce to the codebase so late in the cycle.

cquiroz: That is a very good observation. I'd think the probes will interfere with each other if using the same star. Perhaps in the particular case of oiwfs and altair it doesn't happen?

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/MultiProbeStrategy.scala
```diff
@@ -0,0 +1,128 @@
...
+
+      val brightestMag = sel.assignments.collect {
+        case AgsStrategy.Assignment(_, t) if qualitiesByTarget.get(t).exists(_ === bestQuality) =>
+          probeBands.extract(t).map(_.value)
+      }.collect {
+        case Some(v) => v
+      }.minimum.getOrElse(Double.MaxValue)
+
+      (qualityIdx, vignettingIdx, bestQuality, brightestMag)
+    }
```
swalker2m: Were there requirements for how to pick the best option of multiple options?

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/altair/AltairParams.java
swalker2m: Never mind, I think I see what is going on here.  We're listing all the guide probes that, if otherwise present in an `ObsContext`, are usable for the given mode.  The `GuideProbeConsumer` then takes that into consideration when deciding which guide probes to remove.

sraaphorst: Yes, exactly :-).

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/MultiProbeStrategy.scala
sraaphorst: I think that this would be a good idea. I'm not sure how we want to do it, though.
Why is this a sealed trait anyways?

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/MultiProbeStrategy.scala
sraaphorst: Sure, I like that idea much better.

sraaphorst: Although comparing the four tuple is much easier... I guess we will have to implement a small ordering if we use a case class.

swalker2m:
I'm going to close this one out, since the multi-probe strategy turned out to be more complicated than we anticipated.  Before we can do it, we'll need PWFS guide probe vignetting in the model at the very least.

PR #926, opened 2016-05-05
REL-2749 change 2016B import to discard existing scheduling blocks

tpolecat:
This is a trivial change that was really hard to figure out because it intersected with the Canada issue and a UI bug (REL-2762) that causes all the ephemeris data to disappear if the window is too small.

The underlying problem is that if you have an observation with an existing scheduling block (which will be the case for executed observations that use average parallactic angle) then the scheduling block's start and the target's valid-at date will probably not coincide. This means the imported target's 1-point ephemeris won't be sufficient and you end up with no "default" coordinates. So the solution is to discard any existing scheduling block and always make it align with the old valid-at date. Andy agrees this is ok.


cquiroz:
Looks good to me :+1:

PR #927, opened 2016-05-05
OCSADV-383-I: RA/Dec formatting updates

swalker2m:
This PR introduces a few minor RA/Dec formatting updates to help with TCS parsing.

* zero pads single digit hour/degree values
* adds a + sign to positive declinations
* removes a space from the whitespace separating the julian date from the RA


sraaphorst:
Seems straightforward enough and fine. 👍

jluhrs:
The original files have non negative values padded with a space, but a + will do. (wait, what about 0? will it be +00 ?

PR #929, opened 2016-05-05
REL-2700 improve renaming logic in horizons lookup

tpolecat:
Currently the Horizons lookup code updates the target name with the common name of the target, which isn't always desirable. This changes the logic as follows:

- If the search term is a substring of the common name, it is expanded to the full common name. So `hal` would be expanded `Hale-Bopp`.
- If the search term is a substring of the primary designation (normally this will be an exact match) the primary designation is retained; `67p` would become `67P`.
- If neither case holds (should be impossible but with HORIZONS you never know) the user-provided name is unchanged.
- The name is not updated when refreshing the ephemeris.


cquiroz:
Looks good to me :+1:

PR #930, opened 2016-05-06
REL-2315: IFU Expanded Description

spakzad:
Added a short note under the "Analysis Method for Integral Field Unit (IFU) spectroscopy" section of the GMOS-N and GMOS-S ITC webpages.  The format of the note and the corresponding link was copied from other notes and links on the same page.


cquiroz:
Seems fine to me, I'll merge

PR #931, opened 2016-05-06
REL-2751 disable target node rename

tpolecat:
One of the hugely irritating things about the POT model is that nodes have arbitrary settable names that in many cases should be computed, and the computed name ends up being set as a side-effect. So we end up in the position of looking at the name and trying to decide whether it had been computed and should thus be re-computed if properties change; or was set by hand and should thus be left alone.

So in 16B the target naming scheme has changed to reflect the new model, which makes all the old target names appear to have been set by hand via the "Edit Item Title…" menu item in the OT. This means if you change the target the label in the tree view in the OT doesn't change. Possible fixes are:

- Encode a representation of the old naming scheme forever in OCS so we can figure out whether the name was computed in a prior version, or
- Just disable custom names for target environments (since nobody actually does that) and compute them every time.

As you may have guessed, I chose option B. Any thoughts?

### DIFF
--- a/bundle/jsky.app.ot/src/main/java/jsky/app/ot/viewer/action/EditItemTitleAction.java
+++ b/bundle/jsky.app.ot/src/main/java/jsky/app/ot/viewer/action/EditItemTitleAction.java
@@ -2,6 +2,7 @@ package jsky.app.ot.viewer.action;
 
 import edu.gemini.pot.sp.ISPNode;
 import edu.gemini.spModel.data.ISPDataObject;
+import edu.gemini.spModel.target.obsComp.TargetObsComp;
 import jsky.app.ot.viewer.SPViewer;
 import jsky.util.gui.DialogUtil;
 
@@ -41,6 +42,10 @@ public final class EditItemTitleAction extends AbstractViewerAction {
 
     @Override
     public boolean computeEnabledState() throws Exception {
-        return isEditableContext() && getContextNode(ISPNode.class) != null;
+        if (isEditableContext()) {
+            ISPNode n = getContextNode(ISPNode.class);
+            return !(n.getDataObject() instanceof TargetObsComp);
+        }
+        return false;
     }
 }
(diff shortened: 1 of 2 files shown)


bundle/jsky.app.ot/src/main/java/jsky/app/ot/viewer/action/EditItemTitleAction.java
```diff
@@ -41,6 +42,10 @@ public void actionPerformed(ActionEvent evt) {
 
     @Override
     public boolean computeEnabledState() throws Exception {
-        return isEditableContext() && getContextNode(ISPNode.class) != null;
+        if (isEditableContext()) {
+            ISPNode n = getContextNode(ISPNode.class);
```
cquiroz: On the previous code there was a check for null on `n`. Why don't we need it anymore?

tpolecat: Yep, good catch. Should never happen but I'll add a check.

tpolecat:
Fixed, thx.

cquiroz:
I don't really have a strong comment about the options you presented. If nobody uses custom names, we should probably remove that option, and keeping old data on the OCS is not very appealing

Perhaps @andrewwstephens should weigh in?

Codewise this seems fine 👍

tpolecat:
Yeah I mean, you can rename any node in the tree but I would be surprised if anyone ever used that feature. You can name your Observing Log but why would you?

cquiroz:
feature creep? What happens with existing programs with customized names?

tpolecat:
If anyone actually renamed their target environment that name will be lost and it will be rendered as the target name. Other custom names are unaffected.

tpolecat:
I'm going to merge this now because I want to get it in for the test build. Can revert if it's too controversial,

andrewwstephens:
I think this will be fine.  I looked through the current targets, and while some were customized, it was usually just to remove the "Target:" label before the target name.

PR #932, opened 2016-05-07
REL-491: GMOS ITC should discourage spatial binning for IFU modes

spakzad:
The GMOS team requested a more detailed explanation for why binning in the spatial direction by either 2 or 4 is not recommended.  Changing the warning message required recreating the baseline.txt file.  I tested it in IntelliJ and both BaselineAllSpec and BaselineTest.checkAll passed.


cquiroz:
Seems fine to me, I'll merge right away

PR #933, opened 2016-05-09
REL-2765: Made changes for ephemeris output.

sraaphorst:
Previously, the WDBA XML generated a single ephemeris filename per observation if the science target was nonsidereal. This should have been done per target, with the filename associated with each nonsidereal target instead of with the observation root node.

This fixes this so that the ephemeris parameter is associated with each target if applicable.

I am not certain that we want to commit this until the change for nonsidereal targets is made throughout the OCS: as things currently stand, everything is simply done for the science target, such as the fetching of ephemeris files. I am still submitting this PR for consideration and for review by @swalker2m

### DIFF
--- a/bundle/edu.gemini.wdba.xmlrpc.server/src/test/scala/edu/gemini/wdba/tcc/TcsConfigTest.scala
+++ b/bundle/edu.gemini.wdba.xmlrpc.server/src/test/scala/edu/gemini/wdba/tcc/TcsConfigTest.scala
@@ -21,14 +21,6 @@ final class TcsConfigTest extends TestBase {
   }
 
   @Test
-  def testNonSiderealWithHorizonsDesignation(): Unit = {
-    val titan = HorizonsDesignation.MajorBody(606)
-    setBase(NonSiderealTarget("Titan", ==>>.empty, Some(titan), List.empty, None, None))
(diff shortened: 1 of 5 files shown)


bundle/edu.gemini.wdba.xmlrpc.server/src/test/scala/edu/gemini/wdba/tcc/TcsConfigTest.scala
```diff
@@ -21,14 +21,6 @@ final class TcsConfigTest extends TestBase {
   }
 
   @Test
-  def testNonSiderealWithHorizonsDesignation(): Unit = {
```
cquiroz: Is this test not needed anymore? It sounds like this change sounds should be accompanied by a test(s) case(s)

sraaphorst: I agree that a test case is absolutely needed, but I think we should make a new test file for this with regards to WDBA writing of targets specifically.

I'm thinking something like a `TargetConfigTest`, but I would like to hear back from @swalker2m before I invest any time in this.

jluhrs:
:thumbsup::thumbsup::thumbsup:

PR #934, opened 2016-05-09
REL-2730: NullPointerException opening GPI programs

cquiroz:
This PR fixes some cases where GPI programs produce an NPE while doing p2 checks. This happens because for NONSTANDARD obs mode the bright limits were not set to any value. This is fixed on 	b9bea2e, the other commit are generic code cosmetic improvements


sraaphorst:
This looks fine. 👍

PR #936, opened 2016-05-09
OCSADV-383: add non-sidereal guide targets

swalker2m:
The functor that finds non-sidereal targets in order to gather and write ephemeris files was only considering the science target.  It turns out that non-sidereal targets can be used as guide stars as well, so this PR updates the functor to include any target in an observation's target environment.  Accordingly, it also renames `NonSiderealObservation` to `NonSiderealTargetRef`.

Unfortunately this may not be the best code I've ever written, since I'm doing it in a hurry in order to make the next test release...


bundle/edu.gemini.spdb.reports.collection/src/main/scala/edu/gemini/dbTools/ephemeris/NonSiderealTargetRef.scala
```diff
@@ -8,20 +8,20 @@ import edu.gemini.spModel.target.env.TargetEnvironment
 import edu.gemini.spModel.target.obsComp.TargetObsComp
 import edu.gemini.spModel.util.SPTreeUtil
 
-/** A reference to an individual non-sidereal observation. */
-case class NonSiderealObservation(oid: SPObservationID, hid: HorizonsDesignation, targetName: String)
+/** A reference to an individual non-sidereal target reference. */
+case class NonSiderealTargetRef(oid: SPObservationID, hid: HorizonsDesignation, targetName: String)
```
cquiroz: Why is it renamed?

swalker2m: Before it would find just the science target for the observation and be done.  Now it will look at all the targets, base position + guide targets and include all that are non-sidereal.  Non-sidereal target reference seemed more appropriate for that since each one no longer corresponds to a single observation.

sraaphorst:
I didn't examine the test case in great detail, but the logic in the code seems to be correct to me. 👍

jluhrs:
:+1:

cquiroz:
👍

jluhrs:
What about the directory where the files will be written ?

sraaphorst:
@jluhrs Aren't these a setting in the spdb build.sbt file? Maybe I'm wrong.

swalker2m:
> What about the directory where the files will be written ?

@jluhrs Sebastian is right, it's a property you set in the spdb `build.sbt` file.  Do we know what it should be yet?

jluhrs:
/gemsoft/var/ephemerides

andrewwstephens:
FYI, the mount on hbfcon03 is
/gemsoft/var/data/ephemerides
which points to
pisces:/export/gemini/files/ephemerides

PR #937, opened 2016-05-09
REL-2769 add guide ephemerides to tpe

tpolecat:
If you're guiding on a non-sidereal target it is now displayed on the TPE in gray (here guiding on Titan while observing Dione).

image

@swalker2m the guide environment remains a mystery to me (even after the big refactor) so there may be a better way to implement `getGuideEphemerides`.


cquiroz:
Looks good to me :+1:

sraaphorst:
Looks good to me. 👍

swalker2m:
It seems like it should work.  You're using the old model which was re-implemented in terms of the new model to do this.  That may be fine though.  If you want to get to the underlying `GuideEnv` and go from there it might be clearer what is going on but it seems like this should work. 👍

jluhrs:
That looks so cool.

PR #939, opened 2016-05-09
REL-2767 show setup time for nonsidereal targets in TPE

tpolecat:
This does three unrelated things that happened to appear in the same issue:
- It adds the setup time to the path shown in the TPE.
- It adds a "Now +5 min" option to the sched block and PA controls.
- It removes the "Now + Setup" and "Now + Reacq" options from the sched block controls.


cquiroz:
Looks good to me :+1:

PR #940, opened 2016-05-11
REL-1204: Modify NIRI + Altair max guide star distance

spakzad:
Originally, Altair.java checked that the guide star separation was <= 25" for all instruments.
However, with NIRI, the guide star separation can be <=45" when guiding off-axis.  The check in Altair.java has been removed and is now implemented separately for each of the instruments:
 -Gnirs.java and Nifs.java (limit is <= 25")
 -Niri.java (limit is <= 45")
 -Gmos.java (limit is <= 45", but has been commented out since GMOS cannot yet use Altair).


cquiroz:
Looks good to me :+1:

PR #942, opened 2016-05-13
REL-2776: TPE image lookup fails mysteriously

cquiroz:
At certain targets on the sky, Gemini's image catalogue fails, It was decided that this won't be fixed at the catalogue level, but instead we produce a better error message

The HTTP query to load the image has an OK result so we need to inspect the body to tell that it fails

Here's the error message

screenshot 2016-05-13 11 19 18


swalker2m:
👍

sraaphorst:
Much better. 👍

PR #943, opened 2016-05-13
OCSADV-383-K

swalker2m:
A small update to ephemeris formatting to meet TCS expectations.

* Adds a small header which is expected and parsed by the TCS to obtain the coordinate reference frame.

* Adds a space to positive RAs and replaces the declination `+` with a space so that they line up in the expected columns.


cquiroz:
Looks good to me :+1:

sraaphorst:
The build failed, it appears. Provided that that wasn't a fluke or something to do with a test case requiring updating, 👍

PR #944, opened 2016-05-13
REL-2771 include higher-res data around scheduling block start

tpolecat:
This PR updates the ephemeris fetching logic to fetch fewer data points across the whole semester, and then add extra points for the observing night in which the schedule time falls, in order to allow accurate guidestar selection for rapidly moving/accelerating targets. A somewhat contrived example here, for Enceladus.

image

This also adds a P2 error if the base position is more than 10 arcminutes from the nearest ephemeris point, and removes the "resolution" value from the ephemeris editor since it's non-uniform now.


cquiroz:
Looks good to me, nice screenshot  :+1:

PR #945, opened 2016-05-16
REL-2701: Fix UH affiliate support

cquiroz:
Small fix, when importing UH proposals the partner affiliation was lost because there is no iso code for the University of Hawaii

I suppose the reason this is not present is because UH is not an iso code but I don't see downsides to this change


tpolecat:
Curious. I wonder if we should rename it from ISO code to something else, just to avoid confusion.

In any case 👍

cquiroz:
That probably makes sense, something like code is probably better

PR #946, opened 2016-05-16
REL-2762 fix magical disappearing ephemeris

tpolecat:
After literally hours of doing things at random it turns out that setting a minimum size for the labels was causing them to disappear when the window was sized down.


cquiroz:
Sorry you had to suffer that, Looks good 👍

PR #949, opened 2016-05-17
REL-2748 attempted fix for target detail update issue

tpolecat:
Before this change the target details were initialized as a side-effect of the initial selection in the target list, but this doesn't always happen for some reason. In particular if you change the scheduling block and click back to the target while the little time/date button still has focus then usually the target details would not be re-initialized. So this change adds an explicit update in the editor's `init` method to force the update.


cquiroz:
Looks good to me :+1:

PR #950, opened 2016-05-18
OCSADV-383: fix ephemeris file directory

swalker2m:
Sets the ephemeris file directory for testing and production.  Will these point to different places on g[ns]odb vs g[ns]odbtest2?


cquiroz:
Looks good, I'll merge as soon as Travis gives me the OK

PR #952, opened 2016-05-18
REL-2334: Fix OT Position Editor color map and intensity options

cquiroz:
This PR adds missing LUT files for the colormaps on the TPE and includes a few small code fixes


tpolecat:
👍

PR #954, opened 2016-05-18
REL-2781: Allow selecting any target in executed observations

cquiroz:
Once an obs is 'Observed' you cannot change it so everything is disabled. As a side effect you cannot select items on the targets' table

This PR disables the enabling of the table


tpolecat:
You can change primary guider by double-clicking rows in the table ... is this correctly disabled?

cquiroz:
Yes, the rows have a different mechanisms for enabling/disabling
I guess this change should be tested quite a bit

tpolecat:
Ok. Well 👍  for the code.

PR #955, opened 2016-05-19
OCSADV-383: fix ephemeris file header for TCS parsing

swalker2m:
The ephemeris file header comment must align with the columns expected by the TCS.  The header I added in the previous PR was not properly aligned and so the TCS couldn't parse the reference frame.  This fixes that issue.

The only remaining inconsistency with horizons output is the RA/Dec tracking rate where the robot is using a fixed precision and horizons changes the precision in order to make the values always occupy the same number of characters in the text file.  According to @jluhrs, the TCS is able to handle this particular inconsistency so I'll leave it as is.


cquiroz:
Looks good to me :+1:

PR #956, opened 2016-05-19
REL-2334-B: Fix OT Position Editor color map and intensity options

cquiroz:
The previous PR fixed the color maps, this one fixes the intensity maps


swalker2m:
I bet this has been broken since we made the OT an OSGi app. 👍

cquiroz:
Exactly, the files come from the svn ocs1.5 repo

arturog8m:
The fact that nobody has noticed this before really makes me think that the right solution to this problem is to remove this feature altogether....

tpolecat:
I am always in favor of removing features, so 👍  to that.

PR #958, opened 2016-05-20
REL-2768: Require a time when converting from arbitrary SPTarget to SiderealTarget

swalker2m:
This PR addresses a bug with using non-sidereal targets as guide stars.  In particular the AGS analysis of a guide star works only with `SiderealTarget` so we convert arbitrary guide star targets to `SiderealTarget` if necessary before performing the analysis.  Unfortunately the conversion process would lose the coordinates of non-sidereal targets at the scheduling block start and always use (0,0) coordinates.  The analysis then reported that the guide star is out of reach.

To fix the conversion we need to know the coordinates at the scheduling block start time, which requires adding that time as a parameter.  This parameter then bubbles up a bit through a few APIs and in particular in the ITC, which accounts for most of the changes.

I also renamed the `SPTarget => SiderealTarget` conversion method from `toNewModel` to `toSiderealTarget` to better reflect what it does.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/scala/edu/gemini/pot/ModelConverters.scala
+++ b/bundle/edu.gemini.pot/src/main/scala/edu/gemini/pot/ModelConverters.scala
@@ -88,10 +88,17 @@ object ModelConverters {
       sp.getTarget.fold(
         too => SiderealTarget.empty.copy(name = too.name),
         sid => sid,
-        ns  => {
-          val coords = when.flatMap(ns.coords).getOrElse(Coordinates.zero)
-          SiderealTarget.empty.copy(name = ns.name, coordinates = coords, magnitudes = ns.magnitudes, spectralDistribution = ns.spectralDistribution, spatialProfile = ns.spatialProfile)
-        }
+        ns  =>
+          SiderealTarget(
+            name                 = ns.name,
+            coordinates          = when.flatMap(ns.coords).getOrElse(Coordinates.zero),
+            properMotion         = None,
+            redshift             = None,
+            parallax             = None,
+            magnitudes           = ns.magnitudes,
+            spectralDistribution = ns.spectralDistribution,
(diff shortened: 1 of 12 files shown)


bundle/edu.gemini.pot/src/main/scala/edu/gemini/pot/ModelConverters.scala
```diff
@@ -78,12 +81,17 @@ object ModelConverters {
...
+       toSiderealTarget(when.asScalaOpt.map(_.longValue))
+
+    def toSiderealTarget(when: Option[Long]): SiderealTarget =
       sp.getTarget.fold(
         too => SiderealTarget.empty.copy(name = too.name),
         sid => sid,
-        ns  => SiderealTarget.empty.copy(name = ns.name, magnitudes = ns.magnitudes, spectralDistribution = ns.spectralDistribution, spatialProfile = ns.spatialProfile)
+        ns  => {
+          val coords = when.flatMap(ns.coords).getOrElse(Coordinates.zero)
+          SiderealTarget.empty.copy(name = ns.name, coordinates = coords, magnitudes = ns.magnitudes, spectralDistribution = ns.spectralDistribution, spatialProfile = ns.spatialProfile)
```
cquiroz: At some moment we discussed not to use `empty.copy` in this particular case. The rationale was that if we change `SideralTarget`, e.g adding a new field, this conversion will be forgotten, by using the explicit construct wo should avoid that. What do you think?

swalker2m: Right, that makes sense to me.  I'll update it.

cquiroz:
When I did this I never considered using a non-sidereal target as a GS, didn't know it was possible. looks good to me 👍

tpolecat:
Thanks a bunch for chasing this down. 👍

PR #959, opened 2016-05-20
REL-2760: Fix target and guide star creation from request

swalker2m:
This PR addresses an issue with handling URL ToO requests.  In particular, we were trying to update ToO targets with information in the request.  Since they are ToO targets, you cannot set the RA/Dec and magnitude information. Rather than updating the target and guide environment, the change is to create simple new `SiderealTarget`s from the information in the request and make a new guide environment set up for manual guiding.


cquiroz:
What's a URL ToO?

swalker2m:
It's a feature that allows users to trigger ToO observations with an HTTP request.

cquiroz:
Looks good to me :+1:

PR #960, opened 2016-05-20
REL-2790: Multiple guide probes with GMOS-N + Altair

sraaphorst:
This implements the possibility of guiding using multiple probes with GMOS-N + Altair.
The multiple probe combinations, depending on the Altair mode, are listed in REL-2323.
The result is that the target editor now allows guide stars of the type listed in REL-2323 to be added.
Note that there is no AGS strategy for this yet, as we are missing some details to allow us to implement that.


bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/Strategy.scala
```diff
@@ -86,7 +86,7 @@ object Strategy {
         ao.get.asInstanceOf[InstAltair].getMode match {
           case AltairParams.Mode.LGS_P1 => List(Pwfs1North)
           case AltairParams.Mode.LGS_OI => List(GmosNorthOiwfs)
-          case _                        => List(AltairAowfs, Pwfs1North, GmosNorthOiwfs)
```
swalker2m: Shouldn't we put `Pwfs1North` and `GmosNorthOiwfs` back in the list?

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/altair/InstAltair.java
```diff
@@ -362,16 +357,16 @@ public ISysConfig getSysConfig() {
...
         guiders.addAll(GuideProbeMap.instance.values());
-        guiders.remove(guideProbe);
+        guideProbes.foreach(guiders::remove);
         return guiders;
     }
 
     public Collection<GuideProbe> getConsumedGuideProbes() {
         if (getGuideStarType() == GuideStarType.LGS) {
-            return allMinus(getMode().guider());
+            return allMinus(getMode().guiders());
```
cquiroz: It is not for this PR to fix, but this is quite counter-intuitive to me

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/altair/AltairParams.java
```diff
@@ -244,24 +240,24 @@ static public GuideStarType getGuideStarType(String name, GuideStarType value) {
...
-        NGS     ("NGS",     GuideStarType.NGS, FieldLens.OUT, AltairAowfsGuider.instance),
-        NGS_FL  ("NGS+FL",  GuideStarType.NGS, FieldLens.IN,  AltairAowfsGuider.instance),
-        LGS     ("LGS",     GuideStarType.LGS, FieldLens.IN,  AltairAowfsGuider.instance),
-        LGS_P1  ("LGS+P1",  GuideStarType.LGS, FieldLens.IN,  PwfsGuideProbe.pwfs1),
-        LGS_OI  ("LGS+OI",  GuideStarType.LGS, FieldLens.IN,  GmosOiwfsGuideProbe.instance)
+        NGS     ("NGS",     GuideStarType.NGS, FieldLens.OUT, DefaultImList.create(AltairAowfsGuider.instance)),
+        NGS_FL  ("NGS+FL",  GuideStarType.NGS, FieldLens.IN,  DefaultImList.create(AltairAowfsGuider.instance)),
+        LGS     ("LGS",     GuideStarType.LGS, FieldLens.IN,  DefaultImList.create(AltairAowfsGuider.instance, GmosOiwfsGuideProbe.instance)),
+        LGS_P1  ("LGS+P1",  GuideStarType.LGS, FieldLens.IN,  DefaultImList.create(PwfsGuideProbe.pwfs1, GmosOiwfsGuideProbe.instance)),
+        LGS_OI  ("LGS+OI",  GuideStarType.LGS, FieldLens.IN,  DefaultImList.create(GmosOiwfsGuideProbe.instance))
```
swalker2m: I know I'm out of sync on these changes, but I'm curious why OI would be available in LGS or LGS+P1 mode.  Also, are these modes limited to GMOS?  What happens when using Altair with another instrument?

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/altair/InstAltair.java
sraaphorst: Agreed. This is incredibly confusing and took me awhile to figure out what was happening / needed to be done.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/altair/AltairParams.java
sraaphorst: As far as I know, these modes are indeed limited to GMOS as per REL-2323.
I have no idea about the guider combinations and why these were chosen to be permitted.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/Strategy.scala
sraaphorst: By the info in REL-2323, I don't think that that is what we want. For all combinations other than those listed, we want only Altair and Altair + GMOS OI. Since we don't want GMOS OI by itself, we can't include it in this list; however, the probe handling changes I made to `AltairParams.Mode` takes care of this and provides the option to add the appropriate guide stars.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/altair/AltairParams.java
sraaphorst: Oh, as for using Altair with other instruments, since the only things changed in this transition to lists was the addition of `GmosOiwfsGuideProbe`, it has no effect on other instruments. I have tested this with NIRI, at least, and the probe choices / AGS strategy choices are the same between this code and the 2016A release.

swalker2m: Okay thanks, sounds good. 👍

PR #961, opened 2016-05-20
REL-2785 - WIP on auto ephemeris update

tpolecat:
This PR updates the OT as follows:
- When the user moves the scheduling block into another observing night all ephemerides in the observation are updated.
- Refactored the existing ephemeris lookup a bit.
- Added logging for Horizons errors.
- Fixed a dumb bug in `SPTarget` that was returning the scheduling block duration rather than its start.


tpolecat:
Ok this is ready for review now.

swalker2m:
Looks great. 👍

PR #962, opened 2016-05-22
REL-2518: new pointing origin and instrument types for GMOS-N + Altair

swalker2m:
This PR updates WDBA GMOS support to add new pointing origin and instrument types for GMOS-N + Altair.  The changes are relatively simple despite the size of the diff.  The meat of the change is in `GMOSSupport` in the `getTccConfigInstrument` and `getTccConfigInstrumentOrigin` methods.  For the most part the rest is code cleanup and updating test cases.  It also renames `GuideConfigSouth` to just `GuideConfig` since in reality it is used for both sites.

The WDBA is arguably the worst remaining code in the OCS, though the sequence model vies for that title.  This PR doesn't do much to improve the situation unfortunately. It does however introduce an important change to the way that we determine whether a given guide probe is "used" in a target environment.  Instead of just having associated targets for a guide probe, there now must be an _active_ target for the probe.  In addition to being necessary to implement REL-2518 correctly, this fixes REL-2789 in passing.

### DIFF
--- a/bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/GMOSSupport.java
+++ b/bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/GMOSSupport.java
@@ -6,7 +6,7 @@ import edu.gemini.spModel.gemini.gmos.InstGmosNorth;
 import edu.gemini.spModel.target.obsComp.PwfsGuideProbe;
 import edu.gemini.spModel.telescope.IssPort;
 
-public class GMOSSupport implements ITccInstrumentSupport {
+public final class GMOSSupport implements ITccInstrumentSupport {
 
     private String _wavelength;
     private ObservationEnvironment _oe;
@@ -56,11 +56,17 @@ public class GMOSSupport implements ITccInstrumentSupport {
     }
 
     public String getTccConfigInstrument() {
-        String side = _oe.isNorth() ? "5" : "3";
-        String port = (((InstGmosCommon) _oe.getInstrument()).getIssPort() == IssPort.UP_LOOKING) ? "" : side;
-        String ao   = (_oe.isNorth() && _oe.isAltair()) ? "AO2" : "";
-        String p2   = (_oe.containsTargets(PwfsGuideProbe.pwfs2)) ? "_P2" : "";
-        return ao + "GMOS" + port + p2;
+        final String side = _oe.isNorth() ? "5" : "3";
+        final String port = (((InstGmosCommon) _oe.getInstrument()).getIssPort() == IssPort.UP_LOOKING) ? "" : side;
+
+        if (_oe.isNorth() && _oe.isAltair()) {
+            final String oi = _oe.containsTargets(GmosOiwfsGuideProbe.instance) ? "_OI" : "";
+            final String p1 = _oe.containsTargets(PwfsGuideProbe.pwfs1) ? "_P1" : "";
+            return "AO2GMOS" + port + p1 + oi;
+        } else {
+            final String p2 = (_oe.containsTargets(PwfsGuideProbe.pwfs2)) ? "_P2" : "";
+            return "GMOS" + port + p2;
+        }
     }
 
     /**
@@ -78,7 +84,7 @@ public class GMOSSupport implements ITccInstrumentSupport {
      * @return String that is the name of a TCC config file.  See WDBA-5.
      */
     public String getTccConfigInstrumentOrigin() {
-        InstGmosCommon<?, ?, ?, ?> inst = (InstGmosCommon) _oe.getInstrument();
+        final InstGmosCommon<?, ?, ?, ?> inst = (InstGmosCommon) _oe.getInstrument();
         return (inst instanceof InstGmosNorth) ? northOrigin(inst) : southOrigin(inst);
     }
 
--- a/bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/GMOSSupport.java
+++ b/bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/GMOSSupport.java
@@ -2,6 +2,7 @@ package edu.gemini.wdba.tcc;
 
 import edu.gemini.spModel.gemini.gmos.InstGmosCommon;
 import edu.gemini.spModel.gemini.gmos.GmosOiwfsGuideProbe;
+import edu.gemini.spModel.gemini.gmos.InstGmosNorth;
 import edu.gemini.spModel.target.obsComp.PwfsGuideProbe;
 import edu.gemini.spModel.telescope.IssPort;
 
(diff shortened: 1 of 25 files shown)


bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/GuideConfig.java
```diff
@@ -0,0 +1,177 @@
...
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+
+/**
+ *  Class to evaluate the {@link ObservationEnvironment} and produce a guide config and guide config  name.
+ */
+public final class GuideConfig extends ParamSet {
```
swalker2m: __Disclaimer__: This looks like all new code because I renamed it (from `GuideConfigSouth`).  I did modify this class but for the most part it is the same as the original `GuideConfigSouth`.

bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/ObservationEnvironment.java
```diff
@@ -157,15 +152,16 @@ public TargetEnvironment getTargetEnvironment() {
         return _targetEnv;
     }
 
+    public SortedSet<GuideProbe> usedGuiders() {
+        return _targetEnv.getGuideEnvironment().getPrimaryReferencedGuiders();
+    }
+
     public boolean containsTargets(GuideProbe probe) {
-        final Option<GuideProbeTargets> gtOpt = _targetEnv.getPrimaryGuideProbeTargets(probe);
-        return gtOpt.exists(GuideProbeTargets::containsTargets);
```
swalker2m: This is a key difference from the previous version.  Instead of just checking whether there are targets associated with a guide probe, now require the guide probe to be associated with targets that are configured to be actually _used_.  This is what fixes REL-2789.

bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/GMOSSupport.java
```diff
@@ -55,11 +56,17 @@ public String getPositionAngle() {
...
+        final String port = (((InstGmosCommon) _oe.getInstrument()).getIssPort() == IssPort.UP_LOOKING) ? "" : side;
+
+        if (_oe.isNorth() && _oe.isAltair()) {
+            final String oi = _oe.containsTargets(GmosOiwfsGuideProbe.instance) ? "_OI" : "";
+            final String p1 = _oe.containsTargets(PwfsGuideProbe.pwfs1) ? "_P1" : "";
+            return "AO2GMOS" + port + p1 + oi;
+        } else {
+            final String p2 = (_oe.containsTargets(PwfsGuideProbe.pwfs2)) ? "_P2" : "";
+            return "GMOS" + port + p2;
+        }
```
swalker2m: This part addresses the requirement:

> There are also some additional instrument types needed. These are always all CAPS and are used by the Instrument.xml file. 
>
> AO2GMOS5_P1 
> AO2GMOS5_OI 
> AO2GMOS5_P1_OI

bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/GMOSSupport.java
```diff
@@ -77,12 +84,39 @@ public void addGuideDetails(ParamSet guideConfig) {
...
+        return inst.getFPUnit().isIFU() ? "_ifu" : "";
+    }
+
+    private String northOrigin(InstGmosCommon<?, ?, ?, ?> gmos) {
+        switch (_oe.getAoAspect()) {
+            case ngs: return NGS   + oi();
+            case lgs: return LGS   + p1() + oi();
+            default:  return NO_AO + ifu(gmos);
+        }
+    }
```
swalker2m: This part addresses the requirement:

> Keeping the existing point origin values, please add the following: 
> 
> ngs2gmos_oi 
> lgs2gmos_oi 
> lgs2gmos_p1_oi

bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/GMOSSupport.java
```diff
@@ -55,11 +56,17 @@ public String getPositionAngle() {
     }
 
     public String getTccConfigInstrument() {
-        String side = _oe.isNorth() ? "5" : "3";
-        String port = (((InstGmosCommon) _oe.getInstrument()).getIssPort() == IssPort.UP_LOOKING) ? "" : side;
-        String ao   = (_oe.isNorth() && _oe.isAltair()) ? "AO2" : "";
-        String p2   = (_oe.containsTargets(PwfsGuideProbe.pwfs2)) ? "_P2" : "";
-        return ao + "GMOS" + port + p2;
+        final String side = _oe.isNorth() ? "5" : "3";
```
cquiroz: Unrelated to the PR itself by why does the TCC uses 5 for north and 3 for south?

swalker2m: This is ridiculous but it corresponds to the ISS port on which the instrument is mounted at the two sites.

bundle/edu.gemini.wdba.xmlrpc.server/src/test/java/edu/gemini/wdba/tcc/InstrumentSupportTestBase.java
```diff
@@ -53,6 +62,31 @@ public void store() throws Exception {
...
+        final TargetObsComp toc  = (TargetObsComp) oc.getDataObject();
+        toc.setTargetEnvironment(env);
+        oc.setDataObject(toc);
+    }
+
+    protected void addGuideStar(GuideProbe probe) {
+        TargetEnvironment env = getTargetEnvironment();
+        GuideGroup        grp = env.getPrimaryGuideGroup();
+        if (grp.isAutomatic()) {
+            grp = GuideGroup.create("Manual Group");
```
cquiroz: Just  a tiny detail, on upstream these were renamed to "Manual" using  `GuideGroup.ManualGroupDefaultName`

cquiroz:
Looks good to me :+1: thanks for taking the time to cleanup the tests and other bits of the code. WDBA seems quite ancient

bundle/edu.gemini.wdba.xmlrpc.server/src/test/java/edu/gemini/wdba/tcc/InstrumentSupportTestBase.java
swalker2m: Thanks, I'll fix the `"Manual Group"` thing and merge it.

PR #964, opened 2016-05-23
REL-2763: Rename the manual guide group

cquiroz:
Rename manual guide group naming and set it into a shared constant


swalker2m:
👍

PR #966, opened 2016-05-23
REL-2788: simplify default strategy calculation

swalker2m:
The AGS registrar provides a method to obtain the valid strategies for a given observation, ranked in order of preference.  It also provides a method to obtain the "default" strategy for the observation.  Rather than simply picking the first (if any) ranked valid strategy, the code was performing a convoluted calculation involving the number of guide probes already selected in the observation.  Unfortunately I can't remember why this was done but it is, naturally, picking the wrong default strategy for simple cases. I believe that the separation of Auto/Manual groups makes the previous method obsolete.

In addition to simplifying the default strategy calculation, I updated a couple of labels as requested/approved by Bryan.

### DIFF
--- a/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsRegistrar.scala
+++ b/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsRegistrar.scala
@@ -20,32 +20,8 @@ object AgsRegistrar {
   /**
    * Determines the default or best strategy for the given observation.
    */
-  def defaultStrategy(ctx: ObsContext): Option[AgsStrategy] = {
-    // Figure out which guide probes we're using, preferring those we mark as
-    // selected (i.e., with a primary target)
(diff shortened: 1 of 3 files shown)


bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsRegistrar.scala
```diff
@@ -20,32 +20,8 @@ object AgsRegistrar {
...
-      case h :: t =>
-        val (highestIncidence, _) = ((h, incidence(h))/:t) { case ((s0, i0), s1) =>
-          val i1 = incidence(s1)
-          if (i0 >= i1) (s0, i0) else (s1, i1)
-        }
-        Some(highestIncidence)
-    }
-  }
+  def defaultStrategy(ctx: ObsContext): Option[AgsStrategy] =
+    validStrategies(ctx).headOption
```
cquiroz: Excellent 👍

cquiroz:
Looks good to me :+1:

PR #968, opened 2016-05-23
REL-2726 remove ERR_SCHEDULING_BLOCK_SEM

tpolecat:
Removed P2 check for scheduling block outside the program semester, as requested.


cquiroz:
Very straightforward 👍

PR #969, opened 2016-05-23
REL-2748 reinit target table

tpolecat:
The target table wasn't being updated when the scheduling block changed. This PR fixes that.


cquiroz:
Looks good to me :+1:

PR #970, opened 2016-05-23
REL-2793: Select base target when first viewing Target node

cquiroz:
This PR fixes a bad UI interaction caused by ignoring the case when the index of the item selected on the target environment has no meaning for another observation. The fix will select the first item (Typically Base) instead of ignoring it

### DIFF
--- a/bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/TelescopePosTableWidget.java
+++ b/bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/TelescopePosTableWidget.java
@@ -1137,16 +1137,17 @@ final class TelescopePosTableWidget extends JTable implements TelescopePosWatche
      * Selects the relevant row in the table.
      */
     private void _setSelectedRow(final int index) {
-        if ((index < 0) || (index >= getRowCount())) return;
-        getSelectionModel().setSelectionInterval(index, index);
+        if ((index < 0) || (index >= getRowCount())) {
+            getSelectionModel().setSelectionInterval(0, 0);
+        } else {
+            getSelectionModel().setSelectionInterval(index, index);
+        }
     }
 
     void setIgnoreSelection(boolean ignore) {
         _ignoreSelection = ignore;
     }
 
-    /**
-     */
     public String toString() {
         final String head = getClass().getName() + "[\n";
         String body = "";


bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/TelescopePosTableWidget.java
```diff
@@ -1137,16 +1137,17 @@ void setSelectedRow(final Point location) {
      * Selects the relevant row in the table.
      */
     private void _setSelectedRow(final int index) {
-        if ((index < 0) || (index >= getRowCount())) return;
-        getSelectionModel().setSelectionInterval(index, index);
+        if ((index < 0) || (index >= getRowCount())) {
+            getSelectionModel().setSelectionInterval(0, 0);
+        } else {
+            getSelectionModel().setSelectionInterval(index, index);
+        }
```
swalker2m: What if the new observation _does_ have an item at that index?  Is the goal not to select the base always when you move to a new target environment?

cquiroz: Wouldn't a new observation have only base at pos 0?

swalker2m: I take this to mean that when you first select any target node it should begin with the base position table row selected.

cquiroz: Your're right. this produces some undesired effects, will keep working on it

cquiroz:
The selected index is a shared value so it would be difficult to get every effect as desired, to get to that point there should be one index per observation. That's a bit bigger change than I'm willing to do at this moment.
The version in this PR will solve the immediate problem at the cost that in some circumnstances the base will be selected when you'd expect some of the targets to be selected instead
Note if you have 2 obs with an AGS target, you select the target in one obs, then move to the other obs, it will do the right thing, selecting the AGS target.
In other situations it will be a bit strange, like if you have an obs with an AGS target, another with AGS+manual and you have selected the manual target, then move to the one with AGS it will take you to base
I still think this change is an improvement

swalker2m:
I may be confused about the ultimate goal but my understanding is that it we always select the base when a target node is selected.  Ideally it will not try to remember you were looking at an AGS star or manual guide star or whatever and just select the base as the starting point always.  Is that correct, at least in terms of the goal?

cquiroz:
The goal is that when you create a new obs the base is selected (index 0). Right now the code keeps the index of the last selected item, e.g. in AGS, it will be 2. If you go to another obs it will try to select the one with the same index. For a new obs, there is no index 2 so it will not do anything.
By setting 0 it will select the base
We don't select the base when the target node is selected, we select the item in the list with the same index as the last selection

andrewwstephens:
In the 2016A OT the selected target is remembered for every observation in every program, which is very nice.  If you are looking at User(1) in program 1 observation 1, this target will be selected every time you come back to this Target, even when switching between several programs.  When the OT is restarted the default view of every observation reverts to the Base target.

cquiroz:
hmm, is that so? Can you try having one observation with AGS, then another with AGS and MGS, select MGS and switch to AGS and back, I don't think it will be remembered exactly, unless things have changed a lot in that area since 2016A

andrewwstephens:
@cquiroz - There is no distinction between AGS and MGS in the 16A OT, but every case I have tested does remember which target was selected.

cquiroz:
What if in one obs you have only base and then another base + gs, select gs and go back to the one without gs

andrewwstephens:
Yes, this works as expected.  It remembers that you were looking at the base in obs 1, and the guide star in obs 2, and the user star in obs 3, etc.

cquiroz:
I reviewed the history of the code and there has been many changes due to BAGS. From the point of view of the code there is a single position indicator that changes every time you change obs. There is no per-obs memory of the last position, the best it can do at the moment is to attempt to reuse the last known position

andrewwstephens:
So no hope of recovering target selection memory?

@cquiroz - Your changes look better, but it appears to still remember if the Auto group was the last target selected, in which case the Target of a new observation will default to displaying the Auto group.

cquiroz:
It is possible to keep the selection memory but I think it is a much larger change especially on the context of the already complex target environment component, let's see if I can find a way around in a reasonable time

swalker2m:
The target component UI is pretty delicate so at this point I think this may be the best we can do. 👍

PR #971, opened 2016-05-23
REL-2724:  BagsManager Refactor

swalker2m:
There were a few issues with the `BagsManager` that I found hard to solve by tweaking the existing version.  In particular:

1. When the `BagsManager` determines that an observation is not eligible for BAGS, it clears the `Auto` guide group.  It does not however remove the `AgsHash` value which is used to determine whether anything going into the AGS computation has changed.  If the observation should later become eligible again (say due to an observation class change) a new lookup is not performed. 

2. The above also applies when an AGS lookup fails, say due to a network issue.  The hash value is updated before we know whether the lookup will ultimately be successful.  By the time we try again we've stored the hash value and draw the conclusion that there is nothing to do.

3. The cache of AGS hashes is being manipulated in multiple threads without synchronization.

My naive attempt to fix these issues lead to deadlock on testing.  I don't think I'm capable of fixing it within the existing design so this PR introduces a rather radical reworking of the `BagsManger` for discussion.  It solves these issues by transforming `BagsManager` into an explicit state machine.  All state transitions run in the Swing event thread.  Only a timer wake up and the actual AGS search are performed in separate threads.  It's still a bit more complex than I'd like but to my mind at least it is an improvement.  On the other hand it is late in the dev cycle for such a big change.  Discuss.

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
@@ -1,19 +1,18 @@
 package jsky.app.ot.ags
 
+import jsky.app.ot.ags.BagsState.{IdleState, ErrorTransition, StateTransition}
+
 import java.beans.{PropertyChangeEvent, PropertyChangeListener}
 import java.time.Instant
-import java.util
 import java.util.concurrent.TimeoutException
 import java.util.concurrent._
-import java.util.logging.{Level, Logger}
+import java.util.logging.Logger
 
-import edu.gemini.ags.api.AgsStrategy.Assignment
 import edu.gemini.ags.api.{AgsHash, AgsRegistrar, AgsStrategy}
 import edu.gemini.catalog.votable.{CatalogException, GenericError}
 import edu.gemini.pot.sp._
-import edu.gemini.spModel.core.SPProgramID
 import edu.gemini.spModel.guide.GuideProbe
-import edu.gemini.spModel.obs.{ObsClassService, ObservationStatus, SPObservation}
+import edu.gemini.spModel.obs.{ObsClassService, ObservationStatus}
 import edu.gemini.spModel.obs.context.ObsContext
 import edu.gemini.spModel.obsclass.ObsClass
 import edu.gemini.spModel.rich.shared.immutable._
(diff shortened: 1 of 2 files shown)


cquiroz:
I don't know enough about BAGS to comment but it does seems a fairly big this close to the release

arturog8m:
Wasn't something like this proposed to address the feedback part we couldn't do this time around? I like the new structure, as there is only one place where things change.  I would like this to be reviewed by @sraaphorst who was involved in the previous incarnation of this. If we do it, needs to be released ASAP for testing.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -25,219 +24,396 @@ import jsky.app.ot.gemini.tpe.TpePWFSFeature
...
 
-  // Setup an LRU cache with max entries BagsManager.CacheSize
-  private val hashes = new util.LinkedHashMap[SPNodeKey, Int](math.ceil((BagsManager.CacheSize + 1) * 1.334).toInt, 0.75f, true) {
-    override def removeEldestEntry(e: java.util.Map.Entry[SPNodeKey, Int]): Boolean =
-      size > BagsManager.CacheSize
+
+object BagsState {
+  type AgsHashVal      = Long
+  type StateTransition = (BagsState, IO[Unit]) // Next BagsState, side-effects
+  val ErrorTransition  = (ErrorState, ioUnit)  // Default transition for unexpected events
```
cquiroz: What's ioUnit? Is there some documentation on `IO`?

swalker2m: It's "an `IO` action that does nothing".  A constant value for `IO(())`.  Here we need an `IO` action to conform to the `StateTransition` type but we don't actually have any actions to take.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -25,219 +24,396 @@ import jsky.app.ot.gemini.tpe.TpePWFSFeature
...
-    // observation classes that do not have guiding (e.g. daytime calibration).
-    def isEligibleForBags(ctx: ObsContext): Boolean = {
-      val enabledGroup = ctx.getTargets.getGuideEnvironment.guideEnv.auto match {
-        case AutomaticGroup.Initial | AutomaticGroup.Active(_,_) => true
-        case _                                                   => false
-      }
-      enabledGroup &&
-        AgsRegistrar.validStrategies(ctx).nonEmpty &&
-        ObsClassService.lookupObsClass(observation) != ObsClass.DAY_CAL
+  case class FailureState(obs: ISPObservation, why: String) extends BagsState {
```
sraaphorst: Do we have any intention to do anything with `why`?
From what I can tell, it is passed around, but never used.
Will this just be saved for feedback handling or should it be logged?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -25,219 +24,396 @@ import jsky.app.ot.gemini.tpe.TpePWFSFeature
...
+    override def wakeUp: StateTransition = {
+      def notObserved: Boolean =
+        ObservationStatus.computeFor(obs) != ObservationStatus.OBSERVED
+
+      // Get the ObsContext and AgsStrategy, if possible.  If not possible,
+      // we won't be able to do AGS for this observation.
+      val tup = for {
+        c <- ObsContext.create(obs).asScalaOpt
+        s <- agsStrategy(obs, c)
+      } yield (c, s)
```
sraaphorst: Do we even need `s` here? It looks like it is never used from what I can tell.
I think we could get away with replacing `tup` with an `Option[ObsContext]` and do so only `if agsStrategy(obs,c).isDefined`, or just change the `agsStrategy` line to `_ <- agsStrategy(obs, c)` and then `yield c`.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -25,219 +24,396 @@ import jsky.app.ot.gemini.tpe.TpePWFSFeature
...
+
+      // Compute the new hash value of the observation, if possible.  It may
+      // be possible to (re)run AGS but not necessary if the hash hasn't
+      // changed.
+      val newHash = tup.map { case (ctx, _) => hashObs(ctx) }
+
+      // Here we figure out what the transition to the RunningState should be,
+      // assuming we aren't already observed and that the hash differs.
+      val runningTransition = for {
+        (c, s) <- tup
```
sraaphorst: Again, here, is looks like neither `c` nor `s` is used, and we only care if `tup.isDefined`.
If we change the above from `tup` to `ocOpt: Option[ObsContext]`, then couldn't this just become `_ <- ocOpt`?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -25,219 +24,396 @@ import jsky.app.ot.gemini.tpe.TpePWFSFeature
...
+    stateTuple.foreach { case (obsMap, state) =>
+      val (newState, sideEffect) = update(state)
+      val newStatemap            = stateMap.insert(progKey, obsMap.insert(obsKey, newState))
+
+      val action = for {
+        _ <- IO { stateMap = newStatemap }
+        _ <- sideEffect
+      } yield ()
+
+      action.unsafePerformIO()
```
cquiroz: Will this be calling AGS and thus the catalog? Would it block the UI while doing that?

cquiroz:
I think the code looks really great with the new state machine model being very appropriate for this case. The comments help a lot too. I don't know much about BAGS but even then. you can get a fairly good  idea of what it does reading the code.
I wish I could know more why we are using `IO` here. I think it is to separate the actions changing doing the actual lookups and changing the targets from the rest. Am I close?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
swalker2m: I should probably log it.  It was put there for the feedback originally of course but we're not giving feedback yet.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -25,219 +24,396 @@ import jsky.app.ot.gemini.tpe.TpePWFSFeature
...
+
+      // Place all observations in the IdleState and record them in our stateMap.
+      val obsList = prog.getAllObservations.asScala.toList
+      val obsMap = ==>>.fromList(obsList.map(o => ObsKey(o) -> (IdleState(o, None): BagsState)))
+
+      stateMap = stateMap + (ProgKey(prog) -> obsMap)
+
+      // Now mark all the observations "edited" in order to kick off AGS searches
+      // and hash calculations.
+      obsList.foreach(edited)
```
sraaphorst: Just a quick and possibly stupid question: is there a reason we start in `IdleState` and then call `edited` on each observation, instead of just starting in `Pending`?

swalker2m:
> I wish I could know more why we are using IO here. I think it is to separate the actions changing doing the actual lookups and changing the targets from the rest. Am I close?

@cquiroz Right, it separates out the actions that must be taken from the state machine transitions and allows us to run them when we're ready.

sraaphorst:
Okay, apart from my small observations and the possible `StructureChangeListener`, I really like the way that this turned out. I'm very impressed, and it taught me about some areas of Scala in which I was somewhat lacking in knowledge.

So definite 👍  from me.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
swalker2m: Good question, it is probably possible to do it differently.

`PendingState` requires that it receive a `wakeUp` event to transition either back to `IdleState` if nothing need be done or else to `RunningState` if an AGS lookup is needed.  So we'd be stuck in `PendingState` forever if there isn't a subsequent `wakeUp` event. When an `IdleState` is `edited`, the action associated with the transition is to schedule the `wakeUp`.

Perhaps it would be just as well to start in `PendingState` and explicitly call `wakeUp` instead but I preferred having only the timer calling `wakeUp`.

tpolecat:
Hang on before merging please, I will look at this in the next few hours.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
swalker2m: Thanks for pointing this out, along with the other comment you made along similar lines.  I should pass the context and strategy down to the `triggerAgsAction` instead of having that figure them out again.  I'll make that update.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
swalker2m: There is an action that will trigger an AGS search, but it will be done in a separate thread.  See `triggerAgsAction`.  There `ags.select` returns a `Future`.  If the future gets a successful result, it will make a `succeed` event which from `RunningState` moves to `IdleState`, etc.  If it fails, it will make a `fail` event, etc.

tpolecat:
I'd like to make sure all the updates are forced onto the EDT, so we should either do that or assert that we're on the EDT in `watch`, `unwatch`, and `enqueue`.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -25,219 +24,389 @@ import jsky.app.ot.gemini.tpe.TpePWFSFeature
...
+    } yield (obsMap, state)
+
+    stateTuple.foreach { case (obsMap, state) =>
+      val (newState, sideEffect) = update(state)
+      val newStatemap            = stateMap.insert(progKey, obsMap.insert(obsKey, newState))
+
+      val action = for {
+        _ <- IO { stateMap = newStatemap }
+        _ <- sideEffect
+      } yield ()
```
tpolecat: Can we add a note explaining why the state update needs to happen before the side-effect, and also add a `_ <- IO { log.info(...) }` to this action as a sanity check for the next release? I would feel better with an indication of what's going on. We can turn it down or remove it once we're confident.

tpolecat:
I think this looks good and is worth trying out. There are only two files affected so it's fairly local. If you squash your commits we will then be able to revert it in one shot if it turns out to cause problems.

At some point I think this probably deserves a document with some circle-and-arrow drawings.

👍

swalker2m:
The plan is also to ask Andy to try it out a bit before merging.

swalker2m:
@tpolecat Added explicit EDT shunting for `watch`, `unwatch`, and `enqueue` and transition logging if you want to take a peek before I squash the commits.

arturog8m:
Can we yell at Andy from here? Anyway, @swalker2m could you ask him if he can build a version for him to give this a try before we merge it in?

andrewwstephens:
Building...

swalker2m:
@arturog8m if someone will look at the PR for REL-2799, i'd kind of like to merge that in and rebase so it is included before handing it off

tpolecat:
@swalker2m yep looks good.

PR #972, opened 2016-05-23
REL-2645: Update GSAOI guided offset overhead

cquiroz:
Small PR updating an offset time. Also changed some numbers in the code for constants


swalker2m:
The code seems good. 👍 pending whatever the build break is.

PR #973, opened 2016-05-23
REL-2658: Java exception closing OT with multiple windows open

cquiroz:
An exception can be produced when closing the OT and having multiple SPViewer instances. This PR replaces the list with `CopyOnWriteArrayList` which is safer for concurrent modification

### DIFF
--- a/bundle/jsky.app.ot/src/main/java/jsky/app/ot/viewer/SPViewer.java
+++ b/bundle/jsky.app.ot/src/main/java/jsky/app/ot/viewer/SPViewer.java
@@ -923,7 +923,7 @@ public final class SPViewer extends SPViewerGUI implements PropertyChangeListene
         if (!shouldClose(allPrograms())) {
             return;
         }
-        _orderedInstances.forEach(SPViewer::closeViewer);
+        instances().forEach(SPViewer::closeViewer);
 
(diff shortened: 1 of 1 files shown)


swalker2m:
It worries me that there are multiple threads involved here.  Do we know why?  Closing OT windows should only happen in the Swing event thread.

cquiroz:
Not really, checking the code it is clear that when doing exit it will remove the items one-by-one thus exposing to a concurrent modification exception. How the other changes are done is unclear. Do you want me to dig into that?

swalker2m:
What is the REL number for this?  2648 refers to the QPT icon.

swalker2m:
This should be REL-2658.  I don't think there are multiple threads or anything tricky going on here.  We're just mutating a `List` that we're iterating over:

````java
 _orderedInstances.forEach(SPViewer::closeViewer);
````

where `closeViewer` mutates the `_orderedInstances` list.  I think the most straightforward thing to do would be to copy the `_orderedInstances` and `forEach` that?  For example using the provided `instances()` copier:

````java
instances().forEach(SPViewer::closeViewer)
````

cquiroz:
Updated according to comments on the PR

arturog8m:
Looks good to me 👍

PR #974, opened 2016-05-24
REL-2799: update hash value if altair mode differs

swalker2m:
The AGS hash algorithm is missing a component for the Altair mode.  The mode impacts not only the default `AgsStrategy`, which already is taken into consideration, but also the magnitude limits for the catalog search.

The bug reported in this task happens because the initial Altair mode is NGS, which has different magnitude limits than when in LGS mode despite using the same AOWFS probe.   An initial search is performed with the NGS limits but later when the mode is switched to LGS the hash value computes the same because the AGS strategy is the same.


tpolecat:
👍

cquiroz:
Looks fine

PR #975, opened 2016-05-24
REL-2794: Fix target morphing

sraaphorst:
There was an issue as described in REL-2794 where attempting to morph one target into another type in the target editor was causing issues and not behaving properly: targets were being added to the Auto group, which is incorrect.

This fixes that by moving targets correctly now: auto targets are unmorphable, and guide probe targets remain in their manual group when morphed. User targets are removed and placed in the first manual group, or if there is no first manual group, then one is created.


sraaphorst:
Ooops... there seems to be a conflict. Give me a moment to fix it.

PR #976, opened 2016-05-24
REL-2794: Fix target morphing

sraaphorst:
There was an issue as described in REL-2794 where attempting to morph one target into another type in the target editor was causing issues and not behaving properly: targets were being added to the Auto group, which is incorrect.

This fixes that by moving targets correctly now: auto targets are unmorphable, and guide probe targets remain in their manual group when morphed. User targets are removed and placed in the first manual group, or if there is no first manual group, then one is created.

### DIFF
--- a/bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
+++ b/bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
@@ -229,6 +229,7 @@ public final class EdCompTargetList extends OtItemEditor<ISPObsComponent, Target
         _w.pasteButton.setEnabled(editable && notAutoTarget);
         _w.duplicateButton.setEnabled(editable && notAutoTarget);
         updateDetailEditorEnabledState(editable && notAutoTarget);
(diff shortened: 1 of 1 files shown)


bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
```diff
@@ -491,7 +493,7 @@ private void refreshAll() {
         ptA[index++] = BasePositionType.instance;
         for (GuideProbe guider : guidersList) {
             final boolean guiderNotIllegal = !guider.equals(illegal);
-            ptA[index++] = new GuidePositionType(guider, guiderNotIllegal);
+            ptA[index++] = new GuidePositionType(guider, guiderNotIllegal, _w.positionTable);
```
swalker2m: Unrelated to this PR, I don't follow the logic that sets `illegal` and therefore `guiderNotIllegal` (which seems better named `guiderIsLegal` or something).  My issue is that there can be multiple "illegal" guiders and yet we only keep up with the last one we saw?  (See line 481.)  It seems a bit random whether `guiderNotIllegal` if there is more than one illegal guider.  Does it make sense to you?

swalker2m:
👍

PR #977, opened 2016-05-24
REL-2786 fix PA calculation for zero-duration obs

tpolecat:
The average PA calculation was blowing up for observations with no scheduling block duration or remaining plan time. This forces the duration to be 1ms for the purposes of computing the PA.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/obs/ObsTargetCalculatorService.scala
+++ b/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/obs/ObsTargetCalculatorService.scala
@@ -57,7 +57,11 @@ object ObsTargetCalculatorService {
       // If the duration is going to be smaller than the default step size of 30 seconds used by the
       // target calc, we will have divide by 0 issues, so take this into account.
       val stepSize = if (duration >= TimeUtils.seconds(30)) TimeUtils.seconds(30) else duration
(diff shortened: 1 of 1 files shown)


bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/obs/ObsTargetCalculatorService.scala
```diff
@@ -51,7 +51,7 @@ object ObsTargetCalculatorService {
       // Ideally, if you hover over duration box in GUI, should say acquisition + science time OR not.
 
       // Since we need start < end explicitly, if the duration is None, we cannot use it.
-      val duration = b.duration getOrElse plannedTime.totalTime
+      val duration = b.duration getOrElse plannedTime.totalTime max 1 // REL-2786
```
swalker2m: It would blow up in the `TargetCalculator`?  I'm not that familiar with it but it looks like a provision was made for a "single value" target calculator.  I wonder if the duration is 0 if the `TargetCalculator` should be `TargetCalculator(s, st, b.start)` at line 60.

tpolecat: Ok that might be a better way to do it. I will try that.

tpolecat:
Updated as suggested by @swalker2m.

swalker2m:
👍

PR #978, opened 2016-05-25
REL-2803: don't reinitialize target coordinates for BAGS updates

swalker2m:
When BAGS finds a new guide star, the target editor will reinitialize to show it.  Great care is taken to minimize the amount of work done updating the target table and to maintain the selected target in this case.  Nevertheless the target editor always reinitializes the target details for whatever target is being edited.  When that target is the base position, say because the user is typing in the RA or dec, it resets the text editor with a formatted version of the coordinate and loses the cursor position.

This change simply skips the target detail update when responding to a BAGS update.  This is suboptimal in that when parked over a BAGS-managed star the user must select a different position in the table and come back to the BAGS star to see the updates.  It is however a cheap way to avoid a greater evil.

### DIFF
--- a/bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
+++ b/bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
@@ -421,7 +421,9 @@ public final class EdCompTargetList extends OtItemEditor<ISPObsComponent, Target
             _w.positionTable.reinit(toc, false);
         });
 
-        updateTargetDetails(newTOC.getTargetEnvironment());
+        if (!autoGroupChanged) {
+            updateTargetDetails(newTOC.getTargetEnvironment());
+        }
     }
 
     // OtItemEditor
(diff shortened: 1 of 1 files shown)


bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
```diff
@@ -421,7 +421,9 @@ private boolean autoGroupChanged(final ISPObsComponent oldObsComp, final TargetO
             _w.positionTable.reinit(toc, false);
         });
 
-        updateTargetDetails(newTOC.getTargetEnvironment());
+        if (!autoGroupChanged) {
+            updateTargetDetails(newTOC.getTargetEnvironment());
+        }
```
swalker2m: You should read `autoGroupChanged` as `onlyAutoGroupChanged`.

cquiroz:
I feel like in REL-2973 we are doing ad-hoc patching some underlying design problems on the UI for the target environment. I think is fine though 👍

sraaphorst:
This looks good. Thanks for fixing this. 👍

swalker2m:
I added a commit to rename `autoGroupChanged` to `isBagsUpdate` to better reflect what it actually determines.  I also added a comment to explain what we're doing there.  @cquiroz I agree this is all a bit ad-hoc.

sraaphorst:
This is much clearer. Thanks for making this more readable.

PR #979, opened 2016-05-25
REL-2792: skip header parsing

swalker2m:
The issue in question is that we don't know how to interpret a horizons query response header in the case of Saturn's moon Albiorix.  The header in question is

````
*******************************************************************************
 Revised: Aug 11, 2014               Albiorix / (Saturn)                    626

 For MEAN orbital elements, see  http://ssd.jpl.nasa.gov/sat_elem.html
 Recently discovered Saturnian irregular.
*******************************************************************************
````

which doesn't correspond to anything we are expecting.  This PR introduces an idea for how to handle these situations: skip the header processing when we only care about the ephemeris anyway.

I'm not sure if there are other ramifications of this change, but it does allow us to work with Albiorix and comet Schaumasse...

### DIFF
--- a/bundle/edu.gemini.horizons.api/src/main/scala/edu/gemini/horizons/server/backend/HS2.scala
+++ b/bundle/edu.gemini.horizons.api/src/main/scala/edu/gemini/horizons/server/backend/HS2.scala
@@ -1,5 +1,7 @@
 package edu.gemini.horizons.server.backend
 
+import edu.gemini.spModel.core.HorizonsDesignation.{MajorBody, AsteroidOldStyle, AsteroidNewStyle, Comet}
+
 import java.util.logging.{Level, Logger}
 
 import edu.gemini.spModel.core._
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: bundle/edu.gemini.horizons.api/src/main/scala/edu/gemini/horizons/server/backend/HS2.scala [hunk 1]
```diff
@@ -160,8 +162,15 @@ object HorizonsService2 {
...
 
+    val replyType = target match {
+      case Comet(des)            => HorizonsReply.ReplyType.COMET
+      case AsteroidNewStyle(des) => HorizonsReply.ReplyType.MINOR_OBJECT
+      case AsteroidOldStyle(num) => HorizonsReply.ReplyType.MINOR_OBJECT
+      case MajorBody(num)        => HorizonsReply.ReplyType.MAJOR_PLANET
+    }
+
     def buildEphemeris(m: GetMethod): IO[Long ==>> E]  =
-      IO(CgiReplyBuilder.buildResponse(m.getResponseBodyAsStream, m.getRequestCharSet)).map(toEphemeris)
```
cquiroz: is `buildResponse` not used anymore then?

swalker2m: It appears that the PIT still uses it, via the `CgiBuilder.execute` method.

bundle/edu.gemini.horizons.api/src/main/scala/edu/gemini/horizons/server/backend/HS2.scala
```diff
@@ -160,8 +162,15 @@ object HorizonsService2 {
         TIME_DIGITS      -> FRACTIONAL_SEC
       )
 
+    val replyType = target match {
+      case Comet(des)            => HorizonsReply.ReplyType.COMET
+      case AsteroidNewStyle(des) => HorizonsReply.ReplyType.MINOR_OBJECT
+      case AsteroidOldStyle(num) => HorizonsReply.ReplyType.MINOR_OBJECT
+      case MajorBody(num)        => HorizonsReply.ReplyType.MAJOR_PLANET
+    }
```
swalker2m: I'm not sure whether this is the correct mapping, or if there is a correct mapping.  It appears to be irrelevant since we just ignore everything but the ephemeris data in the `HorizonsReply` we get back from `CgiReplyBuilder`.

bundle/edu.gemini.horizons.api/src/main/scala/edu/gemini/horizons/server/backend/HS2.scala
cquiroz: I guess we'll need to update the PIT when the time comes

cquiroz:
Looks good to me :+1:

bundle/edu.gemini.horizons.api/src/main/scala/edu/gemini/horizons/server/backend/HS2.scala
tpolecat: This is correct, modulo nothing in HORIZONS ever actually being correct.

tpolecat:
Thanks! 👍

PR #980, opened 2016-05-25
REL-2790B: Fixed P2 check and general cleanup.

sraaphorst:
The change allowing Altair + GMOS to use multiple probes broke the P2 check and required some changes to the way the P2 check for LGS was handled.

This was made, along with some basic custodial cleanup.

### DIFF
--- a/bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/general/GeneralRule.java
+++ b/bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/general/GeneralRule.java
@@ -43,6 +43,7 @@ import edu.gemini.spModel.too.TooType;
 
 
 import java.util.*;
+import java.util.stream.Collectors;
 
 /**
  * General Rules are rules that are not instrument-specific
@@ -57,19 +58,17 @@ public class GeneralRule implements IRule {
 
     private static final GeneralRuleHelper helper = new GeneralRuleHelper();
 
-    public IP2Problems check(ObservationElements elements)  {
-        P2Problems problems = new P2Problems();
-        for (IRule rule : GENERAL_RULES) {
-            problems.append(rule.check(elements));
-        }
+    public IP2Problems check(final ObservationElements elements)  {
+        final P2Problems problems = new P2Problems();
+        GENERAL_RULES.stream().forEach(rule -> problems.append(rule.check(elements)));
         return problems;
     }
 
     // Currently Altair is only supported on NIRI, GNIRS and NIFS, and GMOS-N
-    private static boolean isAltairSupported(ObservationElements elements){
-        SPInstObsComp instObsComp = elements.getInstrument();
+    private static boolean isAltairSupported(final ObservationElements elements){
+        final SPInstObsComp instObsComp = elements.getInstrument();
         if (instObsComp == null) return false;
-        SPComponentType type = instObsComp.getType();
+        final SPComponentType type = instObsComp.getType();
         return (type.equals(InstNIRI.SP_TYPE) || type.equals(InstNIFS.SP_TYPE)
             || type.equals(InstGNIRS.SP_TYPE) || type.equals(SPComponentType.INSTRUMENT_GMOS));
     }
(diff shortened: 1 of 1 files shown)


bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/general/GeneralRule.java
```diff
@@ -103,61 +102,49 @@ public IP2Problems check(ObservationElements elements)  {
 
         private static final String WFS_EMPTY_NAME_TEMPLATE = "%s name field should not be empty";
         private static final String P2_PREFERRED = "PWFS2 is the preferred peripheral wavefront sensor.";
-        private static final String NO_AO_OTHER = "Altair cannot use both %s and %s.";
+        private static final String NO_AO_OTHER = "Altair cannot use %s and %s.";
```
sraaphorst: This was changed because this list may now contain more than two guiders due to Altair now supporting multiple guiders. The first %s may become [AOWFS, GMOS OI] and the second something else, although I don't think this P2 warning can occur in new programs.

bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/general/GeneralRule.java
```diff
@@ -103,61 +102,49 @@ public IP2Problems check(ObservationElements elements)  {
...
-            while (iter.hasNext()) {
-                buf.append(", ").append(iter.next().getKey());
-            }
-            return buf.toString();
+        private String formatGuiderString(final Collection<GuideProbe> probes) {
+            final String startDelim = probes.size() >= 2 ? "[" : "";
+            final String endDelim   = probes.size() >= 2 ? "]" : "";
+            final String sep        = ",";
+            final List<String> probeKeys = probes.stream().map(GuideProbe::getKey).collect(Collectors.toList());
+            return startDelim + String.join(sep, probeKeys) + endDelim;
```
sraaphorst: I believe I added the previous `formatGuiderString` in my original PR for 2790, but decided to replace it with this one for cleanliness.

bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/general/GeneralRule.java
```diff
@@ -103,61 +102,49 @@ public IP2Problems check(ObservationElements elements)  {
...
-                badGuiders.foreach(otherGuiders::remove);
+        private void reportAltairLgsGuideIssues(final P2Problems problems, final Set<GuideProbe> usedGuiders,
+                                                final AltairParams.Mode mode, final ISPObsComponent targetComp) {
+            // The "bad guiders", i.e. the used guiders that cannot be used with the mode.
+            final Set<GuideProbe> badGuiders = new TreeSet<>(GuideProbe.KeyComparator.instance);
+            badGuiders.addAll(usedGuiders);
+            mode.guiders().foreach(badGuiders::remove);
+
+            // We must be using the TTF guider, which is the head of the mode guider list.
+            final GuideProbe ttfGuider = mode.guiders().head();
```
sraaphorst: I made a mistake in my understanding of the P2 check when the Altair changes were made, so this should fix that and properly require the TTF target.

bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/general/GeneralRule.java
```diff
@@ -103,61 +102,49 @@ public IP2Problems check(ObservationElements elements)  {
...
+            if (!usedGuiders.contains(ttfGuider)) {
+                problems.addError(PREFIX + "LGS_WFS", String.format(LGS_WFS, mode.displayValue(), ttfGuider), targetComp);
+            } else if (!badGuiders.isEmpty()) {
+                // The good guiders, i.e. the ones that can be used with the mode and are in use.
```
sraaphorst: This only happens if guiders are in use that shouldn't be.
We don't allow this in the new implementation, I don't think, so it should never come up except possibly in old programs.

swalker2m:
Looks good to me. 👍

PR #981, opened 2016-05-25
REL-2783 scheduling block editor enablement

tpolecat:
This PR makes two changes, as requested:

- It removes the scheduling block editor from `EdObservation2`.
- It updates the enablement logic for the scheduling block editor in `EdCompTargetList` such that it is enabled if and only if there is at least one nonsidereal target in the target environment.

Tested with switching program state and keys and the global magical enablement stuff seems to work correctly.


swalker2m:
Looks good to me. 👍

PR #982, opened 2016-05-25
REL-552 Added "Coadds" column to the GNIRS ITC imaging table in the O…

olesjasmirnova:
…T. Also disabled calculations with Altair in Spectroscopy mode for now, the calculations are not good.


cquiroz:
Looks fine to me 👍

PR #985, opened 2016-05-26
REL-2804: ITC table tooltips

sraaphorst:
As per the JIRA task:

1. Added new and changed existing tooltips.
2. Made tooltips appear for table header.

The second task is actually nontrivial, and was done based on a Java example as provided here:
http://docs.oracle.com/javase/tutorial/uiswing/examples/components/TableToolTipsDemoProject/src/components/TableToolTipsDemo.java

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcTable.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcTable.scala
@@ -6,7 +6,6 @@ import javax.swing.{Icon, ListSelectionModel}
 
 import edu.gemini.ags.api.AgsRegistrar
 import edu.gemini.itc.shared._
-import edu.gemini.pot.ModelConverters
 import edu.gemini.pot.sp.SPComponentType._
 import edu.gemini.spModel.`type`.DisplayableSpType
 import edu.gemini.spModel.config2.{Config, ConfigSequence, ItemKey}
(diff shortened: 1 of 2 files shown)


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcTable.scala
```diff
@@ -59,7 +58,6 @@ trait ItcTable extends Table {
   peer.setRowSelectionAllowed(true)
   peer.setColumnSelectionAllowed(false)
   peer.setSelectionMode(ListSelectionModel.SINGLE_SELECTION)
-  peer.getTableHeader.setReorderingAllowed(false)
```
cquiroz: Why do you disable reordering?

sraaphorst: There's no reason for it, I guess, apart from the fact that the default ordering makes sense.
I was just maintaining what was already in place.

cquiroz: Did you try if the tooltips are ok after you reorder? Sometimes the table indexes get out of whack when you reorder

sraaphorst: Yes, I initially had forgotten to turn reordering back on, and the tooltip positions were maintained as they should be. I believe that the call to `getModelIndex` handles this.

cquiroz:
Looks good, I relaunched travis as it failed but I think it is unrelated

PR #987, opened 2016-05-27
REL-2788: label change

swalker2m:
This PR makes the AGS strategy labels consistent between the target list form editor and the TPE.


cquiroz:
👍

PR #988, opened 2016-05-27
REL-2806: Reinstate missing target V-band magnitude option

cquiroz:
On the magnitude editor the ordering used removed `V` as it has the same wavelength as `AP`. the PR instead uses the default band ordering

screenshot 2016-05-27 11 04 15


swalker2m:
👍

PR #991, opened 2016-05-27
REL-2783 update tooltips

tpolecat:
This updates the tooltip when the PA controls (which should really be renamed at this point) are used for editing the scheduling block start.


cquiroz:
Looks good to me :+1:

PR #993, opened 2016-05-30
REL-2797: duplicate observation numbers

swalker2m:
This PR addresses an issue in which it was possible to create two observations with the same observation number.  A little background is required to understand the issue.  Each program maintains a `ProgramData` object which tracks the maximum observation number ever created in the program.  When a new observation is created by the factory, this number is incremented and used for the observation.  Since programs can be edited independently, it is possible for observations with the same numbers to be created in distinct databases.  The sync process takes care of renumbering duplicate observation numbers though so that all observations in the merged update have a unique observation number once again.  It was not, however, updating the max observation number counter but instead leaving it set incorrectly for subsequent new observations.

In addition, the check we performed to detect duplicate observation numbers was not complete.  It would correctly avoid adding an observation with a duplicate number to the same list of observations in a program, group, or template group, but would not check all observations anywhere in the program.

This PR:

* fixes the max observation number increment issue
* checks the entire program when observations are added to make sure there are no duplicates
* removes an older, now unused observation renumbering method in `MemProgram`
* adds a number of test cases

It is possible that existing observations in the ODB have duplicate numbers.  Those will now be rejected on import and we'll have to fix them manually.

@jluhrs @sraaphorst @cquiroz @tpolecat @arturog8m  All test cases run but since this is a fairly sensitive and fairly late change I'd like to ask everyone to carefully review it and comment.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/pot/sp/ISPProgram.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/pot/sp/ISPProgram.java
@@ -48,17 +48,5 @@ public interface ISPProgram extends ISPGroupContainer, ISPObservationContainer,
 
     VersionVector<LifespanId, Integer> getVersions(SPNodeKey key);
     void setVersions(SPNodeKey key, VersionVector<LifespanId, Integer> vv);
-
-    /**
-     * Renumbers the observations in this program to match those in
-     * <code>that</code> program where a matching observation is determined by
-     * observation key.  This is used when merging two versions of the same
-     * program together.
-     * <p>
-     * This is really a horrible method, an embarrassment, and a reflection of
-     * how bad this model really is.  Please don't call this method.
-     * </p>
-     */
-    void renumberObservationsToMatch(ISPProgram that);
 }
 
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/pot/sp/memImpl/MemObservation.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/pot/sp/memImpl/MemObservation.java
@@ -23,7 +23,7 @@ public final class MemObservation extends MemAbstractContainer implements ISPObs
     public static final String OBS_EXEC_LOG_PROP = "ObsExecLog";
 
     private final MemProgram _program;
-    private int _obsNumber;
+    private final int _obsNumber;
 
     // The list of observation components
     private final List<ISPObsComponent> _compList = new ArrayList<>();
@@ -106,16 +106,6 @@ public final class MemObservation extends MemAbstractContainer implements ISPObs
 
     public int getObservationNumber() {
         return _obsNumber;
-    }
-
-    public void setObservationNumber(int number) {
-        getProgramWriteLock();
-        try {
-            _obsNumber = number;
-            ((ProgramData) getProgram().getDocumentData()).ensureMaxEqualToOrGreaterThan(number);
-        } finally {
-            returnProgramWriteLock();
-        }
     }
 
     public SPObservationID getObservationID() {
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/pot/sp/memImpl/MemObservation.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/pot/sp/memImpl/MemObservation.java
@@ -109,7 +109,13 @@ public final class MemObservation extends MemAbstractContainer implements ISPObs
     }
 
     public void setObservationNumber(int number) {
-        _obsNumber = number;
+        getProgramWriteLock();
+        try {
+            _obsNumber = number;
+            ((ProgramData) getProgram().getDocumentData()).ensureMaxEqualToOrGreaterThan(number);
+        } finally {
+            returnProgramWriteLock();
+        }
     }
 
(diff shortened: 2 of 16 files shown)


bundle/edu.gemini.pot/src/main/java/edu/gemini/pot/sp/ISPProgram.java
```diff
@@ -48,17 +48,5 @@
...
-     * Renumbers the observations in this program to match those in
-     * <code>that</code> program where a matching observation is determined by
-     * observation key.  This is used when merging two versions of the same
-     * program together.
-     * <p>
-     * This is really a horrible method, an embarrassment, and a reflection of
-     * how bad this model really is.  Please don't call this method.
-     * </p>
-     */
-    void renumberObservationsToMatch(ISPProgram that);
```
swalker2m: To be clear, this method was already unused.  This PR didn't remove any uses of it.

bundle/edu.gemini.pot/src/main/java/edu/gemini/pot/sp/memImpl/MemObservation.java
```diff
@@ -109,7 +109,13 @@ public int getObservationNumber() {
...
 
     public void setObservationNumber(int number) {
-        _obsNumber = number;
+        getProgramWriteLock();
+        try {
+            _obsNumber = number;
+            ((ProgramData) getProgram().getDocumentData()).ensureMaxEqualToOrGreaterThan(number);
+        } finally {
+            returnProgramWriteLock();
+        }
```
swalker2m: This is the actual bug fix.  Setting observation numbers explicitly after they are created is only done by the merge process (except for test cases).

bundle/edu.gemini.pot/src/main/java/edu/gemini/pot/sp/memImpl/MemTemplateFolder.java
```diff
@@ -86,6 +86,7 @@ public void setTemplateGroups(List<? extends ISPTemplateGroup> newGroupList) thr
         List<ISPTemplateGroup> newCopy = new ArrayList<>(newGroupList);
         getProgramWriteLock();
         try {
+            SPAssert.setsNoDuplicateObs(this, newCopy);
```
cquiroz: I guess this throws an exception, Will it be displayed to the user, and if so what can he do to solve the problem?

swalker2m: If the assertion throws an exception it represents a logic error.  It isn't supposed to be possible for this to happen. I believe the OT will catch exceptions at a high level and dump them into a dialog box, even if they are logic errors.

There's nothing the user can really do to fix the problem, though each time you create a new observation the counter is bumped so they will get a new observation number.  If you keep at it long enough, it will eventually get to a new unused observation number that can be added.

Nevertheless, I believe we should keep the assertion because the alternative, allowing a program with duplicate observation numbers to exist because of a logic error, can have worse consequences.

cquiroz: sounds good

bundle/edu.gemini.pot/src/main/java/edu/gemini/pot/sp/memImpl/MemObservation.java
tpolecat: Still kind of scary that this is public. The comment on `ISPObservation` says

> It is the caller's responsibility to ensure that observation numbers are not duplicated.

but it seems like it might make sense to check internally ... is that possible? Or does merge have to deal with temporarily inconsistent programs?

tpolecat:
This looks good to me. The fix is straightforward and the additional checks and tests can only help. 👍

cquiroz:
Looks fine to me too, I don't know enough about the details to comment but the battery of test should be more than enough.

bundle/edu.gemini.pot/src/main/java/edu/gemini/pot/sp/memImpl/MemObservation.java
swalker2m: I agree.  It's kind of horrible for this to be public.  I was worried about temporarily inconsistent states.  If the observation will replace an existing observation with the same number, for example, then it is okay for it to be duplicated.  They just can't both be in the tree at the same time.  It's only when you add an observation or something that contains an observation, I think, that you can demand unique numbers.

arturog8m:
Looks good 👍  What's the process to fix programs in the database during the next import?

swalker2m:
Hopefully there isn't anything to fix, but if there is a program with duplicate numbers then the program import will reject it.  If they are both have datasets then I'm not sure what to do, since there will be duplicate labels.  If one or the other isn't observed then we will have to manually fix it in the XML.

arturog8m:
Ok. Sounds that we will know soon... let's make a test release?

swalker2m:
@tpolecat I figured out a way to remove `setObservationNumber` from the API altogether.  In the merge code when building up the merged program, if there is a need to renumber the observation then we create a new empty observation with the correct observation number.

tpolecat:
Oh very nice. 🐔

PR #994, opened 2016-05-30
REL-2740 p2 check for obs site and ephemeris site consistency

tpolecat:
Science wanted a P2 check for the case where you fetch an ephemeris for one site, then the instrument changes or the target is copy/pasted such that the ephemeris is no longer consistent with the site where the observation will take place. This ended up being quite a lot of work.

- `Ephemeris` is now a data type with a `site` field and a `data` field (which contains the time/coord map). For an empty ephemeris the site is irrelevant and is arbitrarily set to `GN`. This was a simpler change than introducing an `Option` or `Either` and is really ok.
- This changes both binary and PIO serialization.
- A P2 error is raised if the observation's site is know and inconsistent with the ephemeris.

👎  for now but you can go ahead and review; I need to add a unit test.

### DIFF
--- a/bundle/edu.gemini.sp.vcs.log/src/test/scala/edu/gemini/sp/vcs/log/impl/PersistentVcsLog2Spec.scala
+++ b/bundle/edu.gemini.sp.vcs.log/src/test/scala/edu/gemini/sp/vcs/log/impl/PersistentVcsLog2Spec.scala
@@ -46,28 +46,28 @@ object PersistentVcsLog2Spec extends Specification {
       _   <- checkSchema("«in memory»")
     } yield ()
 
-  "compatibility" should {
-
-    "query old database with identical result" in go {
-
-      // result as constructed by the old implementation
-      val expected = {
-        val pid = SPProgramID.toProgramID("GN-2015B-Q-10")
-        (List(
-          VcsEventSet(2061 to 2061, Map(OpFetch -> 1), (1449531008701L,1449531008701L), pid, Set(ProgramPrincipal(pid))), 
-          VcsEventSet(1982 to 1988, Map(OpFetch -> 2, OpStore -> 1), (1449523267535L,1449523650543L), pid, Set(StaffPrincipal.Gemini)),
-          VcsEventSet(1960 to 1965, Map(OpFetch -> 2, OpStore -> 2), (1449516911922L,1449517556842L), pid, Set(ProgramPrincipal(pid))) //,
-        ), true)
-      }
-
-      for {
-        _ <- initTestData
-        x <- doSelectByProgram(SPProgramID.toProgramID("GN-2015B-Q-10"), 1, 3)
-      } yield x must_== expected
-
-    }.pendingUntilFixed("See REL-2682")
-
-  }
+//  "compatibility" should {
+//
+//    "query old database with identical result" in go {
+//
+//      // result as constructed by the old implementation
+//      val expected = {
+//        val pid = SPProgramID.toProgramID("GN-2015B-Q-10")
+//        (List(
+//          VcsEventSet(2061 to 2061, Map(OpFetch -> 1), (1449531008701L,1449531008701L), pid, Set(ProgramPrincipal(pid))),
+//          VcsEventSet(1982 to 1988, Map(OpFetch -> 2, OpStore -> 1), (1449523267535L,1449523650543L), pid, Set(StaffPrincipal.Gemini)),
+//          VcsEventSet(1960 to 1965, Map(OpFetch -> 2, OpStore -> 2), (1449516911922L,1449517556842L), pid, Set(ProgramPrincipal(pid))) //,
+//        ), true)
+//      }
+//
+//      for {
+//        _ <- initTestData
+//        x <- doSelectByProgram(SPProgramID.toProgramID("GN-2015B-Q-10"), 1, 3)
+//      } yield x must_== expected
+//
+//    }
+//
+//  }
 
   "checkSchema" should {
 
--- a/bundle/edu.gemini.spModel.io/src/main/scala/edu/gemini/spModel/io/impl/migration/to2016B/To2016B.scala
+++ b/bundle/edu.gemini.spModel.io/src/main/scala/edu/gemini/spModel/io/impl/migration/to2016B/To2016B.scala
@@ -103,7 +103,13 @@ object To2016B extends Migration {
     }
 
   // Update targets to the new model
-  def updateTargets(d: Document): Unit =
+  def updateTargets(d: Document): Unit = {
+
+    // For normal programs the site is known from the program ID. For weirdo programs like `Andy`
+    // we'll just pick GN arbitrarily. It's the best we can do for now; some observations will end
+    // up with a 1-element ephemeris (which needs to be refreshed anyway) with the wrong site.
+    val site = programSite(d).getOrElse(Ephemeris.empty.site)
+
     allTargets(d).foreach { t =>
 
       // In order to determine the target type we need the system as well as the coordinates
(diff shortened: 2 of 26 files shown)


bundle/edu.gemini.sp.vcs.log/src/test/scala/edu/gemini/sp/vcs/log/impl/PersistentVcsLog2Spec.scala
```diff
@@ -46,28 +46,28 @@ object PersistentVcsLog2Spec extends Specification {
...
+//      }
+//
+//      for {
+//        _ <- initTestData
+//        x <- doSelectByProgram(SPProgramID.toProgramID("GN-2015B-Q-10"), 1, 3)
+//      } yield x must_== expected
+//
+//    }
+//
+//  }
```
tpolecat: This test is timezone specific and there's no easy way to fix it, so I'm just turning it off for now.

cquiroz: Can't you set the timezone explicitly before running the test? Perhaps it affects other tests if you do so

cquiroz:
Looks fine to me, it makes a lot of sense that `Ephemeris` is its own data type

bundle/edu.gemini.sp.vcs.log/src/test/scala/edu/gemini/sp/vcs/log/impl/PersistentVcsLog2Spec.scala
tpolecat: That's a possibility although yes it might affect other tests running in parallel. I think in the end it doesn't matter; this is testing an SPDB migration that has already happened.

cquiroz: 👍

bundle/edu.gemini.spModel.io/src/test/scala/edu/gemini/spModel/io/impl/migration/to2016B/TargetMigrationTest.scala
```diff
@@ -102,7 +102,7 @@ class TargetMigrationTest extends Specification with MigrationTest {
     "Migrate MPC Minor Planets" in withTestProgram2("targetMigration.xml") { p =>
       p.findBaseByObsTitle("mpc-minor-planet") must_== Some(NonSiderealTarget(
         "beer",
-        IMap(1457994485000L -> Coordinates.fromDegrees(88.769, 21.08299999999997).get),
+        Ephemeris(Site.GN, IMap(1457994485000L -> Coordinates.fromDegrees(88.769, 21.08299999999997).get)),
```
sraaphorst: This is probably a stupid question, but is there any difference between using `IMap` and `==>>`?
I don't see one, but am curious why `IMap` would be essentially a type alias for `==>>`.

swalker2m:
This looks good to me.  I agree with Carlos, it's nice to have an `Ephemeris` type as well. 👍 

There is still some debate about whether there is time to include this for a test for the release on Thursday, but the code looks great.

bundle/edu.gemini.spdb.reports.collection/src/test/java/edu/gemini/dbTools/ephemeris/TestSupport.scala
```diff
@@ -33,7 +32,7 @@ trait TestSupport extends ProgramTestSupport {
 
   val genNonSiderealTarget: Gen[NonSiderealTarget] =
     Gen.oneOf(nonSids).map { case (hid, name) =>
-      NonSiderealTarget(name, ==>>.empty, Some(hid), List.empty, None, None)
```
sraaphorst: Here's an example as to my confusion:
I see `==>>.empty` and `IMap.empty` both used.
Why the inconsistency?

sraaphorst:
Seems straightforward enough.
👍

bundle/edu.gemini.spModel.io/src/test/scala/edu/gemini/spModel/io/impl/migration/to2016B/TargetMigrationTest.scala
tpolecat: No, they're the same thing. Just aliases.

tpolecat:
@swalker2m why would there not be time to include it for Thursday? I want to add a testcase but should be able to merge this morning.

bundle/edu.gemini.spModel.io/src/test/scala/edu/gemini/spModel/io/impl/migration/to2016B/TargetMigrationTest.scala
swalker2m: I assume the "I" in `IMap` stands for invariant and I'm guessing it is providing an option that looks more natural if you instantiate the type with standard prefix notation `IMap[A, B]` as opposed to infix `A ==>> B`.

tpolecat: Yeah the alias was added pretty recently and it's by analogy with `IList` (in which the `I` means invariant).

swalker2m: `ISet` as well of course.

arturog8m:
@tpolecat because I wasn't sure what time you were going to merge, I didn't know exactly what was pending, and I wasn't expecting you to be up so early :) . The packages to be installed on the ops workstations will be signed in tonight - so if this is merged this morning we should be okay to meet that deadline. This does not sound like a _mandatory_ thing to have, but rather a nice to have thing, but since it's ready and everyone seem okay with the level of risk involved in this change, let's ship it.

tpolecat:
@arturog8m ok. It was marked as Major and was the only thing left on my list so I went ahead and did it.

sraaphorst:
Thanks, @tpolecat and @swalker2m for answering my questions! I appreciate it.

arturog8m:
@tpolecat sounds good. We'll make a final release then once this is merged, and we will prepare the final packages for the operational workstations by the end of our day here in Chile.

bundle/edu.gemini.spModel.io/src/main/scala/edu/gemini/spModel/io/impl/migration/to2016B/To2016B.scala
```diff
@@ -103,7 +103,13 @@ object To2016B extends Migration {
     }
 
   // Update targets to the new model
-  def updateTargets(d: Document): Unit =
+  def updateTargets(d: Document): Unit = {
+
+    // For normal programs the site is known from the program ID. For weirdo programs like `Andy`
+    // we'll just pick GN arbitrarily. It's the best we can do for now; some observations will end
+    // up with a 1-element ephemeris (which needs to be refreshed anyway) with the wrong site.
+    val site = programSite(d).getOrElse(Ephemeris.empty.site)
```
tpolecat: So, nobody picked up on this but after sleeping on it I think it may be a deal-breaker. What would we think if I only did the P2 check for ephemerides with >1 point?

swalker2m: The ODB is started with a site option, I believe.  What if you use that instead of defaulting?  See `SiteProperty`.

tpolecat:
Ok I changed the behavior for trivial (1-element) ephemerides and added testcases.

swalker2m:
I think this is fine 👍.  It's not super-intuitive that a one-element ephemeris wouldn't need to have the correct site but it should be okay for this release.

tpolecat:
I'll merge when it turns green.

PR #995, opened 2016-05-31
REL-2797: fully initialize renumbered observations

swalker2m:
This is an addendum to the REL-2797 observation renumbering PR.  There is a bug in that renumbered observations are not being fully initialized with the suite of non-"data object" client data.  Instead of using a `null` node initializer in the `MergePlan` I should have used an instance that does everything except add child nodes.


tpolecat:
GAH! It's been ten years and I'm still amazed how subtle and complicated this stuff is.

Good catch. 👍

bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs2/MergePlan.scala
```diff
@@ -105,7 +106,7 @@ case class MergePlan(update: Tree[MergeNode], delete: Set[Missing]) {
               // with the correct number.  Set the "initializer" to null though
               // since we'll be setting up the observation ourselves below.
               if (num === o.getObservationNumber) o
-              else f.createObservation(p, num, null, o.getNodeKey)
+              else f.createObservation(p, num, ObservationNI.NO_CHILDREN_INSTANCE, o.getNodeKey)
```
cquiroz: `nulls` do funny things

PR #996, opened 2016-06-01
REL-2812: minimum angular separation

swalker2m:
This PR fixes an embarrassing, gross oversight in my fix for REL-2791.  When computing angular separation, we need a value between 0 and PI.  D'oh!  In other words, if the RA is more than 18 hours away, it is actually closer than 6 hours if you go in the other direction of course.

### DIFF
--- a/bundle/jsky.coords/src/main/java/jsky/coords/platepos.java
+++ b/bundle/jsky.coords/src/main/java/jsky/coords/platepos.java
@@ -170,16 +170,18 @@ public class platepos {
         // Difference between RA and plate_ra in radians
         double xdiff = xr - wcs.plate_ra;
 
-        // If the RA of the point is further away than 6 hours from the RA of
-        // the image center then punt because the (x,y) projection wraps around.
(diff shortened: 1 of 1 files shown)


bundle/jsky.coords/src/main/java/jsky/coords/platepos.java
```diff
@@ -170,16 +170,18 @@
...
-
         sypos = Math.sin(yr);
         cypos = Math.cos(yr);
         syplate = Math.sin(wcs.plate_dec);
         cyplate = Math.cos(wcs.plate_dec);
         sxdiff = Math.sin(xdiff);
         cxdiff = Math.cos(xdiff);
+
+        // If the RA of the point is further away than 6 hours from the RA of
+        // the image center then punt because the (x,y) projection wraps around.
```
cquiroz: punt?

cquiroz:
Looks good to me :+1:

arturog8m:
👍  Good!

PR #997, opened 2016-06-01
REL-2813 check for null dataset disposition

tpolecat:
Trivial fix for REL-2813 … I didn't notice it was backlogged.

### DIFF
--- a/bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/EdObservation2.java
+++ b/bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/EdObservation2.java
@@ -582,7 +582,7 @@ public final class EdObservation2 extends OtItemEditor<ISPObservation, SPObserva
         final boolean override = getDataObject().isOverrideQaState();
 
         // Format the dataset disposition ("dataflow step").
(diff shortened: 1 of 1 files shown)


cquiroz:
👍

swalker2m:
I'd like to investigate a bit why the obslog is not present.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/EdObservation2.java
```diff
@@ -582,7 +582,7 @@ private void _showStatus() {
         final boolean override = getDataObject().isOverrideQaState();
 
         // Format the dataset disposition ("dataflow step").
-        final String statusString = (dispo.isEmpty()) ? "No Data" : dispo.get().description();
+        final String statusString = (dispo == null || dispo.isEmpty()) ? "No Data" : dispo.get().description();
```
swalker2m: Since the `lookupDatasetDisposition` method already returns an `Option`:

````scala
final scala.Option<DataflowStatus> dispo = DatasetDispositionService.lookupDatasetDisposition(getNode());
````

we might want to make it return `None` instead of `null` for this case (no obslog).

tpolecat:
I'm going to close this since it's not a new problem and not important for 16B.

PR #998, opened 2016-06-01
REL-2814: add execution context parameter

swalker2m:
This PR adds an implicit `ExecutionContext` parameter to all the AGS methods (and also to the `VoTableClient` catalog search).  This makes it possible to run the BAGS searches in the BAGS execution context instead of the default global execution context where everything else runs.  The BAGS execution context runs with lower-priority threads so manual catalog searches and target name resolution can continue to be responsive while BAGS is running.  The execution context parameter bubbled up through the AGS APIs so a number of files had to be modified.

I also lowered the priority of BAGS searches a bit so as to not compete with the worker that updates serialized program files in the database.

### DIFF
--- a/bundle/edu.gemini.ags.client.api/src/main/scala/edu/gemini/ags/client/api/AgsClient.scala
+++ b/bundle/edu.gemini.ags.client.api/src/main/scala/edu/gemini/ags/client/api/AgsClient.scala
@@ -3,8 +3,7 @@ package edu.gemini.ags.client.api
 import edu.gemini.model.p1.immutable.Observation
 import java.net.URL
 
-import scala.concurrent.Future
-import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.{ExecutionContext, Future}
 
 /**
  * Base trait for AGS client implementations.
@@ -22,7 +21,7 @@ trait AgsClient {
    * Performs an asynchronous estimation request, providing the result to the
    * given callback function when available.
    */
-  def estimate(obs: Observation, time: Long)(callback: Callback) {
+  def estimate(obs: Observation, time: Long)(callback: Callback)(implicit ec: ExecutionContext) {
     Future.apply { callback(estimateNow(obs, time)) }
   }
 
--- a/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsVoTableCatalog.scala
+++ b/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsVoTableCatalog.scala
@@ -8,8 +8,7 @@ import edu.gemini.spModel.gemini.gems.GemsInstrument
 import edu.gemini.spModel.gemini.obscomp.SPSiteQuality.Conditions
 import edu.gemini.spModel.obs.context.ObsContext
 
-import scala.concurrent.{Await, Future}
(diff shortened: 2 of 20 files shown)


bundle/edu.gemini.ags.client.api/src/main/scala/edu/gemini/ags/client/api/AgsClient.scala
```diff
@@ -22,7 +21,7 @@ trait AgsClient {
    * Performs an asynchronous estimation request, providing the result to the
    * given callback function when available.
    */
-  def estimate(obs: Observation, time: Long)(callback: Callback) {
+  def estimate(obs: Observation, time: Long)(callback: Callback)(implicit ec: ExecutionContext) {
```
cquiroz: Why isn't this method with a return type `:Future[_]`?

cquiroz:
Looks good to me :+1:. I'm glad that we could fix this. Could you do a sanity check on the catalog tool searches and GEMS manual search?

bundle/edu.gemini.ags.client.api/src/main/scala/edu/gemini/ags/client/api/AgsClient.scala
swalker2m: Good question.  I didn't look at it much other than to add the implicit parameter.  I guess the idea is that it returns right away and then informs the `Callback` when it is done.  Just having it return a `Future` seems like a better idea to me too but I won't go into refactoring that now.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsVoTableCatalog.scala
```diff
@@ -38,8 +37,8 @@ case class GemsVoTableCatalog(backend: VoTableBackend = ConeSearchBackend, catal
    * @param timeout      Timeout in seconds
    * @return list of search results
    */
-  def search4Java(obsContext: ObsContext, basePosition: Coordinates, options: GemsGuideStarSearchOptions, nirBand: Option[MagnitudeBand], timeout: Int = 10): java.util.List[GemsCatalogSearchResults] =
-    Await.result(search(obsContext, basePosition, options, nirBand), timeout.seconds).asJava
+  def search4Java(obsContext: ObsContext, basePosition: Coordinates, options: GemsGuideStarSearchOptions, nirBand: Option[MagnitudeBand], timeout: Int = 10, ec: ExecutionContext): java.util.List[GemsCatalogSearchResults] =
```
swalker2m: Right, exactly.

tpolecat:
Ugh, implicit thread pools are the worst. `Task` has the same issue. Looks good. 👍

swalker2m:
I did run a manual catalog search to confirm that it finishes quickly while BAGS is still working.  I just tried a GeMS search.  It "works" but I find it a bit odd that the manual GeMS search comes up with something with the default settings while the AGS GeMS strategy finds nothing.  I don't think the algorithm results can be changed by the thread context it runs on though?

cquiroz:
I think the default parameters on the manual search are less restrictive than on AGS thus giving you more results. that's the point of manual I guess. I'll give it a quick check to confirm

swalker2m:
Testing with the 1.12.1 release (before this change) reveals the same behavior. I'll merge.

PR #999, opened 2016-06-01
REL-2815 fix sched block update

tpolecat:
In the fix for REL-2785 I introduced a bug in which the scheduling block would only be updated if the observing night changed. This fixes that.


swalker2m:
Looks good. 👍

cquiroz:
Looks good to me too :+1:

PR #1000, opened 2016-06-02
REL-2816: log NonSiderealTargetRefFunctor exceptions and continue

swalker2m:
This PR addresses a problem wherein an exception in computing the observation status of any observation in the ODB would stop the `NonSiderealTargetRefFunctor` (part of the "ephemeris robot") from finding other non-sidereal targets.  Specifically there are two changes:

- Handle exceptions in the observation status computation explicitly, since smart-gcal makes the computation error-prone.  Assume if we don't know the status because of an exception, the observation is still active.

- Handle other random exceptions that may happen while processing a particular program so that other programs without problems in the database can at least be included in the results.

(I struggled to come up with an automated test quickly so I just sanity checked it in my build.  I'll keep working on it but wanted to get this out for feedback now.)


cquiroz:
Looks good to me :+1:

sraaphorst:
Seems alright to me, too. 👍

PR #1001, opened 2016-06-02
use special executor for blocking catalog queries

tpolecat:
Ok so this modifies @swalker2m's work on #998 by making all the `implicit ec: ExecutionContext` parameters **ex**plicit, and only using the passed `ec` at the very bottom, when doing the blocking call to look up targets. Everything else runs on the global executor, except the BAGS worker itself, which is on its own pool still but may not need to be. I didn't mess with it.

In any case the end result is that BAGS results trickle in rather than batching up, and don't interfere with high-priority manual searches (and sidereal name lookups).

Reviewers need to pull this branch and test locally; the code itself isn't going to tell you much.

### DIFF
--- a/bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableClient.scala
+++ b/bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableClient.scala
@@ -12,6 +12,7 @@ import org.apache.commons.httpclient.methods.GetMethod
 
 import scala.annotation.tailrec
 import scala.concurrent.{ExecutionContext, Promise, Future}
+import scala.concurrent.ExecutionContext.Implicits.global
 import scala.util.{Failure, Success}
 import scala.math.min
 
@@ -20,7 +21,7 @@ import Scalaz._
 
 trait VoTableBackend {
   val catalogUrls: NonEmptyList[URL]
-  protected [votable] def doQuery(query: CatalogQuery, url: URL)(implicit ec: ExecutionContext): Future[QueryResult]
+  protected [votable] def doQuery(query: CatalogQuery, url: URL)(ec: ExecutionContext): Future[QueryResult]
 }
 
 trait CachedBackend extends VoTableBackend {
@@ -129,11 +130,12 @@ trait CachedBackend extends VoTableBackend {
   protected def query(e: SearchKey): QueryResult
 
   // Cache the query not the future so that failed queries are executed again
-  override protected [votable] def doQuery(query: CatalogQuery, url: URL)(implicit ec: ExecutionContext): Future[QueryResult] = Future {
+  override protected [votable] def doQuery(query: CatalogQuery, url: URL)(ec: ExecutionContext): Future[QueryResult] = Future {
+    Logger.getLogger(getClass.getName).info(s"Starting BAGS lookup on ${Thread.currentThread}")
     val qr = cachedQuery(SearchKey(query, url))
     // Filter on the cached query results
     qr.copy(query = query, result = qr.result.filter(query))
-  }
+  } (ec)
 
 }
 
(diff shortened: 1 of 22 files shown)


tpolecat:
Oddly it doesn't compile but it does run. IntelliJ is dumb. Hang on.

tpolecat:
Ok it compiles now.

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableClient.scala
```diff
@@ -129,11 +130,12 @@ trait CachedBackend extends VoTableBackend {
...
 
   // Cache the query not the future so that failed queries are executed again
-  override protected [votable] def doQuery(query: CatalogQuery, url: URL)(implicit ec: ExecutionContext): Future[QueryResult] = Future {
+  override protected [votable] def doQuery(query: CatalogQuery, url: URL)(ec: ExecutionContext): Future[QueryResult] = Future {
+    Logger.getLogger(getClass.getName).info(s"Starting BAGS lookup on ${Thread.currentThread}")
     val qr = cachedQuery(SearchKey(query, url))
     // Filter on the cached query results
     qr.copy(query = query, result = qr.result.filter(query))
-  }
+  } (ec)
```
swalker2m: This is the one place where the execution context is actually used?

swalker2m:
This makes sense 👍 , but in a month or two I'd probably need to refer back to the PR to figure out what is going on.  Maybe we should rename the `ec` arguments everywhere something more verbose like "catalogContext" or "blockingContext"?  And I suppose there isn't any reason to put it in a second argument list if it isn't implicit.  All of that can wait until after the patch release though.

Speaking of the patch release, I think @arturog8m wants to hold off on an OT patch release until we have a bit of time with the 16B OT in action generating new bugs to include.  So @arturog8m, we should hold off on merging this correct?

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableClient.scala
tpolecat: Yep.

tpolecat:
@swalker2m yeah I didn't want to rock the boat too much so I did the minimal amount of work and tried to keep it as mechanical as possible.

arturog8m:
@swalker2m @tpolecat Right - let's not merge this one yet. We are upgrading Gemini South today and Gemini North on Monday, so let's figure out after that when to make a patch release, with any other potential things that real testing might reveal.

tpolecat:
Has anyone pulled this branch to try it out locally?

swalker2m:
@tpolecat I've been a bit busy but I'll do that now.

cquiroz:
I can try but I'm not sure what to look for

tpolecat:
Load up a big program like GS-2016A-Q-80 ... you should see bags working in the background in log messages, but still be able to do manual guidestar lookups and sidereal name lookups (currently these are blocked until bags finishes).

swalker2m:
I just tried it out and it is working great.

cquiroz:
You beat me to it. I'm still not sure how removing the `implicit` changes the equation. Is it that your are just forced to be explicit about where to run ags/catalog?

swalker2m:
Removing the `implicit` keyword from the parameter keeps the context that we're passing around from being used in any `Future`s that are created.  We want everything except the actual catalog lookup to run in the implicit execution context as before, but we want the catalog lookup to be done in the context that we're passing in.  Since the explicit passed-in context uses lower-priority threads, when a catalog lookup finishes, any continuing tasks (say in subsequent `flatMap` calls) are executed in high-priority thread execution context.  In other words, once the catalog results become ready we use them right away in AGS instead of having the AGS evaluation of the results get in line behind all the remaining catalog lookups.

tpolecat:
@cquiroz as Shane says, but I think it's more about queueing than priority.

Every time you `map` or `flatMap` a `Future` the computation gets suspended and re-queued on whatever the implicit execution context is at that point in your code. So if that happens to be the same context that's processing network requests you end up with fast code getting blocked by slow code. We want to use a dedicated execution context for the blocking IO requests; however because it was being passed implicitly it was also being used by the fast code as well. So that's what I fixed.

arturog8m:
what JIRA issue is this one?

PR #1003, opened 2016-06-03
REL-2820: QPT  java.lang.ClassCastException

cquiroz:
Sometimes wildcards `?` are your friend


tpolecat:
Doh! Yeah [that change](https://github.com/gemini-hlsw/ocs/commit/5663d21efeae58f4b3f3c84aa6d03408565aff2b?w=1#diff-a8ba7ee9174d85f6d67c5152e18a2e42) happened after the December release but I didn't look back that far.

It really should not have typechecked with those changes ... Java is very unsafe.

Anyway 👍

PR #1004, opened 2016-06-03
REL-2824: remove call to unimplemented method

swalker2m:
The daily ODB backup fails when it tries to archive the fetch/store (i.e. "vcs") log. I traced this down to a call to `VcsLog.archive` which is unimplemented:

````scala
  def archive(f: File): Unit = ???
````

I'm not sure how much work is involved in implementing the method but we can get around this for the next few days and allow the backup to complete by just skipping the fetch/store log archive for now, I think.


cquiroz:
That's funny 👍

tpolecat:
Well that's weird.

swalker2m:
I think it just got overlooked when we switched from slick to doobie.

tpolecat:
I'll do this today, it's just a couple lines.

swalker2m:
@tpolecat sounds good.  I just merged this update so we can get on with the ODB update now.  At the rate we're going in this release, there will be a patch release soon and we can include this with that.

PR #1006, opened 2016-06-06
REL-2825: fix duplicate target feature

swalker2m:
Before the new target model was created, the only way to duplicate a target was via XML export followed by import.  Anything else required too much special case copying of individual fields and carried the risk of missing information as the target model evolved over time.  With the new target model all target classes are immutable except the `SPTarget` wrapper itself.  This makes duplication trivial and avoids any rounding errors introduced by passing the data through a `String` representation.


cquiroz:
👍

PR #1007, opened 2016-06-07
REL-2262: fix CAL/ENG ProgramId formatting

swalker2m:
CAL and ENG program ids incorporate the date to which they correspond in format YYYYMMDD.  Unfortunately `ProgramId`, our program id parser/formatter, incorrectly formats CAL and ENG months and dates without a leading 0 in the case of single digit months and days.  Since the OT "Open" dialog works with `ProgramId` in order to accomplish its filtering, CAL and ENG programs are not handled correctly.  In particular, the "Remote" filter determines whether a program is local by asking a `ProgramId` for its corresponding `SPProgramID` and using that to lookup the program in the local ODB.  Since the `SPProgramID` is not correct, the filter fails.


cquiroz:
Looks good to me :+1:

PR #1008, opened 2016-06-08
BAGS feedback mechanism

sraaphorst:
This comprises the BAGS feedback mechanism, which appears at the bottom of the target editor, in the area shared with the guiding quality for targets.

In order to accomplish this, we register target editors with the `BagsManager` and process changes to `BagsState`, which are translated into a minimal `BagsStatus` which is all that is required to display the status for an observation.

In most cases, the lookups are nearly immediate to the point that the actual status information does not seem to have the opportunity to display: I'm not sure if we want to do anything about that, but I would imagine not.

I've included an example of output where BAGS has failed to execute due to the network being disconnected, which is a state that persists long enough to be viewed.
bags

Also note that `GuidingFeedback` components were renamed since this information now comprises BAGS feedback as well as guiding feedback, and thus is no longer an accurate name.

Finally, the spinner has been replaced to allow for transparency.


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -32,8 +33,18 @@ import Scalaz._
...
+/** The status of BAGS for a given observation, for UI feedback purposes. */
+sealed class BagsStatus
+
+object BagsStatus {
+  case object ErrorStatus                extends BagsStatus
+  case object PendingStatus              extends BagsStatus
+  case object RunningStatus              extends BagsStatus
+  case class  FailureStatus(why: String) extends BagsStatus
+}
+
```
tpolecat: I don't quite see why we need this new data type. It looks like all the information is already in `BagsState`.

swalker2m: That was my suggestion, but maybe I steered @sraaphorst  wrong on this one.  I didn't want the state transition methods being called in the UI but thinking about it for a second there's no real reason to worry -- they don't _do_ anything other than return a new state and an unexecuted action.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
```diff
@@ -103,6 +103,12 @@ public void textBoxKeyPress(final TextBoxWidget tbwe) {
         });
 
         _agsPub.subscribe((obs, oldOptions, newOptions) -> updateGuiding());
+
+        BagsManager.addBagsStatusListener((key, oldStatus, newStatus) -> {
+            // Need to compare by reference here.
+            if (key.equals(getContextObservation().getNodeKey()))
+                updateGuiding();
```
swalker2m: I wonder if `updateGuiding` is overly broad in that it also configures other widgets.  The state change event can happen at any time, even while editing some other control which would then be reset?  Perhaps it would make sense to separate the feedback update into its own method.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
tpolecat: Could we make the transition methods `private[BagsManager]` or something if we're concerned about that? As you say it's just a potential confusion, not dangerous.

cquiroz:
Could we squash or give a proper message to the `WIP` commits?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
swalker2m: Yeah, I think that's a better idea than introducing a new class.  Sorry about that Sebastian!

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -265,6 +277,36 @@ object BagsManager {
...
+      obsMap <- stateMap.lookup(ProgKey(p))
+      s      <- obsMap.lookup(ObsKey(o))
+      st     <- s.status
+    } yield st
+
+  // Listeners for changes to the Bags Status.
+  // Management must be done on the Swing EDT.
+  private var listeners: List[BagsStatusListener] = Nil
+
+  def addBagsStatusListener(l: BagsStatusListener): Unit = {
```
cquiroz: Could it make sense that instead of checking if we are on the EDT we just run it on `Swing.onEDT`?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
sraaphorst: I think all of that sounds reasonable. I'll make those changes and push a new commit.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -305,7 +347,7 @@ object BagsManager {
         _ <- logTransition(state, newState)
         _ <- sideEffect
       } yield ()
-
+      fireBagsStatusChanged(obs.getNodeKey, state.status, newState.status)
       action.unsafePerformIO()
```
swalker2m: This may be a bit nitpicky but the status change hasn't actually happened until after the action is performed.  It might be best to move the event triggering after the action?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/TargetFeedbackEditor.scala
```diff
@@ -0,0 +1,70 @@
...
+
+      // If the BAGS row is defined, then use it. If not, create the rows corresponding to the analysis.
+      // NOTE that is target.isTooTarget, we don't want an analysis.
+      bagsRow.fold(
+        if (target.isTooTarget) Nil
+        else ctxOpt.asScalaOpt.map(TargetGuidingFeedback.targetAnalysis(_, mt, target)).getOrElse(Nil)
+      )(List(_))
+    }
+
+    if (rows.isEmpty)
```
cquiroz: I guess only this last bit needs to run on EDT. I'm wondering if the bits above like reading the magnitude table and  doing target analysis could block the ui

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
sraaphorst: No worries :-). Should be a fairly minor change!

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
swalker2m: I think that is probably okay but the semantics are a bit different.  `Swing.onEDT` will add an event to the queue and return immediately so the listener isn't added yet at the time the method returns.  I don't know whether that matters here though.  We could at least break out the code for checking whether we're on the EDT so it isn't repeated.

````scala
private def assertingSwingEDT[A](message: String)(block: => A): A = 
  if (SwingUtilities.isEventDispatchThread) block
  else throw new IllegalStateException(message)
````

then calling it like

````scala
def addBagsStatusListener(l: BagsStatusListener): Unit =
  assertingSwingEDT("BagsStatusListeners can only be added on the Swing EDT.") {
    if (!listeners.contains(l)) listeners = l :: listeners
  }
````

or something.

Also, the `statusLookup` (which will become `stateLookup` I suppose) should also guarantee that it is being done on the EDT.

File: bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/TargetFeedback.scala [hunk 6]
```diff
@@ -155,26 +147,62 @@ object GuidingFeedback {
...
+    layout(feedbackLabel) = new Constraints {
+      weightx = 1.0
+      fill = Fill.Both
+    }
+  }
+
+  case object ErrorStatusRow   extends BagsStatusRow(LIGHT_SALMON, "An error has occurred when trying to run BAGS.", None)
+  case object PendingStatusRow extends BagsStatusRow(BANANA, "Waiting for BAGS to run...", spinner)
+  case object RunningStatusRow extends BagsStatusRow(BANANA, "BAGS is running...", spinner)
+  case class  FailureStatusRow(why: String) extends BagsStatusRow(LIGHT_SALMON, s"BAGS failed: $why", None)
```
swalker2m: I suppose we could use the "eclipse/error.gif" icon for the error cases.

swalker2m: This is more of a question for science but I would guess that if the state is idle and yet there is no AGS star for an eligible observation we should report that.  "Automatic guide star lookup found nothing." or something. Otherwise as a user I think I'd be left wondering if something is broken.

swalker2m:
I think this is looking good.  One way to test it is to open up a large program in which it takes a while for BAGS to work through all the observations.  See GS-2016A-Q-80 for example.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
sraaphorst: The `private[BagsManager]` idea generates compiler errors because `BagsManager` is not an enclosing class. Not sure what the best way to work around this is: any advice?

swalker2m: `private[ags]` is probably fine.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -33,7 +33,7 @@ import scalaz.effect.IO
 import scalaz.effect.IO.ioUnit
 
 /** The state of AGS lookup for a single observation. */
-sealed trait BagsState {
+sealed class BagsState {
```
tpolecat: Either `sealed trait` or `sealed abstract class` pls.

sraaphorst: Agreed. I already changed it to sealed abstract class, but it should be sealed trait, I suppose.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
```diff
@@ -103,6 +103,12 @@ public void textBoxKeyPress(final TextBoxWidget tbwe) {
         });
 
         _agsPub.subscribe((obs, oldOptions, newOptions) -> updateGuiding());
+
+        BagsManager.addBagsStateListener((key, oldStatus, newStatus) -> {
+            // Need to compare by reference here.
+            if (key.equals(getContextObservation().getNodeKey()))
+                updateTargetFeedback();
+        });
```
swalker2m: Isn't the comment wrong there?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
swalker2m: Yes, I think this should just be a `sealed trait`.

swalker2m: We were going to make the transitions like `edit` `private[ags]` or something?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/TargetFeedback.scala
```diff
@@ -155,26 +146,67 @@ object GuidingFeedback {
...
+      horizontalAlignment = Alignment.Left
+    }
+
+    layout(feedbackLabel) = new Constraints {
+      weightx = 1.0
+      fill = Fill.Both
+    }
+  }
+
+  case object NoStarsRow      extends BagsStateRow(BANANA, "Automatic guide star lookup found nothing.", warningIcon)
```
swalker2m: Nice.

swalker2m:
Except for a couple of trivial tweaks mentioned in the comments, I think this is looking good.  👍

sraaphorst:
Sorry, @swalker2m ... I made a lot of the changes recommended here but got distracted (so many shiny things) by the build improvements and forgot to commit / push. I have done that now, and I think that all recommended changes have been made except for collapsing the WIP commits from earlier as per the recommendation by @cquiroz - I just need to figure out how to do that and then I will.

We still need to make branches as appropriate.

PR #1009, opened 2016-06-09
REL-2831: compress ephemerides

tpolecat:
This is an experimental PR to see if we can make it less impossible to work with gigantic programs with many ephemerides. This stores data as a compressed blob, expanded lazily into a transient. The blob is written as Base64 in the XML.

If this seems ok we will need to write an updater to read earlier 16A files which have big ephemerides in their XML.

A test program with 1000 copies of IO's ephemeris results in ~24m of XML, which is much better than before. Currently this would be ~500m of XML.

### DIFF
--- a/bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/Deflated.scala
+++ b/bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/Deflated.scala
@@ -4,25 +4,35 @@ import java.io._
 import java.util.Arrays
 import java.util.zip.{ Deflater, Inflater }
 
-/**
- * A compressed value.
- * @param data the underlying byte array
- */
-final class Deflated[A] private (val data: Array[Byte]) extends Serializable {
+import scalaz.Functor
+
+/** A value that has been serialized and compressed. */
+final class Deflated[A] private (
+  private val blob: Array[Byte], 
+  private val k: Object => A
+) extends Serializable {
+
+  /** Return a *reference* to the deflated blob. */
+  def unsafeToByteArray: Array[Byte] =
+    blob
+
+  /** Return a *copy* of the deflated blob. */
+  def toByteArray: Array[Byte] =
+    blob.clone
 
   /** Size of the compressed blob in bytes. */
   def size: Int =
-    data.length
+    blob.length
 
   /** 
    * Inflate and return the value. Note that this operation is not memoized; each call will
-   * require re-inflating the compressed blob.
+   * re-inflate the compressed blob.
    */
   def inflate: A = {
 
     // First inflate
     val inflater = new Inflater();
-    inflater.setInput(data);
+    inflater.setInput(blob);
     val baos = new ByteArrayOutputStream();
     val buf = new Array[Byte](Deflated.BUF_SIZE);
     while (!inflater.needsInput) {
@@ -36,23 +46,27 @@ final class Deflated[A] private (val data: Array[Byte]) extends Serializable {
     val ois  = new ObjectInputStream(bais)
 
     // Done
-    ois.readObject.asInstanceOf[A]
+    k(ois.readObject)
 
   }
 
+  /** Return a new `Deflated` that applies `f` to its inflated value. */
+  def map[B](f: A => B): Deflated[B] =
+    new Deflated(blob, k andThen f)
+
   override def equals(a: Any): Boolean =
     a match {
-      case d: Deflated[_] => Arrays.equals(data, d.data)
+      case d: Deflated[_] => Arrays.equals(blob, d.blob)
       case _              => false
     }
 
   override def hashCode: Int =
-    Arrays.hashCode(data)
+    Arrays.hashCode(blob)
 
 }
 
 object Deflated {
-  val BUF_SIZE = 1024 * 16
+  private val BUF_SIZE = 1024 * 16
 
   /** 
    * Construct a deflated object by serializing and compressing the passed value. This method will
@@ -80,19 +94,29 @@ object Deflated {
       val len = deflater.deflate(buf)
       baos.write(buf, 0, len)
     }
-    deflater.end()           // (1) these two lines are important; see note in CompressedString.java
-    System.runFinalization() // (2)
+    // If you don't call end() the deflater leaks system heap on Linux. This is documented in 
+    // 6293787. It still leaks slowly but running finalizers seems to fix it.
+    deflater.end()
+    System.runFinalization()
     
     // Done
-    new Deflated(baos.toByteArray)
+    unsafeFromByteArray(baos.toByteArray)
   
   }
 
-  /**
-   * Construct a deflated object from the given byte array, which is *not* copied, under the
-   * assertion that its can be decompressed into a value of type A.
-   */
-  def unsafeFromBytes[A](data: Array[Byte]): Deflated[A] =
-    new Deflated[A](data)
+  /** Construct a deflated object using a *reference* to the passed byte array. */
+  def unsafeFromByteArray[A](blob: Array[Byte]): Deflated[A] =
+    new Deflated(blob, _.asInstanceOf[A])
+
+  /** Construct a deflated object using a *copy* of the passed byte array. */
+  def fromByteArray[A](blob: Array[Byte]): Deflated[A] =
+    unsafeFromByteArray(blob.clone)
+
+  /** Deflated is a covariant functor. */
+  implicit val DeflatedFunctor: Functor[Deflated] =
+    new Functor[Deflated] {
+      def map[A, B](fa: Deflated[A])(f: A => B): Deflated[B] =
+        fa.map(f)
+    }
 
 }
--- a/bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/Deflated.scala
+++ b/bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/Deflated.scala
@@ -1,0 +1,98 @@
+package edu.gemini.spModel.core
+
+import java.io._
+import java.util.Arrays
+import java.util.zip.{ Deflater, Inflater }
+
+/**
+ * A compressed value.
+ * @param data the underlying byte array
+ */
+final class Deflated[A] private (val data: Array[Byte]) extends Serializable {
+
+  /** Size of the compressed blob in bytes. */
+  def size: Int =
+    data.length
+
+  /** 
+   * Inflate and return the value. Note that this operation is not memoized; each call will
+   * require re-inflating the compressed blob.
+   */
+  def inflate: A = {
+
+    // First inflate
+    val inflater = new Inflater();
+    inflater.setInput(data);
+    val baos = new ByteArrayOutputStream();
+    val buf = new Array[Byte](Deflated.BUF_SIZE);
+    while (!inflater.needsInput) {
+      val len = inflater.inflate(buf);
+      baos.write(buf, 0, len);
+    }
+    val bs = baos.toByteArray
+
+    // Now deserialize
+    val bais = new ByteArrayInputStream(bs)
+    val ois  = new ObjectInputStream(bais)
+
+    // Done
+    ois.readObject.asInstanceOf[A]
+
+  }
+
+  override def equals(a: Any): Boolean =
+    a match {
+      case d: Deflated[_] => Arrays.equals(data, d.data)
+      case _              => false
+    }
+
+  override def hashCode: Int =
+    Arrays.hashCode(data)
+
+}
+
+object Deflated {
+  val BUF_SIZE = 1024 * 16
+
+  /** 
+   * Construct a deflated object by serializing and compressing the passed value. This method will
+   * throw if the value is not serializable.
+   */
+  def apply[A](a: A): Deflated[A] = {
+
+    // First serialize
+    val bs: Array[Byte] = {
+      val baos = new ByteArrayOutputStream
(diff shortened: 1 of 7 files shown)


tpolecat:
👎  can't read back XML

bundle/edu.gemini.spModel.pio/src/main/scala/edu/gemini/spModel/pio/codec/ParamCodec.scala
```diff
@@ -1,5 +1,8 @@
 package edu.gemini.spModel.pio.codec
 
+import edu.gemini.spModel.core.Deflated
+import sun.misc.{BASE64Decoder, BASE64Encoder}
```
cquiroz: In Java 8 there are Base64 Encoder/Decoders:
https://docs.oracle.com/javase/8/docs/api/java/util/Base64.html

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/Deflated.scala
```diff
@@ -0,0 +1,98 @@
...
+
+import java.io._
+import java.util.Arrays
+import java.util.zip.{ Deflater, Inflater }
+
+/**
+ * A compressed value.
+ * @param data the underlying byte array
+ */
+final class Deflated[A] private (val data: Array[Byte]) extends Serializable {
```
tpolecat: Need to make the `data` field private; I can't cope with it being available directly. The only time it's used externally is for serialization so a copy then is ok.

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/Ephemeris.scala
```diff
@@ -2,7 +2,19 @@ package edu.gemini.spModel.core
...
+    ==>>.fromList(compressedData.inflate.map { case (t, r, d) =>
+      t -> Coordinates.fromDegrees(r, d).getOrElse(sys.error(s"corrupted ephemeris data: $t $r $d"))
+    })
+
+  /** True if the ephemeris data is compressed (will be uncompressed by need). */
+  def isCompressed: Boolean = {
+    val f = getClass.getDeclaredField("data")
+    f.setAccessible(true)
+    f.get(this) == null
+  }
```
tpolecat: This is only for testing and it's a side-effect, so move it into the tests.

bundle/edu.gemini.spModel.pio/src/main/scala/edu/gemini/spModel/pio/codec/ParamCodec.scala
tpolecat: Cool, thanks. I didn't even pay attention to what got imported. I'll switch.

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/Deflated.scala
```diff
@@ -0,0 +1,98 @@
...
+    deflater.setInput(bs)
+    deflater.finish()
+    val baos = new ByteArrayOutputStream()
+    val buf = new Array[Byte](BUF_SIZE)
+    while (!deflater.finished) {
+      val len = deflater.deflate(buf)
+      baos.write(buf, 0, len)
+    }
+    deflater.end()           // (1) these two lines are important; see note in CompressedString.java
+    System.runFinalization() // (2)
```
tpolecat: Copy/paste the comment here, in case `CompressedString` goes away.

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/Deflated.scala
```diff
@@ -0,0 +1,98 @@
...
+    new Deflated(baos.toByteArray)
+  
+  }
+
+  /**
+   * Construct a deflated object from the given byte array, which is *not* copied, under the
+   * assertion that its can be decompressed into a value of type A.
+   */
+  def unsafeFromBytes[A](data: Array[Byte]): Deflated[A] =
+    new Deflated[A](data)
```
tpolecat: This is only used for deserialization, so go ahead and make a copy to keep things pure.

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/Ephemeris.scala
```diff
@@ -2,7 +2,19 @@ package edu.gemini.spModel.core
 
 import scalaz._, Scalaz._
 
-final case class Ephemeris(site: Site, data: Long ==>> Coordinates) {
+final class Ephemeris(val site: Site, val compressedData: Deflated[List[(Long, Float, Float)]]) extends Serializable {
+
+  @transient lazy val data: Long ==>> Coordinates =
+    ==>>.fromList(compressedData.inflate.map { case (t, r, d) =>
+      t -> Coordinates.fromDegrees(r, d).getOrElse(sys.error(s"corrupted ephemeris data: $t $r $d"))
+    })
```
tpolecat: Add a comment here about the behavior of `@transient lazy val`.

tpolecat:
Ok I added code that times out the inflated ephemeris data after three seconds, primarily to allow us to reclaim memory after the initial P2 check. The target editor pokes the ephemeris once a second to update the "now" coordinates, so the ephemeris for the "current" target ends up being available until you switch to another one.

There is logging so you can see what's going on, but it should be turned down/off before merging because it's not actually useful. (Might be worth turning down the BagsManager logging at the same time.)

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/Deflated.scala
```diff
@@ -4,25 +4,35 @@ import java.io._
...
- * A compressed value.
- * @param data the underlying byte array
- */
-final class Deflated[A] private (val data: Array[Byte]) extends Serializable {
+import scalaz.Functor
+
+/** A value that has been serialized and compressed. */
+final class Deflated[A] private (
+  private val blob: Array[Byte], 
+  private val k: Object => A
```
cquiroz: How can you serialize `Object => A`?

tpolecat: Functions are serializable but problematic, so that's a good point. I'm not using the functor so I'll get rid of it.

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/Deflated.scala
```diff
@@ -0,0 +1,122 @@
...
+  def fromByteArray[A](blob: Array[Byte]): Deflated[A] =
+    unsafeFromByteArray(blob.clone)
+
+  /** Deflated is a covariant functor. */
+  implicit val DeflatedFunctor: Functor[Deflated] =
+    new Functor[Deflated] {
+      def map[A, B](fa: Deflated[A])(f: A => B): Deflated[B] =
+        fa.map(f)
+    }
+
```
swalker2m: Maybe an `Equals` instance?

tpolecat:
Andy is trying this out locally. I'm going on vacation in a few hours and won't be back until July 5 so someone else will need to steer this onto the runway. The PIO converter will be a pain, sorry. I'll at least try to stub it out before I leave.

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/Deflated.scala
tpolecat: Good idea.

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/Ephemeris.scala
```diff
@@ -1,9 +1,46 @@
...
+    }
+
+    // Reset the cache timer to a point the future
+    if (_task != null) _task.cancel()
+    _task = new TimerTask {
+      def run(): Unit = {
+        logger.info("Discarding inflated ephemeris.")
+        Ephemeris.this.synchronized(_data = null)
+      }
+    }
```
swalker2m: This is quite a clever idea, and one that probably wouldn't have occurred to me.  I wonder about the performance in large programs in the OT though.  The OT is insistent about running P2 checks on every node in the program (probably multiple times) whenever anything changes.  Since there must be a P2 check or two that look for coordinates this means we'll be constantly deflating ephemeris data, I'm guessing.  Something similar may happen for observation query functors for the ODB.  For example, when building up the mini-model for QPT.

We'll need memory enough for the original deflated ephemeris data as well as the compressed data during those times.  It is probably worth profiling memory use while editing a large program with ephemeris data.

andrewwstephens:
This looks good.  I don't have much to add, but here are the numbers I got:

2016B.1.1.2      ODB,   XML
1 sidereal       16K,    9K
1 nonsidereal   238K,  1.1M
10 nonsidereal  2.2M,   11M
20 nonsidereal  4.5M,   21M
30 nonsidereal  6.7M,   32M
Per ephemeris ~ 114K,  546K

Test OT          ODB,   XML
1 sidereal       16K,    9K
1 nonsidereal    34K,   56K
10 nonsidereal  174K,  551K
20 nonsidereal  545K,  1.1M
30 nonsidereal  922K,  1.7M
Per ephemeris ~  15K,   29K

swalker2m:
@andrewwstephens what do these numbers correspond to?

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/Ephemeris.scala
swalker2m: I think the vast majority of the time we're just asking for the coordinates at the scheduling block start time.  I would like to experiment with this PR as it is vs. one additional tweak.  The coordinate lookup method could just remember the time and coordinate of the last value it was asked to find.  That way if the next request is for the same time it can return the coordinates without inflating the ephemeris data.  If the request is for a different time then it can go ahead and inflate and remember the new (time, coordinate) pair.  My suspicion is that this will save a lot of churn that will otherwise be caused by P2 checking and ODB queries.

tpolecat: That thought occurred to me last night also, so yeah worth a try.

andrewwstephens:
@swalker2m - The first size is that of the ODB program (as displayed in the OT "Open" dialog).  The second size is of the exported XML file.

The most relevant values are "Per ephemeris", which show that in the database each ephemeris is ~114K uncompressed vs 15K compressed (13%), and in the exported XML each ephemeris is ~546K uncompressed vs 29K compressed (5%).

tpolecat:
Hi, I'm back. I thought this was going to be merged while I was on vacation. Have we decided not to do this?

swalker2m:
Ultimately we (@arturog8m and I) decided it was too big of a change for the patch release.  Given how even small updates have unintended consequences in this codebase, we worried about including it without any time for testing.

I think we can merge it now into the `develop` branch though 👍 . In the meantime, there is the lame menu option work around that trims ephemeris data.  There is still an [open task](http://jira.gemini.edu:8080/browse/REL-2831) about ephemeris size that we might also want to consider for December.

tpolecat:
Rebased. This is on `develop` already so I'll merge when CI goes green (without objection).

swalker2m:
> Rebased. This is on develop already so I'll merge when CI goes green (without objection).

That's what I do too.

tpolecat:
Hm, so it looks like the compression from  `Double` to `Float` introduces an error on the order of ~1mas; in the `PlutoDemotionTest` the RA value `15:41:38.380` becomes `15:41:38.379` which caused the test to fail. @andrewwstephens is this precise enough or do we need to use a higher-precision type?

andrewwstephens:
0.001 second corresponds to  ~0.015 arcseconds,  but that should be fine for the low-medium resolution ephemerides stored in the OT.  Supposedly the high resolution ephemerides stored in files by the ODB remain unaffected.

bryanmiller:
Is this a one-time error or is there a chance that the blob can get de/recompressed multiple times?

tpolecat:
That's a good question. I think `Double -> Float -> Double` is idempotent but I need to ask a pro.

tpolecat:
```scala
scala> def f(d: Double) = d.toFloat.toDouble
f: (d: Double)Double

scala> forAll { (d: Double) => f(d) == f(f(d)) } .check
+ OK, passed 100 tests.
```

tpolecat:
@bryanmiller existing ephemerides going into 17A will be truncated to 32 bits of precision, as will new ephemerides fetched from Horizons by the OT. As stated above this does not affect the full-precision TCS files maintained by the ODB.

swalker2m:
@tpolecat Do we have a missing data migration for 2017A or should this be taken care of by the code in this PR?  It appears to me that this might have fallen through the cracks and the 2017A nonsidereal target PIO [cannot read 2016B nonsidereal target ephemeris data](http://jira.gemini.edu:8080/browse/REL-2971). 

I can work on this tomorrow if necessary.  Should be easy-ish.

tpolecat:
> If this seems ok we will need to write an updater to read earlier 16A files which have big ephemerides in their XML.

So yes, the conversion is missing, and yes should be easy-ish. I can work on it tomorrow if you don't get a chance or don't feel like it. Let me know.

PR #1010, opened 2016-06-09
REL-2831: purge ephemeris action

swalker2m:
This PR introduces a new "Edit" menu item that, when selected, purges ephemeris data in all non-sidereal observations that the user has permission to edit.  In particular, it replaces the full ephemeris table with a single-element entry at the scheduling block start.  This can reduce the amount of memory occupied by the program in an observing database, including the OT's local database.  This is, hopefully, a stop-gap option for 16B.  The idea is that this action will only be needed in very special cases where the program has many non-sidereal target instances.

~~The usefulness of this item is somewhat questionable though because it can only update observations that the user can edit.  Otherwise, the updated observations will only be rejected by the merge algorithm upon sync.  Unfortunately that means observed observations cannot be purged even by staff members despite the fact that observed observations are precisely those for which full ephemeris data is least useful.  Sad!~~

__UPDATE:__ The sync allows staff members to "edit" observed observations because the OT allows them to toggle observation status, make a change, and toggle back to observed.  We can allow the ephemeris purge to reset observed observation ephemeris after all.  Sorry for the confusion.

### DIFF
--- a/bundle/edu.gemini.spdb.shell/src/main/java/edu/gemini/spdb/shell/osgi/Commands.java
+++ b/bundle/edu.gemini.spdb.shell/src/main/java/edu/gemini/spdb/shell/osgi/Commands.java
@@ -6,9 +6,12 @@ import edu.gemini.pot.spdb.DBAbstractQueryFunctor;
 import edu.gemini.pot.spdb.IDBDatabaseService;
 import edu.gemini.pot.spdb.IDBQueryFunctor;
 import edu.gemini.pot.spdb.IDBQueryRunner;
+import edu.gemini.shared.util.immutable.ImEither;
 import edu.gemini.spModel.core.SPBadIDException;
 import edu.gemini.spModel.core.SPProgramID;
 import edu.gemini.spModel.io.SpImportService;
+import edu.gemini.spdb.shell.misc.EphemerisPurgeCommand;
+import static edu.gemini.spdb.shell.misc.EphemerisPurgeCommand.*;
 import edu.gemini.spdb.shell.misc.ExportXmlCommand;
 import edu.gemini.spdb.shell.misc.ImportXmlCommand;
 import edu.gemini.spdb.shell.misc.LsProgs;
@@ -41,17 +44,9 @@ class Commands {
         IDBQueryFunctor doQuery(IDBQueryRunner run, IDBQueryFunctor fun);
     }
 
-    private static final Query PROG_QUERY = new Query() {
-        public IDBQueryFunctor doQuery(final IDBQueryRunner run, final IDBQueryFunctor fun) {
-            return run.queryPrograms(fun);
-        }
-    };
+    private static final Query PROG_QUERY = IDBQueryRunner::queryPrograms;
 
-    private static final Query PLAN_QUERY = new Query() {
-        public IDBQueryFunctor doQuery(final IDBQueryRunner run, final IDBQueryFunctor fun) {
-            return run.queryNightlyPlans(fun);
-        }
-    };
+    private static final Query PLAN_QUERY = IDBQueryRunner::queryNightlyPlans;
 
     // return all the program ids formatted in columns
     public String lsprogs() {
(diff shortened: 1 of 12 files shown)


swalker2m:
We could also add a shell command to purge programs from the telnet command line.  That wouldn't be limited by permission constraints.

tpolecat:
Hm I thought the intent was that you could purge ephemerides from observed observations since the information is not needed anymore. When would you want to do this for active observations?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/viewer/action/PurgeEphemerisAction.scala
```diff
@@ -0,0 +1,108 @@
...
+    | have permission to edit?  This will decrease the amount of memory required to edit
+    | the program by keeping only target coordinates at the scheduling block time.
+    |
+    | Ephemeris data can always be re-fetched later for each observation in its Target
+    | Environment component.
+  """.stripMargin
+
+  putValue(AbstractViewerAction.SHORT_NAME, "Purge Ephemeris")
+  putValue(Action.SHORT_DESCRIPTION, "Purge ephemeris data in all non-sidereal observations.")
+  setEnabled(true)
```
cquiroz: Will it enable it for all users?

cquiroz:
I can't comment on the reason to do this but codewise this seems 👍

swalker2m:
@tpolecat I only realized that it had to be limited to editable observations after I got started but decided to finish it regardless. I'm not sure how useful it will be in practice, or whether we should just switch to a shell command. Part of the purpose of this PR is to have that discussion.

That said, the specific case of program GN-16A-Q-32 came about because someone loaded up an entire semester's worth of yet-to-be-executed observations with ephemeris data and then had no way of exporting the program w/o raising the max heap size.

tpolecat:
Ok, seems reasonable as a safety net. Certainly useful as a shell command that [only] does completed observations.

btw with compression my test program with 1000 distinct ephemerides uses 17M in the database and 24M as XML, which is a lot of data but not obviously problematic.

tpolecat:
Also this is probably necessary just to fix Q-32 since it can't be exported or imported without special VM settings at the moment.

swalker2m:
I added a plethora of ephemeris purging options:

1. Updated the menu item in the OT to allow staff to choose between purging only observed observations vs. all observations.
2. Added a `purgeEphemeris [all | observed]` shell command (defaulting to `observed` only).
3. Added a `purgeEphemeris` cron job which purges ephemeris data from all observed observations in the whole database.

Now we can decide whether to ever run the shell command or cron job.

bundle/edu.gemini.spdb.shell/src/main/java/edu/gemini/spdb/shell/osgi/Commands.java
```diff
@@ -190,4 +186,18 @@ public String rmprog(String[] programIds) {
     public String rmprog(List<String> programIds) {
         return rmprog(programIds.toArray(new String[programIds.size()]));
     }
+
+    public String purgeEphemeris(String programId) {
+        return purgeEphemeris(programId, ObservedOnly$.MODULE$.displayValue());
```
cquiroz: So the abstract class trick didn't work?

cquiroz:
Looks good to me :+1:

arturog8m:
👍  looks good.

bundle/edu.gemini.spdb.shell/src/main/scala/edu/gemini/spdb/shell/misc/EphemerisPurgeCommand.scala
```diff
@@ -0,0 +1,52 @@
...
+
+    def fromDisplayValue(s: String): Option[PurgeOption] =
+      AllValues.find(_.displayValue === s)
+
+    def usageString: String =
+      AllValues.map(_.displayValue).mkString("[", " | ", "]")
+  }
+
+  /** Option to purge ephemeris data from all observations in the program. */
+  case object All extends PurgeOption
```
sraaphorst: Is there a reason why you don't just have `include` as abstract in `PurgeOption` and then implement it in the case objects?

swalker2m: I think either way is fine.  I guess I like doing it that way when the implementations are short so you can easily see them all at once.  It keeps the definitions of the `case object`s short too so it's easier to see them all as well.

PR #1012, opened 2016-06-14
Improve build specification for itc

cquiroz:
PR to include these changes on the release branch

### DIFF
--- a/app/itc/build.sbt
+++ b/app/itc/build.sbt
@@ -18,7 +18,10 @@ ocsAppManifest := {
     configs = List(
       common(v),
         with_gogo(v),
-          itc(v)
+          itc(v),
+          with_remote_gogo(v),
+            itctest(v),
+            itcproduction(v)
     )
   )
 }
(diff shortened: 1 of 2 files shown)


app/itc/build.sbt
```diff
@@ -91,3 +108,31 @@ def itc(version: Version) = AppConfig(
...
+  props = Map(
+    "org.osgi.service.http.port" -> "9080",
+    "osgi.shell.telnet.ip"       -> "172.16.55.80"
+  )
+) extending List(with_remote_gogo(version))
+
+// ITC production
+def itcproduction(version: Version) = AppConfig(
+  id = "itcproduction",
+  distribution = List(Linux32),
```
arturog8m: Why the distribution is Linux 32 and not Linux 64?

cquiroz: As I mentioned in hipchat the itc hosts are **suprisingly** still Linux 32

arturog8m: thanks, ok. We should plan to upgrade those next. 👍

cquiroz: certainly, I thought they were all 64 bits already

PR #1013, opened 2016-06-15
REL-2841: Fix PPMXL proper motion units

cquiroz:
This PR properly converts proper motion read from the PPMXL catalog

### DIFF
--- a/bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/AngularVelocity.scala
+++ b/bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/AngularVelocity.scala
@@ -22,10 +22,18 @@ case class DeclinationAngularVelocity(velocity: AngularVelocity)
 
 object AngularVelocity {
 
+  /**
+    * Builds an AngularVelocity expressed on degrees per year, converting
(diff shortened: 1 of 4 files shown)


bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/AngularVelocity.scala
```diff
@@ -22,10 +22,17 @@ case class DeclinationAngularVelocity(velocity: AngularVelocity)
 
 object AngularVelocity {
 
+  /**
+    * Builds an AngularVelocity expressed on degrees per year
+    */
+  def fromDegreesPerYear(dpy: Double): AngularVelocity =
+    AngularVelocity(dpy * DegreesYearToArcSecsYear)
```
swalker2m: Doesn't the `AngularVelocity` constructor expect a `Double` representing mas, not arcsec?

cquiroz: Yes, bad constant name, it should be `DegreesYearToMilliArcSecYear`

cquiroz:
Updated according to PR

PR #1014, opened 2016-06-16
Fix for REL-2823 

jdnavarro:
Change warning to error when there is no scheduling block for nonsidereal targets

Additionally, the error is now reported next to scheduling form, on the `Major
Body` node, instead of the `Observation` parent node.

### DIFF
--- a/bundle/edu.gemini.p2checker/src/main/scala/edu/gemini/p2checker/target/NonSiderealTargetRules.scala
+++ b/bundle/edu.gemini.p2checker/src/main/scala/edu/gemini/p2checker/target/NonSiderealTargetRules.scala
@@ -40,14 +40,14 @@ final class NonSiderealTargetRules extends IRule {
 
   def checkNoSchedulingBlock(es: ObservationElements): IP2Problems =
     new P2Problems <| { p2p =>
-      es.getTargetObsComp.asScalaOpt.foreach { toc =>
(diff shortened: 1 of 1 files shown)


swalker2m:
Looks great. 👍 Just an FYI, the "Major Body" node is the target component node in general.  It says "Major Body" because the science target happens to be a non-sidereal (i.e., local to the solar system) target that is considered by the horizons non-sidereal service to be a "major body".  The target component in general will display whatever type of science target the observation has in its label.

I think it was a good idea to move this warning (now an error) down to the target component itself.  I believe it was there on the observation before because it used to be that the scheduling time was specified on the observation node.  When you mark the JIRA ticket resolved, add a comment saying that you moved it just in case Andy or Bryan have any feedback on that.

andrewwstephens:
Would it be possible to show the error on *both* the Target node and the top-level Observation node?  IMHO the top-level should show all warnings and errors in the observation, otherwise users will only see them if they expand the observation.

swalker2m:
@andrewwstephens Right, I believe that's how it will work now.  If you look at an Observation node that contains nodes with issues, they appear also in the Observation node.

andrewwstephens:
Perfect!

jdnavarro:
@andrewwstephens @swalker2m I just confirmed the error also shows up on the top level node, too.

arturog8m:
First merge in your first official day! Well done! 👏

PR #1016, opened 2016-06-17
REL-2456 Report a warning when DTA X offset is not a multiple of y-pixel binning

jdnavarro:
Please make sure the dirtiness of the tests I wrote are tolerable. I felt quite nasty writing them ;)

### DIFF
--- a/bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/gmos/GmosRule.java
+++ b/bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/gmos/GmosRule.java
@@ -419,6 +419,9 @@ public final class GmosRule implements IRule {
         }
     };
 
+    public static final String GMOS_S_DTA_X_RULE_Y1_ID = PREFIX + "Y1";
+    public static final String GMOS_S_DTA_X_RULE_Y2_4_ID = PREFIX + "Y2_4";
+
     private static IConfigRule GMOS_S_DTA_X_RULE = new IConfigRule() {
 
         private static final String Y1   = "The GMOS-S Hamamatsu allowed DTA X range is -4 to +6 for Ybin=1";
@@ -430,9 +433,9 @@ public final class GmosRule implements IRule {
 
             final ISPProgramNode node = SequenceRule.getInstrumentOrSequenceNode(step, elems, config);
             if ((y == Binning.ONE) && (dtaX.intValue() < -4)) {
-                return new Problem(ERROR, PREFIX + "Y1", Y1, node);
+                return new Problem(ERROR, GMOS_S_DTA_X_RULE_Y1_ID, Y1, node);
             } else if ((y != Binning.ONE) && (dtaX.intValue() < -2)) {
-                return new Problem(ERROR, PREFIX + "Y2_4", Y2_4, node);
+                return new Problem(ERROR, GMOS_S_DTA_X_RULE_Y2_4_ID, Y2_4, node);
             } else {
                 return null;
             }
@@ -1611,6 +1614,11 @@ public final class GmosRule implements IRule {
         return (Binning) SequenceRule.getItem(config, Binning.class, CCD_Y_BINNING_KEY);
     }
 
+    private static DTAX getDtaXOffset(Config config) {
+        return (DTAX) SequenceRule.getInstrumentItem(config, InstGmosCommon.DTAX_OFFSET_PROP);
+    }
+
+
     private static Double getExposureTime(InstGmosCommon<?,?,?,?> inst, Config config) {
         // REL-196.  If there are no observes, there will be no observe exposure time.
         final Double obsExp = SequenceRule.getExposureTime(config);
@@ -1849,6 +1857,35 @@ public final class GmosRule implements IRule {
         }
     };
 
+    public static final String DTA_X_Y_MULTIPLE_BINNING_RULE_ID = PREFIX + "DTA_X_Y_MULTIPLE_BINNING_RULE";
+
+    /**
+     * REL-2456: We request a Phase II check to give a warning if the DTA-X
+     * position is not a multiple of the y-pixel binning.
+     */
+    private static IConfigRule DTA_X_Y_MULTIPLE_BINNING_RULE = new IConfigRule() {
+        private static final String errMsg = "The DTA X offset (%d) should be multiple of the Y-binning (%d)";
+
+        @Override
+        public Problem check(Config config, int step, ObservationElements elems, Object state) {
+            final int y = getYBinning(config).getValue();
+            final int dtaX = getDtaXOffset(config).intValue();
+
+            if (dtaX % y != 0) {
+                final String msg = String.format(errMsg, dtaX, y);
+                return new Problem(WARNING, DTA_X_Y_MULTIPLE_BINNING_RULE_ID, msg,
+                        SequenceRule.getInstrumentOrSequenceNode(step, elems));
+            }
+
+            return null;
+        }
+
+        @Override
+        public IConfigMatcher getMatcher() {
+            return IConfigMatcher.ALWAYS;
+        }
+    };
+
     /**
      * Register all the GMOS rules to apply
      */
@@ -1892,6 +1929,7 @@ public final class GmosRule implements IRule {
         GMOS_RULES.add(ROI_INVALID_RULE);
         GMOS_RULES.add(CUSTOM_ROI_NOT_DECLARED_RULE);
         GMOS_RULES.add(IFU_NO_SPATIAL_BINNING_RULE);
+        GMOS_RULES.add(DTA_X_Y_MULTIPLE_BINNING_RULE);
         GMOS_RULES.add(NO_P_OFFSETS_WITH_SLIT_SPECTROSCOPY_RULE);
         GMOS_RULES.add(new MdfMaskNameRule(Problem.Type.ERROR));
         GMOS_RULES.add(new MdfMaskNameRule(Problem.Type.WARNING));
--- a/bundle/edu.gemini.p2checker/src/test/scala/edu/gemini/p2checker/rules/GmosSDtaXRuleTest.scala
+++ b/bundle/edu.gemini.p2checker/src/test/scala/edu/gemini/p2checker/rules/GmosSDtaXRuleTest.scala
@@ -56,31 +56,42 @@ class GmosSDtaXRuleTest extends AbstractRuleTest {
     rule.check(elems).getProblems.asScala.toList
   }
 
-  private def expectSuccess(): Unit = {
-//    println(runCheck.mkString(",\n"))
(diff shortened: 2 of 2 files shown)


bundle/edu.gemini.p2checker/src/test/scala/edu/gemini/p2checker/rules/GmosSDtaXRuleTest.scala
```diff
@@ -58,9 +58,16 @@ class GmosSDtaXRuleTest extends AbstractRuleTest {
 
   private def expectSuccess(): Unit = {
 //    println(runCheck.mkString(",\n"))
-    assertTrue(runCheck.isEmpty)
+    assertTrue(! runCheck.exists { p =>
```
cquiroz: Small thing, from a style point of view perhaps we prefer no space on the negation `!runCheck.ex...`

bundle/edu.gemini.p2checker/src/test/scala/edu/gemini/p2checker/rules/GmosSDtaXRuleTest.scala
```diff
@@ -71,6 +78,11 @@ class GmosSDtaXRuleTest extends AbstractRuleTest {
 
   private def expectFailure2(): Unit = expectFailure("Ybin=2")
 
+  private def expectFailure3(): Unit =
+    assertTrue (runCheck.exists { p =>
```
cquiroz: Also on style, no space on the parameters? `assertTrue(runC...`

bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/gmos/GmosRule.java
```diff
@@ -1850,6 +1855,33 @@ public IConfigMatcher getMatcher() {
...
+            final int y = getYBinning(config).getValue();
+            final int dtaX = getDtaXOffset(config).intValue();
+
+            if (dtaX % y != 0) {
+                final String msg = String.format(errMsg, dtaX, y);
+                return new Problem(WARNING, PREFIX + "DTA_X_Y_MULTIPLE_BINNING_RULE", msg,
+                        SequenceRule.getInstrumentOrSequenceNode(step, elems));
+            }
+
+            return null;
```
cquiroz: I don't recall much of P2, is it correct to return `null` if there are no problems?

bundle/edu.gemini.p2checker/src/test/scala/edu/gemini/p2checker/rules/GmosSDtaXRuleTest.scala
```diff
@@ -71,6 +78,11 @@ class GmosSDtaXRuleTest extends AbstractRuleTest {
 
   private def expectFailure2(): Unit = expectFailure("Ybin=2")
```
cquiroz: Perhaps this could benefit of some more human-friendly names?

bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/gmos/GmosRule.java
arturog8m: unfortunately, yes, that's the convention

cquiroz:
Why do you think the tests are dirty?

bundle/edu.gemini.p2checker/src/test/scala/edu/gemini/p2checker/rules/GmosSDtaXRuleTest.scala
swalker2m: Maybe instead of 1, 2, 3 we could pass the `Set` of problem ids we aren't expecting to `expectSuccess` and set of problem ids we are expecting to `expectProblem`.

bundle/edu.gemini.p2checker/src/test/scala/edu/gemini/p2checker/rules/GmosSDtaXRuleTest.scala
```diff
@@ -71,6 +78,11 @@ class GmosSDtaXRuleTest extends AbstractRuleTest {
...
 
+  private def expectFailure3(): Unit =
+    assertTrue (runCheck.exists { p =>
+      p.getId() == "GmosRule_DTA_X_Y_MULTIPLE_BINNING_RULE"
```
swalker2m: We could expose the problem id in the corresponding GMOS rule so we can just reference it from the test.

jdnavarro:
@cquiroz, mainly because with `initGmos` there is no reference to hold to, it's all very *side-effectish*.

bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/gmos/GmosRule.java
```diff
@@ -1850,6 +1855,33 @@ public IConfigMatcher getMatcher() {
     };
 
     /**
+     * REL-2456: We request a Phase II check to give a warning if the DTA-X
+     * position is not a multiple of the y-pixel binning.
+     */
+    private static IConfigRule DTA_X_Y_MULTIPLE_BINNING_RULE = new IConfigRule() {
+        private static final String errMsg = "The DTA X offset (%d) should be multiple of the Y-binning (%d)";
+
```
swalker2m: Apropos of one of my comments below, for example here:

````java
    public static final String Id = PREFIX + "DTA_X_Y_MULTIPLE_BINNING_RULE"
````

and just make the entire `DTA_X_Y_MULTIPLE_BINNING_RULE` public too.  Then you can reference it with: `DTA_X_Y_MULTIPLE_BINNING_RULE.Id`.

**Comment by swalker2m** [comment]
@jdnavarro right, we're a bit stuck with the mutable science program model for now, but we're working on replacing it!  The test case sets up a science program programmatically (mutating a new science program until it is ready) and then runs the P2 checking on it.

File: bundle/edu.gemini.p2checker/src/test/scala/edu/gemini/p2checker/rules/GmosSDtaXRuleTest.scala (hunk 4 above)
Comment (jdnavarro) [inline bundle/edu.gemini.p2checker/src/test/scala/edu/gemini/p2checker/rules/GmosSDtaXRuleTest.scala]: I'm trying not to modify much of what's already written and just add new code necessary to fix the bug. I still don't feel familiar enough with the code to renames things. However this concrete case looks trivial enough.

File: bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/gmos/GmosRule.java (hunk 6 above)
Comment (jdnavarro) [inline bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/gmos/GmosRule.java]: Right, the value of `PREFIX` is hardcoded in the tests. I have to make public a bunch of methods and attributes.

Shall I do it for the rest of the rules, too?

Comment (swalker2m) [inline bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/gmos/GmosRule.java]: I don't think we need any new public methods.  We can define this id and two others, in the other rule being tested by this same test case:

````java
   public static IConfigRule GMOS_S_DTA_X_RULE = new IConfigRule() {
       public static final IdY1   = PREFIX + "Y1";
       public static final IdY2_4 = PREFIX + "Y2_4";
       ...
   }
````

File: bundle/edu.gemini.p2checker/src/test/scala/edu/gemini/p2checker/rules/GmosSDtaXRuleTest.scala [hunk 7]
```diff
@@ -58,9 +58,16 @@ class GmosSDtaXRuleTest extends AbstractRuleTest {
 
   private def expectSuccess(): Unit = {
 //    println(runCheck.mkString(",\n"))
-    assertTrue(runCheck.isEmpty)
+    assertTrue(! runCheck.exists { p =>
+      Set("GmosRule_X1", "GmosRule_Y2_4")(p.getId())
+    })
   }
```
swalker2m: Then `expectSuccess` becomes:

````scala
private def expectSuccess(hasId: Set[String]): Unit =
  assertTrue(!runCheck.exists(p => hasId(p.getId))
````

and you call it with, for example:

````scala
val DTA_Y_BIN = Set(GmosRule.DTA_X_Y_MULTIPLE_BINNING_RULE.Id)

expectSuccess(DTA_Y_BIN)
````

Comment (jdnavarro) [inline bundle/edu.gemini.p2checker/src/test/scala/edu/gemini/p2checker/rules/GmosSDtaXRuleTest.scala]: I actually implemented slightly different in order to resemble more the code already present, but the idea is the same.

**Comment by jdnavarro** [comment]
I think I now took care of all the comments. Please check the PR again.

**Comment by jdnavarro** [comment]
I screwed up with eac343d, please ignore it and check fcf2aa9 instead. Sorry.

File: bundle/edu.gemini.p2checker/src/test/scala/edu/gemini/p2checker/rules/GmosSDtaXRuleTest.scala [hunk 8]
```diff
@@ -56,31 +56,45 @@ class GmosSDtaXRuleTest extends AbstractRuleTest {
...
+  private def expectSuccessBinY(): Unit =
+    expectSuccess(Set(
+      GmosRule.GMOS_S_DTA_X_RULE_Y1_ID,
+      GmosRule.GMOS_S_DTA_X_RULE_Y2_4_ID)
+    )
+
+  private def expectSuccessBinMultiple(): Unit = {
+    assertTrue(!runCheck.exists { p =>
+      GmosRule.DTA_X_Y_MULTIPLE_BINNING_RULE_ID == p.getId()
+    })
```
swalker2m: It looks like this could be just a call to `expectSuccess` with a singleton `Set` containing the one rule.

bundle/edu.gemini.p2checker/src/test/scala/edu/gemini/p2checker/rules/GmosSDtaXRuleTest.scala
```diff
@@ -56,31 +56,45 @@ class GmosSDtaXRuleTest extends AbstractRuleTest {
...
   }
-
   private def expectFailure(ybin: String): Unit =
     assertTrue(runCheck.exists { p =>
       p.getType == Problem.Type.ERROR &&
       p.getDescription.contains(ybin)
     })
```
jdnavarro: I'll also remove the current parameter that checks for a `String` in the error message and instead check for the problem ID, just like when checking for success.

jdnavarro:
Please check there isn't anything bogus with the refactoring I did with for `expectFailure`.

jdnavarro:
I forgot to check for the problem type in `expectFailure`. Thanks @swalker2m for pointing it out.

swalker2m:
Looks good to me. 👍

PR #1018, opened 2016-06-24
REL-2853: Simbad target name resolution failing

cquiroz:
Sometimes Simbad votable contains several repetitions of the same magnitude and the current parser code wasn't able to cope with that case

This PR makes the parser more robust discarding the repeated values and includes a test case


swalker2m:
This looks fine. 👍 

(Did we need 4,000+ lines of XML just to test it?)

sraaphorst:
Based on your notes on the JIRA issue, which seem sensible to me, 👍

cquiroz:
The XML is exactly as sent by simbad. In principle it could be manually trimmed but that would be very labor intensive and easy to make mistakes given the indexed structure of votable. Also, IMHO it represents exactly what the OT would see

PR #1019, opened 2016-06-24
REL-2857: don't flush BAGS state until closing a program

swalker2m:
This PR updates `SPViewer` to avoid flushing BAGS state unless actually closing a program.  Without this update, every time a user opens more than one program in a single viewer and cycles through programs with the Prev and Next navigation buttons, we redo all BAGS lookups in each program we pass through.

### DIFF
--- a/bundle/jsky.app.ot/src/main/java/jsky/app/ot/viewer/SPViewer.java
+++ b/bundle/jsky.app.ot/src/main/java/jsky/app/ot/viewer/SPViewer.java
@@ -31,6 +31,7 @@ import jsky.app.ot.ui.util.UIConstants;
 import jsky.app.ot.util.History;
 import jsky.app.ot.util.PropertyChangeMultiplexer;
 import jsky.app.ot.util.Resources;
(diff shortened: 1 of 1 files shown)


bundle/jsky.app.ot/src/main/java/jsky/app/ot/viewer/SPViewer.java
```diff
@@ -893,7 +893,9 @@ public void closeProgram(ISPProgram node) {
     /** Close the current viewer window, but don't exit the application, even if it is the last window. */
     private void closeViewer() {
         if (getRoot() != null) {
-            BagsManager.unwatch(getRoot());
+            for (RootEntry e : _history.rootEntriesAsJava()) {
```
sraaphorst: Just a note that this could instead be:
`_history.rootEntriesAsJava().stream().foreach(e -> BagsManager.unwatch(e.root()));`

sraaphorst:
Thanks for fixing this. =embarrassed= 👍

cquiroz:
Looks good to me :+1:

PR #1021, opened 2016-07-05
REL-2874: setting pos angle

swalker2m:
This PR fixes AGS selection application to properly record the position angle when updated, even if the selected guide star itself does not change.  In the version before this update, when AGS would run and find the same guide star that is already recorded, it would not update the target environment even if the position angle differs.

### DIFF
--- a/bundle/edu.gemini.ags/src/test/scala/edu/gemini/ags/impl/Helpers.scala
+++ b/bundle/edu.gemini.ags/src/test/scala/edu/gemini/ags/impl/Helpers.scala
@@ -1,0 +1,32 @@
+package edu.gemini.ags.impl
+
+import edu.gemini.ags.conf.ProbeLimitsTable
+import edu.gemini.skycalc.{DDMMSS, HHMMSS}
+import edu.gemini.spModel.core.{Declination, RightAscension, Angle, Coordinates, MagnitudeBand, Magnitude, SiderealTarget}
+import edu.gemini.spModel.target.SPTarget
+
+/**
+ *
+ */
+trait Helpers {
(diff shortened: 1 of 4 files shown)


bundle/edu.gemini.ags/src/test/scala/edu/gemini/ags/impl/Helpers.scala
```diff
@@ -0,0 +1,32 @@
...
+import edu.gemini.skycalc.{DDMMSS, HHMMSS}
+import edu.gemini.spModel.core.{Declination, RightAscension, Angle, Coordinates, MagnitudeBand, Magnitude, SiderealTarget}
+import edu.gemini.spModel.target.SPTarget
+
+/**
+ *
+ */
+trait Helpers {
+  val mt       = ProbeLimitsTable.loadOrThrow()
+  val zeroBase = basePosition("00:00:00.000 00:00:00.00")
```
cquiroz: Isn't this the same as using `Coordinates.zero?`

cquiroz:
Looks good to me :+1:

bundle/edu.gemini.ags/src/test/scala/edu/gemini/ags/impl/Helpers.scala
swalker2m: Almost, but we need an `SPTarget` here which is what the `basePosition` helper provides.

cquiroz: Yes `SPTarget` at Coordinates 0? Maybe there is even a `SideralTarget.zero?`

PR #1023, opened 2016-07-06
REL-2873: Keep TPE guide star add buttons active

cquiroz:
This PR changes the TPE to let adding Guide Stars always regardless if there is an active manual group. Added guide stars will go to the active manual group or it will create a manual group instead


swalker2m:
Seems fine to me. 👍 I'm not sure I understand why this was being restricted to environments with only the auto guide group before though.  At any rate, these changes look good to me.

cquiroz:
I'm not sure about that either, it seems rather arbitrary

PR #1024, opened 2016-07-06
Retrofit REL-2841/REL-2853

cquiroz:
This PR brings three pull requests into ocs-2016B, these already exist on develop


swalker2m:
Is the squants update needed for one of the other two items?

cquiroz:
No, we could do away with that if you think is best

swalker2m:
I guess if possible we should limit changes unless they are small, low-risk and fixing some bug.  So if the squants update isn't required, it might be best to wait for the December release so it can be part of the regular testing cycle.

cquiroz:
Ok, it has been removed

arturog8m:
Thanks, I agree. We should limit the changes for the patch release as testing time is fairly limited and we want to keep it focused.

swalker2m:
👍

PR #1026, opened 2016-07-08
Feature/rel 2880

andrewwstephens:
Attempt to fix the spectroscopy signal calculation in a 1-pixel aperture which was previously not taking into account the slit width.  This seems to give approximately the correct answer for continuum sources although we will need to do something more elaborate for line sources.


app/spdb/build.sbt
```diff
@@ -374,6 +375,35 @@ def anunez(version: Version) = AppConfig(
   )
 ) extending List(with_gogo(version), anunez_credentials(version))
 
+// ASTEPHENS
+def astephens(version: Version) = AppConfig(
```
cquiroz: 👍 did you add an entry to `ocs-credentials` too?

bundle/edu.gemini.itc/src/main/scala/edu/gemini/itc/operation/Slit.scala
```diff
@@ -26,16 +26,17 @@ sealed trait Slit {
...
-final case class OnePixelSlit(pixelSize: Double) extends Slit {
+/** Slit that is 1-pixel long for calculating the peak pixel flux. */
+final case class OnePixelSlit(width: Double, pixelSize: Double) extends Slit {
 
-  val width                   = pixelSize
-  val length                  = pixelSize
-  val area                    = width * length
+  // width = slit width in arcseconds
+  // pixelSize = pixel size in arcseconds
+  val length                  = pixelSize        // slit length in arcseconds
```
cquiroz: This is not necessary for this PR but we may want to have these in squants-like quantities. It would make some operations safer

cquiroz:
This PR is breaking the build, a test on itc_web is failing

tpolecat:
:sad-trombone:

app/spdb/build.sbt
cquiroz: I add astephens, olesja and spakzad to the svn-based ocs-credentials

cquiroz:
Tests are passing! Looks good to me 👍

PR #1027, opened 2016-07-08
REL-2878: QPT password prompt

swalker2m:
The purpose of this PR is to add a password prompt when publishing a QPT plan.  The account under which plans are published will have a new password soon, requiring a new QPT release.  The IT department is not comfortable with keeping the password in a plain text file in svn, even if only available to staff members.  This PR updates the publish process to request the password from the user.  It will remember the password that the user provides for subsequent publishes until the application is restarted.

Beyond that simple goal, I rewrote the popup(s) associated with publishing.  Since the process already involved clicking through two separate popups, I felt it would be abusive to ask for the password in a new popup.  Instead I combined all the information in a single dialog as shown below.

publishtotheweb

previewlocally

### DIFF
--- a/bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/ui/action/PublishAction.java
+++ b/bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/ui/action/PublishAction.java
@@ -20,7 +20,6 @@ import java.text.DateFormat;
 import java.text.SimpleDateFormat;
 import java.util.Calendar;
 import java.util.TimeZone;
-import java.util.concurrent.CountDownLatch;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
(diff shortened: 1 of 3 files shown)


swalker2m:
Once obvious criticism is that it doesn't say what password is being requested.  You just have to know.  Since there are only a handful of people who publish plans, this may be sufficient anyway.  Or I could explicitly say what the user account is.

cquiroz:
A general question, could this have been handled with ssh keys? You may need to enter a passphrase anyway..

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/ui/action/PublishAction.java
```diff
@@ -82,120 +78,16 @@ protected void asyncActionPerformed(ActionEvent e) {
...
-//                latch.await();
-//            } catch (InterruptedException e1) {
-//                System.out.println("sldjfhskfgh");
-//            }
-//            System.out.println("utc " + utc + " web " + web + " qc " + qcMarkers + " cancel " + cancel);
-//
-//            if (cancel) {
-//                return;
-//            }
+            final scala.Option<PublishDialog.PublishOptions> res = PublishDialog.prompt(shell.getPeer(), opts);
```
cquiroz: It seems this will prompt every time you publish, would it make sense to make it enter the password only once per session?

swalker2m: It prompts but it asks for more information than just the password (the publish destination and time zone).  If you have supplied a password in the past, it is reused so you don't have to type it in again.

This is actually an improvement over the existing implementation which hits the user with two consecutive popups every time you publish!

cquiroz: Excellent!

cquiroz:
Looks good to me :+1:

PR #1028, opened 2016-07-08
REL-2852 improve mean parallactic angle calculation usability

tpolecat:
This PR improves the usability for mean parallactic angle computations as follows:

- The duration used for the computation is retained, even if it is a snapshot of the current remaining time. We do this by storing it as a negative value, in order to maintain binary compatibility with the existing `SchedulingBlock` class.
- When no duration is specified, remaining observe time is used instead and is *not* recorded in the scheduling block. This is a corner case that happens only once in practice, when you first select the mean parallactic angle option. Since this is never the final answer I don't think it matters … it would end up being a potentially risky change to update the scheduling block in this case.

This needs monkey testing.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/obscomp/SPProgram.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/obscomp/SPProgram.java
@@ -45,7 +45,7 @@ public class SPProgram extends AbstractDataObject implements ISPStaffOnlyFieldPr
     // for serialization
     private static final long serialVersionUID = 4L;
 
-    public static final String VERSION = "2016B-1";
+    public static final String VERSION = "2016B-2";
 
     /** This property records the program queue/classical state. */
     public static final String PROGRAM_MODE_PROP = "programMode";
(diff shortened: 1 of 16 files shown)


swalker2m:
Technically we maintain binary compatibility but what happens when the previous OT version edits an observation with a negative scheduling duration?

In the end, another patch release feature will require breaking binary compatibility anyway if we want to take the opportunity to do this in a more explicit way.

cquiroz:
code-wise looks good, I can't comment on the details of the changes

tpolecat:
Ah, good point about the old OT. That's not going to work.

tpolecat:
Ok since this is going to be a breaking release anyway I can re-do this in a better way, but not in time for Monday's test release because it's going to require writing a new data migration. Not sure which way to go.

swalker2m:
My vote would be to hold up the test release until this is ready.  We'll have to do it differently for the `develop` branch anyway so we might as well just merge the same thing to both branches.  Perhaps if it were ready except for the data migration, I could work on that Monday morning and we might be able to do a test release regardless.  If not, then Tuesday.

What do you think @andrewwstephens / @arturog8m ?

tpolecat:
That sounds reasonable. I'll see how much I can get done this afternoon.

tpolecat:
Ok @swalker2m I changed the model such that `SchedulingBlock.duration` is an ADT with three cases. PIO serialization works but there is no handler in `To2016B` and the existing scheduling block migration fails at the moment. I will look at these Monday morning, or you can.

tpolecat:
@swalker2m hey is `SPProgram.VERSION` going to change to be `2016B-2`? I need to distinguish pre and post-patch XML files in the updater.

tpolecat:
Ok updated as follows:
- We now have an ADT `SchedulingBlock.Duration` that distinguishes `Unstated`, `Implicit`, and `Explicit` durations rather than encoding this via negative numbers.
- Changed `SPProgram.VERSION ` to `2016B-2` and added a new migration.

swalker2m:
Looks great. 👍

cquiroz:
Would you also integrate this PR into develop?

PR #1031, opened 2016-07-11
REL-2823 Change warning to error when there is no scheduling block for nonsidereal targets

jdnavarro:
This has been backported from a previous PR to `develop`: https://github.com/gemini-hlsw/ocs/pull/1014

Please be extra careful with this PR since I had removed the branch for this bug and had to do some unconventional checkouts and cherry-picking to be able to reach this point.


cquiroz:
Looks good to me :+1:

PR #1032, opened 2016-07-11
REL-2456 Report a warning when DTA X offset is not a multiple of y-pixel binning

jdnavarro:
This has been backported from a previous PR to `develop`: #1016

Please be extra careful with this PR since I had removed the branch for this bug and had to do some unconventional checkouts and cherry-picking to be able to reach this point.


cquiroz:
Looks good to me :+1:

PR #1033, opened 2016-07-11
REL-2858: Add PA mode option to GSAOI component

cquiroz:
This PR includes:

* Changes GSAOI model to include a flag for position angle constraints
* Update Gsaoi Editor to with the new flag
* Changes GEMS-AGS algorithm to select position angles depending on the flag
* Updates Gems Manual Search Dialog box to sync the old pos angle constraint flag with the model. The sync is two way between both editors

screenshot 2016-07-11 13 47 03

screenshot 2016-07-11 13 48 35

### DIFF
--- a/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/GemsStrategy.scala
+++ b/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/GemsStrategy.scala
@@ -9,12 +9,11 @@ import edu.gemini.catalog.votable._
 import edu.gemini.pot.sp.SPComponentType
 import edu.gemini.pot.ModelConverters._
 import edu.gemini.spModel.core.SiderealTarget
-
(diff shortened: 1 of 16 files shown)


andrewwstephens:
Could we change the label to say "Adjust PA to find optimal guide stars" and move it under the "Pos Angle" field ?

bryanmiller:
I suggest the following layout on the GSAOI component
Position Angle |___0.0_| deg E of N   [ ] Allow Auto Guide Search to select PA    
put it on two lines if necessary
Position Angle-------------
 |___0.0_| deg E of N   [ ] Allow Auto Guide Search to select PA    
so that it is similar to the GMOS component.

cquiroz:
What about this:

screenshot 2016-07-11 15 07 07

swalker2m:
Code and screen shots look good to me.  👍 We can fiddle with the UI more after a test release if Bryan and/or Andy think it would be useful though.

PR #1036, opened 2016-07-19
Linux OT using PIT icon

cquiroz:
This PR sets the icon for OT frames in Linux/Windows instead of the Java default frame

screenshot 2016-07-19 12 23 46


bundle/jsky.app.ot/src/main/java/jsky/app/ot/SplashDialog.java
```diff
@@ -64,6 +65,11 @@ public void windowClosed(WindowEvent e) {
             });
         }
 
+        if (Platform.get() != Platform.osx) {
+            // Set the frame icon, useful for Windows and Linux
+            Resources.setOTFrameIcon(this);
+        }
```
swalker2m: Why does it check for platform here but not in other places where this is used?

swalker2m:
Seems good to me. 👍

bundle/jsky.app.ot/src/main/java/jsky/app/ot/SplashDialog.java
cquiroz: Actually this is very important, I'll change the code. For some reason that call crashes the JVM on OSX, on the other hand you don't need it in OSX since it will use the icon on the dmg

tpolecat: At the very least this deserves a comment about what you just said.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/util/Resources.java
```diff
@@ -96,14 +97,16 @@ public static ImageIcon getIcon(String iconFileName) {
     }
 
     public static void setOTFrameIcon(java.awt.Frame frame) {
-        ImageIcon icon = Resources.getIcon("ot.png");
-        if (icon != null) {
-            Image image = icon.getImage();
-            frame.setIconImage(image);
+        if (Platform.get() != Platform.osx) {
+            // Set the frame icon. This fails in OSX but it is useful on Windows and Linux
```
tpolecat: Can you elaborate on "fail" a little bit? The next person to look at this will be curious.

cquiroz: Sure, basically the JVM crashes. This was on the PIT a while ago. I'll re-run this on OSX to make sure

tpolecat: Sorry, I mean, can you include that in your comment?

bundle/jsky.app.ot/src/main/java/jsky/app/ot/util/Resources.java
```diff
@@ -94,6 +96,19 @@ public static ImageIcon getIcon(String iconFileName) {
         return icon;
     }
 
+    public static void setOTFrameIcon(java.awt.Frame frame) {
+        if (Platform.get() != Platform.osx) {
+            // This has been reported to crash the JVM on OSX when the application
+            // is bundled as a DMG. Starting from the DMG The application fails to start
+            // without any visible indication on why it has failed
+            // However, we want to do this on Windows and Linux
```
cquiroz: Ok, I updated the comment and reproduced the issue. It only happens if the app is in a DMG and it basically doesn't start, There are no messages on the screen or any logs. Clearly is the `setIconImage` call but I can't find a bug report matching this

cquiroz:
This is how it looks in OSX if you call `setIconImage`

screenshot 2016-07-19 15 57 02

cquiroz:
Verified to work on Windows 7 and Linux 64

PR #1039, opened 2016-07-19
REL-2858: Stop PA from being modified by BAGS when it should not be

sraaphorst:
An issue was arising when the PA field was being edited during an AGS lookup, which can easily be done by performing a longer AGS lookup such as in the example in the REL task via GSAOI.

Previously, PA constraints (e.g. fixed, fixed+180, unbounded) were handled by the AGS lookup itself and ignored by BAGS, which indiscriminately took the results of the lookup and applied them to the observation.

If the PA constraint was set to fixed and the PA was being edited, say to contain 45, then an AGS lookup would be triggered for the entry of the 4. If the user typed the 5 before the results for 4 came back, the PA would be set to 45 by the user, and another BAGS lookup would be queued by moving the BAGS state for the observation to `RunningEdited`. Once the results returned for the PA = 4 lookup, they would be indiscriminately applied to the observation and the PA set back to 4; thus, the further queued lookup would not be performed due to the hash value not changing.

(Indeed, even if the hash value did change, the lookup would still be performed for PA = 4 as it would have overwritten the user entered PA of 45.)

This change prevents this from happening by determining if BAGS is trying to make an "illegal change", which occurs when the PA is fixed but the AGS selection results in a change to the PA.

I'm not sure if this is the best way to handle this, but I am open to other suggestions.

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
@@ -131,7 +131,7 @@ object BagsState {
     // come back from the AGS lookup when in RunningEditedState, we store them
     // but move to PendingState to run again.
     override val edit: StateTransition =
-      (RunningEditedState(obs, hash), ioUnit)
+      (RunningEditedState(obs), ioUnit)
 
     // Successful AGS lookup while running (and not edited).  Apply the update
     // and move to IdleState.
@@ -149,7 +149,7 @@ object BagsState {
     * observation that was subsequently edited.  Since it has been edited, the
     * results we're expecting may no longer be valid when they arrive.
     */
-  case class RunningEditedState(obs: ISPObservation, hash: AgsHashVal) extends BagsState {
+  case class RunningEditedState(obs: ISPObservation) extends BagsState {
     // If we're edited again while running, just loop back.  Once the results of
     // the previous edit that got us into RunningState in the first place
     // finish, we'll switch to Pending and go again.
@@ -162,7 +162,7 @@ object BagsState {
     // and as this was causing issues for BAGS overwriting PA edits, we only allow
     // BAGS to apply to unedited observations.
     override def succeed(results: Option[AgsStrategy.Selection]): StateTransition =
-      (PendingState(obs, Some(hash)), BagsManager.wakeUpAction(obs, 0))
+      (PendingState(obs, None), BagsManager.wakeUpAction(obs, 0))
 
     // Failed AGS but we've been edited in the meantime anyway.  Go back to
     // PendingState so we can run again.  Here we pass in no AgsHash value to
@@ -267,12 +267,12 @@ object BagsManager {
 
     def logTransition(from: BagsState, to: BagsState): IO[Unit] = IO {
       def stateString(s: BagsState): String = s match {
-        case BagsState.ErrorState                  => "Error"
-        case BagsState.IdleState(_, hash)          => s"Idle($hash)"
-        case BagsState.PendingState(_, hash)       => s"Pending($hash)"
-        case BagsState.RunningState(_, hash)       => s"Running($hash)"
-        case BagsState.RunningEditedState(_, hash) => s"RunningEdited($hash)"
-        case BagsState.FailureState(_, why)        => s"Failure($why)"
+        case BagsState.ErrorState            => "Error"
+        case BagsState.IdleState(_, hash)    => s"Idle($hash)"
+        case BagsState.PendingState(_, hash) => s"Pending($hash)"
+        case BagsState.RunningState(_, hash) => s"Running($hash)"
+        case BagsState.RunningEditedState(_) => s"RunningEdited"
+        case BagsState.FailureState(_, why)  => s"Failure($why)"
       }
 
       val obsId = Option(obs.getObservationID).getOrElse(obs.getNodeKey)
@@ -457,7 +457,6 @@ object BagsManager {
 
     obs.getProgram.removeStructureChangeListener(StructureListener)
     obs.getProgram.removeCompositeChangeListener(ChangeListener)
-
     applySelection(TpeContext(obs))
 
     // Update the TPE if it is visible
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
@@ -1,7 +1,6 @@
 package jsky.app.ot.ags
 
-import jsky.app.ot.ags.BagsState.{IdleState, ErrorTransition, StateTransition}
-
+import jsky.app.ot.ags.BagsState.{ErrorTransition, IdleState, StateTransition}
 import java.beans.{PropertyChangeEvent, PropertyChangeListener}
 import java.time.Instant
 import java.util.concurrent.TimeoutException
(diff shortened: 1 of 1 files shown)


tpolecat:
Sorry, I don't quite understand how bags handles concurrent changes. Why doesn't the hash value change and invalidate the `4` result, if the current value is `45`?

sraaphorst:
What happens is that the hash result is only calculated when the BAGS state receives the `wakeUp` transition from the `Pending` state, i.e. when BAGS starts its run for a given observation, and then is passed around through the state machine without change.

As the side effects (such as setting PA) are done in `IO`, I'm not sure we could calculate the change to the hash prior to this point, although it may be possible; however, this still would not prevent the issue of BAGS writing to the PA and thus overwriting intermediate changes that the user made.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -459,25 +459,39 @@ object BagsManager {
...
-            case GuideProbe.Type.AOWFS => classOf[Altair_WFS_Feature]
-            case GuideProbe.Type.OIWFS => classOf[OIWFS_Feature]
-            case GuideProbe.Type.PWFS  => classOf[TpePWFSFeature]
+    val ctx = TpeContext(obs)
+
+    // If the PA is not allowed to change by the PA constraint, but does, then this is an intermediate
+    // result that must be ignored: further changes to the PA have been made since this result was obtained,
+    // so do not process this case.
+    val process = (for {
+      paCst <- ctx.obsContext.map(_.getPosAngleConstraint)
```
cquiroz: Is this general enough for all instruments? I guess not all of the are `PosConstraintAware`

sraaphorst: It should be. If the instrument isnt't `PosAngleConstraintAware`, then `PosAngleConstraint.FIXED` is returned by the `ObsContext.getPosAngleConstraint` call, which is precisely the behaviour that we'd want in this case.

cquiroz: Ok, that makes sense or it wouldn't compile

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -459,25 +459,39 @@ object BagsManager {
...
+    if (process) {
+      obs.getProgram.removeStructureChangeListener(StructureListener)
+      obs.getProgram.removeCompositeChangeListener(ChangeListener)
+
+      applySelection(ctx)
+
+      // Update the TPE if it is visible
+      selOpt.foreach { sel =>
+        Option(TpeManager.get()).filter(_.isVisible).foreach { tpe =>
+          sel.assignments.foreach { ass =>
```
cquiroz: Why do we need to change the features on the TPE in this case?

sraaphorst: Keep in mind that this was existing code from aways back that I didn't change. If the TPE is visible, if the guide stars / PA change through AGS, then I believe that this is how the TPE is notified, but I am not certain.

swalker2m:
In `RunningEdited.success`, we always apply the results "in case the edit is not to a field that impacts AGS"

````scala
    // Success, but now the observation has been edited so the AGS selection
    // may not be valid.  Apply the results anyway in case the edit is not
    // to a field that impacts AGS and go to pending to run again if necessary.
    override def succeed(results: Option[AgsStrategy.Selection]): StateTransition = {
      val action = for {
        _ <- BagsManager.applyAction(obs, results)
        _ <- BagsManager.wakeUpAction(obs, 0)
      } yield ()
      (PendingState(obs, Some(hash)), action)
    }
````

Would we accomplish the goal by not applying the action if the current hash value computed from `obs` doesn't match the `hash` value that existed when we decided to do the lookup?

**Comment by swalker2m** [comment]
I suppose that is what your change accomplishes, but looking specifically at the position angle.  I guess we could instead have `applyAction` take the expected hash and recompute it before actually applying.  Then in general we don't touch anything if there is to be a new update.

File: bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala [hunk 3]
```diff
@@ -459,25 +459,39 @@ object BagsManager {
...
+    val ctx = TpeContext(obs)
+
+    // If the PA is not allowed to change by the PA constraint, but does, then this is an intermediate
+    // result that must be ignored: further changes to the PA have been made since this result was obtained,
+    // so do not process this case.
+    val process = (for {
+      paCst <- ctx.obsContext.map(_.getPosAngleConstraint)
+      oldPA <- ctx.instrument.dataObject.map(_.getPosAngleDegrees)
+      newPA <- selOpt.map(_.posAngle.toDegrees)
+    } yield paCst != PosAngleConstraint.FIXED || oldPA == newPA) | false
```
swalker2m: So here, the `process` flag could be set based upon whether the current hash matches the expected one which would be passed into `applyAction`.  It's the same idea but it applies to any changes that impact the AGS calculation rather than just the pos angle constraint.

sraaphorst: Smart. I like it better. I'll give that a try and confirm that it works.

swalker2m: "Smart" is probably a stretch... :-)  At any rate, I think there will be a race condition here unless we change the definition of `RunningEdited.success`.  If we might decide in `applyAction` to skip the update and yet we record the `hash` value before transitioning to `Pending`, we could be in for trouble if the observation is subsequently edited in such a way as to produce the same `hash` value that we just ignored before the pending state wake up is processed.  It pains me, but I guess the safe thing to do then, in addition, would be to change `RunningEdited.success` to just always re-trigger like so:

````scala
override def succeed(results: Option[AgsStrategy.Selection]): StateTransition =
    (PendingState(obs, None), BagsManager.wakeUpAction(obs, 0))
````

Comment (swalker2m) [inline bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala]: In fact, if you just do this, I suppose there's no need to modify `applyAction` at all?  It will only ever be called by `success` on a plain `Running` state which means that no edits happened in the interim.  This does mean that continued editing of an observation will delay the ultimate findings of the AGS lookup.

File: bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala [hunk 4]
```diff
@@ -158,15 +157,12 @@ object BagsState {
...
-        _ <- BagsManager.wakeUpAction(obs, 0)
-      } yield ()
-      (PendingState(obs, Some(hash)), action)
-    }
+    // may not be valid. We could apply the results in case the edits made to the
+    // observation do not change the AGS selection, but in the interests of safety
+    // and as this was causing issues for BAGS overwriting PA edits, we only allow
+    // BAGS to apply to unedited observations.
+    override def succeed(results: Option[AgsStrategy.Selection]): StateTransition =
+      (PendingState(obs, Some(hash)), BagsManager.wakeUpAction(obs, 0))
```
swalker2m: To make sure that it runs the computation again, we will need to pass  in `None` instead of `Some(hash)`.  Otherwise in `Pending` when it wakes up it could calculate the same hash value and skip the update.

swalker2m: Oh, and it looks like that means that the `RunningEditedState` doesn't need the `hash` value any more in its constructor.

sraaphorst: Alright... I squashed down the previous commits due to a bunch of ridiculous cleanups, and I'll make these changes as well! Thanks for the feedback.

sraaphorst:
Okay, all changes made. I'm going to merge.

PR #1040, opened 2016-07-20
REL-2886: Include OT installation link in incompatibility warning

cquiroz:
This changes the error display shown when the OT/SPDB have incompatible versions to show on one hand the actual versions involved and an link to let the users download the current version of the OT

This PR requires quite a bit of swing dark magic. The second commit d393b59 is the relevant one

The new box looks like:

screenshot 2016-07-19 21 50 56


swalker2m:
Hooray! 👍 (except for whatever is making the `spModel.io` tests fail which is surely unrelated?)

PR #1043, opened 2016-07-20
Reduce the amount of logging on Gems AGS

cquiroz:
This PR reduces the amount of logging produced by Gems AGS which pollutes the build output

I don't think this output is very useful to users  but perhaps @andrewwstephens can chime in about that

There is more output related to AGS that could be removed but I'm not sure, in particular in [SingleProbeStrategy](https://github.com/gemini-hlsw/ocs/blob/develop/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/SingleProbeStrategy.scala#L88)

### DIFF
--- a/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsResultsAnalyzer.scala
+++ b/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsResultsAnalyzer.scala
@@ -79,7 +79,7 @@ object GemsResultsAnalyzer {
           gemsGuideStars
         }
       }
-      sortResultsByRanking(gemsGuideStars.toList).asJava
+      sortResultsByRanking(gemsGuideStars).asJava
     }.getOrElse(List.empty[GemsGuideStars].asJava)
   }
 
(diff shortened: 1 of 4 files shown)


bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsResultsAnalyzer.scala
```diff
@@ -125,14 +125,14 @@ object GemsResultsAnalyzer {
...
+      sortResultsByRanking(gemsGuideStars)
     }
   }
 
   private def printResults(result: List[GemsGuideStars]) {
-    Log.info("Results:")
+    Log.fine("Results:")
     result.zipWithIndex.foreach { case (s, i) =>
-      Log.info(s"result #$i : $s")
+      Log.fine(s"result #$i : $s")
```
swalker2m: I suppose we could wrap the whole thing in a check for the logging level and avoid the zipping and iteration if we're not going to use it.

````scala
if (Log.isLoggable(Level.FINE)) {
  Log.fine("Results:")
  ...
}
````

swalker2m:
I'm for this change 👍 but if @andrewwstephens feels the output is useful we could create an AGS logger and he could twiddle the log level for it in configuration?

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsResultsAnalyzer.scala
cquiroz: Right, this is very old stuff so I'm not sure if we should just remove it altogether

sraaphorst: I'm all for this. GeMS AGS lookups are painfully noisy.

sraaphorst:
👍

andrewwstephens:
:+1:  I'm happy changing this from INFO to FINE, as long as it is possible to run my local build with level = FINE.

cquiroz:
Any final opinions about this PR, should I merge it?

tpolecat:
Needs rebase but 👍  from me.

PR #1044, opened 2016-07-21
REL-2888: Made changes to bump PIT version to 2017A.1.1

sraaphorst:
Changes needed to prepare for PIT 2017A.1.1.

### DIFF
--- a/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/transform/UpConverter.scala
+++ b/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/transform/UpConverter.scala
@@ -307,6 +307,21 @@ case object SemesterConverter2014BTo2015A extends SemesterConverter {
 }
 
 /**
+  * This converter only changes the schema version but retains the semester
+  */
+case object SchemaVersionConverter extends SemesterConverter {
+  val current = Semester.current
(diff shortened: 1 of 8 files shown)


bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/transform/UpConverter.scala
```diff
@@ -295,20 +305,6 @@ case object SemesterConverter2014BTo2015A extends SemesterConverter {
   }
   override val transformers = List(gsSubmission)
 }
-/**
- * This converter only changes the schema version but retains the semester
- */
-case object SchemaVersionConverter extends SemesterConverter {
```
cquiroz: We shouldn't delete this. even if it is unused, it is useful when we have bugfix releases inside the same semester

cquiroz:
Looks good to me :+1:
Great Sebastian, the PIT is yours now 😄

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/transform/UpConverter.scala
sraaphorst: Sure! I'll re-add it.

sraaphorst:
I am filled with excitement and a mild sense of terror 😄

PR #1046, opened 2016-07-22
REL-2798: Added Japanese institutions.

sraaphorst:
Subject line says it all.


swalker2m:
Seems fine.  Is there any rule about the content of the `Country` element relative to the partner list or can it be anything at all?

sraaphorst:
I had the exact same question, and asked Bryan about this, because I wasn't sure if it would figure into anything relevant such as what TAC a proposal would be submitted to.

Apparently, it just tacks it on the end of the address for the PI, so really, it could be anything at all.

swalker2m:
At one point, as I vaguely recall, we used to try to divine the partner country based on the institution address but maybe that is no longer necessary.  👍

cquiroz:
There is a link between the institution and the NGO. In this particular case it is valuable to test this as Japan is linked to Subaru as there is no Japanese NGO

Did you try using these institutions on the PIT? Can you show how the UI looks like?

sraaphorst:
Hi Carlos,

Sure! Note that I just realized that Subaru is a duplicate institution and was already listed with the same address, albeit spread across multiple lines. I am guessing that it is probably preferable to keep the old entry rather than the newer one (which is cleaner)? Or do we want one entry linked to the US (the old one) and one to Japan? I spoke with Bryan and he said the US affiliation for country should be correct, but given what you say here, I now have my doubts and maybe we should have two entries, i.e. one for the US and one for Japan.

Please see attached PNGs for Subaru and Tohoku University.
japan2
japan1

cquiroz:
I think Subaru with US affiliation should be right while others on the Japanese list should be affiliated to Japan as in your screenshots

Perhaps @andrewwstephens or Bryan can comment on whether we need a Subaru/Japan combination.

Can you test how this behaves if your proposal class is Fast Turnaround. If you select a Japanese institution, does the PI affiliation become Japan too?

sraaphorst:
Yes, it appears that the PI affiliation for FT does indeed become Japan!
japan3

cquiroz:
Looks good to me :+1:

sraaphorst:
Several notes on this PR:
1. The `PI's Affiliation` does not get set properly in many instances (tested with 2016B.2.2 - see REL-2901).
2. Even though we will need to fix this, I am still hoping to get from @andrewwstephens or Bryan a list of the exception cases where `Country` and `PI's Affiliation` are different.
3. Additionally, I have some confusion: for the Japanese institutions, should there be a new `PI's Affiliation` of Japan, or should all of these be associated as exchange time with Subaru?

sraaphorst:
While there is some work to do on this still, it extends beyond just this issue, and thus I am merging this and have created a separate JIRA task to account for the remaining work, namely REL-2907.

PR #1049, opened 2016-07-27
REL-2697: GeMS requires IQ 85

sraaphorst:
Previously, the PIT would issue a warning for use of AO with IQ that was worse than 70.
New requirements allow GeMS based instruments to accept IQ up to 85.
This code checks for GeMS use (GSAOI) and implements the new rule.

Additionally:
1. Small cleanup of exception handling in `CgiQueryExecturor`; and
2. `Problem` messages were previous inconsistent with regards to punctuation. This has been fixed so that all problems now have a trailing period.
3. `Problem`s from catalog lookups that generate an `Error` did not do anything with the error message, but the format of the `Problem` message suggests that it should be included, so this has been added.

### DIFF
--- a/bundle/edu.gemini.horizons.api/src/main/java/edu/gemini/horizons/server/backend/CgiQueryExecutor.java
+++ b/bundle/edu.gemini.horizons.api/src/main/java/edu/gemini/horizons/server/backend/CgiQueryExecutor.java
@@ -94,10 +94,6 @@ public enum CgiQueryExecutor implements IQueryExecutor {
             client.executeMethod(method);
             reply = CgiReplyBuilder.buildResponse(method.getResponseBodyAsStream(), method.getRequestCharSet());
             method.releaseConnection();
-        } catch (HorizonsException ex) {
-            throw ex;
-        } catch (HttpException ex) {
-            throw HorizonsException.create(ex);
         } catch (IOException ex) {
             throw HorizonsException.create(ex);
         }
--- a/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/catalog/Horizons.scala
+++ b/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/catalog/Horizons.scala
@@ -56,10 +56,10 @@ class Horizons private (/*host: String, port: Int,*/ site: Site, start: Date, en
           // Ambiguous
           case MUTLIPLE_ANSWER =>
             val table = hr.getResultsTable
-            val header = table.getHeader.asScala.toSeq
+            val header = table.getHeader.asScala
 
             val options = for {
-              row <- table.getResults.asScala.map(_.asScala.toSeq).toSeq
+              row <- table.getResults.asScala.map(_.asScala.toSeq)
               map = Map(header.zip(row): _*)
               id <- map.get("ID#")
               name <- map.get("Name")
--- a/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
+++ b/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
@@ -88,8 +88,8 @@ class ProblemRobot(s: ShellAdvisor) extends Robot {
           TimeProblems.partnerZeroTimeRequest(p, s) ++
           TacProblems(p, s).all ++
           List(incompleteInvestigator, missingObsElementCheck, cfCheck, emptyTargetCheck, emptyEphemerisCheck, initialEphemerisCheck, finalEphemerisCheck,
-            badGuiding, badVisibility, iffyVisibility, singlePointEphemerisCheck, minTimeCheck, wrongSite, band3Orphan2, gpiCheck, lgsCC50Check, lgsIQCheck,
-            texesCCCheck, texesWVCheck, gmosWVCheck, gmosR600Check, band3IQ, band3LGS, band3RapidToO, sbIrObservation).flatten
+            badGuiding, badVisibility, iffyVisibility, singlePointEphemerisCheck, minTimeCheck, wrongSite, band3Orphan2, gpiCheck, lgsIQ70Check, lgsGemsIQ85Check,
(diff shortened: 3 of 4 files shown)


bundle/edu.gemini.horizons.api/src/main/java/edu/gemini/horizons/server/backend/CgiQueryExecutor.java
```diff
@@ -94,10 +94,6 @@ public HorizonsReply execute(HorizonsQuery query) throws HorizonsException {
             client.executeMethod(method);
             reply = CgiReplyBuilder.buildResponse(method.getResponseBodyAsStream(), method.getRequestCharSet());
             method.releaseConnection();
-        } catch (HorizonsException ex) {
-            throw ex;
-        } catch (HttpException ex) {
-            throw HorizonsException.create(ex);
         } catch (IOException ex) {
             throw HorizonsException.create(ex);
```
sraaphorst: Note that since `HorizonException` is just passed through, there is no need to include it in the `try`-`catch` block.
Aslo, `HttpException`s are `IOException`s, so no special handling is needed of these since they are treated identically to `IOException`s.

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/catalog/Horizons.scala
```diff
@@ -56,10 +56,10 @@ class Horizons private (/*host: String, port: Int,*/ site: Site, start: Date, en
           // Ambiguous
           case MUTLIPLE_ANSWER =>
             val table = hr.getResultsTable
-            val header = table.getHeader.asScala.toSeq
+            val header = table.getHeader.asScala
 
             val options = for {
-              row <- table.getResults.asScala.map(_.asScala.toSeq).toSeq
+              row <- table.getResults.asScala.map(_.asScala.toSeq)
```
sraaphorst: IDEA reported that the `toSeq` calls here were redundant, so I removed them for cleanness.

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
```diff
@@ -205,59 +205,73 @@ class ProblemRobot(s: ShellAdvisor) extends Robot {
...
+
+    private val lgsGemsIQ85Check = for {
+      o <- p.observations
+      c <- o.condition
+      b <- o.blueprint
+      if bpIsGemsLgs(b) && (!List(ImageQuality.IQ85, ImageQuality.IQ70, ImageQuality.BEST).contains(c.iq))
+    } yield new Problem(Severity.Error, s"GeMS LGS requires IQ85 or better.", "Observations", s.inObsListView(o.band, _.Fixes.fixConditions(c)))
 
-    private val lgsIQCheck = for {
+    private val lgsCC50Check = for {
```
sraaphorst: Renamed since the name seemed previously erroneous, as the `Problem` had nothing to do with IQ but instead was focused on CC.

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
```diff
@@ -155,8 +155,8 @@ class ProblemRobot(s: ShellAdvisor) extends Robot {
       (t, Some(f)) <- s.catalogHandler.state
       (sev, msg) = f match {
         case Offline      => (Severity.Error, s"Catalog lookup failed for ${t.name} due to network connectivity problems.")
-        case NotFound(n)  => (Severity.Error, s"""Catalog lookup returned no results for target "$n" """)
-        case Error(e)     => (Severity.Error, s"Catalog lookup failed for ${t.name} due to an unexpected error: ")
+        case NotFound(n)  => (Severity.Error, s"""Catalog lookup returned no results for target "$n".""")
+        case Error(e)     => (Severity.Error, s"Catalog lookup failed for ${t.name} due to an unexpected error: ${e.getMessage}.")
```
cquiroz: Did you test this? Does it look reasonable on screen? There maybe no point on displaying a very obscure error message that the user cannot understand

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
```diff
@@ -205,59 +205,73 @@ class ProblemRobot(s: ShellAdvisor) extends Robot {
...
+        case _                 => false
+      }
+
+    private val lgsIQ70Check = for {
       o  <- p.observations
       c  <- o.condition
       b  <- o.blueprint
-      if bpIsLgs(b) && (c.iq != ImageQuality.IQ70 && c.iq != ImageQuality.BEST)
-    } yield new Problem(Severity.Error, s"LGS requires IQ70 or better", "Observations", s.inObsListView(o.band, _.Fixes.fixConditions(c)))
+      if bpIsLgs(b) && (!bpIsGemsLgs(b)) && (!List(ImageQuality.IQ70, ImageQuality.BEST).contains(c.iq))
```
cquiroz: There are some unneeded parenthesis here, but regardless is this true. Should this check explicitly not apply to gems?

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
sraaphorst: It seems, from what I can tell, that the ONLY time that this ever comes up is in `Catalog.empty`.
So you are probably right: we could eliminate the error and just put a period.

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
sraaphorst: The new `Problem` requires IQ85 or better for GeMS. Seems like it would be erroneous to display this problem since IQ70 is no longer the requirement for GeMS?

That is why I separated the two.

cquiroz:
Looks good to me :+1:

PR #1050, opened 2016-07-28
REL-2879: Warning for missing PI phone number (and other investigator improvements)

sraaphorst:
REL-2879 requests that a warning be generated when the PI phone number is missing, and that double-clicking on the warning launch the PI editor with the focus in the Phone field.

This PR takes care of that (and more) through the following:
1. A bug has been fixed in the phone list parsing, which resulted in a `List("")` being returned when no phone numbers were specified.
2. Warning added to `ProblemRobot` as required.
3. Cleaned up problems generated by `ProblemRobot` to ensure that we never display empty strings as investigator names.
4. Removed hard-coded strings from `ProblemRobot` for PIs.
5. Changed `CoiEditor` and `PIEditor` to allow for setup code so that the `Phone` text field can request the focus.

I'm not sure that (5) was done in the best way possible, so if anyone has alternative suggestions, I would much appreciate this.

### DIFF
--- a/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/editor/CoiEditor.scala
+++ b/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/editor/CoiEditor.scala
@@ -14,10 +14,14 @@ import edu.gemini.pit.ui.util._
 import swing.event.ValueChanged
 
 object CoiEditor {
-  def open(coi: CoInvestigator, editable:Boolean, parent: UIElement) = new CoiEditor(coi, editable).open(parent)
+  def open(coi: CoInvestigator, editable: Boolean, parent: UIElement, setup: CoiEditor => Unit = _ => {}) = {
+    val editor = new CoiEditor(coi, editable)
+    setup(editor)
+    editor.open(parent)
+  }
 }
 
-class CoiEditor private (coi: CoInvestigator, editable:Boolean) extends StdModalEditor[CoInvestigator]("Edit Co-Investigator") {
+class CoiEditor(coi: CoInvestigator, editable:Boolean) extends StdModalEditor[CoInvestigator]("Edit Co-Investigator") {
 
   // Our main editor component
   object Editor extends GridBagPanel with Rows {
--- a/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/editor/CoiEditor.scala
+++ b/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/editor/CoiEditor.scala
@@ -14,7 +14,7 @@ import edu.gemini.pit.ui.util._
 import swing.event.ValueChanged
 
 object CoiEditor {
-  def open(coi: CoInvestigator, editable: Boolean, parent: UIElement, setup: CoiEditor => Unit = _ => {}) = {
+  def open(coi: CoInvestigator, editable: Boolean, parent: UIElement, setup: CoiEditor => Unit = _ => ()): Option[CoInvestigator] = {
     val editor = new CoiEditor(coi, editable)
     setup(editor)
     editor.open(parent)
--- a/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/editor/PiEditor.scala
+++ b/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/editor/PiEditor.scala
@@ -13,10 +13,14 @@ import swing._
 import event.ValueChanged
 
 object PiEditor {
-  def open(pi: PrincipalInvestigator, editable:Boolean,  parent: UIElement) = new PiEditor(pi, editable).open(parent)
+  def open(coi: PrincipalInvestigator, editable: Boolean, parent: UIElement, setup: PiEditor => Unit = _ => {}) = {
+    val editor = new PiEditor(coi, editable)
+    setup(editor)
+    editor.open(parent)
+  }
 }
 
-class PiEditor private (pi: PrincipalInvestigator, editable:Boolean) extends StdModalEditor[PrincipalInvestigator]("Edit Principal Investigator") {
+class PiEditor(pi: PrincipalInvestigator, editable:Boolean) extends StdModalEditor[PrincipalInvestigator]("Edit Principal Investigator") {
 
   // Our editor
   object Editor extends GridBagPanel with Rows {
--- a/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/editor/PiEditor.scala
+++ b/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/editor/PiEditor.scala
@@ -13,7 +13,7 @@ import swing._
 import event.ValueChanged
 
 object PiEditor {
-  def open(coi: PrincipalInvestigator, editable: Boolean, parent: UIElement, setup: PiEditor => Unit = _ => {}) = {
+  def open(coi: PrincipalInvestigator, editable: Boolean, parent: UIElement, setup: PiEditor => Unit = _ => ()): Option[PrincipalInvestigator] = {
     val editor = new PiEditor(coi, editable)
     setup(editor)
     editor.open(parent)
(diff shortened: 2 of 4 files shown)


bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/editor/PiEditor.scala
```diff
@@ -112,7 +116,7 @@ class PiEditor private (pi: PrincipalInvestigator, editable:Boolean) extends Std
       address = Institution.Address.text,
       country = Institution.Country.text),
     status = Status.selection.item.asInstanceOf[InvestigatorStatus], // :-/
-    phone = Phone.text.split(",").map(_.trim).toList,
+    phone = Phone.text.split(",").map(_.trim).filter(_.nonEmpty).toList,
```
cquiroz: 👍

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/editor/CoiEditor.scala
```diff
@@ -14,10 +14,14 @@ import edu.gemini.pit.ui.util._
 import swing.event.ValueChanged
 
 object CoiEditor {
-  def open(coi: CoInvestigator, editable:Boolean, parent: UIElement) = new CoiEditor(coi, editable).open(parent)
+  def open(coi: CoInvestigator, editable: Boolean, parent: UIElement, setup: CoiEditor => Unit = _ => {}) = {
```
cquiroz: So open returns `Unit` now? If so maybe it should be explicit

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
```diff
@@ -586,21 +587,32 @@ class ProblemRobot(s: ShellAdvisor) extends Robot {
       new Problem(Severity.Todo, "Please create observations with conditions, targets, and resources.", "Observations", ())
     }
 
+    private def investigatorFullName(i: Investigator, default: String = "Investigator"): String = {
+      val piEmptyName = PrincipalInvestigator.empty.fullName
+      i.fullName.trim match {
+        case ""            => default
+        case `piEmptyName` => "PI"
```
cquiroz: I think that instead of the quotes you can make the variable be capitalized and get the same effect

``
...
   case PiEmptyName => ...
...
```

File: bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/view/proposal/ProposalView.scala [hunk 4]
```diff
@@ -407,8 +406,8 @@ class ProposalView(advisor:ShellAdvisor) extends BorderPanel with BoundView[Prop
 
   }
 
-  // public edit method for quick fix
-  def edit(i:Investigator) = investigators.listViewer.edit(i)
-
-
+  // public edit methods for quick fixes
+  def editPi(setup: PiEditor => Unit = _ => {}) = investigators.listViewer.editPi(setup)
+  def editCoi(i: CoInvestigator, setup: CoiEditor => Unit = _ => {}) = investigators.listViewer.editCoi(i, setup)
```
cquiroz: It seems `editCoi` is never called with a non-default setup method, Do we need it?

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/editor/CoiEditor.scala
sraaphorst: It should still be returning `Option[CoInvestigator]`, as that is what is returned by the `open` method on `CoiEditor` instances. The `setup` function simply allows configuration of the `CoiEditor` UI so that we can do things like set the focus on certain UI elements, e.g. the `Phone` text field as was requested by this REL task.

I do think it would be helpful to make the return type explicit, so I am going to add that to the method signature for both `CoiEditor.open` and `PIEditor.open` (which, of course, returns `Option[PrincipalInvestigator]`.

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
sraaphorst: Nice. I did not know this! I verified that it works and I'll commit the change!

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/view/proposal/ProposalView.scala
sraaphorst: Probably not. It just felt right from the perspective of completeness 😺

sraaphorst:
It seems as well that I can replace the default horribleness of `_ => {}` with `identity`, although IDEA complains and moans about it. It appears to compile and work fine, and is much cleaner, I think, so I'm going to do that as well.

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
```diff
@@ -586,21 +587,32 @@ class ProblemRobot(s: ShellAdvisor) extends Robot {
...
-      })
+    private val nonUpdatedInvestigatorName =
+      when(p.investigators.pi.fullName.trim === PrincipalInvestigator.empty.fullName || p.investigators.pi.fullName.trim.isEmpty) {
+        new Problem(Severity.Todo, s"Please provide PI's full name.", "Overview", s.inOverview{_.edit(p.investigators.pi)})
+      }
+
+    private val noPIPhoneNumber = when (p.investigators.pi.phone.isEmpty) {
+      new Problem(Severity.Warning, s"No phone number given for ${investigatorFullName(p.investigators.pi, "PI")}. This is for improved user support.",
+        "Overview", s.inOverview{
+          _.editPi(_.Phone.requestFocus)
```
cquiroz: It seems this is the only use of `setup`? Does it make sense to make it the default behavior? or would it mess the UI flow?

cquiroz:
IDEA can fail in places and show false positives, as long as it compiles it should be fine to use `identity` though `_ => {}` doesn't exactly match `identity`

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
sraaphorst: Yeah, this is currently the only use of `setup`. I'm really not sure a better way to allow for a setup on the investigator editors, but if you have any suggestions, I would love to hear them.

I can see this possibly coming in handy in the future, or possibly in other existing warnings in order to set focused fields. Alternatively, if we just dropped the requirement from the focus to be set to the Phone field in the editor as per the REL task, all of this would be unnecessary. It does seem like a lot of trouble just to get the Phone field start with the focus.

cquiroz: I think is fine, it makes sense to focus on the field when you double-click on the problem. It is a bit cumbersome but the requirement is too 😄

swalker2m:
Identity is `A => A`.  I'm not sure why `identity` would compile, and I wonder if it would if we turned on enough strictness testing?

The only instance of `Unit` is `()`.  To be clear, I think it might be better as `_ => ()`.

tpolecat:
Scala goes to great length to insert `()` for you, if you ask.

```scala
scala> reflect.runtime.universe.reify { identity : (Int => Unit) }
res2: reflect.runtime.universe.Expr[Int => Unit] =
Expr[Int => Unit](({
  ((x) => {
    Predef.identity(x);
    ()
  })
}: Function1[Int, Unit]))
```

sraaphorst:
I always got the impression that `A => B` could always be supplied as `A => Unit`, and the return type was simply ignored. (At least, that's been my experience, but my Scala knowledge is far less than the three of you.)

I have no objection to `_ => ( )`; I just figured `identity` was a bit less verbose. It's a shame that Scala doesn't have something like a `def noOp[A](_: A) {}` for this purpose (at least as far as I know).

cquiroz:
Looks good to me :+1:
Thanks for going the extra mile and fix the other details

sraaphorst:
Thanks to all for the feedback (especially @cquiroz of course)!
Merging.

PR #1051, opened 2016-07-28
REL-2811: Changed icon for 2017A.

sraaphorst:
Subject line says it all.


cquiroz:
Is it just me or the icon is very ugly?

cquiroz:
So the png has a transparent part, Could that cause troubles in some of the platforms?

sraaphorst:
LOL anything is better than the hideous icon from 2016A (IIRC). That looked like a train wreck.

Agree that this icon still isn't great, and it seems that the icons have become more of a "show off piece" of Gemini accomplishments than actual thought put into UI attractiveness, which is what the website should be for and not the icons.

I have searched around, and I think that transparent icons should be fine on Windows and Linux (and they are definitely okay on Mac), but the white PIT writing extending into the transparency has me concerned about readability: on any remotely light background, a substantial T in PIT will be unreadable.

IIRC, @swalker2m and I have had a similar conversation with science in the past about icons (2016A), and in the end, I think we just ended up going along with the request.

cquiroz:
Looks good to me :+1:

PR #1052, opened 2016-07-29
REL-2890: Catch SIMBAD error response

cquiroz:
It was noted that Simbad sometimes returns a bogus response including a java exception. This breaks the parser that doesn't handle this case and instead tells the user that target is not found

This PR addresses that and simplifies some bits of code. The solution is less clean than ideal since Simbad votable is not really compliant, so this particular error needs to be handled in a special case

### DIFF
--- a/bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/api/CatalogQuery.scala
+++ b/bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/api/CatalogQuery.scala
@@ -19,8 +19,10 @@ case class MagnitudeQueryFilter(mc: MagnitudeConstraints) extends QueryResultsFi
 
 sealed abstract class CatalogName(val id: String, val displayName: String) {
   def supportedBands: List[MagnitudeBand] = Nil
-  // Indicates what is the band used when a generic R band is requried
+  // Indicates what is the band used when a generic R band is required
   def rBand: MagnitudeBand = MagnitudeBand.UC
+  // Whether we should validate the xml on parsing
+  def checkValidity: Boolean = true
 }
 
 case object SDSS extends CatalogName("sdss9", "SDSS9 @ Gemini")
@@ -34,11 +36,12 @@ case object UCAC4 extends CatalogName("ucac4", "UCAC4 @ Gemini") {
 }
 case object TWOMASS_PSC extends CatalogName("twomass_psc", "TwoMass PSC @ Gemini")
 case object TWOMASS_XSC extends CatalogName("twomass_xsc", "TwoMass XSC @ Gemini")
-case object SIMBAD extends CatalogName("simbad", "Simbad")
+case object SIMBAD extends CatalogName("simbad", "Simbad") {
+  override def checkValidity = false // Simbad sometimes returns non-valid XML, in particular in errors
+}
 
 object CatalogName {
   implicit val equals = Equal.equal[CatalogName]((a, b) => a.id === b.id)
-
 }
 
 /**
--- a/bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/api/CatalogQuery.scala
+++ b/bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/api/CatalogQuery.scala
@@ -21,8 +21,6 @@ sealed abstract class CatalogName(val id: String, val displayName: String) {
   def supportedBands: List[MagnitudeBand] = Nil
   // Indicates what is the band used when a generic R band is required
   def rBand: MagnitudeBand = MagnitudeBand.UC
-  // Whether we should validate the xml on parsing
-  def checkValidity: Boolean = true
 }
 
 case object SDSS extends CatalogName("sdss9", "SDSS9 @ Gemini")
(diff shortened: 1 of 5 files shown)


bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/api/CatalogQuery.scala
```diff
@@ -34,11 +36,12 @@ case object UCAC4 extends CatalogName("ucac4", "UCAC4 @ Gemini") {
 }
 case object TWOMASS_PSC extends CatalogName("twomass_psc", "TwoMass PSC @ Gemini")
 case object TWOMASS_XSC extends CatalogName("twomass_xsc", "TwoMass XSC @ Gemini")
-case object SIMBAD extends CatalogName("simbad", "Simbad")
+case object SIMBAD extends CatalogName("simbad", "Simbad") {
+  override def checkValidity = false // Simbad sometimes returns non-valid XML, in particular in errors
```
swalker2m: What would happen if you did check validity and there were an exception element?  Would it pass as valid anyway?  Is that the issue?

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableParser.scala
```diff
@@ -244,6 +254,12 @@ case object SimbadAdapter extends CatalogAdapter {
     }
     \/-((band.flatten |@| MagnitudeSystem.fromString(p._2))((b, s) => (b, s)))
   }
+
+  def containsExceptions(xml: Node): Boolean = {
+    // The only case known is with java.lang.NullPointerException but let's make the check
+    // more general.
+    (xml \\ "INFO" \ "@value").text.matches("java\\..*Exception")
```
swalker2m: I suppose there could there could be `*Error`s or exceptions particular to some library other than the standard library?  I guess detecting this is tricky though.

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/api/CatalogQuery.scala
```diff
@@ -19,8 +19,10 @@ case class MagnitudeQueryFilter(mc: MagnitudeConstraints) extends QueryResultsFi
 
 sealed abstract class CatalogName(val id: String, val displayName: String) {
   def supportedBands: List[MagnitudeBand] = Nil
-  // Indicates what is the band used when a generic R band is requried
+  // Indicates what is the band used when a generic R band is required
   def rBand: MagnitudeBand = MagnitudeBand.UC
+  // Whether we should validate the xml on parsing
+  def checkValidity: Boolean = true
```
swalker2m: Instead of a flag here and special case handling `VoTableParser`, perhaps each catalog should have an opportunity to define its own validity checks?  Would that be feasible?  If so, then the peculiar simbad logic could at least be kept with the definition of the catalog and the parsing could would not need a SIMBAD special case.

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/api/CatalogQuery.scala
cquiroz: Simbad returns the exception and the not found case on a non compliant xml. If we check validity in both cases, they will become errors, which is not correct, only the first case is an actual error

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/api/CatalogQuery.scala
cquiroz: I considered that, it would be possible to create a "pre-check" method that does the validation check. I went against that because the most logical place to do that is on the `CatalogAdapters`.

However those are created at runtime based on the content of an already validated xml producing  a chicken an egg problem

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableParser.scala
cquiroz: Note that this was discovered by accident. Right now the backend is working fine so it is difficult to have a generic definition of what the value can be

There is no open documentation of the Simbad Votable format

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableParser.scala
```diff
@@ -49,14 +49,24 @@ object VoTableParser extends VoTableParser {
...
     // Load in memory (Could be a problem for large responses)
     val xmlText = Source.fromInputStream(is, "UTF-8").getLines().mkString
 
-    (checkValidity, validate(xmlText)) match {
-      case (true, -\/(e))  => \/.left(ValidationError(catalog))
-      case (false, -\/(e)) => \/.right(parse(XML.loadString(xmlText)))
-      case (_, \/-(r))     => \/.right(parse(XML.loadString(r)))
+    validate(xmlText) match {
+      case -\/(e) if catalog.checkValidity => \/.left(ValidationError(catalog))
+      case -\/(e) if catalog == SIMBAD     =>
```
swalker2m: Maybe it has to be this way, but there's something about this that isn't totally satisfying.  There's a `checkValidity` flag on the catalog as if we can handle this behavior generically and yet we need the special case `catalog == SIMBAD` anyway.  We could remove the `checkValidity` flag altogether and just check whether the catalog is SIMBAD directly and not lose anything as far as I can see.

cquiroz: I agree it is not satisfying 😄 . The `checkValidity` is there for the cases of other catalogs that are not compliant, though we don't have that case except for `Simbad` at the moment (At a certain time I think ppmxl wasn't compliant either.

I guess I could remove the flag and return to this in case we find another case of funny standard compliance. Now that I think about I could reorg to skip validation altogether for simbad saving some processing time

cquiroz:
Updated according to PR comments

cquiroz:
Somehow I closed this PR by accident, rebased and reopened

PR #1053, opened 2016-07-29
REL-2807: Changed GN LGS Dec limits in PIT.

sraaphorst:
The Dec limits for GN LGS have been requested to be changed in the PIT as per the JIRA task.
This sets the new limits and updates the test cases.

### DIFF
--- a/bundle/edu.gemini.model.p1/src/test/scala/edu/gemini/model/p1/visibility/TestTargetVisibilityTest.scala
+++ b/bundle/edu.gemini.model.p1/src/test/scala/edu/gemini/model/p1/visibility/TestTargetVisibilityTest.scala
@@ -233,20 +233,20 @@ class TestTargetVisibilityTest {
   }
 
   @Test def testDecForGNLgs() {
-    // between higher than -27 and less or equal than -22 is warning
-    gnLgsB(Limited, ("9:00:00",  "-26:59:59.999"))
-    gnLgsB(Limited, ("9:00:00",  "-22:00:00"))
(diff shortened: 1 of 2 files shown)


bundle/edu.gemini.model.p1/src/test/scala/edu/gemini/model/p1/visibility/TestTargetVisibilityTest.scala
```diff
@@ -233,20 +233,20 @@ class TestTargetVisibilityTest {
...
     gnLgsB(Bad, ("9:00:00",  "-90:00:00"))
-    // More or or equal to 68 is Bad
-    gnLgsB(Bad, ("9:00:00",  "68:00:00"))
+    // More or or equal to 70 is Bad
+    gnLgsB(Bad, ("9:00:00",  "70:00:00"))
     gnLgsB(Bad, ("9:00:00",  "90:00:00"))
-    // Between more than -22 and less than 65 is Good
-    gnLgsB(Good, ("0:00:00",  "22:00:00.001"))
+    // Between more than -25 and less than 65 is Good
+    gnLgsB(Good, ("0:00:00",  "-24:59:59.999"))
```
sraaphorst: Note that this also improves this test case.
Previously, despite the lower limit being greater than -22, for some reason, 22 was used.
The new lower limit is anything greater than -25, so I moved the test to the smallest acceptable value greater than this.

cquiroz:
Looks good to me :+1:

PR #1054, opened 2016-08-01
REL-1389: Change AGS step size for F2OI

sraaphorst:
The PIT uses a servlet to communicate with a running instance of the ODB (gnodb) to obtain estimates for being able to use AGS to find guide stars, which is accomplished by trying a variety of position angles through a series of step sizes.

The default step size for most single probes is in 10 degree increments, as is the default in `SingleProbeStrategyParams`. Flamingoes 2 OIWFS overrode this to instead us 90 degree increments, which was found to be too large and give unsatisfactory estimates.

This reduces the step size to the default of 10 degree increments by removing the overriding step size of 90 degrees. Speed comparisons for 10 degree versus 90 degree increments on a test program provided by Andy were done from GS on an ODB instance installed on `gnodbtest2` and were found to be unremarkable in difference.


cquiroz:
Looks good to me :+1:
I guess we need to redeploy the ODB to let the PIT use it right?

sraaphorst:
Yes, that is correct! I don't know what kind of rush there is for this but I will discuss with @andrewwstephens . Since there are no model changes, it should consist solely of a stop of 2016B-2.1.1 and a start of a new build 2016B-2.1.2.

If we want to do that, obviously I will create a PR for the 2016B branch.

PR #1055, opened 2016-08-03
REL-1217: Fixed PIT ephemeris check issues and reporting.

sraaphorst:
As per the REL task, the PIT was mis-reporting missing ephemeris data.
This PR fixes that, and a number of other bugs that were discovered in the process.

1. One of the main issues was misuse of `java.util.Calendar`, which was being used with the assumption that the `Calendar.DAY_OF_MONTH` field is indexed starting at 0, when it is actually indexed starting at 1. (Using a `DAY_OF_MONTH` of 0 will actually give the last day of the previous month, delightfully confusingly.) Thus, `Semester.firstDay` and `Semester.lastDay` both occurred one day too early.
2. Additionally, `Calendar.HOUR` was used instead of `Calendar.HOUR_OF_DAY`, which made the `Semester.firstDay` and `Semester.lastDay` flip from time `00:00:00` to `12:00:00` after noon.
3. There was a contradiction in the `emptyEphemerisCheck` so that it was never being generated: the `for` loop had `if !t.isEmpty` (for `t: NonSiderealTarget`) and `if e.isEmpty` (for `e: List[EphemerisElement]` for `t`) as conditions for execution, and `NonSiderealTarget.isEmpty` simply checks that its `List[EphemerisElement]` is empty, so these calls could never both be true. This was fixed and the list is now only checked for emptiness.
4. General cleanup: redundant variables were either removed or used instead of alternatives, and ephemeris checks were grouped together in the list of problems.
5. `TargetEditor.defaultDay` would add a day to `Semester.firstDay`, I assume, to work around the incorrect use of `Calendar` that made `Semester.firstDay` one day too early. This was removed and the code simplified.

### DIFF
--- a/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Semester.scala
+++ b/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Semester.scala
@@ -1,7 +1,7 @@
 package edu.gemini.model.p1.immutable
 
-import edu.gemini.model.p1.{ mutable => M }
-import java.util.{TimeZone, Calendar}
+import edu.gemini.model.p1.{mutable => M}
+import java.util.{Calendar, TimeZone}
 
 object Semester {
 
@@ -65,8 +65,8 @@ case class Semester(year: Int, half: SemesterOption) {
     cal.setTimeZone(TimeZone.getTimeZone("UTC"))
     cal.set(Calendar.YEAR, year)
     cal.set(Calendar.MONTH, half.firstMonth)
-    cal.set(Calendar.DAY_OF_MONTH, 0)
-    cal.set(Calendar.HOUR, 0)
+    cal.set(Calendar.DAY_OF_MONTH, 1)
+    cal.set(Calendar.HOUR_OF_DAY, 0)
     cal.set(Calendar.MINUTE, 0)
     cal.set(Calendar.SECOND, 0)
     cal.set(Calendar.MILLISECOND, 0)
@@ -81,8 +81,8 @@ case class Semester(year: Int, half: SemesterOption) {
       case SemesterOption.B => year + 1
     })
     cal.set(Calendar.MONTH, half.lastMonth)
-    cal.set(Calendar.DAY_OF_MONTH, 0)
-    cal.set(Calendar.HOUR, 0)
+    cal.set(Calendar.DAY_OF_MONTH, 1)
+    cal.set(Calendar.HOUR_OF_DAY, 0)
     cal.set(Calendar.MINUTE, 0)
     cal.set(Calendar.SECOND, 0)
     cal.set(Calendar.MILLISECOND, 0)
--- a/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/EphemerisElement.scala
+++ b/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/EphemerisElement.scala
@@ -1,6 +1,6 @@
 package edu.gemini.model.p1.immutable
 
-import edu.gemini.model.p1.{ mutable => M }
+import edu.gemini.model.p1.{mutable => M}
 import java.util.GregorianCalendar
 
 import edu.gemini.spModel.core.Coordinates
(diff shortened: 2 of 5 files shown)


cquiroz:
Excellent! Sorry you had to deal with `java.util.Calendar`. I'm surprised b the bad check on the `ephemeris` check. I thought somebody would have spotted it already

Looks good to me

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Semester.scala
```diff
@@ -65,8 +65,8 @@ case class Semester(year: Int, half: SemesterOption) {
     cal.setTimeZone(TimeZone.getTimeZone("UTC"))
     cal.set(Calendar.YEAR, year)
     cal.set(Calendar.MONTH, half.firstMonth)
-    cal.set(Calendar.DAY_OF_MONTH, 0)
-    cal.set(Calendar.HOUR, 0)
+    cal.set(Calendar.DAY_OF_MONTH, 1)
+    cal.set(Calendar.HOUR_OF_DAY, 0)
```
swalker2m: I thought, technically, the semester starts at 2pm of the previous day.  So for example the A semester starts on Jan 31 at 14:00.  That's the first observing night of the semester, Jan 31 - Feb. 1

cquiroz: That's a good observation, would then the semester also at the previous night of the obvious end date?

sraaphorst: Interesting. I didn't know this. Should be easy enough to fix, of course, and it makes more sense than starting at midnight given observing would never start that late under acceptable conditions.

As Carlos said, how does this figure into the end of semester? I am assuming the same, too, then, regarding 2:00 PM on July 31?

swalker2m: In the core `Semester` implementation the start date is defined as mentioned above and the end date is calculated by adding 6 months to the start date via `Calendar.add`.  Would it be possible to just use the core `Semester` instead of having a different version?  I think the P1 version was split off back when the PIT was a separate codebase but presumably there's no need to keep different `Semester` implementations now?

sraaphorst: I don't see why it isn't doable, but I think given that the code freeze is Friday and I still have a couple tasks that need to be done, it probably isn't something we can do before then, at least, unfortunately. I did a code search and the P1 `Semester` implementation is widely used, so it would require probably a couple days of work, I would imagine.

cquiroz: P1 `Semester` is fairly entrenched in the p1 model. I'd make a JIRA task to remind us to do it in the next iteration.

I reviewed ITAC and P1 `Semester` is barely used, instead relying on the ITAC version of `Semester` so this task should be doable

sraaphorst: I'll go ahead and make a JIRA task for this now.

andrewwstephens: Not sure this matters, but in practice the first night of the semester is local (not UT)  Feb 1 (A) and Aug 1 (B).

swalker2m: @andrewwstephens  The core semester [gets the time right](https://github.com/gemini-hlsw/ocs/blob/develop/bundle/edu.gemini.spModel.core/src/main/java/edu/gemini/spModel/core/Semester.java#L69-L73), by the way.  I agree though that it's not a good time to go about replacing the PIT `Semester` class.

sraaphorst:
From my discussion (via e-mail) with Andy:


> For the PIT, we are currently representing semester dates with UTC midnight
> from Feb 1 - July 31, and Aug 1 - Jan 31, so that may require some changes.
> I'm not sure how important it is, though, to have this be fully accurately
> represented in the PIT.

This sounds okay for the PIT. I think there was just confusion about
why the PIT was requesting ephemeris data for before the semester
stars and why it didn't want it for the last day of the semester.

Thus, I propose that we go ahead with these changes as are for this release, unless anyone has significant objection.

cquiroz:
I agree, it would be good at some moment to merge both `Semester` implementations though but I'll leave that for the next release.

PR #1056, opened 2016-08-04
REL-2691: Added Korean institutions and required NGO / partner information

sraaphorst:
Pretty straightforward from the subject.
Note that since Korea is an inaccurate country name, as per the data provided in the JIRA task, I changed the code internally to use "Republic of Korea" as well.

### DIFF
--- a/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Partners.scala
+++ b/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Partners.scala
@@ -12,7 +12,7 @@ object Partners {
     NgoPartner.BR -> "Brazil",
     NgoPartner.CA -> "Canada",
     NgoPartner.CL -> "Chile",
-    NgoPartner.KR -> "Korea",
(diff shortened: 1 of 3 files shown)


bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Partners.scala
```diff
@@ -12,7 +12,7 @@ object Partners {
     NgoPartner.BR -> "Brazil",
     NgoPartner.CA -> "Canada",
     NgoPartner.CL -> "Chile",
-    NgoPartner.KR -> "Korea",
+    NgoPartner.KR -> "Republic of Korea",
```
cquiroz: Actually the name `Korea` was provided by Andy Adamson as part of the agreements with Korea, I'd say we shouldn't change it. If we do change it we may need to change the upconversion process

sraaphorst: Sure. I'll go ahead and change this back and modify the XML.

cquiroz:
Looks good to me :+1:

PR #1058, opened 2016-08-12
REL-2909: Changed PIT configuration handling of debug flags

sraaphorst:
Originally, a test release of the PIT would be built if `app/pit/build.sbt` had the property `edu.gemini.pit.test` defined at all. This was confusing because in `app/pit/build.sbt`, it was set to `true` and commented out to create an official release: its actual contents were never looked at, so setting it to `false` would make it defined and thus also result in a test release being built.

This cleans that up by ONLY creating a test release / running in test mode if the property is set to `true`.
If it is set to `false`, it is not run in test mode.
If it is set to anything else or undefined, a warning is logged and it is also not run in test mode.

### DIFF
--- a/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/model/AppMode.scala
+++ b/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/model/AppMode.scala
@@ -7,13 +7,11 @@ import scala.util.Try
 object AppMode {
   private val Log = Logger.getLogger(AppMode.getClass.getName)
 
-  lazy val isTAC  = isDefinedAndTrue("edu.gemini.pit.tac")
-  lazy val isTest = isDefinedAndTrue("edu.gemini.pit.test")
-
-  private def isDefinedAndTrue(prop: String): Boolean = Try {
-    System.getProperty(prop).toBoolean
+  val TestProperty = "edu.gemini.pit.test"
+  lazy val isTest = Try {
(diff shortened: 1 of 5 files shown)


bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/model/AppMode.scala
```diff
@@ -1,10 +1,19 @@
...
+
+import scala.util.Try
 
-  lazy val isTAC = isDefined("edu.gemini.pit.tac")
-  lazy val isTest = isDefined("edu.gemini.pit.test")
+object AppMode {
+  private val Log = Logger.getLogger(AppMode.getClass.getName)
 
-  private def isDefined(prop:String) = Option(System.getProperty(prop)).isDefined
+  lazy val isTAC  = isDefinedAndTrue("edu.gemini.pit.tac")
```
cquiroz: It seems `isTAC` isn't used? the TAC mode is a user option rather than an application property

sraaphorst: It certainly isn't set anywhere in the code, so yes, it would have to be a user option.
It looks to be used in the `UndoAction` and `RedoAction`, though, so I was hesitant to make any changes to it.

cquiroz:
Looks good to me :+1:

PR #1059, opened 2016-08-12
REL-2910: Update radial range in Catalog Query Tool when WFS changes

cquiroz:
This PR adds a missing feature to the Catalog Query Tool. At the moment, when the user selects a different guider, the search form magnitudes is updated but not the radius range, thus making it necessary to set it manually

This PR acts when the guider's selection is updated and will set the radius range appropriately

screenshot 2016-08-12 11 54 59


sraaphorst:
Looks good to me. 👍

PR #1060, opened 2016-08-16
REL-2918: Added more AU institutions and fixed spacing issues

sraaphorst:
This PR does two things:
1. Adds four new Australian institutions to the PIT; and
2. Removes some double spaces that were in institution names.


cquiroz:
Looks good to me :+1:

PR #1061, opened 2016-08-16
REL-2879: Added warning for invalid phone numbers

sraaphorst:
A new task came in requesting that the PIT issue a warning for PI phone numbers with less than 10 digits.

### DIFF
--- a/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
+++ b/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
@@ -81,7 +81,7 @@ class ProblemRobot(s: ShellAdvisor) extends Robot {
 
     lazy val all = {
       val ps =
-        List(noObs, nonUpdatedInvestigatorName, noPIPhoneNumber, titleCheck, band3option, abstractCheck, tacCategoryCheck,
+        List(noObs, nonUpdatedInvestigatorName, noPIPhoneNumber, invalidPIPhoneNumber, titleCheck, band3option, abstractCheck, tacCategoryCheck,
           keywordCheck, attachmentCheck, attachmentValidityCheck, attachmentSizeCheck, missingObsDetailsCheck,
           duplicateInvestigatorCheck, ftReviewerOrMentor, ftAffiliationMismatch, band3Obs).flatten ++
           TimeProblems(p, s).all ++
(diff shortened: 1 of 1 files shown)


bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
```diff
@@ -612,12 +612,16 @@ class ProblemRobot(s: ShellAdvisor) extends Robot {
...
-        "Overview", s.inOverview{
-          _.editPi(_.Phone.requestFocus)
-        })
+      new Problem(Severity.Warning,
+        s"No phone number given for ${investigatorFullName(p.investigators.pi, "PI")}. This is for improved user support.",
+        "Overview", s.inOverview(_.editPi(_.Phone.requestFocus)))
     }
 
+    private val invalidPIPhoneNumber = (for {
+      badPhone <- p.investigators.pi.phone if badPhone.count(_.isDigit) < 10
```
cquiroz: Should we also trigger an error if e.g. the phone number contains letters? I guess `+` should be allowed

sraaphorst: I feel very hesitant to do something like that. Extensions, for example, contain the letter x, typically, and although I don't think there are any other cases in the institutions list now of anything else, it is entirely possible that something might come up. Additionally, how complicated would we want to take this? Parentheses (in which case, do we only allow one set, require them to be balanced, etc), a multitude of separators [.-/], and all kinds of things could crop up, as could an optional leading +, as you mention.

I was trying to find a regex online for phone numbers as a curiosity, and even one for just North America was ridiculously complicated. I don't think we want to take things that far, at least until science requests it.

cquiroz: That sounds fine, I think you're right

cquiroz:
Looks good to me :+1:

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
swalker2m: I'm even a little surprised that there isn't a country that uses 9 or fewer digits.

sraaphorst: Actually, looking online, the assumption that a phone number will contain at least 10 digits is wrong. Even with country code, Swedish phone numbers can have nine digits, and phone numbers for the Solomon Islands, eight.

Out of the partner countries, they will all have 10+, but this seems like a poor assumption to make. I'm going to make note on the JIRA task for this.

sraaphorst:
Okay, spoke with @andrewwstephens and we agreed to set it to eight digits, which is the minimum legal number for any international phone number. I also simplified the code to use `find` instead of the ridiculous `for`...`headOption` I was going before.

PR #1062, opened 2016-08-16
REL-2911: Set the OT icon image for all child windows

cquiroz:
We have been setting manually the icon for all OT frames as required for Windows/Linux but a few have been missing. This PR fixes that setting the icon for the remaining dialog/frames

This task turned out harder as I had to move the `Resources` class to `jsky.util.gui` and fix the OSGI exports to let classes in other bundles be able to use the icon


bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/QvTool.scala
```diff
@@ -22,27 +23,28 @@ import scala.swing.{Frame, Swing}
...
+  Resources.setOTFrameIcon(peer)
   visible = true
 }
 
 /**
  * Make utility services used in QV available throughout the plugin.
  */
 object QvTool {
-  var defaultsFile: File = null
+  var defaultsFile: File = _
```
sraaphorst: Just out of curiosity, could someone explain the difference between setting a `var` to `_` versus `null` and why it is preferable to use `_`?

bundle/edu.gemini.util.security/src/main/scala/edu/gemini/util/security/auth/ui/AuthDialog.scala
```diff
@@ -106,7 +107,7 @@ class AuthDialog(ac: KeyChain, detailText: String =  "Database keys allow you to
     def apply {
       DatabaseTab.Table.selectedItem.foreach { db =>
         val keysExist: Boolean =
-          ac.keys.map(_.get(db).filterNot(_.isEmpty).isDefined).unsafeRun.fold(throw _, identity)
+          ac.keys.map(!_.get(db).forall(_.isEmpty)).unsafeRun.fold(throw _, identity)
```
sraaphorst: `ac.keys.map(_.get(db).exists(_.nonEmpty))` might be easier?

bundle/jsky.app.ot/src/main/java/jsky/app/ot/progadmin/AdminDialog.java
```diff
@@ -44,8 +37,9 @@ ISPProgram show(IDBDatabaseService database, ISPProgram prog)  {
 
         pane.setMessage(ui);
 
-        JDialog dialog = pane.createDialog(owner, title);
+        JDialog dialog = pane.createDialog(null, title);
```
sraaphorst: Why is the `owner` suddenly unimportant? Do we want an OS-wide modal dialog, or does this have some other behaviour?

Indeed, according to the Java API docs, it doesn't seem that the use of `null` here is even clearly stated. Can we always assume it will work properly?

If we are going to go this route, it seems that using `pane.createDialog(title)` might the correct method call to make?

sraaphorst:
As for everything else, good stuff. 👍

bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/QvTool.scala
cquiroz: In this case it doesn't make much of a difference, `_` will assign the default value which is `null` anyway. It just looks nicer.

swalker2m: I'm not sure how or where this is used but perhaps we should switch it to an `Option[File]` anyway.

cquiroz: Right, I'll give it another review, perhaps these can be initialized to `val`s directly

bundle/jsky.app.ot/src/main/java/jsky/app/ot/progadmin/AdminDialog.java
cquiroz: If you follow the original code upstream, owner was always `null`, so I removed the parameter and set it explicitly to `null`.

Using `null` basically tells that the box has no parent but as you said `pane.createDialog(title)` may be better in this case

bundle/edu.gemini.util.security/src/main/scala/edu/gemini/util/security/auth/ui/AuthDialog.scala
cquiroz: 👍

cquiroz:
Added fixes according to the PR comments

PR #1063, opened 2016-08-16
REL-2811: Replaced PIT icon as per request.

sraaphorst:
The subject says it all.


cquiroz:
Looks fine, what's the difference with the previous one?
We have been doing the rebase locally on the command line rather than using the github UI, which is apparently what you did?

sraaphorst:
Ahhh, I didn't realize that there was a significant difference between rebasing on the command line versus here in github. The issue, of course, is that I'll make a PR, and then have to wait ages for travis to build, and in the interim, make another PR, at which point, there is nothing to rebase, but github is insistent that I need to update on the previous PR pending the travis build, I think.

The new icon is square instead of a circular transparent PNG.

cquiroz:
I'm not sure if there is a difference in practice one one way or the other though the history looks different

sraaphorst:
Good to know. I'll keep that in mind and stick to rebasing, then.

PR #1064, opened 2016-08-16
REL-2888: Cleaned up UpConverter

sraaphorst:
This is a major cleanup of `UpConverter`, since some of the conversions in the past simply bumped version numbers, which should no longer be necessary for `2017A` as the later version was handled identically to the earlier version in terms of later upgrading.

Thus, this should simplify things substantially while still allowing full proposal upgrades.

### DIFF
--- a/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/transform/UpConverter.scala
+++ b/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/transform/UpConverter.scala
@@ -33,19 +33,16 @@ object UpConverter {
     convert(node).map(toConversionResult)
   }
 
-  // Sequence of conversions for proposals from a given semester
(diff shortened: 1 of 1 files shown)


bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/transform/UpConverter.scala
```diff
@@ -56,22 +53,16 @@ object UpConverter {
   def convert(node: XMLNode):Result = node match {
     case p @ <proposal>{ns @ _*}</proposal> if (p \ "@schemaVersion").text == Proposal.currentSchemaVersion =>
       StepResult(Nil, node).successNel[String]
-    case p @ <proposal>{ns @ _*}</proposal> if (p \ "@schemaVersion").text == "2016.2.2"                    =>
+    case p @ <proposal>{ns @ _*}</proposal> if (p \ "@schemaVersion").text.matches("2016.2.[12]")           =>
```
cquiroz: nice!

cquiroz:
This looks good. I guess there is no need to keep the intermediate special versions if they don't have any relevant changes

Looks good to me :+1:

PR #1065, opened 2016-08-17
REL-2914: Instrument changes for PIT 2017A.

sraaphorst:
This PR changes the instrument availability in the PIT for 2017A proposals as follows:
= Allow TEXES for GN only.
= Allow DSSI for both GN and GS (formerly was only allowed for GN).


cquiroz:
Looks good to me :+1:

PR #1066, opened 2016-08-17
REL-2920: Fixing parallactic angle formatting and warning

sraaphorst:
When the parallactic angle mode is chosen, the parallactic angle is calculated based on the scheduling block and the angle entered for the PA.

In the past, if the user "overrode" this PA, a warning would be issued in the form of a small yield sign next to the parallactic angle calculation in the display (see first figure).

If the user restored the value to the parallactic angle calculation, the warning would disappear (see second figure).

This functionality was broken by code changes for the target model (I suspect) along the way, as the number formatter used for parallactic angle display was altered in the code (and used direct string formatting to one decimal place instead of two) and thus was no longer consistent with the internal check. I have restored the code to use the formatter: thus, any changes to angle truncation or formatting are limited to one place.

This simply fixes that to restore the warning to the proper behaviour.
parangle1
parangle2


cquiroz:
Seems funny that a warning depends on the formatting of  a string. Shouldn't it use the actual value instead?

sraaphorst:
We played around with both options at the time, but since we were setting / getting values from a text field, I think we went with string formatting.

I think there was another reason, too, but I can't remember precisely what it was.

cquiroz:
Sounds good 👍

PR #1067, opened 2016-08-18
REL-2922: Parallactic angle setting PA at too high a precision

sraaphorst:
When using the parallactic angle mode for observations, the resultant PA is calculated at an extremely high precision. This resulted in issues with the instrument configuration: for example, the parallactic angle would be calculated and set on the instrument to a high level of precision, and then BAGS would run, and then set the PA to the "almost equals" angle at a lower precision, which would cause issues with the parallactic angle once again to be calculated and determine itself to be different from the instrument PA. Lather, rinse, repeat.

This change comprises:
1. Using a more standard precision for setting the PA on the instrument for parallactic angle calculations;
2. Not recognizing a 180 flip from the calculated parallactic angle (which can be used by BAGS in GS selection and is permitted) as a new PA; and
3. General cleanup.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/parallacticangle/ParallacticAngleSupportInst.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/parallacticangle/ParallacticAngleSupportInst.java
@@ -4,7 +4,8 @@ import edu.gemini.pot.sp.ISPObservation;
 import edu.gemini.pot.sp.SPComponentType;
 import edu.gemini.shared.util.immutable.ImOption;
 import edu.gemini.shared.util.immutable.Option;
-import edu.gemini.skycalc.Angle;
+import edu.gemini.spModel.core.Angle;
+import edu.gemini.spModel.core.Angle$;
 import edu.gemini.spModel.inst.ParallacticAngleSupport;
 import edu.gemini.spModel.obs.ObsTargetCalculatorService;
 import edu.gemini.spModel.obscomp.SPInstObsComp;
@@ -29,7 +30,6 @@ public abstract class ParallacticAngleSupportInst extends SPInstObsComp implemen
     public Option<Angle> calculateParallacticAngle(final ISPObservation obs) {
         return ImOption.fromScalaOpt(ObsTargetCalculatorService.targetCalculation(obs))
                 .flatMap(targetCalculator -> ImOption.fromScalaOpt(targetCalculator.weightedMeanParallacticAngle()))
-                .map(angleObj -> (double) angleObj)
-                .map(angle -> (new edu.gemini.skycalc.Angle(angle, edu.gemini.skycalc.Angle.Unit.DEGREES)).toPositive());
+                .map(angleObj -> Angle$.MODULE$.fromDegrees((double)angleObj));
     }
 }
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/parallacticangle/ParallacticAngleSupportInst.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/parallacticangle/ParallacticAngleSupportInst.java
@@ -2,14 +2,12 @@ package edu.gemini.spModel.gemini.parallacticangle;
 
 import edu.gemini.pot.sp.ISPObservation;
 import edu.gemini.pot.sp.SPComponentType;
-import edu.gemini.shared.util.immutable.None;
+import edu.gemini.shared.util.immutable.ImOption;
 import edu.gemini.shared.util.immutable.Option;
-import edu.gemini.shared.util.immutable.Some;
 import edu.gemini.skycalc.Angle;
 import edu.gemini.spModel.inst.ParallacticAngleSupport;
 import edu.gemini.spModel.obs.ObsTargetCalculatorService;
 import edu.gemini.spModel.obscomp.SPInstObsComp;
-import edu.gemini.util.skycalc.calc.TargetCalculator;
 
 /**
  * A superclass for instruments that support the parallactic angle feature with most of the implementation in place.
@@ -19,7 +17,7 @@ public abstract class ParallacticAngleSupportInst extends SPInstObsComp implemen
     /**
      * Constructor and methods for parallactic angle support.
      */
-    protected ParallacticAngleSupportInst(SPComponentType type) {
+    protected ParallacticAngleSupportInst(final SPComponentType type) {
         super(type);
     }
 
@@ -28,28 +26,10 @@ public abstract class ParallacticAngleSupportInst extends SPInstObsComp implemen
      * Instruments who need specific modifications based on settings should override this, e.g. GMOS.
      */
     @Override
-    public Option<Angle> calculateParallacticAngle(ISPObservation obs) {
-        final Option<TargetCalculator> targetCalculatorOption = ObsTargetCalculatorService.targetCalculationForJava(obs);
-        if (!targetCalculatorOption.isEmpty()) {
-            final TargetCalculator targetCalculator = targetCalculatorOption.getValue();
-
-            // Calculate the weighted angle.
-            scala.Option<Object> angleOption = targetCalculator.weightedMeanParallacticAngle();
-            if (angleOption.nonEmpty()) {
-                // Calculate the parallactic angle.
-                final double dAngle = (Double) angleOption.get();
-                final edu.gemini.skycalc.Angle angle = (new edu.gemini.skycalc.Angle(dAngle, edu.gemini.skycalc.Angle.Unit.DEGREES)).toPositive();
-                return new Some<>(angle);
-            } else return None.instance();
-        } else return None.instance();
-    }
-
-    /**
-     * By default, assume that the instrument is compatible with the parallactic angle feature unless indicated
-     * otherwise.
-     */
-    @Override
-    public boolean isCompatibleWithMeanParallacticAngleMode() {
-        return true;
+    public Option<Angle> calculateParallacticAngle(final ISPObservation obs) {
+        return ImOption.fromScalaOpt(ObsTargetCalculatorService.targetCalculation(obs))
+                .flatMap(targetCalculator -> ImOption.fromScalaOpt(targetCalculator.weightedMeanParallacticAngle()))
+                .map(angleObj -> (double) angleObj)
+                .map(angle -> (new edu.gemini.skycalc.Angle(angle, edu.gemini.skycalc.Angle.Unit.DEGREES)).toPositive());
     }
 }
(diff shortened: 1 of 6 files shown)


sraaphorst:
Note that this, by (2) in the list above, also solves REL-2923, as they are linked.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/parallacticangle/PositionAnglePanel.scala
```diff
@@ -221,15 +217,26 @@ class PositionAnglePanel[I <: SPInstObsComp with PosAngleConstraintAware,
...
   /**
+    * The actual copying of a given pos angle to the data object.
+    * This is done explicitly to avoid overwriting based on minor differences in double precision and to avoid
+    * adding 180 to the position angle, which causes issues.
+    **/
+  private def setInstPosAngle(newAngleDegrees: Double): Unit = editor.foreach { e =>
+    val Precision = 0.005
+    val oldAngleDegrees = e.getDataObject.getPosAngle
+    if (Math.abs(oldAngleDegrees - newAngleDegrees) >= Precision && Math.abs(Math.abs(oldAngleDegrees - newAngleDegrees) - 180) >= Precision) {
+      e.getDataObject.setPosAngle(newAngleDegrees)
```
swalker2m: Could you explain a bit more here?  From the point of view of the guide probe range, often there is a big difference when adding 180 degrees.  For example with GMOS-OI, the guide stars reachable at 0 degrees are not the same as those at 180.  But perhaps this setter is only called by the parallactic angle calculation?

sraaphorst: Well, technically, this IS called by both the parallactic angle code AND whenever the contents of the text field are manually changed. My initial thought was that every time a keystroke is typed in the field, this is called, and it is impossible, I believe, with one numerical keystroke to increment a PA by 180 so different processing for regular PA and parallactic angle shouldn't be required.

Then I realized, though, that of course, copy-pasting could change the angle by 180 with a single key combination, so that nixes that idea.

I think maybe the best thing is to do the 180 equivalence / ignore change check IF the pos angle constraint is parallactic angle, and otherwise ignore it.

sraaphorst: I also spotted some more opportunity to clean up cruft (unused variables, unnecessary use of `for` constructs, etc), so I'm going to do that as well.

swalker2m: I still don't quite get what is going on here.  What is sequence of events with the 180 flip that is problematic?  If we've chosen to observe at the average parallactic angle, is the position angle still user editable?  If so why?

sraaphorst: Yes, the PA is still user-editable. We decided at the time of parallactic angle implementation that the user should be able to select parallactic angle mode, have the parallactic angle displayed in the calculation area, but then manually override the PA. I think it was Andy that specifically requested this functionality.

As for the 180 flip, this is just a horror story of BAGS and parallactic angle fighting each other.

A number of things trigger the parallactic angle calculation, which since the target model change, now reconfigures the entire widget with whatever the parallactic angle is calculated to be. Say this is x. Then BAGS will run for x, and find, say, an optimal GS at x+180, setting the PA to x+180. This will trigger a new parallactic angle calculation by the way the code was changed, which will revert the PA to x. Then BAGS does not run, because the hash is still set to PA of x since that was where the last lookup was done. This results in a GS that is then reported out of bounds.

swalker2m:
> Yes, the PA is still user-editable. We decided at the time of parallactic angle implementation that the user should be able to select parallactic angle mode, have the parallactic angle displayed in the calculation area, but then manually override the PA. I think it was Andy that specifically requested this functionality.

@andrewwstephens Is this correct?  This seems unnecessarily confusing.  If the user wants to explicitly set the position angle, then perhaps s/he should select the fixed position angle setting instead?  Sorry if you've already explained this 10x, I've forgotten the rationale but it makes the UI and implementation more complicated.

> A number of things trigger the parallactic angle calculation, which since the target model change, now reconfigures the entire widget with whatever the parallactic angle is calculated to be. Say this is x. Then BAGS will run for x, and find, say, an optimal GS at x+180, setting the PA to x+180. This will trigger a new parallactic angle calculation by the way the code was changed, which will revert the PA to x.

This is maybe where the fix should be applied?  Only when the average parallactic angle calculation comes up with a value that is the same as the existing value + 180, call it good and don't update the angle?  You wouldn't want to do this for explicit user input or BAGS updates in any case.

> Then BAGS does not run, because the hash is still set to PA of x since that was where the last lookup was done. This results in a GS that is then reported out of bounds.

If the average parallactic angle calculation (or anything else) actually changes the position angle BAGS *should* run again because the hash value should differ.  If what you are saying is correct, this could be another bug.

sraaphorst:
The BAGS / parallactic angle interactions were quite a head-on-desk banging.

From my notes regarding the observation in the fault report on which this is based:
= When parallactic angle mode is selected, the parallactic angle is calculated as 61.
= BAGS runs, finds a GS at 61+180 = 241, and then sets the PA to 241.
= BAGS then moves the observation state from `Running` to `Idle`, but with the same original hash that reflects the PA being 61.
= This change to the PA text field triggers the parallactic angle controls `init` method, which sees that the angle has changed.
= The parallactic angle is recalculated, and set from 241 back to 61.
= BAGS is then triggered due to the change, but calculates the same hash, and determines that it doesn't need to run again.

So my previous description may not have been entirely correct.
The parallactic angle controls - I think based on changes made to them since their original implementation - have become a bit "over-exuberant" in their reinitialization, which is one potential place where things could be looked at, but I suspect this would require a substantial amount of effort.

Certainly, solutions could also include:
= Moving the ignore of a 180 degree difference to the parallactic angle code.
= Disallowing the editing of the PA when parallactic angle is selected.
(The second option would also eliminate REL-2924, since this functionality is no longer working anyway.)

sraaphorst:
For reference, the observation where this is easily reproducible and to which I refer above is:
GS-2016B-Q-64-55

Also note that there is REL-2921, which requires moving some of the parallactic angle controls methods to run on the Swing EDT so that they occur in the right order with BAGS, as right now, it is easy to make things happen out of order due to BAGS only running on the EDT.

swalker2m:
The BAGS/parallactic angle calculation duel sounds plausible especially if they are running in separate threads.  It's a race condition.  In the absence of the parallactic angle calculation if BAGS finds a guide star at pos angle + 180 and does the corresponding update, it runs again because the recorded hash value doesn't match the current hash.  Presumably it comes up with the same result as the previous though so only the hash value is updated.

sraaphorst:
Yes, what you're saying makes complete sense. The parallactic angle calculation being triggered by the BAGS update, which occurs before the next BAGS run, is what seems to be causing the problem by flipping the angle back.

Apart from ripping apart the parallactic angle code, I'm not sure how to address this save ignoring 180 degree changes to PA by the parallactic angle `init`.

swalker2m:
> Apart from ripping apart the parallactic angle code, I'm not sure how to address this save ignoring 180 degree changes to PA by the parallactic angle controls init.

If that's where the calculation is done, and the +180 comparison can be confined to the calculation, then that makes sense.  I skimmed thorough the parallactic angle controls code (which isn't part of this PR so far) and found a couple of issues:

1. When it sets the position angle it explicitly turns off events to avoid triggering BAGS?  That could also account for why BAGS doesn't run.

2. Updates to the UI are happening in a non-swing thread which is just asking for mysterious problems.

sraaphorst:
I believe these are both things I addressed in the next PR I was planning on pulling out of my general FR mucking, i.e. for REL-2921, in which case, the order of events is as described above: BAGS runs for x and maintains hash for x, sets PA to x+180, parallactic angle runs, sets x, BAGS runs, sees x, does nothing.

The parallactic angle calculation is called, IIRC, from a couple places, but I think it would make sense to limit the 180 comparison... the reason I placed it where I did was because of handling the precision issue at the same time, which will have to be done when checking the 180 and when checking whether the parallactic angle calculation is "almost equal" to the PA.

tpolecat:
My work on this stuff was a journey through several distinct states of rage and confusion, so I apologize for any screwups that I introduced. We kind of painted ourselves into a corner I think.

sraaphorst:
Understood. I'm sure it was probably almost entirely fine, too, until BAGS fully joined the fray.
Due to all the things both BAGS and parallactic angle mess with, it's not really surprising, I think, that this situation was pretty much bound to arise.

sraaphorst:
@swalker2m From my conversation with @andrewwstephens about this:
"We requested that the PA value remain editable when set to the average parallactic angle so that the night crew does not need to change the PA mode, thereby erasing the PI's desire to be at the average parallactic angle, in order to tweak the PA when a guide star is not available at exactly the average parallactic angle.
It is not uncommon to tweak the PA by a few degrees to get a guide star."

sraaphorst:
Please note that I made a number of changes in a recent commit to this PR that I think address the comments made by all.

Bryan was hoping to have an OT test release out by Monday that would fix the bulk of the parallactic angle vs. BAGS issues, so I am hoping to have this merged by then.

(There will be one more PR regarding handling parallactic angle features on the Swing EDT to work with BAGS.)

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/parallacticangle/ParallacticAngleSupportInst.java
```diff
@@ -28,28 +26,10 @@ protected ParallacticAngleSupportInst(SPComponentType type) {
...
-     * otherwise.
-     */
-    @Override
-    public boolean isCompatibleWithMeanParallacticAngleMode() {
-        return true;
+    public Option<Angle> calculateParallacticAngle(final ISPObservation obs) {
+        return ImOption.fromScalaOpt(ObsTargetCalculatorService.targetCalculation(obs))
+                .flatMap(targetCalculator -> ImOption.fromScalaOpt(targetCalculator.weightedMeanParallacticAngle()))
+                .map(angleObj -> (double) angleObj)
+                .map(angle -> (new edu.gemini.skycalc.Angle(angle, edu.gemini.skycalc.Angle.Unit.DEGREES)).toPositive());
```
cquiroz: Couldn't we use the new target model Angle here?

sraaphorst: I suppose we should get in the habit of switching things over to that.
I made the changes and everything appears to be working well!

If someone could 👍  this this morning or offer more feedback, that would be excellent so I can move ahead with the `Swing.EDT` issues for the `PositionAnglePanel` and `ParallacticAngleControls` so that we can attempt to get a test release out today.

cquiroz:
I reviewed the last 2 commits and it looks good to me 👍

PR #1068, opened 2016-08-22
REL-2691: Changed Korea to ROK.

sraaphorst:
As per discussions, Korea was changed to Republic of Korea.
This will not affect earlier proposals, as there were previously no Korean institutions available for selection, and setting the affiliate manually to Korea simply wrote `kr` in the proposal XML, which should not be affected by this, especially in light of REL-2930.


cquiroz:
I think there can be earlier proposals with time requested for korea. do they need handling to be upconverted? Or do they have `kr` anyway?

sraaphorst:
2016A and 2016B both code Korea as `kr` in the proposal XMLs, and I see no upconverting for anything before that point, so I think it is safe to assume that it is always `kr`?

cquiroz:
`NgoPartner` in the schema tells us that we use `kr` to refer to the partner so this change doesn't require an upconversion

Looks good to me :+1:

PR #1069, opened 2016-08-24
REL-2930: Fixed FT affiliate overriding

sraaphorst:
In previous versions of the PIT, when an institution for the PI is selected, the address is used for FT programs to determine the affiliate.

This can be overridden, in which case, the affiliate should be set to the chosen affiliate in the combobox in `PartnerView`; however, this code was broken, and even if a new affiliate was chosen, when a proposal was saved to XML (which did contain the proper affiliate code in the XML) and then reloaded, the affiliate combo box defaulted to the affiliate of the PI's institution.

There was an attempt to work around this by using a `previousPartnerAffiliation` with the FT program class, but the code to use this was broken in `PartnerView`, and the value was not persistent, so the information was lost when the proposal was saved and then later re-opened.

This fixes that by introducing a simple `overrideAffiliate` flag that does persist. It functions as follows:

1. If set to `true`, it assumes that the current affiliate has been overridden from the PI institution affiliate, in which case:
a. If the current affiliate is different from any new PI institution affiliate, the current affiliate is maintained as the override; or
b. If the current affiliate is the same as the new PI institution affiliate, the override is considered terminated.
Thus, by default, `overrideAffiliate` is set to `true` as this is the desired behaviour since the `PartnerView` initializer will properly maintain this, which would be extremely difficult to do from the `UpConverter`.

2. If set to `false`, it assumes that the current affiliate has not been manually overridden from the PI institution affiliate. Thus, changes to the PI institution affiliate are made to the current affiliate.

### DIFF
--- a/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/transform/UpConverter.scala
+++ b/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/transform/UpConverter.scala
@@ -128,7 +128,13 @@ case class LastStepConverter(semester: Semester) extends SemesterConverter {
  * This converter will upgrade to 2017A
  */
 case object SemesterConverter2016BTo2017A extends SemesterConverter {
-  override val transformers = Nil
+  // Set overrideAffiliate as true by default. It will automatically set to false if current FT affiliate is undefined
+  // or the same as the PI institution affiliate.
+  val overrideAffiliate:TransformFunction = {
+    case m @ <meta>{ns @ _*}</meta> if (m \ "@overrideAffiliate").isEmpty =>
+      StepResult("Affiliate override flag is missing", <meta overrideAffiliate="true">{ns}</meta>).successNel
+  }
+  override val transformers = List(overrideAffiliate)
 }
 
 /**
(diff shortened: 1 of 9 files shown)


sraaphorst:
Just noticed that my changes to `UpConverter` failed the spec... I'll fix them and commit.

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/transform/UpConverter.scala
```diff
@@ -128,7 +128,13 @@ case class LastStepConverter(semester: Semester) extends SemesterConverter {
  * This converter will upgrade to 2017A
  */
 case object SemesterConverter2016BTo2017A extends SemesterConverter {
-  override val transformers = Nil
+  // Set overrideAffiliate as true by default. It will automatically set to false if current FT affiliate is undefined
+  // or the same as the PI institution affiliate.
+  val overrideAffiliate:TransformFunction = {
+    case m @ <meta>{ns @ _*}</meta> if (m \ "@overrideAffiliate").isEmpty =>
+      StepResult("Affiliate override flag is missing", <meta overrideAffiliate="true">{ns}</meta>).successNel
```
cquiroz: It is probably overkill but should the `overrideAffiliate` be decided based on the PI's institution and affiliate on upconversion.
I think this maybe difficult to determine at the xml level, is it?

cquiroz:
This looks excellent, great work ironing this. It would be good to add test to verify that the upconversion is doing the correct thing for older proposals

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/transform/UpConverter.scala
sraaphorst: I get the impression that it would be quite difficult to determine without some serious manipulation, unfortunately.

sraaphorst:
Working on the `UpconverterSpec` changes now.

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/transform/UpConverter.scala
cquiroz: I agree, you'd need to parse the pi's institution and the affiliation. Defaulting to `true` is probably correct

cquiroz:
Looks good :+1:

PR #1070, opened 2016-08-25
REL-2907: Affiliate for FT can now be specified independently from country

sraaphorst:
In the past, the PIT used the country of location of an institution to do a lookup for the affiliate for FT programs.

This was problematic in several instances, because there are a small number of institutions where the country in which an institution is located should not dictate the affiliate. (Example: Gemini South is located in Chile, but the affiliate should be US.)

This PR introduces an optional `affiliate` XML tag for the `institutions.xml` file that allows an affiliate to be specified that, if given, is used instead of the `country` tag when doing affiliate lookups.

The appropriate institutions have been given `affiliate` tags in `institutions.xml`, along with a few typo fixes, as well as some cosmetic name changes as per the comments in the JIRA task. As well, `University of Hawaii` has been re-added as an affiliate as per Bryan's list.

A test case for CFT was also added to the `InstitutionsSpec` to make sure that the country of location is USA but the affiliate is Canada.

### DIFF
--- a/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/editor/Institutions.scala
+++ b/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/editor/Institutions.scala
@@ -13,6 +13,12 @@ object Institutions {
     res.sortBy(_.name)
   }
 
+  // Institutions whose affiliation is different from the one dictated by their location.
+  lazy val alternateAffiliates = (for {
+    i <- all
+    a <- i.affiliate
+  } yield (i.name, a)).toMap
+
   private def strList(n: Node, tag: String): List[String] =
     (n \ tag).map(_.text).toList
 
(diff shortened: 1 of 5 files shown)


bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/editor/Institutions.scala
```diff
@@ -74,6 +81,6 @@ object Contact {
 case class Contact(phone: List[String], email: List[String])
 
 object Institution {
-  val empty = Institution("", Nil, "", Contact.empty)
+  val empty = Institution("", Nil, "", None, Contact.empty)
 }
-case class Institution(name: String, addr: List[String], country: String, contact: Contact)
+case class Institution(name: String, addr: List[String], country: String, affiliate: Option[String], contact: Contact)
```
cquiroz: Could/Should affiliate be `Option[NgoPartner (or the correct type for an affiliate)]`?

cquiroz: Perhaps it should be `Option[FtPartner]`

sraaphorst:
Good call, @cquiroz - I made the change.
As we need to do a PIT test release today (against the real backends), if I don't hear from anyone before the travis checks are completed, I am going to merge.

PR #1071, opened 2016-08-26
REL-2888: Updated p1-monitor build file to include 2017A.

sraaphorst:
Subject says it all.

### DIFF
--- a/app/p1-monitor/build.sbt
+++ b/app/p1-monitor/build.sbt
@@ -18,9 +18,9 @@ ocsAppManifest := {
       common(ov, pv),
         dev(ov, pv),
         staging(ov, pv),
-        v2015B(ov, pv),
         v2016B(ov, pv),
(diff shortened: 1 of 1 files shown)


app/p1-monitor/build.sbt
```diff
@@ -18,9 +18,9 @@ ocsAppManifest := {
       common(ov, pv),
         dev(ov, pv),
         staging(ov, pv),
-        v2015B(ov, pv),
         v2016B(ov, pv),
-        v2016B_2(ov, pv)
+        v2016B_2(ov, pv),
+	v2017A(ov, pv)
```
arturog8m: just to be pedantic - why v2017A isn't aligned with the others?.... other than that.... 👍  ship it

sraaphorst: LOL probably emacs.
I should have noticed :-).

PR #1072, opened 2016-08-29
REL-2888: Add config file for p1-monitor or 2017A.

sraaphorst:
The 2017A release of the PIT needs a configuration file in order to build and function properly.
This is that configuration file.


cquiroz:
Looks good to me :+1:

PR #1073, opened 2016-08-29
REL-2907: Modified warning message regarding affiliates

sraaphorst:
Bryan suggested modifying the warning displayed when an affiliate for an FT program differs from the affiliate associated with the PI's institution.

This makes more sense as not every affiliate is a country.


cquiroz:
Looks good to me :+1:

PR #1077, opened 2016-09-05
REL-2893: GPI TLC wrong name in TLC APODIZER entry

cquiroz:
This PR updates GPI marking the `APOD_star` apodizer and `H_star` observing mode as obsolete and adds the `ND3` apodizer, which can be set on the engineering mode

Note that there are no existing programs using the obsolete options, so there is no need to convert existing data.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gpi/Gpi.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gpi/Gpi.java
@@ -44,7 +44,7 @@ import edu.gemini.spModel.type.*;
 import java.beans.PropertyDescriptor;
 import java.io.Serializable;
 import java.util.*;
-import java.util.stream.Stream;
+import java.util.function.IntFunction;
 
 import static edu.gemini.spModel.seqcomp.SeqConfigNames.INSTRUMENT_KEY;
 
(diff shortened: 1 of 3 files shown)


bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gpi/Gpi.java
```diff
@@ -320,6 +324,11 @@ public String logValue() {
             return _logValue;
         }
 
+        @Override
+        public boolean isObsolete() {
+            return false;
+        }
+
```
sraaphorst: This is not in the scope of this PR, but I think at some point we should get rid of all of these default implementations of the `ObsoletableSPType` interface `isObsolete` method that return `false` and just change `ObsoletableSpType` to have:

`default boolean isObsolete() { return false; }`

There are so many of them.
Maybe I'll do this if I have a few minutes.

cquiroz: That would be great indeed

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gpi/Gpi.java
```diff
@@ -792,6 +818,10 @@ public static Apodizer valueOf(String name, Apodizer value) {
             }
             return None.instance();
         }
+
+        public static Apodizer[] validValues() {
+            return Arrays.stream(values()).filter(a -> !a.isObsolete()).toArray(Apodizer[]::new);
+        }
```
sraaphorst: Just out of curiosity... why `Apodizer.validValues`, but `ObservingMode.nonObsoleteValues`?
They seem to be very similar so I was just wondering about the name difference.

sraaphorst:
This all looks great to me. 👍

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gpi/Gpi.java
cquiroz: No reason really, I'll adjust to make them consistent. `nonObsoleteValues` seems the clearer

PR #1079, opened 2016-09-05
REL-2921: Making BAGS and parallactic angle play nicely.

sraaphorst:
This task comprises two elements:

1. Making code for the `PositionAnglePanel` and `ParallacticAngleControls` run on the Swing EDT so that it plays nicely with BAGS, which runs on the Swing EDT. Formerly, there were issues where, IIRC, BAGS lookups were not triggered because of this discrepancy.

2. BAGS now triggers changes to instrument configuration, which causes the `PositionAnglePanel.init` method to be called to reinitialize the panel. This causes the `ParallacticAngleControls.init` to be called, which leads to a call to `ParallacticAngleControls.updateSchedulingBlock`. This causes the parallactic angle to be recalculated, which sends an event to the `PositionAnglePanel` to indicate that the parallactic angle has changed. The `PositionAnglePanel` previously naively assumed that this event came only when the `PosAngleConstraint` on the instrument was set to `PARALLACTIC_ANGLE`, which is not always the case. Now when the parallactic angle is recorded as having changed, we explicitly check before we do anything that we are in `PARALLACTIC_ANGLE` mode.

I'm not sure if (2) is the best way to handle this, but this code has become fragile from all of the changes made over time and I think that this is the simplest and best way to make it safe and keep it safe in the future.

Independently, it may be worth trying to minimize the calls to `PositionAnglePanel.init` and `ParallacticAngleControls.init` as this situation is hardly ideal, but I believe that this is probably best left for a later time when the FR this addresses is at least working and is somewhat outside of the scope of this PR.

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/parallacticangle/ParallacticAngleControls.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/parallacticangle/ParallacticAngleControls.scala
@@ -74,7 +74,7 @@ class ParallacticAngleControls(isPaUi: Boolean) extends GridBagPanel with Publis
       icon                   = Resources.getIcon("eclipse/menu-trimmed.gif")
       margin                 = new Insets(-1, -10, -1, -5)
 
-      def rebuild(): Unit = {
+      def rebuild(): Unit = Swing.onEDT {
         contents.clear()
 
         // menu items that don't depend on the context
@@ -152,7 +152,7 @@ class ParallacticAngleControls(isPaUi: Boolean) extends GridBagPanel with Publis
    * Initialize the UI and set the instrument editor to allow for the parallactic angle updates.
    * The `Runnable` is a callback that will be invoked on the EDT after the target is updated.
    */
-  def init(e: OtItemEditor[_, _], s: Option[Site], f: Format, c: Runnable): Unit = {
+  def init(e: OtItemEditor[_, _], s: Option[Site], f: Format, c: Runnable): Unit = Swing.onEDT {
     editor    = Some(e)
     site      = s
     formatter = Some(f)
@@ -244,7 +244,7 @@ class ParallacticAngleControls(isPaUi: Boolean) extends GridBagPanel with Publis
    * Triggered when the date time button is clicked. Shows the ParallacticAngleDialog to allow the user to
    * explicitly set a date and duration for the parallactic angle calculation.
    */
-  private def displayParallacticAngleDialog(): Unit =
+  private def displayParallacticAngleDialog(): Unit = Swing.onEDT {
     for {
       e <- editor
       o <- editor.map(_.getContextObservation)
@@ -259,6 +259,7 @@ class ParallacticAngleControls(isPaUi: Boolean) extends GridBagPanel with Publis
       dialog.visible = true
       updateSchedulingBlock(dialog.schedulingBlock)
     }
+  }
 
 
   /**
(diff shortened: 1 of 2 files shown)


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/parallacticangle/ParallacticAngleControls.scala
```diff
@@ -259,22 +259,23 @@ class ParallacticAngleControls(isPaUi: Boolean) extends GridBagPanel with Publis
...
+  def positionAngleChanged(positionAngleText: String): Unit = Swing.onEDT {
     // We only do this if the parallactic angle can be calculated, and is different from the PA.
     for {
       e     <- editor
       angle <- parallacticAngle
       fmt   <- formatter
     } {
-      val explicitlySet = !fmt.format(ParallacticAngleControls.angleToDegrees(angle)).equals(positionAngleText) &&
-                          !fmt.format(ParallacticAngleControls.angleToDegrees(angle + Angle.fromDegrees(180))).equals(positionAngleText)
+      val explicitlySet = !fmt.format(angle.toDegrees).equals(positionAngleText) &&
```
cquiroz: Are we comparing the angles as text? Could we compare them as `Angle` instead?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/parallacticangle/PositionAnglePanel.scala
```diff
@@ -141,16 +142,19 @@ class PositionAnglePanel[I <: SPInstObsComp with PosAngleConstraintAware,
       }
       peer.add(positionAngleFeedbackPanel.peer, positionAngleFeedbackPanel.cardId)
 
-      def showParallacticAngleControls(): Unit =
+      private def showParallacticAngleControls(): Unit =
```
cquiroz: These calls to `cardLayout.show` should probably be on the EDT too

cquiroz:
Code-wise this PR seems fine, I'm not totally sure if I understand the core problem though

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/parallacticangle/ParallacticAngleControls.scala
sraaphorst: =grins=
I think we had this exact same discussion before when I made a PR specifically regarding this.
There were several reasons that we did this as text: one because the input was coming from a text field and we wanted to indicate any differences between the angle and text field and not just numerical differences, and the other because the actual parallactic angle is at an extremely high precision, so we use the formatter to chop it to two decimal places.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/parallacticangle/PositionAnglePanel.scala
sraaphorst: Good point. I'll do that.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/parallacticangle/ParallacticAngleControls.scala
cquiroz: 👍

sraaphorst:
I do think that this whole issue needs to be looked at with a bit more detail and some rewriting, but this solves the FR issue, which is causing grief for the science staff, so I'm going to merge, track back to the release branch, and make a test release today.

PR #1080, opened 2016-09-05
REL-2943: ObsoletableSpType cleanup and other warnings

sraaphorst:
As per my discussion with @cquiroz earlier on https://github.com/gemini-hlsw/ocs/pull/1077, I decided tonight to go ahead and eliminate the many redundant implementations of `ObsoletableSpType.isObsolete` by using Java implementation defaults, which allowed to to remove many unnecessary boolean variables from enums.

I also cleaned up a number of things along the way, trying to eliminate many warnings about `static` being used with `enum` and `private` being used with enum constructors. Furthermore, I eliminated a bunch of redundant simple maps, using lambda expressions instead, and in several places, moved to a stream based solution instead of using loops. 

I also cleaned up what seemed like some ridiculous uses of `Option`.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/calunit/CalUnitParams.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/calunit/CalUnitParams.java
@@ -1,33 +1,26 @@
-// Copyright 1997 Association for Universities for Research in Astronomy, Inc.,
-// Observatory Control System, Gemini Telescopes Project.
-// See the file LICENSE for complete details.
-//
-// $Id: CalUnitParams.java 38751 2011-11-16 19:37:18Z swalker $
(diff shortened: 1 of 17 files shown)


cquiroz:
This looks great thanks. My only worry is with some of the deleted fields. We have been hit a few times when we delete fields marked as unused by idea but that the spdb uses anyway via reflection.

Could you try starting the applications (spdb and itc I think) check that they run and in particular ensure that the obslog still runs?

Otherwise 👍

sraaphorst:
Tested with spdb, itc, and the obslog, and everything seems to be working as expected.

I came across many unused methods, but I left those in as I knew we had had issues in the past with reflection breaking as a result. I think almost all the fields I deleted were private.

I rebased and then will merge once Travis finishes.

cquiroz:
sounds good 👍

PR #1082, opened 2016-09-07
REL-2944: Undo changes to classes that break serializability with 2016B production ODBs

sraaphorst:
I had updated the `ParallacticAngleSupport` interface and `ParallacticAngleSupportInst` class to take advantage of Java interface `default`s and to use the new target model `Angle` class, but this, of course, broke serializability with the 2016B production ODB.

Thus, for the patch release for 2016B, we will revert back to the old `Angle` class and implementation of the aforementioned. This solves the problem and allows production ODB programs to be properly serialized / deserialized.

For 2017A, we can safely go ahead with the new and improved implementations, which I have left intact in the develop branch.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -4,7 +4,7 @@ name := "ocs"
 
 organization in Global := "edu.gemini.ocs"
 
-ocsVersion in ThisBuild := OcsVersion("2016B", false, 2, 1, 1)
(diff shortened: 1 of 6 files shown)


cquiroz:
Looks good to me :+1:

tpolecat:
👍  it will be nice when we don't have this issue to worry about

PR #1083, opened 2016-09-08
REL-2945 add new result handler

tpolecat:
This adds a handler for a previously-unknown HORIZONS header line that gives the record number but no common name, followed by the primary designation in prens. The examples given (`2009 MS9`, `2013 LU28`, and `2015 DB216`) now resolve properly.

/cc @andrewwstephens

### DIFF
--- a/bundle/edu.gemini.horizons.api/src/main/scala/edu/gemini/horizons/server/backend/HS2.scala
+++ b/bundle/edu.gemini.horizons.api/src/main/scala/edu/gemini/horizons/server/backend/HS2.scala
@@ -280,12 +280,18 @@ object HorizonsService2 {
               List(Row(HorizonsDesignation.AsteroidNewStyle(m.group(1)) : HorizonsDesignation.Asteroid, m.group(1)))
             } \/> "Could not match '(2016 GB222)' header pattern."
 
+          // Single result with form: JPL/HORIZONS        418993 (2009 MS9)            2016-Sep-07 18:23:54
+          lazy val case4 =
+            """  +\d+\s+\((.+?)\)  """.r.findFirstMatchIn(header).map { m =>
+              List(Row(HorizonsDesignation.AsteroidNewStyle(m.group(1)) : HorizonsDesignation.Asteroid, m.group(1)))
+            } \/> "Could not match '418993 (2009 MS9)' header pattern."
 
           // First one that works!
           case0 orElse
           case1 orElse
           case2 orElse
-          case3 orElse "Could not parse the header line as an asteroid".left
+          case3 orElse
+          case4 orElse "Could not parse the header line as an asteroid".left
 
 
         case Search.MajorBody(_) =>
@@ -358,5 +364,3 @@ object HorizonsService2 {
     }
 
 }
-
-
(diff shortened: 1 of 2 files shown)


tpolecat:
I have been out of the loop on this patch release so I'm not sure which branch to merge against; can someone advise? I will be traveling for the next week but will be online periodically.

bundle/edu.gemini.horizons.api/src/main/scala/edu/gemini/horizons/server/backend/HS2.scala
```diff
@@ -280,12 +280,18 @@ object HorizonsService2 {
               List(Row(HorizonsDesignation.AsteroidNewStyle(m.group(1)) : HorizonsDesignation.Asteroid, m.group(1)))
             } \/> "Could not match '(2016 GB222)' header pattern."
 
+          // Single result with form: JPL/HORIZONS        418993 (2009 MS9)            2016-Sep-07 18:23:54
+          lazy val case4 =
+            """  +\d+\s+\((.+?)\)  """.r.findFirstMatchIn(header).map { m =>
```
cquiroz: Should the whitespace at the beginning and end also be done using `\s`?

cquiroz:
Looks good to me 👍 
If it is needed for the patch you should also make a PR against `release-2016B`

arturog8m:
We aren't planning on a patch release to the ODB or its services - so
unless this is an emergency that can't wait until December, we shouldn't
merge with release-2016B.

bundle/edu.gemini.horizons.api/src/main/scala/edu/gemini/horizons/server/backend/HS2.scala
sraaphorst: Could someone explain the `\/>` operator? Does it simply create a right `\/[A,B]`?

cquiroz: it takes the `Option` produced by `findFirstMatchIn` and if it is `Some(a)` it gives you a `\/.right(a)` and if `None` it gives you a `\/.left` with the error message

sraaphorst: Excellent. Thanks @cquiroz for the very clear explanation!

sraaphorst:
Also looks good to me, although I am also curious about Carlos' comment about the regexes. 👍

bundle/edu.gemini.horizons.api/src/main/scala/edu/gemini/horizons/server/backend/HS2.scala
tpolecat: @cquiroz It's explicitly looking for two spaces because there is leading and trailing text in the header line with space padding (the text we're looking for is centered). So it could really be two or more but `  ` is easier to read/write than `\s{2,}`.

cquiroz: Interesting format 😄 very FITS-ish

tpolecat:
@arturog8m the package says backend for historical (i.e., bad) reasons but it runs locally on the OT now.

tpolecat:
I'll rebase and move this to `release-2016B` when I get to Toronto. The airplane wifi is crapping out.

arturog8m:
@tpolecat thanks, ok - if we can release this as a pure OT thing, I'm all for it. @andrewwstephens could you include this in the testing you are doing ?

andrewwstephens:
No problem.

PR #1084, opened 2016-09-08
REL-2924: Make PA text field uneditable in parallactic angle mode

sraaphorst:
Since the functionality to override the parallactic angle is currently broken and may be somewhat complicated to fix, for the patch release, we are going to leave this functionality out.

Bryan sensibly suggested that we thus make the PA text field uneditable in parallactic angle mode to indicate this. This PR simply does that.


cquiroz:
Looks good to me :+1:

PR #1085, opened 2016-09-08
REL-2922B: Set the minimum number of fractional digits to 2.

sraaphorst:
Bryan requested that the minimum number of fractional digits for the PA text field be set to two.


cquiroz:
Looks good to me :+1:

PR #1088, opened 2016-09-08
REL-2922B: Fix change to revert to old target model angle

sraaphorst:
I missed one aspect when reverting to the old target model, which was, bizarrely, compiling, but causing an issue with regards to running the formatter.

This fixes that and again, will only be in the patch release branch.


cquiroz:
Looks good to me :+1:
I guess this is UI code right? I mean there are no tests that could catch things like this?

sraaphorst:
Yeah, UI code, so I don't think there's much we can do with regards to test cases, unfortunately.

PR #1089, opened 2016-09-13
REL-2655: LchQueryFunctor overhaul

sraaphorst:
This PR is a major overhaul of `LchQueryFunctor` and friends to avoid the earlier hideousness that was about to get much more complicated through the addition of extra parameters.

Mainly:

- No more passing around a ridiculous number of `String`s to constructors: matching is now done by checking servlet parameters to `LchQueryParam`, which contains a parameter `name` and a `ValueMatcher` that extracts a value from either an `ISPProgram` or an `ISPObservation` and then compares it against a regular expression.

- Real regular expressions are now used instead of the horrible `StringUtil.match` method.

- Almost all uses of `null` have been eliminated.

Test cases pass, but more testing will need to be done on the test servers, so this is essentially a preview PR to elicit feedback. There will still also need to be modifications to LCH itself in order to accommodate the new parameters and to complete this PR, with corresponding code added to `LchQueryFunctor`.


bundle/edu.gemini.lchquery.servlet/src/main/java/edu/gemini/lchquery/servlet/LchQueryFunctor.java
```diff
@@ -1,549 +0,0 @@
...
-    private String _observationName;
-    private String _observationStatus;
-    private String _observationInstrument;
-    private String _observationAo;
-    private String _observationClass;
-
-    public LchQueryFunctor(QueryType queryType, String programSemester, String programTitle, String programReference,
-                           String programActive, String programCompleted, String programNotifyPi, String programRollover,
-                           String observationTooStatus, String observationName, String observationStatus, String observationInstrument,
-                           String observationAo, String observationClass) {
```
sraaphorst: One of the main purposes of this PR was to get rid of horrible things like this...

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
```diff
@@ -0,0 +1,230 @@
...
+import edu.gemini.spModel.target.env.TargetEnvironment
+import edu.gemini.spModel.too.Too
+
+import scala.collection.JavaConverters._
+import scala.util.{Failure, Success, Try}
+
+
+class LchQueryFunctor(queryType: LchQueryFunctor.QueryType,
+                      programParams: List[(LchQueryParam[ISPProgram], String)],
+                      observationParams: List[(LchQueryParam[ISPObservation], String)])
```
sraaphorst: ...and instead replace them with an easily extendible set of generic query parameters like this.

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryParam.scala
```diff
@@ -0,0 +1,233 @@
...
+        prog.semester
+    }
+  )
+
+  private[servlet] val ProgramTitleParam = LchQueryParam("programTitle",
+    new StringValueMatcher[ISPProgram] {
+      override protected def extractor(prog: ISPProgram): Option[String] =
+        prog.toSPProg.map(_.getTitle)
+    }
+  )
```
sraaphorst: Now parameters, instead of being mere `String`s that are manually extracted, are self-contained `LchQueryParams` with their `name`, and the `ValueMatcher` that extracts the value from the appropriate data object for comparison with the supplied pseudo-RE as per the ODB Browser page.

bundle/edu.gemini.lchquery.servlet/src/main/java/edu/gemini/lchquery/servlet/LchQueryServlet.java
```diff
@@ -1,201 +0,0 @@
...
-            return;
-        }
-
-        String programSemester = request.getParameter(LchQueryService.PARAMETER_PROGRAM_SEMESTER);
-        String programTitle = request.getParameter(LchQueryService.PARAMETER_PROGRAM_TITLE);
-        String programReference = request.getParameter(LchQueryService.PARAMETER_PROGRAM_REFERENCE);
-        String programActive = request.getParameter(LchQueryService.PARAMETER_PROGRAM_ACTIVE);
-        String programCompleted = request.getParameter(LchQueryService.PARAMETER_PROGRAM_COMPLETED);
-        String programNotifyPi = request.getParameter(LchQueryService.PARAMETER_PROGRAM_NOTIFY_PI);
-        String programRollover = request.getParameter(LchQueryService.PARAMETER_PROGRAM_ROLLOVER);
```
sraaphorst: Formerly, to extract parameters in `LchQueryServlet`, we had to deal with nonsense like this.

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryServlet.scala
```diff
@@ -0,0 +1,147 @@
...
+final case class LchQueryServlet(odb: IDBDatabaseService, user: Set[Principal]) extends HttpServlet {
+  override def doGet(request: HttpServletRequest, response: HttpServletResponse): Unit = {
+    import LchQueryServlet.QueryResultToXml
+
+    // Get the params from the HTTP request that exist and correspond to LchQueryParams.
+    def extractParams[A <: LchQueryParam[_]](lst: List[A]): List[(A,String)] =
+      for {
+        p <- lst
+        v <- Option(request.getParameter(p.name))
+      } yield (p,v)
```
sraaphorst: Now, this is all that is needed to extract all parameters.

cquiroz:
I'll check this in more detail later but it looks like a great start. Any reason the servlet produces xml instead of e.g. json?

sraaphorst:
Thanks! I appreciate the encouragement.

I'm not sure if there's a reason why we use XML instead of JSON: perhaps simply because the code was old (and I just stuck with what was being done), but at this point it is used by LTTS, FT, and GSA, so I don't think it's possible to just go changing it without major consequences.

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
```diff
@@ -0,0 +1,230 @@
...
+      def targetType: Option[String] = {
+        if (env.isBasePosition(target)) Some("Base")
+        else if (env.isUserPosition(target)) Some("User")
+        else if (env.isGuidePosition(target)) {
+          for {
+            gg  <- env.getGroups.asScala
+            gpt <- gg.getAll.asScala
+            if gpt.containsTarget(target)
+          } yield gpt.getGuider.getKey
+        }.headOption
```
sraaphorst: There's probably a better way to write this as it isn't using a `Stream`, but I'm not sure.
Suggestions are welcome.

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
```diff
@@ -0,0 +1,230 @@
...
+                    } foreach { tgts =>
+                      setTargetsNode(new TargetsNode() {
+                        tgts.map(tgt => makeTargetNode(tgt, env)).foreach(getTargets.add)
+                      })
+                    }
+                  }
+
+                }
+              }
+            } foreach getObservations.add
```
sraaphorst: This is probably confusing and could be simplified, perhaps with some more methods.

swalker2m:
If I recall the existing `StringUtil` matching was kind of quirky.  It would be a good idea, I think, to do some property testing that demonstrates that the new regex matching provides equivalent results.

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
```diff
@@ -0,0 +1,230 @@
...
+import scala.util.{Failure, Success, Try}
+
+
+class LchQueryFunctor(queryType: LchQueryFunctor.QueryType,
+                      programParams: List[(LchQueryParam[ISPProgram], String)],
+                      observationParams: List[(LchQueryParam[ISPObservation], String)])
+  extends DBAbstractQueryFunctor with IDBParallelFunctor {
+
+  // The current QueryResult. Must be mutable.
+  var queryResult: QueryResult = new QueryResult() {
```
cquiroz: Is this a requirement of being  an spdb functor?

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
```diff
@@ -0,0 +1,230 @@
...
+    }
+    def observationMatches(obs: ISPObservation): Boolean = observationParams.forall {
+      case (LchQueryParam(_, valueMatcher), paramValue) => valueMatcher.matches(paramValue, obs)
+    }
+
+    Try {
+      // See if an ISPProgram matches the query specifications.
+      if (programMatches) {
+        val matchingObs = prog.getAllObservations.asScala.toList.filter(observationMatches)
+        if (matchingObs.nonEmpty) {
```
swalker2m: What happens if `programMatches` but there are no observations?  That's indistinguishable from no matching observations and results in skipping the program.  Is that intended?

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
```diff
@@ -0,0 +1,230 @@
...
+          addProgram(prog, if (queryType == LchQueryFunctor.QueryType.ProgramQuery) Nil else matchingObs)
+        }
+      }
+    } match {
+      case Success(_) =>
+      case Failure(t) =>
+        LchQueryFunctor.Log.log(Level.SEVERE, "problem running LchQueryFunctor", t)
+        throw new RuntimeException(t)
+    }
```
swalker2m: We do the exception wrapping in a number of different ways in the codebase, and I've personally contributed to doing it in every different way possible.  These days though I've settled on using `\/.fromTryCatchNonFatal` since we make extensive use of `\/` anyway.  Does everyone agree it would be good to handle this in the same way everywhere and that `\/` should be that way?  An argument against is that eventually presumably we'll replace scalaz `\/` with cats `Xor` or something.  Still that will be a more or less mechanical switch that we have to do anyway.

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
```diff
@@ -0,0 +1,230 @@
...
+            gg  <- env.getGroups.asScala
+            gpt <- gg.getAll.asScala
+            if gpt.containsTarget(target)
+          } yield gpt.getGuider.getKey
+        }.headOption
+        else None
+      }
+
+      if (target.isNonSidereal) {
+        Some(new NonSidereal() {
```
cquiroz: What are those `Sidereal`/`NoSidereal` classes, other target classes?

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
cquiroz: I agree on using `\/.fromTryCatchNonFatal` in general, though sometimes there are good reasons to use `Try` (e.g. talking to some third-party lib)

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
sraaphorst: Yes. At least, that is how things worked in the previous code. I'll indicate the relevant section with an additional comment and flag you in it. It seems, though, that if there are no observations / no matching observations, the program is simply skipped.

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
```diff
@@ -0,0 +1,230 @@
...
+  }
+}
+
+object LchQueryFunctor {
+  private[LchQueryFunctor] val Log = Logger.getLogger(classOf[LchQueryFunctor].getName)
+  private[LchQueryFunctor] val MsPerSecond = 1000
+  private[LchQueryFunctor] val MsPerMinute = MsPerSecond * 60
+  private[LchQueryFunctor] val MsPerHour   = MsPerMinute * 60
+
+  private[LchQueryFunctor] val dateFormat = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss z") {
```
cquiroz: Should we use the new `java.time` api

bundle/edu.gemini.lchquery.servlet/src/main/java/edu/gemini/lchquery/servlet/LchQueryFunctor.java
```diff
@@ -1,549 +0,0 @@
...
-                        obsList.add(obs);
-                    }
-                }
-                if (obsList.size() != 0) {
-                    if (_queryType == QueryType.PROGRAMS) {
-                        addProgram(prog, new ArrayList<ISPObservation>());
-                    } else {
-                        addProgram(prog, obsList);
-                    }
-                }
```
sraaphorst: @swalker2m You can see from the preceding nested `if` that if the `obsList` (the list of `ISPObservation`s that match the query parameters is empty, then `addProgram` is never called, and the `ISPProgram` is simply ignored.

Unless I am missing something, this should correspond with what I did in the rewritten code.

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
```diff
@@ -0,0 +1,230 @@
...
+import edu.gemini.spModel.too.Too
+
+import scala.collection.JavaConverters._
+import scala.util.{Failure, Success, Try}
+
+
+class LchQueryFunctor(queryType: LchQueryFunctor.QueryType,
+                      programParams: List[(LchQueryParam[ISPProgram], String)],
+                      observationParams: List[(LchQueryParam[ISPObservation], String)])
+  extends DBAbstractQueryFunctor with IDBParallelFunctor {
```
swalker2m: `IDBParallelFunctor` is defunct and should be `@deprecated`.  It comes from the days when we were planning on distributing the ODB across multiple processes on multiple servers.  You can safely delete the `mergeResults` method below and just extend `IDBFunctor`.  Nobody is calling `mergeResults` in this code base.

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
sraaphorst: I think it's a consequence of being an `IDBParallelFunctor`, which requires the implementation of the `mergeResults` method, which has to recreate the object.

Honestly, I'm not even convinced that we need this, because based on what I saw in the old code, I am thinking that this method would fail with a `NullPointerException` so it was probably never called: when you create new instance of `QueryResult`, if you don't also set it to have an instance of `ProgramsNode`, this is set to `null` by default.

In the old implementation, a new `QueryResult` was created but no new `ProgramsNode` was set to it, so the call `queryResult.getProgramsNode.getPrograms` would have thrown a `NullPointerException` and failed.

Thus, maybe we can remove this and the implementation of `IDBParallelFunctor` and instead change `queryResult` to an immutable instance of `QueryResult`, which is obviously mutable.

@swalker2m - do you have any thoughts on this?

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
```diff
@@ -0,0 +1,230 @@
...
+          } yield gpt.getGuider.getKey
+        }.headOption
+        else None
+      }
+
+      if (target.isNonSidereal) {
+        Some(new NonSidereal() {
+          setName(target.getName)
+          setType(targetType.orNull)
+        })
```
swalker2m: It took me a minute to figure out but I guess we're hamstrung by the the ODB browser library supplied and used by the LCH?  That is `NonSidereal` and `Sidereal` comes from that and they require a `String` target type, etc.

cquiroz: Are we really that constrained? We can deploy an updated LCH if needed. Perhaps as part of @sraaphorst refactoring, we could replace those classes with the new ones

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
```diff
@@ -0,0 +1,230 @@
...
+            setPeriod(formatPeriod.orNull)
+            setTimes(formatTimes.orNull)
+          })
+        }
+      }
+    }
+
+    queryResult.getProgramsNode.getPrograms.add {
+      val spProg = prog.toSPProg.get
+      new Program() {
```
cquiroz: I guess this is also an LCH class?

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
```diff
@@ -0,0 +1,230 @@
...
+        setActive(spProg.getActive.displayValue)
+        setCompleted(spProg.isCompleted.toYesNo.displayValue)
+        Option(prog.getProgramID).foreach(id => setReference(id.stringValue()))
+        setSemester(prog.semester.orNull)
+        setTitle(spProg.getTitle)
+        setContactScientistEmail(spProg.getContactPerson)
+        setNgoEmail(spProg.getNGOContactEmail)
+        setNotifyPi(spProg.getNotifyPi.displayValue)
+        setPiEmail(Option(spProg.getPIInfo).map(_.getEmail).orNull)
+        setRollover(spProg.getRolloverStatus.toYesNo.displayValue)
```
swalker2m: Still, it's a bit sad that we throw away types for no apparent gain.  I guess it was done so that there is no dependency from LCH on the program model?  A long term goal would be to incorporate LCH into the same architecture as everything else and make it use the same program model.  But for now, I guess this is what it is.

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryParam.scala
```diff
@@ -0,0 +1,233 @@
...
+
+
+sealed trait ValueMatcher[A] {
+  def matches(expression: String, x: A): Boolean
+}
+
+object ValueMatcher {
+  private[servlet] implicit class ToRegex(val expression: String) extends AnyVal {
+    def toRegex: Regex = ("^(?i)" +
+      (expression.contains("|") ? s"($expression)" | expression).
```
cquiroz: 👍

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
```diff
@@ -0,0 +1,230 @@
...
+  def getResult: QueryResult = queryResult
+
+  override def mergeResults(functorCollection: util.Collection[IDBFunctor]): Unit = {
+    queryResult = new QueryResult() {
+      setProgramsNode(new ProgramsNode())
+    }
+    functorCollection.asScala.map(_.asInstanceOf[LchQueryFunctor]).foreach { f =>
+      queryResult.getProgramsNode.getPrograms.addAll(f.getResult.getProgramsNode.getPrograms)
+    }
+  }
```
swalker2m: As mentioned above, this method can go away since `IDBParallelFunctor` isn't really used.

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryParam.scala
```diff
@@ -0,0 +1,233 @@
...
+
+
+  private[servlet] implicit class ISPProgramExtractors(val prog: ISPProgram) extends AnyVal {
+    def toSPProg: Option[SPProgram] =
+      Option(prog).map(_.getDataObject.asInstanceOf[SPProgram])
+
+    def semester: Option[String] = for {
+      s <- Option(prog.getProgramID).map(_.stringValue)
+      i = s.indexOf('-')
+      if i != -1 && s.length > i+6 && s.charAt(i+1) == '2'
```
cquiroz: This looks a bit risky, could/should we split the program id with a regex?

sraaphorst:
@swalker2m Yes, absolutely. I have already done some testing in this regard, both independently and through the test cases, and found an issue with `+` and case insensitivity, both of which I have fixed. I am going to, however, install this on `gsodbtest2` today and do some serious testing against `gsodb` results to ensure that everything works as should with or clauses and a good number of queries.

Until that point, I was just hoping to give everyone some time to provide some general comments on the code quality, and any suggestions for improvement.

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
sraaphorst: Good. I'll make these changes then and make `queryResult` an immutable variable, and ensure everything still compiles and runs on `gsodbtest2` then, as we would expect.

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
cquiroz: Should LCH depend on the program model? I don't see issues on importing `spModel.core` into LCH. we even do that in ITAC

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryServlet.scala
```diff
@@ -0,0 +1,147 @@
+package edu.gemini.lchquery.servlet
+
+import java.io.{BufferedOutputStream, ByteArrayOutputStream, IOException, PrintWriter}
+import java.rmi.RemoteException
```
cquiroz: ??? I thought RMI was deprecated in the odb

bundle/edu.gemini.lchquery.servlet/src/test/java/edu/gemini/lchquery/servlet/LchQueryFunctorTest.java
```diff
@@ -131,127 +114,45 @@ protected void addAltair(ISPProgram prog, AltairParams.Mode mode) throws SPUnkno
...
-        LchQueryFunctor functor = new LchQueryFunctor(LchQueryFunctor.QueryType.OBSERVATIONS,
-                programSemester, programTitle, programReference,
-                programActive, programCompleted, programNotifyPi, programRollover,
-                observationTooStatus, observationName, observationStatus, observationInstrument,
-                observationAo, observationClass);
-        functor.execute(null, prog, Collections.<Principal>emptySet());
-        QueryResult result = functor.getResult();
-        List<Program> programs = result.getProgramsNode().getPrograms();
+    private boolean matches(final ISPProgram prog, final int expectedPrograms, final int expectedObservations) {
+        final LchQueryFunctor functor = new LchQueryFunctor(LchQueryFunctorTestFuncs.obsQuery(),
```
sraaphorst: Also, there HAS to be a better way to do this than `LchQueryFunctorTestFuncs.obsQuery()`, which just returns `LchQueryFunctory.QueryType.ObservationQuery`, but I cannot for the life of me figure out how to access `LchQueryFunctory.QueryType.ObservationQuery` from Java.

Help?

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryServlet.scala
```diff
@@ -0,0 +1,147 @@
...
+  val Log = Logger.getLogger(LchQueryServlet.getClass.getName)
+
+  sealed abstract class HttpResponseCode(val code: Int)
+  object HttpResponseCodes {
+    case object AllOK extends HttpResponseCode(200)
+    case object InvalidRequest extends HttpResponseCode(400)
+    case object ServerError extends HttpResponseCode(500)
+  }
+
+  def apply(odb: IDBDatabaseService, user: java.util.Set[Principal]): LchQueryServlet =
```
cquiroz: any reason to use the java collections?

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
sraaphorst: They are from a library produced by LCH: auto-generated classes, I believe, that use XML bindings.
It is probably going to be something of a pain to modify.

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
sraaphorst: I'll look into it, if it seems appropriate. (I know pretty much nothing about it, but may as well learn about it at some point.)

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
cquiroz: So, the LCH project generates these classes out of a XML schema generation via jaxb so the functor converts from existing classes to the xml model. Probably it is not trivial to change, specially if some clients rely on the schema as well

bundle/edu.gemini.lchquery.servlet/src/main/java/edu/gemini/lchquery/servlet/LchQueryFunctor.java
swalker2m: Yes, it seems like it would have the same behavior. 👍

This may have been the cause of a recent issue though, if I remember correctly.  This functor is used to gather up emails or other program-wide values(?).  Since programs often start life without any observations (assuming the template has no baseline calibrations) they can be skipped at the beginning of the semester before template are instantiated.  On the other hand, I'm not sure what the ramifications of fixing it would be on LCH.

swalker2m: Here's the corresponding [JIRA issue](http://jira.gemini.edu:8080/browse/REL-2929).

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
sraaphorst: Yes... all the classes used internally in `LchQueryFunctor`, I believe, are from an LCH library.

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryParam.scala
sraaphorst: Alright, switched to a regex here as suggested. I think this is a good idea.
All test cases pass.

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
swalker2m: I was guessing that it is tricky because we use a library supplied by LCH when building the OCS codebase.  `SPProgram` and friends aren't in `spModel.core` but even if they were it sets up a cross dependency between the two projects.

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryServlet.scala
sraaphorst: I guess we don't need this anymore. The bundle `Activator` was previously using a constructor with this signature, but we can just `.toSet` it with `scala.collection.JavaConversions` and be done with it. Removing it and updating the `Activtor`.

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
swalker2m: One reason to switch is that `SimpleDateFormat` is not thread safe.  I don't know if that's an issue here but we might as well use the safe equivalent in `java.time.format.DateTimeFormatter` and not worry about it.

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
sraaphorst: Would it be helpful if I made a small JIRA task and PR to remove this interface?
Or at least mark it deprecated here?

cquiroz: yes, let's delete deprecated stuff 👍

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryParam.scala
```diff
@@ -0,0 +1,232 @@
...
+  }
+
+
+  private[servlet] implicit class ISPProgramExtractors(val prog: ISPProgram) extends AnyVal {
+    def toSPProg: Option[SPProgram] =
+      Option(prog).map(_.getDataObject.asInstanceOf[SPProgram])
+
+    def semester: Option[String] = {
+      val semesterRe = "(?i)\\-(2\\d{3}[AB])".r
+      Option(prog.getProgramID).flatMap(id => semesterRe.findFirstMatchIn(id.stringValue()).map(_.group(1)))
```
swalker2m: Let's take advantage of `edu.gemini.spModel.core.ProgramId.parse(s: String)` so that you don't have to define another regular expression.

swalker2m: Though one thing that has to be evaluated is whether CAL and ENG semesters should be considered.  `ProgramId` will get a semester out of a CAL or ENG program id whereas this regex won't.  In which case, I suppose you could match on `ProgramId.Science` if that is what is intended.

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
sraaphorst: Okay. I have attempted to migrate to `java.time`. If someone more familiar than me with this API (first time using it) has a moment and can check the `LchQueryFunctor` to see if my usage is correct, that would be excellent.

swalker2m:
This is very much improved vs. the original version with all the `String`s.  Thanks for cleaning this up.  If there are additional changes/parameters in the future (util the whole ODB is replaced with a relational database anyway), they should be much easier.

sraaphorst:
I've been playing around with the ODB Browser (production on `gsodb` and develop + the changes in this PR on `gsodbtest2`) and I _seem_ to be getting comparable results, although it is difficult to do direct `diff` comparisons of large data sets due to the import issue that I came across and @swalker2m looked into regarding nonsidereal targets.

I am going to play around some more with the various parameters. The regex conversion looks to be working properly and I'm feeling reasonably confident that this is working as it should.

sraaphorst:
As the LCH changes can be started independently, I'll begin working on those and hopefully we can reach a point in the interim where we feel comfortable merging this.

swalker2m:
Are there unit test cases for the regex patterns?  It would be good to see them side-by-side -- regex vs. old string util and validate that they get the same results.

sraaphorst:
@swalker2m - there are test cases that do search by regex and expect a (exactly one) match from the `LchQueryFunctor` on a wide variety of fields.

No reason why I couldn't add a test case to compare `.toRegex` matching with `StringUtil` matching, though. Shouldn't be hard.

sraaphorst:
Note that I added a spec file for testing RegEx conversions as per @swalker2m's recommendation.
This could probably be expanded a bit in the future, but I am pressed for time.

I leave tomorrow for three weeks and would like to at least merge this into develop so that we can have an ODB Query browser on the test ODB servers, so if someone could review this and approve, that would be great.

cquiroz:
Note that  travis failed on a test on one of the lch bundles

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryParam.scala
```diff
@@ -94,7 +94,14 @@ object LchQueryParam {
         Option(inv).map(i => s"${i.getFirst} ${i.getLast}").filterNot(_.trim.isEmpty)
 
       val spProg = toSPProg
-      (investigatorName(spProg.getGsaPhase1Data.getPi) :: spProg.getGsaPhase1Data.getCois.asScala.map(investigatorName).toList).
+      val piName = for {
+        f <- Option(spProg.getPIFirstName)
+        l <- Option(spProg.getPILastName)
+        n = s"$f $l".trim
+        if !n.isEmpty
```
cquiroz: I guess you can do `if n.nonEmpty` but regardless I think n is never empty as you have an space in the expression

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryParam.scala
```diff
@@ -94,7 +94,14 @@ object LchQueryParam {
...
+      } yield n
```
swalker2m: Maybe it should work if either name is defined (or both of course)?

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryParam.scala
swalker2m: The `.trim` should reduce it to an empty `String` if both `f` and `l` are empty or spaces.  If one or the other is `null` though, the entire thing is skipped.  I'm not sure if that's intended.

sraaphorst: Alright, I changed the PI name extraction to use `spProg.getPIInfo.getFirstName` and `getLastName`, which never return `null`: they either return the name or an empty string. Thus, I eliminated the `for` and the PI name should be valid if only a first name or a last name is specified.

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryParam.scala
sraaphorst: Huh... weird about Travis. I had to rebuild the project, I think, to get things to work properly. Perhaps it has something to do with this since the version number on the lch bundle is different now?

swalker2m: Perhaps the test case works most of the time but not for all randomly generated input.
 
````
[error] (bundle_edu_gemini_lchquery_servlet/test:test) sbt.TestsFailedException: Tests unsuccessful
````

sraaphorst:
Okay, I see what I did wrong (I removed handling of '%' in query expressions, since it was never even mentioned on the ODB Browser page, I don't think).

I restored this and now all test cases and the spec pass locally. Let's see what Travis has to say.

sraaphorst:
Alright, all checks have passed 😄

PR #1090, opened 2016-09-14
REL-2890: Catch SIMBAD error response

cquiroz:
This PR fixes a border case when the simbad catalog returns a NullPointerException on its response. This PR was already opened but it was accidentally closed, see #1052


swalker2m:
👍

PR #1092, opened 2016-09-15
REL-2948: Unused OT preferences

cquiroz:
While updating the OT preferences it was noted that one user preference "Show alternative guide star options" was not used in the OT, this is a legacy option that is unused since we have BAGS

This PR removes this option and cleans some of the old preferences-related classes

Here's the option removed:

screenshot 2016-09-15 11 35 18

### DIFF
--- a/bundle/jsky.app.ot/src/main/java/jsky/app/ot/userprefs/ui/PreferenceButtonPanelBuilder.java
+++ b/bundle/jsky.app.ot/src/main/java/jsky/app/ot/userprefs/ui/PreferenceButtonPanelBuilder.java
@@ -1,11 +1,6 @@
-//
-// $
-//
-
 package jsky.app.ot.userprefs.ui;
 
 import edu.gemini.shared.gui.SeparateLineBorder;
-import edu.gemini.shared.util.immutable.ApplyOp;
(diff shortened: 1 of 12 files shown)


bundle/jsky.app.ot/src/main/java/jsky/app/ot/userprefs/ui/PreferenceButtonPanelBuilder.java
```diff
@@ -18,14 +13,14 @@
...
+    static final Color SELECTED_BORDER_COLOR = new Color(128, 128, 128);
 
-    /** Default {@link #setSelectedColor(java.awt.Color)} selected color}. */
-    public static final Color SELECTED_COLOR        = new Color(157, 157, 157);
+    /** Default selected color}. */
+    static final Color SELECTED_COLOR        = new Color(157, 157, 157);
 
-    /** Default {@link #setBackgroundColor(java.awt.Color) background color}. */
-    public static final Color BACKGROUND_COLOR      = new Color(191, 191, 191);
+    /** Default background color}. */
```
swalker2m: I suppose we could remove the trailing `}`.

PR #1093, opened 2016-09-15
REL-2949: fix smart gcal initialization from TestLauncher

swalker2m:
A small update to smart gcal initialization that will permit it to work from `TestLauncher`.  The issue is that the storage directory we use is in `/tmp/ot-testlauncher/ot-storage/smartgcal` and neither `ot-storage` nor `smartgcal` exist at startup.


swalker2m:
Merging this appears to be hopeless.  I'll just incorporate it in another PR.

PR #1096, opened 2016-09-26
REL-403: adding timing windows to OT plot

swalker2m:
This PR adds the long requested feature of timing windows to the OT target plot.  Elevation constraints and timing windows are independently selectable.  When selected, the timing windows are shown in reverse in that invalid times are crossed through while valid times are not painted.  This follows the task description in [REL-403](http://jira.gemini.edu:8080/browse/REL-403).

To accomplish this task, I worked with the existing elevation plot framework used by the OT.  I did move existing classes `TargetDesc` and `ObsTargetDesc` since they

1. are only needed by the elevation plot,
2. don't appear to be useful outside of that context, and
3. now require access to skycalc classes (`Union` and `Interval`) that weren't available otherwise.

The following screenshot displays one hour timing windows with a period of three hours.  The blank gaps are valid observing times according to the windows.

timingwindows1

The next screenshot shows timing windows and elevation constraints at the same time.  The green gaps without the crossed out area are valid observing times according to both constraints.  From this we can conclude the only valid time to observe this target at night is from 05 to 06 UT.

timingwindows2


bundle/jsky.elevation.plot/src/main/java/jsky/plot/ElevationPanel.java
```diff
@@ -76,6 +77,23 @@
     // Color used to display elevation constraints
     private static final Color CONSTRAINT_COLOR = new Color(0.0F, 1.0F, 0.0F, 0.5F);
 
+    private static final Paint TIMING_WINDOW_PAINT;
+
+    static {
+        final BufferedImage bi = new BufferedImage(4, 4,  BufferedImage.TYPE_4BYTE_ABGR);
+        final int rgb = new Color(164, 160, 203).getRGB();
```
cquiroz: Does this define the texture of the bars to be painted?

bundle/jsky.app.ot/src/main/java/jsky/app/ot/plot/ObsTargetDesc.java
```diff
@@ -100,10 +105,16 @@ public static TargetDesc getTargetDesc(IDBDatabaseService db, ISPObservation obs
...
         }
 
-        if (useTargetName)
-            return new ObsTargetDesc(targetName, pos, prio, category, obsId, targetName, timeStr, elType, min, max);
-        else
-            return new ObsTargetDesc(obsId, pos, prio, category, obsId, targetName, timeStr, elType, min, max);
+        // Function (start timestamp, end timestamp) => Union<Interval>.
+        // The Union indicates the time intervals where the target can be
+        // observed according to the observation's timing windows.
+        final BiFunction<Long, Long, Union<Interval>> timingWindows = (s, e) ->
```
cquiroz: What's a `BiFunction`, would it be equivalent to `Function2` in scala?

bundle/jsky.elevation.plot/src/main/java/jsky/plot/TargetDesc.java
```diff
@@ -5,11 +5,15 @@
...
  */
 
-package jsky.coords;
+package jsky.plot;
 
 
 import edu.gemini.shared.util.immutable.Option;
+import edu.gemini.skycalc.Interval;
+import edu.gemini.skycalc.Union;
+import jsky.coords.WorldCoords;
```
cquiroz: I've been trying to avoid `WorldCoords`. could/should we use the new `Coordinates` instead?

tpolecat:
Looks good. I'm unhappy in principle with the path this is taking since they will eventually want the whole QPT display here, but the implementation seems fine. 👍

bundle/jsky.elevation.plot/src/main/java/jsky/plot/TargetDesc.java
swalker2m: Here I'm not actually introducing any new use of `WorldCoords`.  The import happens only because `TargetDesc` was moved out of the lower-level `jsky.coords` package.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/plot/ObsTargetDesc.java
swalker2m: That's my understanding, yes.

bundle/jsky.elevation.plot/src/main/java/jsky/plot/ElevationPanel.java
swalker2m: Right, this defines how the area outside of the defined timing windows will be painted.

PR #1098, opened 2016-09-29
REL-2955: BAGS/Sync bug

swalker2m:
This PR addresses a bug with BAGS and sync.  There are two related changes:

1. The major issue is in the `BagsManager.watch` method.  `watch` is called whenever a new program is opened, or when a sync that has to incorporate remote changes is executed.  When there are incoming changes the current program is replaced wholesale by an updated copy. The `watch` method wasn't taking this into account and would skip these updates.  Future change events would be ignored since `BagsManager` would continue watching the previous/unchanged version of the program.

2. The `BagState`s all kept a reference to the corresponding `ISPObservation`.  This is problematic for sync because the entire program is copied and the copy is updated.  Much of the refactoring work went into avoiding the need to hold on to `ISPObservation` references and instead fetch them from keys when needed.

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
@@ -1,6 +1,6 @@
 package jsky.app.ot.ags
 
-import jsky.app.ot.ags.BagsState.{ErrorTransition, IdleState, StateTransition}
+import jsky.app.ot.ags.BagsState._
 import java.beans.{PropertyChangeEvent, PropertyChangeListener}
 import java.time.Instant
 import java.util.concurrent.TimeoutException
@@ -188,7 +188,7 @@ object BagsState {
       (PendingState(obs, None), BagsManager.wakeUpAction(obs, 0))
   }
 
-  private def hashObs(ctx: ObsContext): AgsHashVal = {
+  def hashObs(ctx: ObsContext): AgsHashVal = {
     val when = ctx.getSchedulingBlockStart.asScalaOpt | Instant.now.toEpochMilli
     AgsHash.hash(ctx, when)
   }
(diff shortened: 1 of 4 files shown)


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -282,97 +301,115 @@ object BagsManager {
...
+  /** Add a program to our watch list and attach listeners. Enqueue modified
     * observations for consideration for a BAGS lookup.
     */
   def watch(prog: ISPProgram): Unit = Swing.onEDT {
-    // Check that the program isn't being managed already. If it is, do nothing.
+    Log.info("watch")
+
+    // Remove existing listeners, if any.
+    prog.removeStructureChangeListener(StructureListener)
+    prog.removeCompositeChangeListener(ChangeListener)
```
tpolecat: Is it possible to miss updates since we're not listening during the computation below?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -282,97 +301,115 @@ object BagsManager {
...
+  private def fail(o: ObsKey, why: String, delay: Long): Unit =
+    transition(o, _.fail(why, delay))
 
-  /** Add a program to our watch list and attach listeners. Enqueue all
+  /** Add a program to our watch list and attach listeners. Enqueue modified
     * observations for consideration for a BAGS lookup.
     */
   def watch(prog: ISPProgram): Unit = Swing.onEDT {
-    // Check that the program isn't being managed already. If it is, do nothing.
+    Log.info("watch")
```
tpolecat: Probably should remove this or improve the message and turn down to `fine`.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -282,97 +301,115 @@ object BagsManager {
...
+
+    // Watch the program for future edits.
+    prog.addCompositeChangeListener(ChangeListener)
+    prog.addStructureChangeListener(StructureListener)
   }
 
   /** Remove a program from our watch list and remove listeners.
     */
   def unwatch(prog: ISPProgram): Unit = Swing.onEDT {
+    Log.info("unwatch")
```
tpolecat: See comment above re: logging `"watch"`.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -486,25 +527,29 @@ object BagsManager {
...
-      Option(TpeManager.get()).filter(_.isVisible).foreach { tpe =>
-        sel.assignments.foreach { ass =>
-          val clazz = ass.guideProbe.getType match {
-            case GuideProbe.Type.AOWFS => classOf[Altair_WFS_Feature]
-            case GuideProbe.Type.OIWFS => classOf[OIWFS_Feature]
-            case GuideProbe.Type.PWFS  => classOf[TpePWFSFeature]
+    def doApply(obs: ISPObservation): Unit = {
+      obs.getProgram.removeStructureChangeListener(StructureListener)
+      obs.getProgram.removeCompositeChangeListener(ChangeListener)
+      applySelection(TpeContext(obs))
```
tpolecat: Same question here about racing. I guess we do this a lot in the OT and just assume all updates happen on the EDT.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
swalker2m: Right thanks, that wasn't supposed to have been left there.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -65,6 +68,20 @@ object BagsState {
...
+  def hashObs(ctx: ObsContext): AgsHashVal = {
+    val when = ctx.getSchedulingBlockStart.asScalaOpt | Instant.now.toEpochMilli
+    AgsHash.hash(ctx, when)
+  }
+
+  def hashObs(o: ISPObservation): Option[AgsHashVal] =
+    ObsContext.create(o).asScalaOpt.map(hashObs)
+
+  def fetchObs(k: ObsKey): Option[ISPObservation] =
+    Option(SPDB.get().lookupProgram(k.prog.key)).flatMap { p =>
```
cquiroz: Unrelated to this PR but is the call `SPDB.get().lookupProgram()` the canonical way to read the program from the `SPNodeKey`? would it read the local db only?

tpolecat: Yeah this bothers me but it's the OT so I'm ok with compromised ethics.

swalker2m: It's kind of a hack, admittedly.  `SPDB` is assigned with the local ODB on startup.  This is used throughout the `OT` so I don't feel *too* guilty about it.  But I do feel a bit guilty about it.

PR #1099, opened 2016-10-06

cquiroz:
This PR implements a new OT feature to download automatically the images for the TPE. The following features have been implemented:

* Automatic download of target images when loading or creating a program
* Track of target changes via e.g. dragging or manual edit, to redownload images as needed
* Automatic selection of catalog based on observing wavelength
* Catalogs are not fully reliable on providing the image center. The fits files will be inspected to extract the true center and size
* Gemini's catalogs at GN and GS are presented as a single one, using the fastest one to respond
* On-disk images cache. The cache can be reused across OT sessions
* The cache max drive space is configurable and a UI has been provided to change the size and to manually clean the cache
* Catalog selection is automatic but it can be overridden in the TPE. The overrides are preserved across OT sessions
* The TPE shows when an image is being downloaded if not already present. In case of errors, it is possible to request a new download

I consider this code ready for testing. However, there are some improvements that could be done if there is a bit more time (the 2 first may be a MustDo though)

- [x] Several tests broke when the equation to calculate if a target belongs to an image changed. We could try to recover them, though the change makes testing much harder.
- [ ] There are few problematic spots on the sky where the catalog does not return an image containing the required target. If this is a common occurrence we may be able to workaround by relaxing the criteria to include an image
- [ ] Compressing of DSS images
- [ ] Support a mosaic based catalog instead of 2MASS
- [ ] The overrides file may need to be kept up to date removing old programs
- [ ] Improve parsing of Gemini DSS fits file to calculate size based on the FITS WCS matrix

Sorry for the PR size, the feature grew as some new ideas appeared and bugs were discovered


arturog8m:
Holy cow, this is massive!. I think it will be useful if you can point people to specific areas of the implementation that are important in your opinion to be reviewed. Good to see this in place, great job!

cquiroz:
I think the scala side of the code is worth checking. The java side is mostly the hooks needed on the TPE and OT to use the new code and some refactoring plus code cleanups when appropriate

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/image/ImageCacheWatcher.scala
```diff
@@ -0,0 +1,124 @@
...
+
+    for {
+      _  <- rmTempFiles
+      ds <- initStream
+      ab <- readFiles(ds).onFinish(_ => closeStream(ds)) // Make sure the stream is closed
+    } yield ab
+  }
+
+  /**
+    * Observe the file system to detect when files are modified and update the cache accordingly
```
swalker2m: Outside of the OT itself, how would files on disk be modified?  In other words, is this necessary?

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/image/ImageCatalog.scala
```diff
@@ -2,85 +2,121 @@ package edu.gemini.catalog.image
...
+sealed abstract class ImageCatalog(val id: String, val displayName: String, val shortName: String) {
+  /** Returns the urls that can load the passed coordinates */
+  def queryUrl(c: Coordinates): NonEmptyList[URL]
+
+  /** Size of the default requested image */
+  def imageSize: AngularSize
+
+  /** Overlap angle between adjacent images */
+  // TODO Should be corrected for dec?
+  def adjacentOverlap: Angle
```
swalker2m: Maybe add a note about what this is used for?

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/image/ImageCatalogClient.scala
```diff
@@ -0,0 +1,266 @@
...
+
+case object AngularSize {
+  val zero = AngularSize(Angle.zero, Angle.zero)
+
+  /** @group Typeclass Instances */
+  implicit val equal = Equal.equalA[AngularSize]
+
+  /** @group Typeclass Instances */
+  implicit val order: Order[AngularSize] =
+    Order.orderBy(a => (a.ra.toRadians * a.dec.toRadians))
```
swalker2m: Maybe this doesn't matter (?) but I expected to see (ra, dec) instead of (ra * dec).  With the multiplication version two instances that are not equals could compare `EQ`.

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/image/ImageCacheWatcher.scala
cquiroz: This would detect if somebody modifies the files outside the OT, it is a bit overkill since we will check for deletions anyway before displaying in the TPE, At some moment I thought about hashing the files to detect modifications and the watcher would have been more useful then, anyway it doesn't hurt

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/image/ImageCatalogClient.scala
cquiroz: Good catch. I struggled with this, and it was `(ra, dec)` for a while. You're right that `equals` would be violated but it didn't seem right that you'd consider an image with ra small but a large dec smaller than one with ra slightly bigger but dec small.

This is mostly academic since it is only used to discern what image to pick if the angular distance from target to center is the same between 2 images, and images are mostly equal size for a given catalog. In fact this is only a problem in tests

I'll change it as suggested

cquiroz:
I think the integration with the TPE maybe wasted but the feature of automatic downloads and the improved handling of images will be used in the future OT. I expect all the code in the `catalog` bundle to survive, though the code under `jsky.app.ot` probably won't

tpolecat:
I'm looking through this now. Give me an hour or two.

Re: the checklist above, I would prefer to get this merged and then do follow-up work rather than adding to this PR. I fear we will hit the point where GH will refuse to show the diff because it's too big.

cquiroz:
About the checklist, the only one that may have some relevance is the first (depending on user testing), the rest look more like candidates for future tasks (if ever needed)

cquiroz:
I'm adding a screenshot of the TPE with this PR. It shows some of the visible changes.

* No Images button
* Preferences dialog box
* Catalog override selection on the tpe

screenshot 2016-10-07 14 18 49

tpolecat:
How does it decide which images to remove if you shrink the cache size? This may be obvious, I haven't gotten very far in the code yet.

cquiroz:
there is no immediate action when you change cache size. The next time you download an image it will prune the cache to size, keeping the most recently used as much as space allows.

To that effect, there is an in-memory structure keeping when an image was last used, and this is initialized at startup reading (as much as possible) the last access time of the images

tpolecat:
So it touches the image files when you use them, to keep the LRU accurate?

cquiroz:
Implicitly. TPE will read the file before displaying, updating the access time

tpolecat:
I won't have my review done until later today, sorry. It's bigger than I thought. Looks good so far though ;-)

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/image/StoredImagesCache.scala
```diff
@@ -0,0 +1,161 @@
...
+  */
+object ImageCacheOnDisk {
+
+  /**
+    * Method to prune the cache, limiting the space used
+    */
+  def pruneCache(maxSize: Information): Task[Unit] = Task.fork {
+    // Remove files from the in memory cache and delete from drive
+    def deleteOldFiles(files: List[ImageInFile]): Task[Unit] =
+      Task.gatherUnordered(files.map(StoredImagesCache.remove)) *> Task.delay(files.foreach(_.file.toFile.delete()))
```
tpolecat: The tasks in this list aren't associated with an executor so they will run sequentially when you do this. But all they're doing is an in-memory update so maybe you just want `files.traverseU(StoredImagesCache.remove)`?

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/image/ImageCacheWatcher.scala
```diff
@@ -0,0 +1,124 @@
...
+      Files.walkFileTree(cacheDir, new SimpleFileVisitor[Path] {
+        override def visitFile(file: Path, attrs: BasicFileAttributes): FileVisitResult = {
+          super.visitFile(file, attrs)
+          if (matcher.matches(file.getFileName)) {
+            file.toFile.delete()
+          }
+          FileVisitResult.CONTINUE
+        }
+      })
+    }
```
cquiroz: Yeah, IO could be much better too. I'm anyway avoiding new libraries on the OT (Like I'd like http4s-client) especially to avoid dealing with OSGi

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/image/ImageCacheWatcher.scala
```diff
@@ -0,0 +1,124 @@
...
+
+  /**
+    * Observe the file system to detect when files are modified and update the cache accordingly
+    */
+  private def watch(cacheDir: Path)(implicit B: Bind[Task]): Task[Unit] = {
+    def waitForWatcher(watcher: WatchService) : Task[Unit] = Task.delay {
+      val watchKey = watcher.take()
+      // Called when a file is deleted
+      val tasks = watchKey.pollEvents().asScala.toList.collect {
+          case ev: WatchEvent[Path] if ev.kind() == ENTRY_DELETE =>
```
cquiroz: Right, I get a warning.

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/image/ImageCatalog.scala
```diff
@@ -2,85 +2,121 @@ package edu.gemini.catalog.image
...
+
+import scalaz._
+import scalaz.NonEmptyList._
+import Scalaz._
 
 /** Represents an end point that can load an image for a given set of coordinates */
-sealed abstract class ImageCatalog(val id: String, val displayName: String) {
-  /** Returns the url that can load the passed coordinates */
-  def queryUrl(c: Coordinates): URL
+sealed abstract class ImageCatalog(val id: String, val displayName: String, val shortName: String) {
```
cquiroz: It is a unique ID used in the file name to reparse them into files belonging to a given catalog. I guess it could be a `sealed trait` to limit what you put there

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/image/ImageCatalogClient.scala
```diff
@@ -0,0 +1,266 @@
...
+
+    Δφ <= εφ && Δλ*θ <= ελ
+  }
+}
+
+object ImageInFile {
+  /** @group Typeclass Instances */
+  implicit val equals: Equal[ImageInFile] = Equal.equalA[ImageInFile]
+
+  val fileRegex: Regex = """img_(.*)_ra_(.*)_dec_(.*)_w_(\d*)_h_(\d*)\.fits.*""".r
```
cquiroz: The pattern trusts that only files written by the program will match, you can probably fool it if you want

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/image/ImageCatalogClient.scala
```diff
@@ -0,0 +1,266 @@
...
+    file.getName match {
+      case fileRegex(c, raStr, decStr, w, h) =>
+        for {
+          catalog <- ImageCatalog.byId(c)
+          ra      <- Angle.parseHMS(raStr).map(RightAscension.fromAngle).toOption
+          dec     <- Angle.parseDMS(decStr).toOption.map(_.toDegrees).flatMap(Declination.fromDegrees)
+          width   <- Angle.fromArcsecs(w.toInt).some
+          height  <- Angle.fromArcsecs(h.toInt).some
+        } yield ImageInFile(ImageSearchQuery(catalog, Coordinates(ra, dec), AngularSize(width, height)), file.toPath, file.length())
+      case _                           => None
```
cquiroz: it means that somebody put a non-matching file name in the cache, while not common it cannot be discarded. Those should be just ignored. It would also let us evolve the naming schema if needed

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/image/StoredImagesCache.scala
```diff
@@ -0,0 +1,161 @@
...
+
+  def +(i: ImageInFile): StoredImages             = copy((Instant.now, i) :: entries)
+  def +(i: Instant, e: ImageInFile): StoredImages = copy((i, e) :: entries)
+  def -(i: ImageInFile): StoredImages             = copy(entries.filterNot(_._2 === i))
+
+  /**
+    * Indicates the image was used, update the access time
+    */
+  def touch(i: ImageInFile): StoredImages = copy(entries.collect {
+    case x if x._2 === i => (Instant.now, x._2)
```
cquiroz: really? would the `i` in backquotes match?

tpolecat: Yes. Uppercase and back-quoted identifiers are matched by value.

```scala
case  Foo  => ... matches only the value Foo and binds nothing
case  foo  => ... matches everything and binds foo
case `foo` => ... matches only the value foo and binds nothing
```

It's a little-known feature. Note that it uses `==` which sucks.

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/image/ImageCatalogClient.scala
```diff
@@ -0,0 +1,266 @@
...
+import scalaz._
+import scalaz.concurrent.Task
+
+/**
+  * Angular size of an image. Used to store the real size of images instead to the requested size
+  */
+case class AngularSize(ra: Angle, dec: Angle) {
+  def this(size: Angle) = this(size, size)
+
+  def halfDec = (dec / 2).getOrElse(Angle.zero)
```
cquiroz: I guess it means you can also do `~(dec / 2)` too?

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/image/TaskHelper.scala
```diff
@@ -0,0 +1,35 @@
...
+
+import scalaz.{-\/, INil, NonEmptyList, \/, \/-}
+import scalaz.concurrent.{Future, Task}
+
+object TaskHelper {
+  /**
+   * From a list of task create a new Task that will return the first available result
+   * or a failure if all the tasks fail
+   */
+  def selectFirstToComplete[A](tasks: NonEmptyList[Task[A]])(implicit pool: ExecutorService): Task[A] = {
```
tpolecat: I believe this does what you want.

```scala
def selectFirstToComplete[A](tasks: NonEmptyList[Task[A]])(implicit pool: ExecutorService): Task[A] = {
  val ts = tasks.map(Task.fork(_)(pool))
  Nondeterminism[Task].chooseAny(ts.head, ts.tail.toList).map(_._1)
}
```

For example

```scala
implicit val ex = Strategy.DefaultExecutorService
def t(n: Int) = Task.delay { Thread.sleep(n * 100); println(s"returning $n");  n }
val a = selectFirstToComplete(NonEmptyList(5, 6, 2, 9).map(t))

scala> a.unsafePerformAsync(e => println(s"the answer was $e"))

scala> returning 2
the answer was \/-(2)
returning 5
returning 6
returning 9
```

bundle/jsky.app.ot/src/main/scala/edu/gemini/catalog/ui/image/BackgroundImageLoader.scala
```diff
@@ -0,0 +1,220 @@
...
+}
+
+/**
+  * This interface can be used to listen when the image is being loaded and update the UI
+  */
+trait ImageLoadingListener {
+  def downloadStarts(): Task[Unit]
+  def downloadCompletes(): Task[Unit]
+  def downloadError(): Task[Unit]
+}
```
tpolecat: Why not a data type? 

```scala
case class ImageLoadingListener(start: Task[_], complete: Task[_], error: Task[_])
```

bundle/jsky.app.ot/src/main/scala/edu/gemini/catalog/ui/image/BackgroundImageLoader.scala
```diff
@@ -0,0 +1,220 @@
...
+object BackgroundImageLoader {
+  private val taskUnit = Task.now(())
+
+  private def imageDownloadsThreadFactory(priority: Int) = new ThreadFactory {
+    private val threadNumber: AtomicInteger = new AtomicInteger(1)
+    private val defaultThreadFactory = Executors.defaultThreadFactory()
+
+    override def newThread(r: Runnable): Thread = {
+      val name = s"Background Image Downloads - ${threadNumber.getAndIncrement()} - priority: $priority"
+      defaultThreadFactory.newThread(r) <| {_.setDaemon(true)} <| {_.setName(name)} <| {_.setPriority(priority)}
```
tpolecat: `<|` is deprecated and will be removed so it's best to use a block in cases like this.

bundle/jsky.app.ot/src/main/scala/edu/gemini/catalog/ui/image/BackgroundImageLoader.scala
```diff
@@ -0,0 +1,220 @@
...
+
+  /**
+    * Execution context for lower priority downloads
+    */
+  private val lowPriorityEC = Executors.newFixedThreadPool(ImageCatalog.all.length, imageDownloadsThreadFactory(Thread.MIN_PRIORITY))
+
+  /**
+    * Regular execution context for higher priority tasks, i.e. UI requests
+    */
+  private val highPriorityEC = Executors.newFixedThreadPool(ImageCatalog.all.length, imageDownloadsThreadFactory(Thread.NORM_PRIORITY))
```
tpolecat: Can we factor out a method here?

bundle/jsky.app.ot/src/main/scala/edu/gemini/catalog/ui/image/BackgroundImageLoader.scala
```diff
@@ -0,0 +1,220 @@
...
+
+  /** Called when a program is created to download its images */
+  def watch(prog: ISPProgram): Unit = {
+    // Listen for future changes
+    prog.addCompositeChangeListener(ChangeListener)
+    val targets = prog.getAllObservations.asScala.toList.flatMap(_.getObsComponents.asScala.flatMap(n => requestedImage(TpeContext(n))))
```
tpolecat: I prefer a `for` comprehension when you have more than one `flatMap`.

bundle/jsky.app.ot/src/main/scala/edu/gemini/catalog/ui/image/BackgroundImageLoader.scala
```diff
@@ -0,0 +1,220 @@
...
+
+  // Watches for changes to existing observations, runs BAGS on them when updated.
+  object ChangeListener extends PropertyChangeListener {
+    override def propertyChange(evt: PropertyChangeEvent): Unit =
+      evt.getSource match {
+        case node: ISPNode => Option(node.getContextObservation).foreach { o =>
+          val task = requestedImage(TpeContext(o)).fold(taskUnit)(requestImageDownload(highPriorityEC))
+
+          // Run it in the background as it is lower priority than GUI
+          runAsync(task)(_ => ())(lowPriorityEC)
```
tpolecat: We're swallowing errors here, can we at least log it?

bundle/jsky.app.ot/src/main/scala/edu/gemini/catalog/ui/image/BackgroundImageLoader.scala
```diff
@@ -0,0 +1,220 @@
...
+    * Creates a task to load an image and set it on the tpe
+    */
+  private[image] def requestImageDownload(pool: ExecutorService)(t: TargetImageRequest): Task[Unit] =
+    for {
+      catalog  <- ObservationCatalogOverrides.catalogFor(t.key, t.obsWavelength)
+      image    <- loadImage(ImageSearchQuery(catalog, t.coordinates, catalog.imageSize), ImageCatalogPanel.resetListener)(pool)
+    } yield image match {
+      case Some(e) => updateTpeImage(e) // Try to set it on the UI
+      case _       => // Ignore, some other thread is downloading the image
+    }
```
tpolecat: `updateTpeImage` is a side-effect so you need to capture it with `delay`. Something like

```scala
for {
  catalog  <- ObservationCatalogOverrides.catalogFor(t.key, t.obsWavelength)
  image    <- loadImage(ImageSearchQuery(catalog, t.coordinates, catalog.imageSize), ImageCatalogPanel.resetListener)(pool)
  _        <- image.fold(taskUnit)(e => Task.delay(updateTpeImage(e)))
} yield ()
```

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/image/TaskHelper.scala
tpolecat: Ah, yeah you the first failure in that case.

tpolecat: Your implementation is fine then. 👍

bundle/jsky.app.ot/src/main/scala/edu/gemini/catalog/ui/image/BackgroundImageLoader.scala
cquiroz: I guess that's right. My thinking was that actually `updateTpeImage` delegates to the swing thread anyway, but yes, it should go into delay

bundle/jsky.app.ot/src/main/scala/edu/gemini/catalog/ui/image/BackgroundImageLoader.scala
cquiroz: `<|` is deprecated but we use it extensively in the OCS, particularly in Swing. I personally like it and would rather make our own `unsafeTap` once we move to scalaz 7.3

bundle/jsky.app.ot/src/main/scala/edu/gemini/catalog/ui/image/BackgroundImageLoader.scala
cquiroz: In practice this will never be called. Errors are either handled on the task code, or they run on the EDT thread. Regardless, I think we should put a log entry as suggested

cquiroz:
I updated the code according to the PR comments and also improved the inside check making it more robust on edge cases and thus fixing most of the ignored tests

PR #1101, opened 2016-10-08
REL-2906: Updated F2 templates for 2017A

cquiroz:
Small changes to the phase2 template for F2


cquiroz:
I didn't add a test since this PR removes a feature (darks) from the template. I guess it is worth to add a test confirming that no darks are added

cquiroz:
Added test as suggested

tpolecat:
👍 Thanks! The template expansion stuff has so much machinery it's just kind of reassuring to see a test in there.

cquiroz:
It has been a source of hard to track bugs in the past, so yes tests are nice!

PR #1103, opened 2016-10-15
Remove incorrect logging statement

cquiroz:
This is just a tiny change but had an impact on our last AGS failure. The line is missleading as this is not only called by BAGS but can be called by AGS, the Catalog Query Tool, etc

It was probably added to track the thread usage in BAGS but I think it can be removed

### DIFF
--- a/bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableClient.scala
+++ b/bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableClient.scala
@@ -25,6 +25,8 @@ trait VoTableBackend {
 }
 
 trait CachedBackend extends VoTableBackend {
+  val Log = Logger.getLogger(this.getClass.getName)
+
   case class SearchKey(query: CatalogQuery, url: URL)
 
   case class CacheEntry[K, V](k: K, v: V)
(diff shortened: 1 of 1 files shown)


bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableClient.scala
```diff
@@ -131,7 +131,6 @@ trait CachedBackend extends VoTableBackend {
 
   // Cache the query not the future so that failed queries are executed again
   override protected [votable] def doQuery(query: CatalogQuery, url: URL)(ec: ExecutionContext): Future[QueryResult] = Future {
-    Logger.getLogger(getClass.getName).info(s"Starting BAGS lookup on ${Thread.currentThread}")
```
swalker2m: Perhaps we should fix the message and maybe lower the log level?  I don't know but it seems vaguely useful to have the option of logging this particularly given the issue with AGS estimations.

cquiroz: I'm not sure either, we could lower it to fine. My main complaint is the BAGS mention as this is not a BAGS-only event

swalker2m: Right.  Either way is fine with me.

PR #1104, opened 2016-10-18
REL-2655: Fixed issue with ODB Browser target requests by filtering / unpacking Option.

sraaphorst:
So as it turns out, I was forgetting to filter / unpack a scala `Option` when working with targets in the ODB Browser. This was adding `Option[SPTarget]` to the `TargetsNode`, which blindly accepts anything `Serializable`. This caused the marshaling to fail when converting to XML.

This collects the unpacked `Some[SPTarget]` entries from the list and adds them to the `TargetsNode` instead.

I have tested with three fairly detailed programs and the query runs properly and the targets now appear in the XML output as they did previously.

(Also did some very minor cleanup.)

### DIFF
--- a/bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
+++ b/bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
@@ -191,15 +191,9 @@ class LchQueryFunctor(queryType: LchQueryFunctor.QueryType,
 
                 if (queryType == LchQueryFunctor.QueryType.TargetQuery) {
                   obs.targetEnvironment.foreach { env =>
-                    Option(env.getTargets) collect {
-                      case lst if lst.nonEmpty => lst.asScalaList
-                    } foreach { tgts =>
-                      setTargetsNode(new TargetsNode() {
-                        tgts.map(tgt => makeTargetNode(tgt, env)).collect {
-                          case Some(t) => t
-                        }.foreach(getTargets.add)
-                      })
-                    }
+                    setTargetsNode(new TargetsNode() {
+                      getTargets.addAll(env.getTargets.asScalaList.flatMap(t => makeTargetNode(t, env)).asJavaCollection)
+                    })
                   }
                 }
               }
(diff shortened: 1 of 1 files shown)


bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
```diff
@@ -195,14 +195,15 @@ class LchQueryFunctor(queryType: LchQueryFunctor.QueryType,
                       case lst if lst.nonEmpty => lst.asScalaList
                     } foreach { tgts =>
                       setTargetsNode(new TargetsNode() {
-                        tgts.map(tgt => makeTargetNode(tgt, env)).foreach(getTargets.add)
+                        tgts.map(tgt => makeTargetNode(tgt, env)).collect {
```
cquiroz: if it is a `List[Option]` you can just call `flatten`

tpolecat: If you have a `List[Option[A]]` you can call `.flatten` with the stdlib collections because there is an implicit conversion from `Option` to `Iterable` and `CanBuildFrom` can gloss over the difference between `List` and `Iterable`. But it's a very specific case.

In the general case it's bogus, which is why I prefer `collect`.

sraaphorst:
Thanks for the suggestions and to @tpolecat for the very useful explanation.
Upon a second look, the `Option` was unnecessary, and based on the previous behaviour, seems like you always want to add a `TargetsNode` to be consistent with what is happening on `gnodb` and `gsodb` (i.e. getting an empty `<targets/>` tag in the XML if there are no targets), so I cleaned this up a lot. I did go with `flatMap` for the `makeTargetsNode` transformation as per @cquiroz's suggestion, as that seemed to make more sense in the much cleaner and shortened context.

Tested again with the same programs, one with observations each with targets, and one with observations both with and without targets, and it looks to be returning the proper results.

cquiroz:
Do we have tests for the `LchFunctor` I'd think this change would break tests

sraaphorst:
We do have tests, but nothing for a target node appearing specifically as I guess that probably doesn't really matter much and I'm not sure what the users of the ODB Browser expect and am just going with the old status quo by changing this.

The checks have all passed. I can add some more tests, I guess, but without knowing if a specific behaviour is required, I'm hesitant to do so.

sraaphorst:
So =grins= it would be very useful is someone could approve this, so I can release a new test ODB on gnodbtest2 / gsodbtest2.

andrewwstephens:
@sraaphorst - Should we expect any changes in the ODB  Browser, or are there things that we should specifically test?

PR #1105, opened 2016-10-18
REL-2957: thread pool for AgsServlet

swalker2m:
This PR adds a thread pool for the `AgsServlet` used by the PIT for running AGS estimations. For context, we discovered during the day or two before the last proposal deadline that OCS cron jobs stopped working. This was traced to a flood of `AgsServlet` requests from PIT. Since the servlet was doing AGS estimation in the default execution context, it was competing with other non-AGS work like the daily backup. Many estimation requests were resulting in timeout failures so PIT also wasn't showing the guide star search result probability.

This update doesn't fix the load problem per se, but it does improve the situation. Namely the default execution context thread count is limited to the number of processors available. Since AGS estimations are mostly blocked on the catalog search, throughput is lower than it otherwise should be. Switching to a separate thread pool allows us to make more threads available for simultaneous catalog requests and more importantly isolate them from the rest of the ODB. It should no longer cause cron jobs to fail for example.

Testing with the old configuration, I could reproduce the problem with 100 simultaneous queries. Perhaps even less are required, I didn't try fewer. With this change, 200 simultaneous non-GeMS queries can run without a problem, but it does start to get bogged down and timeout at 300.

The next step, I think, is to examine what the PIT is doing and see whether we can cut down on the number of queries we send to the ODB. We could fairly easily (?) cache results keyed by the query URL since unless the parameters change the result will always be the same anyway.

### DIFF
--- a/bundle/edu.gemini.ags.servlet/src/main/scala/edu/gemini/ags/servlet/AgsServlet.scala
+++ b/bundle/edu.gemini.ags.servlet/src/main/scala/edu/gemini/ags/servlet/AgsServlet.scala
@@ -7,16 +7,23 @@ import edu.gemini.spModel.obs.context.ObsContext
 
 import java.io.{BufferedOutputStream, IOException}
 import java.net.URLDecoder
+import java.util.concurrent.{ThreadFactory, LinkedBlockingDeque, TimeUnit, ThreadPoolExecutor}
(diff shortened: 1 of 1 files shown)


cquiroz:
I think we could make the PIT more cooperative, e.g. by having delays once we fail and improving caching on the client side

PR #1106, opened 2016-10-20

cquiroz:
This PR contains some small bug fixes and improvements requested by users:

* It will create the cache dir if not present (Useful if the user deletes the cache file while the OT is open)
* At program load it will not load images for inactive or observed observations (They will load if the observation is loaded on the TPE)
* Catalog overrides were stored even if it matched the calculated one, polluting the overrides file

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/edu/gemini/catalog/ui/image/BackgroundImageLoader.scala
+++ b/bundle/jsky.app.ot/src/main/scala/edu/gemini/catalog/ui/image/BackgroundImageLoader.scala
@@ -14,6 +14,7 @@ import scala.collection.JavaConverters._
 import edu.gemini.pot.sp._
 import edu.gemini.spModel.core.{Angle, Coordinates, Wavelength}
 import edu.gemini.spModel.obs.ObservationStatus
+import edu.gemini.spModel.rich.pot.sp._
 import jsky.app.ot.tpe.{ImageCatalogPanel, TpeContext, TpeImageWidget, TpeManager}
 import jsky.image.gui.ImageLoadingException
 
@@ -78,19 +79,14 @@ object BackgroundImageLoader {
   /** Called when a program is created to download its images */
   def watch(prog: ISPProgram): Unit = {
     // At startup only load images for active programs
-    def needsImage(ctx: TpeContext): Boolean = {
-      ctx.obsShell.exists(ObservationStatus.computeFor(_) match {
-        case ObservationStatus.INACTIVE | ObservationStatus.OBSERVED => false
-        case _                                                       => true
-      })
-    }
+    def needsImage(ctx: TpeContext): Boolean =
+      ctx.obsShell.exists(ObservationStatus.computeFor(_).isActive)
 
     // Listen for future changes
     prog.addCompositeChangeListener(ChangeListener)
     val targets = for {
-        p      <- prog.getAllObservations.asScala.toList
-        obs    <- p.getObsComponents.asScala
-        tpeCtx  = TpeContext(obs)
+        p      <- prog.allObservations
+        tpeCtx  = TpeContext(p)
         if needsImage(tpeCtx)
         i      <- requestedImage(tpeCtx)
       } yield i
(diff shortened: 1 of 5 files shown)


bundle/jsky.app.ot/src/main/scala/edu/gemini/catalog/ui/image/BackgroundImageLoader.scala
```diff
@@ -76,12 +77,22 @@ object BackgroundImageLoader {
 
   /** Called when a program is created to download its images */
   def watch(prog: ISPProgram): Unit = {
+    // At startup only load images for active programs
+    def needsImage(ctx: TpeContext): Boolean = {
+      ctx.obsShell.exists(ObservationStatus.computeFor(_) match {
+        case ObservationStatus.INACTIVE | ObservationStatus.OBSERVED => false
+        case _                                                       => true
```
sraaphorst: Here, it seems like it might be more logical to use the `ObservationStatus.isActive` return value instead of matching? Seems to correspond to exactly what you need, and I suspect that that is the sensible value you'd want. So:
`ctx.obsShell.exists(ObservationStatus.computeFor(_).isActive)`?

bundle/jsky.app.ot/src/main/scala/edu/gemini/catalog/ui/image/BackgroundImageLoader.scala
```diff
@@ -76,12 +77,22 @@ object BackgroundImageLoader {
...
+      })
+    }
+
     // Listen for future changes
     prog.addCompositeChangeListener(ChangeListener)
     val targets = for {
-        p   <- prog.getAllObservations.asScala.toList
-        obs <- p.getObsComponents.asScala
-        i   <- requestedImage(TpeContext(obs))
+        p      <- prog.getAllObservations.asScala.toList
```
swalker2m: There is a bit of magic that does this exactly in `RichProgram`, if you want to import it.  Not a bit deal.

bundle/jsky.app.ot/src/main/scala/edu/gemini/catalog/ui/image/BackgroundImageLoader.scala
```diff
@@ -76,12 +77,22 @@ object BackgroundImageLoader {
...
+        obs    <- p.getObsComponents.asScala
+        tpeCtx  = TpeContext(obs)
```
swalker2m: Won't this create a bunch of (for our purposes here) identical `TpeContext` objects, one after another?  Why not just create the `tpeCtx` with the observation itself once?

bundle/jsky.app.ot/src/main/scala/edu/gemini/catalog/ui/image/ObservationCatalogOverrides.scala
```diff
@@ -98,11 +99,18 @@ object ObservationCatalogOverrides {
       }
     }
 
+    // Updates the overrides file, removing the existing entries if needed
+    def newOverrides(overrides: Overrides): Overrides =
+      if (ImageCatalog.catalogForWavelength(wv.some) =/= c) {
+          overrides.copy(overrides = CatalogOverride(key, c) :: overrides.overrides.filter(_.key != key))
+      } else {
+          overrides.copy(overrides =overrides.overrides.filter(_.key != key))
+      }
```
swalker2m: What are overrides?  I'm not clear on what is happening here.

bundle/jsky.app.ot/src/main/scala/edu/gemini/catalog/ui/image/BackgroundImageLoader.scala
swalker2m: Good point, I'd forgotten about that.

cquiroz: Thanks, good to know

PR #1109, opened 2016-10-24
REL-2638: Add individual CWFS stars using the TPE

cquiroz:
This PR adds support for adding CWFS guide stars on the TPE once the GEMS Manual Guide Star box has been opened. To make it work you need to get the GEMS Dialog box and using the TPE click on the `+CMFS1/2/3` button and hit on of the hilighted targets

There were at least 2 other alternatives to this problem:
* Integrate GEMS to the Catalog Query Tool
* Add explicit add buttons to GEMS Guide Star Dialog

These were discarded due to the amount of time available

### DIFF
--- a/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/CatalogSearchCriterion.scala
+++ b/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/CatalogSearchCriterion.scala
@@ -65,9 +65,8 @@ case class CatalogSearchCriterion(name: String, radiusConstraint: RadiusConstrai
      * @param obj the SiderealTarget to match
      * @return true if the object matches the magnitude and radius limits
      */
-    def matches(obj: SiderealTarget): Boolean = {
+    def matches(obj: SiderealTarget): Boolean =
       matches(obj.magnitudes) && matches(obj.coordinates)
-    }
 
     private def matches(coords: Coordinates): Boolean = {
       val distance = Coordinates.difference(adjBase, coords).distance
--- a/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/CatalogSearchCriterion.scala
+++ b/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/CatalogSearchCriterion.scala
@@ -107,7 +107,7 @@ case class GemsCatalogSearchResults(criterion: GemsCatalogSearchCriterion, resul
 
   def resultsAsJava: java.util.List[SiderealTarget] = new java.util.ArrayList[SiderealTarget](results.asJava)
 
-  def targetAt(i: Int): GOption[SiderealTarget] = \/.fromTryCatchNonFatal(results(i)).toOption.asGeminiOpt
+  def targetAt(i: Int): GOption[SiderealTarget] = results.lift(i).asGeminiOpt
 }
 
 /**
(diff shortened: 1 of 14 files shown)


bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/CatalogSearchCriterion.scala
```diff
@@ -107,6 +107,8 @@ case class GemsCatalogSearchResults(criterion: GemsCatalogSearchCriterion, resul
   def this(results: java.util.List[SiderealTarget], criterion: GemsCatalogSearchCriterion) = this(criterion, results.asScala.toList)
 
   def resultsAsJava: java.util.List[SiderealTarget] = new java.util.ArrayList[SiderealTarget](results.asJava)
+
+  def targetAt(i: Int): GOption[SiderealTarget] = \/.fromTryCatchNonFatal(results(i)).toOption.asGeminiOpt
```
swalker2m: Might use `lift` here instead to do the same thing.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/tpe/gems/CandidateGuideStarsTableModel.java
```diff
@@ -165,12 +170,12 @@ public String getId() {
      * Returns a list of SkyObjects corresponding to the checked (or unchecked) rows in the table
      * @param checked if true, return the checked rows (candidates), otherwise the unchecked (non-candidates)
```
swalker2m: I suppose we should remove the comment there since `checked` is no longer a param.  Why did we need that before?

bundle/jsky.app.ot/src/main/java/jsky/app/ot/tpe/gems/CandidateGuideStarsTableModel.java
```diff
@@ -162,13 +166,10 @@ public String getId() {
...
             }
         };
-        skycatTable.setCatalog(new SkycatCatalog(entry));
-        return skycatTable;
     }
 
     /**
-     * Returns a list of SkyObjects corresponding to the checked (or unchecked) rows in the table
-     * @param checked if true, return the checked rows (candidates), otherwise the unchecked (non-candidates)
+     * Returns a list of SideralTargets corresponding to the checked (or unchecked) rows in the table
```
swalker2m: At this point, it looks like it's just the unchecked?

cquiroz: Right, the checked variable was always called with the false value se we could as well just remove it

bundle/jsky.app.ot/src/main/java/jsky/app/ot/tpe/gems/CandidateGuideStarsTableModel.java
```diff
@@ -137,16 +140,21 @@ public boolean isCellEditable(int row, int column) {
...
+        final String decPosition = String.valueOf(_columnNames.indexOf(DEC_TITLE));
+        final Properties props = new Properties();
+        props.setProperty(RA_COL, raPosition);
+        props.setProperty(DEC_COL, decPosition);
+        props.setProperty(SERV_TYPE, "catalog");
+        props.setProperty(LONG_NAME, "ucac4");
+        final SkycatConfigEntry entry = new SkycatConfigEntry(props);
+        return new SkycatTable(entry, CandidateGuideStarsTableModel.this.getDataVector(), getFields()) {
+            @Override
+            public Option<SiderealTarget> getSiderealTarget(int i) {
```
cquiroz: This is the key change for the PR. This method is called from the TPE to find out if a given point (translated into a query result row id) correspond to a target. By default this returns `None` making the selection fail

Most of the other changes come down from this change. In particular the order of items in `_model` is critical since we are looking up by row number and getting the order right trickles downstream

PR #1110, opened 2016-10-28
REL-2924: Parallactic angle feedback request

sraaphorst:
This is a first attempt at a partial implementation based on the meeting about this issue on Monday.

I am attempting to get back some feedback, so this should be fairly straightforward and simplistic.
Instruments now implement a `PositionAngleInstrument` subclass based on the position angle modes they support, if any at all.

Obviously, much more extensive changes will be needed, but I was hoping to get some feedback before I made massive changes.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/flamingos2/Flamingos2.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/flamingos2/Flamingos2.java
@@ -23,7 +23,6 @@ import edu.gemini.spModel.gemini.calunit.smartgcal.CalibrationKey;
 import edu.gemini.spModel.gemini.calunit.smartgcal.CalibrationKeyProvider;
 import edu.gemini.spModel.gemini.calunit.smartgcal.keys.CalibrationKeyImpl;
 import edu.gemini.spModel.gemini.calunit.smartgcal.keys.ConfigKeyFlamingos2;
-import edu.gemini.spModel.gemini.parallacticangle.ParallacticAngleSupportInst;
 import edu.gemini.spModel.guide.GuideProbe;
 import edu.gemini.spModel.guide.GuideProbeProvider;
 import edu.gemini.spModel.guide.GuideProbeUtil;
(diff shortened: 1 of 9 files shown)


sraaphorst:
So note that the Travis check will probably fail.

swalker2m:
Sebastian and I have been discussing this PR a bit today.  I had some ideas for changes but regardless of what we do, the work we discussed last week for a better typed position angle is going to be a lot to get done before the end of Friday.

I'd like to float the idea of a less aggressive (and admittedly less attractive) option.  The fundamental problem as I recall is that BAGS is trying to change the position angle while you are typing in the position angle.  Today the position angle is stored in the `SPInstObsComp` base class.  What if we just add an optional `bagsAngle` to store the angle that BAGS wants us to use.  When BAGS reads the position angle for its calculation, it reads the normal position angle.  When it calculates a new position angle it updates only the `bagsAngle`.  When anything else wants to know the observation's position angle, it is essentially `bagsAngle.getOrElse(normalPositionAngle)`.  You would still need to update the UI to show the difference between requested PA vs. actual PA and warn about overriding parallactic angle but it seems like the information needed to do so is all there.

This is more or less what the better typed solution was aiming at without having to change so much code?  I don't know, the downside is that it seems like it would be easy to overlook a necessary change and isn't what we really want going forward.

tpolecat:
How is overriding the PA indicated in this model? Is there another boolean?

swalker2m:
The `bagsAngle` would have to be driven by the existing position angle.  When you make a change to the requested position angle, it should set the `bagsAngle` to `None` at the same time.  The only override is to the parallactic angle.  That seems like it should work the same way it does today.  When the requested value differs from the computed value (and computed value + 180) it has been overridden.

It's probably going to be horrible no matter what we do.

sraaphorst:
The original idea was to calculate the parallactic angle, and if it differs
from position angle specified, we assume it has been overridden, and BAGS
never changes this.

I think this should be possible based on the UI. If the user explicitly
causes a change that would trigger a change to the parallactic angle, such
as changing the scheduling block, then we would set the position angle
field to the parallactic angle. Otherwise, we wouldn't touch it.

Does that sound about right?

tpolecat:
This sounds fine in the sense that it's probably the least disruptive change that's likely to work, at least in the short term. If OCS is really nearing end of life maybe it's not worth it to try to fix things.

sraaphorst:
Alright, the decision moving forward to simplify things entirely is that we are going to add a new position angle mode, namely "Parallactic Override."

Then, "Parallactic" will force the PI to use the calculated parallactic mode, whereas "Parallactic Override" will show the parallactic angle calculation, but will not touch the PA text field or modify the PA in any way, and whatever the user enters will be the PA that will be used.

I'm going to close this PR tomorrow morning, as it's no longer relevant, but I will leave it open until then in case anyone has any more comments regarding this.

PR #1111, opened 2016-11-01
Feature/rel 2251v2

spakzad:
REL-2251: Separate read-mode and well-depth in GNIRS ITC.


bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/ConfigExtractor.scala
```diff
@@ -123,7 +120,7 @@ object ConfigExtractor {
       slitWidth   <- extract[SlitWidth]         (c, SlitWidthKey)
       grating     <- extractDisperser
       camera      <- extractCamera
```
sraaphorst: Is there some reason we're not doing this change in extraction for other things like `camera` as well?
`camera <- extract[Camera](c, CameraKey)` ?

PR #1112, opened 2016-11-02
REL-2646: GNIRS observing wavelength in acquisitions

swalker2m:
## Disclaimer

I must apologize in advance for this PR.  It's a fairly ugly hack.  I didn't see another way of doing this because it deals with a value that is computed during sequence construction.  The computation must work one way for old executed observations and a different way for new observations.  In the new tabular sequence model, this will no longer be an issue.  At least I feel like I've documented it adequately in the code itself.

- [x] Test cases pending; PR update to come

## PR Description

The observing wavelength for GNIRS acquisition observations should be computed based on the imaging filter.  Unfortunately in the past this was not done and instead the observing wavelength was always taken from the grating central wavelength during sequence construction.  Old observations therefore used the wrong observing wavelength and we must continue to compute the value that was actually used for them so that the display matches the value that was used.  To avoid correcting the observing wavelength during sequence construction for old observed observations, we have to check a property that is set during migration.

If "override acquisition observing wavelength" is `true` (which it will be by default for new observations), then the new method of setting the observing wavelength from the filter will be used.  If not, the old method of using the grating central wavelength will be used.

### DIFF
--- a/bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/template/factory/impl/gnirs/GnirsImaging.scala
+++ b/bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/template/factory/impl/gnirs/GnirsImaging.scala
@@ -31,12 +31,13 @@ case class GnirsImaging(blueprint:SpGnirsBlueprintImaging) extends GnirsBase[SpG
   //                 H2 (2.122um) => H2: 2.12um
   //                 PAH (3.295) => PAH: 3.3um
   //             IF FILTER == PAH SET Well Depth = Deep
-  //             SET Central Wavelength according to FILTER in all GNIRS iterators
+  //             REL-2646 removes the following line:
+  //             X - SET Central Wavelength according to FILTER in all GNIRS iterators
 
   val filter = oldFilter match {
     case Filter.ORDER_5 => Filter.J // J (1.25um) => J-MK: 1.25um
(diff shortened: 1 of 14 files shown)


cquiroz:
Both the ITC and the Background images loader calculate the observing wavelength. How is this change related? Should we change those too to match? Or does this change match ITC?

tpolecat:
The code looks fine to me, but I want to quit and open a bait shop now.

👍

swalker2m:
> Both the ITC and the Background images loader calculate the observing wavelength. How is this change related? Should we change those too to match? Or does this change match ITC?

Do they calculate the observing wavelength or just read it from the sequence?

swalker2m:
I've added some test cases.  Assuming everything checks out and nobody has any major objections, I'll merge this.

cquiroz:
I refer to this particular piece of code:

https://github.com/gemini-hlsw/ocs/blob/develop/bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/ConfigExtractor.scala#L284-L298

The config extractor reads the wavelength from the model but it has some special cases for GNIRS

swalker2m:
> I refer to this particular [piece of code](https://github.com/gemini-hlsw/ocs/blob/develop/bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/ConfigExtractor.scala#L284-L298)

> The config extractor reads the wavelength from the model but it has some special cases for GNIRS

It's just papering over the issue that we can't decide what the type of the wavelength data to put in the sequence is so this change should have zero impact on that.

PR #1114, opened 2016-11-04
REL-2833: Add Phoenix support to QPT

cquiroz:
This PR enables phoenix to be usable on the QPT, see below:

screenshot 2016-11-04 13 09 59

While testing an NPE was detected when closing. The second commit adds a fix for that event

### DIFF
--- a/bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/ui/view/variant/VariantDecorator.java
+++ b/bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/ui/view/variant/VariantDecorator.java
@@ -14,44 +14,32 @@ import edu.gemini.ui.gface.GViewer;
 
 public class VariantDecorator implements GSubElementDecorator<Schedule, Variant, VariantAttribute> {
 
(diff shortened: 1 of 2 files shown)


bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/ui/view/variant/VariantDecorator.java
```diff
@@ -14,72 +14,67 @@
...
-			} else {
-				label.setIcon(ICON_VARIANT); 	
-			}
-			
-			// Alignment
-			label.setHorizontalAlignment(SwingConstants.LEFT);
-			
-		} else if (subElement == VariantAttribute.Wind) {
-
-			if (value == null) value = "-";
```
swalker2m: It seems like in a couple of places if `value` is `null` then we would reassign it to `"-"`.   The logic is a bit hard to follow there in the original code but is this no longer necessary?

cquiroz: I'm not sure why it was there in the first place but the reassignment of `value` was never used

PR #1115, opened 2016-11-04
REL-2033: fix split export

swalker2m:
This PR fixes an embarrassing stringly-typed programming issue.  Namely, when exporting an observing log split by program id, we were picking up program multiple program ids based on a `String` prefix comparison.   For example, observation `GS-2016B-Q-20-1` would be considered to be a part of program `GS-2016B-Q-2` because the program id as a `String` is a prefix of the observation id as a `String`.

### DIFF
--- a/bundle/edu.gemini.obslog/src/main/java/edu/gemini/obslog/TextExport/support/InstrumentTextSegmentExporter.java
+++ b/bundle/edu.gemini.obslog/src/main/java/edu/gemini/obslog/TextExport/support/InstrumentTextSegmentExporter.java
@@ -4,6 +4,11 @@ import edu.gemini.obslog.config.model.OlLogItem;
 import edu.gemini.obslog.obslog.ConfigMap;
 import edu.gemini.obslog.obslog.ConfigMapUtil;
 import edu.gemini.obslog.obslog.IObservingLogSegment;
+import edu.gemini.pot.sp.SPObservationID;
+import edu.gemini.shared.util.immutable.ImOption;
+import edu.gemini.shared.util.immutable.Option;
+import edu.gemini.spModel.core.SPBadIDException;
+import edu.gemini.spModel.core.SPProgramID;
 
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


bundle/edu.gemini.obslog/src/main/java/edu/gemini/obslog/TextExport/support/InstrumentTextSegmentExporter.java
```diff
@@ -58,11 +56,16 @@ private void _printComment(StringBuilder sb, ConfigMap row) {
 
     // Check to make the text a little clearer
     private boolean _oneProgramOnly() {
-        return _thisProgramOnly != null;
+        return _thisProgramOnly.isDefined();
     }
     private boolean _useThisRow(ConfigMap rowMap) {
-        String observationID = rowMap.sget(OBSERVATION_ID_PROPERTY_NAME);
-        return observationID.startsWith(_thisProgramOnly);
```
swalker2m: This is where we were bitten by the `String` prefix comparison.

bundle/edu.gemini.obslog/src/main/java/edu/gemini/obslog/TextExport/support/InstrumentTextSegmentExporter.java
```diff
@@ -14,26 +19,19 @@
...
     static protected final String OBSERVATION_ID_PROPERTY_NAME = ConfigMapUtil.OBSLOG_OBSERVATIONID_ITEM_NAME;
-    static protected final String DATASETCOMMENT_PROPERTY_NAME = ConfigMapUtil.OBSLOG_DATASETCOMMENTS_ITEM_NAME;
 
     // When set, only look for observations from  this program
-    private String _thisProgramOnly;
+    private final Option<SPProgramID> _thisProgramOnly;
 
     public InstrumentTextSegmentExporter(IObservingLogSegment segment) {
-        super(segment);
+        this(segment, null);
```
cquiroz: Could it instead be `_thisProgramOnly = None.instance()` to avoid the null?

tpolecat:
image

PR #1116, opened 2016-11-04
REL-2924: Add explicit parallactic override mode

sraaphorst:
This PR solves the issue of BAGS clashing with the parallactic angle controls and restores the functionality of allowing users to override the parallactic angle.

Prior to BAGS, we allowed users to override the parallactic angle simply by typing in the position angle text field. This was problematic with BAGS because BAGS lookups would trigger reinitialization of GUI components, which would recalculate the parallactic angle, which would repopulate the position angle text field, thus clobbering user input.

It was determined that being able to override parallactic angle is a requirement, as there are cases where the user wishes to have the parallactic angle calculation available and specify a different angle to be used.

The solution we decided to implement for the Dec 2016 release is to change the behavior of the "Parallactic angle" position angle constraint, so when it is selected, the parallactic angle must be used as the position angle: the position angle text field is rendered uneditable as can be seen here:
screen shot 2016-11-04 at 3 47 10 pm

To allow the override to be made, the user must now select a new position angle constraint from the combo box, namely "Parallactic override", which behaves precisely like the "Allow 180 flip" constraint, but displays the parallactic angle calculation:
screen shot 2016-11-04 at 3 45 55 pm

This required some minor changes to AGS to accommodate the new mode.

I also took advantage of this to migrate from the old `edu.gemini.skycalc.Angle` to the new `edu.gemini.spModel.core.Angle` class in a number of places, and switched from explicit single-method mappings and expressions to Java 8 lambda expressions.

### DIFF
--- a/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsHash.scala
+++ b/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsHash.scala
@@ -78,7 +78,7 @@ object AgsHash {
 
     // Position Angle
     Option(ctx.getPositionAngle).foreach { a =>
-      a.degrees.## +=: buf
+      a.toDegrees.## +=: buf
     }
 
     // Position Angle Constraint
--- a/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/SingleProbeStrategy.scala
+++ b/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/SingleProbeStrategy.scala
@@ -5,7 +5,6 @@ import edu.gemini.ags.api.AgsMagnitude._
 import edu.gemini.catalog.api.CatalogQuery
 import edu.gemini.catalog.votable.{ConeSearchBackend, VoTableBackend, CatalogException, VoTableClient}
 import edu.gemini.pot.ModelConverters._
-import edu.gemini.skycalc
 import edu.gemini.spModel.ags.AgsStrategyKey
 import edu.gemini.spModel.core.{Coordinates, Angle}
 import edu.gemini.spModel.core.SiderealTarget
@@ -68,8 +67,8 @@ case class SingleProbeStrategy(key: AgsStrategyKey, params: SingleProbeStrategyP
     // If we are unbounded and there are any candidates, we are guaranteed success.
     val pac   = ctx.getPosAngleConstraint(UNBOUNDED)
     val cv    = CandidateValidator(params, mt, candidates)
-    val steps = pac.steps(ctx.getPositionAngle.toNewModel, params.stepSize).toList.asScala
-    val anglesWithResults  = steps.filter { angle => cv.exists(ctx.withPositionAngle(angle.toOldModel)) }
+    val steps = pac.steps(ctx.getPositionAngle, params.stepSize).toList.asScala
+    val anglesWithResults  = steps.filter { angle => cv.exists(ctx.withPositionAngle(angle)) }
     val successProbability = anglesWithResults.size.toDouble / steps.size.toDouble
     AgsStrategy.Estimate.toEstimate(successProbability)
   }
@@ -112,7 +111,7 @@ case class SingleProbeStrategy(key: AgsStrategyKey, params: SingleProbeStrategyP
         // position angle and the Double is the percent of vignetting.
         val ctxResults = analyzed.flatMap { case (ctx0, targets) =>
 
-          feedback { f"Analyzing pos angle: ${ctx0.getPositionAngle.toDegrees.getMagnitude}%.2f degrees" }
+          feedback { f"Analyzing pos angle: ${ctx0.getPositionAngle.toDegrees}%.2f degrees" }
 
           // Filter out anything but the best quality targets.
           val qualityTargets = targets.filter { case (_,_,q) => q === quality }
@@ -142,7 +141,7 @@ case class SingleProbeStrategy(key: AgsStrategyKey, params: SingleProbeStrategyP
 
           // Finally extract the pos angle, sidereal target, vignetting, and mag
           minVig.map { case ((target,mag,_), vignetting) =>
-            (ctx0.getPositionAngle.toNewModel, target, vignetting, mag)
+            (ctx0.getPositionAngle, target, vignetting, mag)
           }
         }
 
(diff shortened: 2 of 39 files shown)


bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/SingleProbeStrategy.scala
```diff
@@ -199,7 +198,7 @@ case class SingleProbeStrategy(key: AgsStrategyKey, params: SingleProbeStrategyP
       base <- ctx.getBaseCoordinates.asScalaOpt.toList
       so <- candidates
       pa = SingleProbeStrategy.calculatePositionAngle(base.toNewModel, so)
-      ctxSo = ctx.withPositionAngle(pa.toOldModel)
```
cquiroz: I wonder if we will soon be able to retire those `toOld/NewModel` methods

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gsaoi/Gsaoi.java
```diff
@@ -850,11 +850,6 @@ public void setPosAngleConstraint(PosAngleConstraint newValue) {
     }
 
     @Override
-    public String getPosAngleConstraintDescriptorKey() {
-        return POS_ANGLE_CONSTRAINT_PROP.getName();
-    }
-
-    @Override
```
cquiroz: Why GSAOI doesn't support `PARALLCTIC_OVERRIDE`?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/parallacticangle/PositionAnglePanel.scala
```diff
@@ -54,11 +56,18 @@ class PositionAnglePanel[I <: SPInstObsComp with PosAngleConstraintAware,
       peer.setColumns(6)
       minimumSize = preferredSize
 
-      def angle: Option[Double] =
-        Try { text.toDouble }.toOption
+      // We don't want to allow things like 0.00d even though this is a legal double, so we specifically
+      // disallow any alphabetic characters.
+      def angle: Option[Double] = for {
+        t <- "^-?\\d*\\.?\\d*$".r.findFirstMatchIn(text.trim)
+        d <- Try { t.group(0).toDouble }.toOption
```
cquiroz: I guess we tend to use `\/.from...` rather than `Try`

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/SingleProbeStrategy.scala
sraaphorst: I think that that's something that should be fairly easy to do, and that we should aim towards.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/parallacticangle/PositionAnglePanel.scala
sraaphorst: Good call. I'll change this.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gsaoi/Gsaoi.java
swalker2m: Observing at the parallactic angle is only important for spectroscopy, as I understand it.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gems/Canopus.java
```diff
@@ -320,24 +306,12 @@ public boolean inRange(ObsContext ctx, Offset offset){
...
     /**
      * Returns a Shape that defines the AO port in arcsecs, centered at
      * <code>(0,0)</code>.
      */
-    public Shape shape(ObsContext ctx) {
-        double offset = getOffsetArcsec(ctx);
-        return new Ellipse2D.Double(AO_PORT.getX()+offset, AO_PORT.getY()+offset,
-                AO_PORT.getWidth(), AO_PORT.getHeight());
-    }
+    private static final Shape AO_PORT_SHAPE = new Ellipse2D.Double(AO_PORT.getX(), AO_PORT.getY(), AO_PORT.getWidth(), AO_PORT.getHeight());
```
swalker2m: I see that the offset arcsec was fixed at 0., but that kind of surprises me.  I wonder what happened to the offset of REL-1042.  It was just commented out without explanation, at least in the code.  At any rate, I suppose we can always add an offset back if necessary and it's good to get rid of unused junk so I support this change.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsHash.scala
```diff
@@ -78,7 +78,7 @@ object AgsHash {
 
     // Position Angle
     Option(ctx.getPositionAngle).foreach { a =>
-      a.degrees.## +=: buf
```
sraaphorst: Oh, this is because `ctx.getPositionAngle` now returns a new spModel core `Angle` instead of an old skycalc `Angle`, so the call to get the value in degrees changed accordingly.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/telescope/PosAngleConstraint.java
```diff
@@ -28,11 +29,6 @@ public String description() {
     /** The provided pos angle, or the pos angle plus 180 deg. */
     FIXED_180() {
         @Override
-        public ImList<Angle> steps(Angle start, Angle stepSize) {
```
sraaphorst: Ahhh... previously, the default was to run through the `stepSize` and calculate all angles, but since this behaviour only occurs in the `UNBOUNDED` case and now in three cases, the default is to use `start` and `start + 180`, I moved the previous default implementation into an overridden method in `UNBOUNDED` and made the `angle` and `angle.flip` the default behaviour.

Does that make sense? I think that explanation was a bit confusing. (Trying to fight cats from laying on my keyboard.)

PR #1117, opened 2016-11-08
REL-2855: Do not allow the BAGS feedback to override the feedback for the currently selected target.

sraaphorst:
In the previous implementation, if a target in the target environment was selected, pending BAGS feedback would clobber the feedback and only the BAGS feedback was shown.

This small modification ensures that both the feedback for the selected target and any BAGS feedback are both shown in the target editor.

I'd like to get this merged in the next 40 mins if possible to have it included in today's test release.

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/TargetFeedbackEditor.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/TargetFeedbackEditor.scala
@@ -11,8 +11,8 @@ import jsky.app.ot.OT
 import jsky.app.ot.ags.{ObsKey, BagsManager}
 import jsky.app.ot.gemini.editor.targetComponent.TargetFeedback.Row
 
+import scala.swing.GridBagPanel
 import scala.swing.GridBagPanel.Fill
-import scala.swing.{GridBagPanel, Swing}
 
 
 class TargetFeedbackEditor extends TelescopePosEditor {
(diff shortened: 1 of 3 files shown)


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/TargetFeedbackEditor.scala
```diff
@@ -21,13 +21,14 @@ class TargetFeedbackEditor extends TelescopePosEditor {
...
-      val mt = OT.getMagnitudeTable
-      ctxOpt.asScalaOpt.map(TargetGuidingFeedback.targetAnalysis(_, mt, target)).getOrElse(Nil)
-    }
-
     // Construct the rows for the table. Optionally a BAGS row, and then a list of AGS analysis rows.
     val rows = {
+      val analysisRows = for {
+        c <- ctxOpt.asScalaOpt.toList
+        a <- TargetGuidingFeedback.targetAnalysis(c, OT.getMagnitudeTable, target)
+        if !target.isTooTarget
```
cquiroz: It is a bit odd that the check is done on something not affected in the loop, or does `TargetGuidingFeedback.targetAnalysis` modify `target`?

sraaphorst: Or, I guess, rather:

val analysisRows = {
        if (target.isTooTarget) Nil
        //else ctxOpt.asScalaOpt.toList.flatMap(TargetGuidingFeedback.targetAnalysis(_, OT.getMagnitudeTable, target))
        else ctxOpt.asScalaOpt.fold(List[TargetFeedback.Row]())(TargetGuidingFeedback.targetAnalysis(_, OT.getMagnitudeTable, target))
      }`

I feel like there must be a cleaner way to write this, though.

cquiroz: maybe something like
```
!target.isTooTarget option ctxOpt.asScalaOpt.toList.flatMap(TargetGuidingFeedback.targetAnalysis(_, OT.getMagnitudeTable, target))

```

Though that will return `Option[List[_]]` (I think)

sraaphorst: Aha... I think I got it, now that you pointed me to those implicits for `Boolean`. I believe that `fold` should do the trick!

`val analysisRows = target.isTooTarget fold (Nil, ctxOpt.asScalaOpt.toList.flatMap(TargetGuidingFeedback.targetAnalysis(_, OT.getMagnitudeTable, target)))`

sraaphorst: `Boolean.fold` returns the evaluation of the first parameter if the `Boolean` is `true`, and if `false`, then the evaluation of the second parameter, which is precisely what we want.

swalker2m:
I think this addresses point 2 in [REL-2956](http://jira.gemini.edu:8080/browse/REL-2956)?

sraaphorst:
Yes, I'm quite certain that @swalker2m is correct.

PR #1119, opened 2016-11-14
REL-2655: ODB Browser XML output modification

sraaphorst:
This small PR simply makes modifications to the XML output from the ODB browser to make investigator data more organized and to fix a bug with regards to obslog output (where previously, only one obslog record was retained per observation).

The changes made to the XML are written up in detail in:
http://jira.gemini.edu:8080/browse/REL-2655

The relevant conversation begins with Bryan's message from Nov 8th.


bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
```diff
@@ -180,11 +182,14 @@ class LchQueryFunctor(queryType: LchQueryFunctor.QueryType,
...
-                        setDataset(lab.toString)
-                        setRecord(rec.comment)
+                // If there are none, no setObsLogNode should be called.
+                obs.obsLogComments.filter(_.nonEmpty).foreach { log =>
+                  setObsLogNode(new ObsLogNode() {
+                    log.toList.foreach { case (lb,rc) =>
+                      getObsLog.add(new ObsLogRecord() {
+                        setId(lb.toString)
+                        setRecord(rc.comment)
+                      })
```
swalker2m: Is the idea that they are only interested in comments?  If there are no comments they don't want dataset information?

sraaphorst:
Here is a link to the relevant comment, with the number of further comments also being essential:
http://jira.gemini.edu:8080/browse/REL-2655?focusedCommentId=32020&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-32020

I am hoping to get this merged for feedback for tonight's test release, so if someone could take a look at it some point during today, that would be very much appreciated.

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
sraaphorst: That was my impression, but good question. I'm going to double-check on this.

sraaphorst: I double-checked, and it seems that yes, the interest is only in the comments.

swalker2m: Ok then. 👍

PR #1121, opened 2016-11-16

cquiroz:
Two small requested changes for REL-2859:

* Updates to the UI to make the catalogs fit in a smaller space:

screenshot 2016-11-16 12 55 32

* Attempt to use only one Gemini image server depending on the OT configuration


cquiroz:
I added more code fixing the following bugs:

* File naming not compatible with Windows
* Fixed bug when not appropriately closing FITS files after parsing
* Improve deletion of temporary files in case of failure
* Fixed a bug when images are not swapped if the same target is present multiple times
* Redesign of the selecting the local Gemini catalog if possible, rather than attempting both catalogues in parallel

I'm not too happy about the last feature. It trades reducing traffic for reliability, I think ideally we would keep the parallel downloads but move the servers off-site

PR #1122, opened 2016-11-18
REL-2941: make GeMS AGS ignore flexure correction assignment

swalker2m:
This PR updates GeMS AGS so that it ignores the flexure star assignment, preventing ODGW stars from being added to the auto guide group.  Correspondingly, it removes any analysis or phase 2 checks related to ODGW stars for GeMS.

This is the bare minimum required for December but there is a tremendous opportunity for simplification here that we cannot pass up.  The entirety of GeMS AGS can be stripped of flexure star selection code and no longer needs to support ODGW for tip tilt correction - tip tilt will always be Canopus.  The NIR band option can be removed as well.  Since the changes are rather large, we'll leave further work for post-December.

__Note__: there are a few comments tagged with TODO: REL-2941.  Ordinarily we should not leave commented out code but here I wanted to do so since work on 2941 will continue post-release.


swalker2m:
The following suggestion from Bryan is all that is happening in this PR:

> Rodrigo pointed out that BAGS is currently causes problems since a ODGW star is chosen if possible but it is impossible to remove that from the Auto group. Therefore, we cannot send the Auto groups to the TCC and almost every asterism has to be copied into a manual group so that the ODGW star can be deleted. Is there a way to turn off the mascot ODGW selection (or at least the insertion into the Auto group) for December?

swalker2m:
I wouldn't say it's "half-way".  It's more like "1/100th way".  There is a good opportunity for simplification here. 

I don't think the PIT estimation is impacted negatively because the analysis will now just ignore missing flexure stars.  They should definitely test it though.

PR #1123, opened 2016-11-18
REL-2655: Added more fields to ODB browser and general cleanup.

sraaphorst:
This is to add some additional information to the ODB browser output, reorganize the partners XML info, and clean up a little bit.


bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryParam.scala
```diff
@@ -80,6 +80,16 @@ object InvestigatorInfo {
   }
 }
 
+sealed case class PartnerInfo(name: String, hoursAllocated: Double)
```
cquiroz: Why are these sealed? Since you can't (or shouldn't) extend case classes it seems unecessary

sraaphorst: Ahhh, good point. No reason to do this. I'll remove the sealed modifiers.

sraaphorst:
There is a test class, namely `LchQueryFunctorTest`, but this only tests that the parameters passed in for the query result in the appropriate number of results (and doesn't actually check the results themselves). Since the query parameters haven't changed, as this test class currently stands, there is nothing to change in it stemming from the addition of new information added to the output.

Maybe a new test case should be added. I'm somewhat hesitant to do so, though, as things currently stand, given that the requirements for the ODB browser changes have been modified about a half dozen times thus far and may be modified again: I don't want to spend a lot of time writing up test cases that will require detailed changes until the requirements are at least stable.

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryParam.scala
swalker2m: Since you can extend case classes, it's probably a good idea to mark them `final` though.

PR #1124, opened 2016-11-22
REL-2988: NullPointer causes persistent OT "Searching" overla

cquiroz:
To build a `GlassPane` you need a reference to the current frame. The original code would get the frame in two different threads eventually making it null.

This PR captures a reference to the frame to be used when the catalog search completes


swalker2m:
I'm actually working on this too.  The problem is that the `GlassPane` is used in a number of places including one in `EphemerisUpdater` where a similar trick was more difficult to apply.   Also, because the glass pane is shown on top of all the widgets, one would naturally think that events wouldn't be passed through to widgets beneath (which is the root cause of the NPE).  I'm updating `GlassPane` to fix that issue as well.

cquiroz:
Superseded by PR #1125

PR #1126, opened 2016-11-22
REL-2990: QPT and parallactic override mode

sraaphorst:
The QPT displays a warning when a parallactic angle mode observation is scheduled.
This was not properly accounted for when the `PARALLACTIC_OVERRIDE` mode was added.

This code cleans up the `PosAngleConstraint`s corresponding to parallactic modes by making a common method that ascertains whether a position angle constraint is parallactic, as this check must be done in multiple places, thus eliminating the need to ensure both modes are checked for.

The QPT now properly displays a warning whether an observation's position angle constraint is `PARALLACTIC_ANGLE` or `PARALLACTIC_OVERRIDE`.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -4,7 +4,7 @@ name := "ocs"
 
 organization in Global := "edu.gemini.ocs"
 
-ocsVersion in ThisBuild := OcsVersion("2016A", true, 1, 1, 1)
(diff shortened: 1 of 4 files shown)


swalker2m:
It seems odd to me that overridden still counts as using the parallactic angle, but that appears to be what Bryan wants. 👍

PR #1127, opened 2016-11-22
Feature/rel 2992 - GMOS Hamamatsu CCD ITC overestimates the peak signal + background

andrewwstephens:
The ITC calculates the peak pixel which should match what is displayed in the Signal plot.  However, for GMOS + Hamamatsu CCDs this value is frequently wrong, often telling users that their data will be saturated.  The peak value is calculated in SpecS2N.java.  The signal and background data arrays are created in SpecS2NSlitVisitor.java, and normally consist of all the data on the detector, however, for GMOS + Hamamatsu CCDs, only the pixels for one CCD at a time are defined (and the routine is looped 3 times), although the arrays are dimensioned for all three CCDs.  It appears that since these arrays are created with "clone" the peak is often in the data from the old cloned array.  I see two options:
1. Only consider the data ranges that are appropriate for each CCD when calculating the peaks.
2. Zero the arrays before using them.

It isn't obvious how to know which CCD should be considered in the peak-finding routine, so here I simply zero the unused values of the cloned arrays in SpecS2NSlitVisitor.java.


bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/operation/SpecS2NSlitVisitor.java
```diff
@@ -195,9 +198,13 @@ private VisitableSampledSpectrum signal(final double throughput) {
 
         final VisitableSampledSpectrum signal = (VisitableSampledSpectrum) sourceFlux.clone();
         final int lastPixel = lastCcdPixel(signal.getLength());
+        Log.fine("Calculating signal with " + throughput + " throughput on detector pixels " + firstCcdPixel + " - " + lastPixel);
+
+        for (int i = 0; i < firstCcdPixel; ++i) { signal.setY(i, 0); }
```
andrewwstephens: If you think that's cleaner, I'd be happy with that.  I wasn't sure which would be faster.

cquiroz: Either way should be fast, it seems cleaner though

andrewwstephens:
The latest commit consolidates the two iterators into one for simplicity.

cquiroz:
Looks good 👍

PR #1131, opened 2016-11-28
REL-2996: Parallactic angle + position angle warning fix

sraaphorst:
When a parallactic angle mode is selected and the PA text field has a value that does not equal the calculated parallactic angle or its 180 flip, a warning icon is displayed in the parallactic angle information label, and a tooltip set.

Andy found an issue where switching between observations caused this to break sometimes, where the warning would be improperly set. The need for displaying the warning needs to be re-evaluated whenever the parallactic angle information label changes or the PA text field contents change.

When the PA contents change, an event is fired that is handled, so this case is fine. However, there were instances where the parallactic angle info label contents would change and the warning display code would not be called, thus resulting possibly in a lack of warning or a warning in improper circumstances.

I now ensure that whenever the parallactic info label text changes, the warning generation / clearing method is called via a `PropertyChangeListener` on the `text` of the label. This is kind of horrible in that it is so Javaesque, but there doesn't appear to be any published events from the Scala `Label` class, so I think it is the only way to make sure this is handled.

This should be preferable to having to call the warning text from anywhere else, as it will void the need to do so should the code change.

Also, there was an issue where if the parallactic angle could not be calculated, the warning would stay the same due to the use of the `for` construct. I have fixed this as well to clear the warning if the parallactic angle cannot be calculated.

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/parallacticangle/ParallacticAngleControls.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/parallacticangle/ParallacticAngleControls.scala
@@ -145,22 +145,24 @@ class ParallacticAngleControls(isPaUi: Boolean) extends GridBagPanel with Publis
         */
       def warningState(): Unit =
         for {
-          e     <- editor
-          fmt   <- formatter
-          inst  <- Option(e.getContextInstrumentDataObject).collect { case s: SPInstObsComp => s }
+          e    <- editor
+          fmt  <- formatter
+          inst <- Option(e.getContextInstrumentDataObject).collect { case s: SPInstObsComp => s }
         } warningStateFromPAString(inst.getPosAngleDegrees.toString)
 
-      def warningStateFromPAString(paStr: String): Unit =
+      /**
+        * Compares the parallactic angle to the supplied position angle.
+        */
+      def warningStateFromPAString(paStr: String): Unit = Swing.onEDT {
         for {
-          e     <- editor
-          fmt   <- formatter
-          inst  <- Option(e.getContextInstrumentDataObject).collect { case s: SPInstObsComp => s }
+          e   <- editor
+          fmt <- formatter
         } {
           // We warn only if the parallactic angle exists and has been explicitly set.
           val warningFlag = parallacticAngle.exists { angle =>
-              val fa = fmt.format(angle.toDegrees)
-              val faFlip = fmt.format(angle.flip.toDegrees)
-              !fa.equals(paStr) && !faFlip.equals(paStr)
+            val fa = fmt.format(angle.toDegrees)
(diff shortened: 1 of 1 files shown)


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/parallacticangle/ParallacticAngleControls.scala
```diff
@@ -142,32 +143,42 @@ class ParallacticAngleControls(isPaUi: Boolean) extends GridBagPanel with Publis
...
-          if (explicitlySet) {
+          // We warn only if the parallactic angle exists and has been explicitly set.
+          val warningFlag = parallacticAngle.exists { angle =>
+              val fa = fmt.format(angle.toDegrees)
+              val faFlip = fmt.format(angle.flip.toDegrees)
+              !fa.equals(paStr) && !faFlip.equals(paStr)
+          }
+
+          if (warningFlag) {
             icon = Resources.getIcon("eclipse/alert.gif")
```
cquiroz: These calls should definitely be done on the EDT. How do you know this is called from inside swing

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/parallacticangle/ParallacticAngleControls.scala
```diff
@@ -142,32 +143,42 @@ class ParallacticAngleControls(isPaUi: Boolean) extends GridBagPanel with Publis
...
           } else {
             icon = Resources.getIcon("eclipse/blank.gif")
             tooltip = ""
           }
         }
-      }
+
+      // This is kind of horrible, but we want to recalculate the warning state whenever the text of the label
+      // changes, so this is the most robust way to ensure that this will happen no matter where the label changes.
+      peer.addPropertyChangeListener("text", new PropertyChangeListener {
```
cquiroz: We discussed this a bit and unfortunately seems the only way to go

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/parallacticangle/ParallacticAngleControls.scala
sraaphorst: 😸  After last week, I think I am still having deep fears about the `Swing.onEDT`. I'm pretty sure these are all invoked through methods scheduled on the EDT, but I will try only doing them on the EDT and see what happens. I'm a bit hesitant about not immediately doing the operation, since it can involve the contents of the PA text field, and again, we have the issue of this updating from multiple sources.

I'll play around and dig around a bit and see.

sraaphorst:
@cquiroz From what I can tell based on my testing and trying to replicate the problem as per Andy's description, invoking the code on the Swing EDT as per your comment seems fine, so I pushed another commit with this modification.

PR #1133, opened 2016-11-28
REL-2998: check for context observation before updating

swalker2m:
A tiny update to make sure that the context observation is still available when BAGS feedback arrives.

### DIFF
--- a/bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
+++ b/bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
@@ -105,9 +105,10 @@ public final class EdCompTargetList extends OtItemEditor<ISPObsComponent, Target
         _agsPub.subscribe((obs, oldOptions, newOptions) -> updateGuiding());
 
         BagsManager.addBagsStateListener((key, oldStatus, newStatus) -> {
-            if (key.equals(getContextObservation().getNodeKey()))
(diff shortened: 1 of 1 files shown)


bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
```diff
@@ -105,9 +105,10 @@ public void textBoxKeyPress(final TextBoxWidget tbwe) {
         _agsPub.subscribe((obs, oldOptions, newOptions) -> updateGuiding());
 
         BagsManager.addBagsStateListener((key, oldStatus, newStatus) -> {
-            if (key.equals(getContextObservation().getNodeKey()))
+            if (Optional.ofNullable(getContextObservation()).map(co -> key.equals(co.getNodeKey())).orElse(false)) {
```
sraaphorst: This has been something I've been wondering about for awhile: do we want to make a decision moving forward regarding consistency about using our own `Option` or Java 8's `Optional`?

sraaphorst: (I ask partly because it seems like the lack of a good `exists` makes for uglier code. For example, this could have been `ImOption.apply(getContextObservation()).exists(co -> key.equals(co.getNodeKey())`.

tpolecat: I don't have a strong opinion on this. Our stuff is probably better than Java's so I'm not inclined to get rid of it, especially if we're phasing out OCS.

PR #1134, opened 2016-11-28
REL-2992: GMOS Hamamatsu CCD ITC overestimates the peak signal + background 

cquiroz:
The OT ITC spectroscopy table was ignoring the fact that some instruments have multiple CCDs. This PR fixes that taking the maximum of the CCD values as requested  and improves the safety of the calculations

### DIFF
--- a/bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/ItcService.scala
+++ b/bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/ItcService.scala
@@ -44,6 +44,11 @@ sealed trait ItcResult extends Serializable {
 
   def ccd(i: Int = 0): ItcCcd            = ccds(i % ccds.length)
   def peakPixelFlux(ccdIx: Int = 0): Int = ccd(ccdIx).peakPixelFlux.toInt
+  def maxPeakPixelFlux: Int              = ccds.map(_.peakPixelFlux).max.toInt
+  def maxAdu: Int                        = ccds.map(_.adu).max
+  def maxPercentFullWell: Double         = ccds.map(_.percentFullWell).max
+  def maxSingleSNRatio: Double           = ccds.map(_.singleSNRatio).max
+  def maxTotalSNRatio: Double            = ccds.map(_.totalSNRatio).max
 
   val warnings: List[ItcWarning] = {
     def concatWarnings =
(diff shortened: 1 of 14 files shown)


bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/ItcService.scala
```diff
@@ -35,50 +35,54 @@ final case class ItcCcd(
...
 sealed trait ItcResult extends Serializable {
-  def ccds:                         List[ItcCcd]
+  def ccds: List[ItcCcd]
 
-  def ccd(i: Int = 0)               = ccds(i % ccds.length)
-  def peakPixelFlux(ccdIx: Int = 0) = ccd(ccdIx).peakPixelFlux.toInt
+  /**
+    * Max value for a property across the CCDs
+    */
+  private def maxP[A](f: ItcCcd => A)(implicit a: scala.Ordering[A]): A = ccds.map(f).max
```
swalker2m: Could also be: `ccds.maxBy(f)`.

bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/ItcService.scala
```diff
@@ -35,50 +35,58 @@ final case class ItcCcd(
...
 
-
 // === SPECTROSCOPY RESULTS
 
 // There are two different types of charts
 sealed trait SpcChartType
-case object SignalChart       extends SpcChartType { val instance = this } // signal and background over wavelength [nm]
-case object S2NChart          extends SpcChartType { val instance = this } // single and final S2N over wavelength [nm]
-case object SignalPixelChart  extends SpcChartType { val instance = this } // single and final S2N over wavelength [nm]
+case object SignalChart       extends SpcChartType { val instance: SpcChartType = this } // signal and background over wavelength [nm]
```
swalker2m: The `instance` methods are done for the benefit of Java access, I'm guessing?

bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/ItcService.scala
cquiroz: doesn't `mabBy` return the maximum `ItcCcd` based on some property? I need the max value of the property `A`

swalker2m: Ah, right.  It'd have to be `f(ccds.maxBy(f))`.  Saves a pass though the `List`, I guess, but probably not worth it.

swalker2m: We know for sure that the list isn't empty, by the way?

cquiroz: This is a pain point on this. The list could be empty, but not in any meaningful use case. I tried using `NonEmptyList` but I hit a wall being non serializable and it lead to a bunch of broken tests. I'd like to make it use `NEL` but it is probably not worth the pain inflicted at this moment

PR #1137, opened 2016-11-29
REL-2655: Fixing ODB Browser remaining time

sraaphorst:
During testing, Andy noticed that the remaining time was not as he expected, and I discovered that I was not calculating it correctly.

This PR fixes the issue by rectifying the calculation.

Since this calculation is also used in `EdProgram`, I moved the calculation into `ObsTimesService` as per Shane's recommendation.

Since the `EdProgram` code was rather muddled and messy, I cleaned up this class considerably as well. I did, however, leave the `try - catch` blocks in place, even though I don't think some of them are ever invoked (unless possibly with a `NullPointerException`) because I didn't want to risk fundamentally altering the behaviour of this class, especially so close to the December release.


swalker2m:
This looks good. 👍

PR #1139, opened 2016-12-01
REL-2997: Fix PA issue with BAGS

sraaphorst:
There was a PA issue with BAGS for OI guide probes where if:

1. the PA was set to `x` and a 180 flip mode was available;
2. BAGS found the best guide star at the flip; and
3. The user then reset, either via dragging in the TPE or pasting in the instrument editor, the PA to `x`;

then BAGS would not rerun for the PA of `x`, and thus the guide star at the 180 flip of `x` was maintained but registered as inaccessible to the patrol field.

This is because BAGS maintained, for the observation, the old `AgsHash` value that it previously used to do the lookup, and noted that it had already last performed the lookup for this PA and determined that it did not need to run.

This modification passes around the `ObsContext` to check if the position angle has changed, and if it has, then the hash is recalculated.

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
@@ -129,7 +129,7 @@ object BagsState {
           (c, s) <- tup
           h      <- newHash
           if !hash.contains(h) && notObserved
-        } yield (RunningState(k, h), BagsManager.triggerAgsAction(k, c, s)): StateTransition
+        } yield (RunningState(k, c, h), BagsManager.triggerAgsAction(k, c, s)): StateTransition
 
         // Returns the new state to switch to, either RunningState if all is well
         // and a new AGS lookup is needed or else IdleState (possibly with an
@@ -147,7 +147,7 @@ object BagsState {
     * since the search began since we will transition to RunningEditedState
     * if something is edited in the meantime.
     */
-  case class RunningState(k: ObsKey, hash: AgsHashVal) extends BagsState {
+  case class RunningState(k: ObsKey, ctx: ObsContext, hash: AgsHashVal) extends BagsState {
     // When edited while running, we continue running but remember we were
     // edited by moving to RunningEditedState.  When the results eventually
     // come back from the AGS lookup when in RunningEditedState, we store them
@@ -157,8 +157,13 @@ object BagsState {
 
     // Successful AGS lookup while running (and not edited).  Apply the update
     // and move to IdleState.
-    private[ags] override def succeed(results: Option[AgsStrategy.Selection]): StateTransition =
-      (IdleState(k, Some(hash)), BagsManager.applyAction(k, results))
+    private[ags] override def succeed(results: Option[AgsStrategy.Selection]): StateTransition = {
+      val h = results.fold(hash) { r =>
+        if (r.posAngle === ctx.getPositionAngle) hash
+        else hashObs(ctx.withPositionAngle(r.posAngle))
+      }
+      (IdleState(k, Some(h)), BagsManager.applyAction(k, results))
+    }
 
     // Failed AGS lookup.  Move to FailedState for a while and ask the timer to
     // wake us up in a bit.  When the timer eventually goes off we will switch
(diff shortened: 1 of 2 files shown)


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -378,7 +383,7 @@ object BagsManager {
       val k = ObsKey(o)
       val s = m.lookup(k).fold(IdleState(k, None): BagsState) {
         case IdleState(_, Some(hash)) => IdleState(k, hashObs(o).filter(_ === hash))
-        case RunningState(_, h)       => if (hashObs(o).contains(h)) RunningState(k, h) else RunningEditedState(k)
+        case RunningState(_, c, h)    => if (hashObs(o).contains(h)) RunningState(k, c, h) else RunningEditedState(k)
```
cquiroz: Should you move the if  into the pattern match?

sraaphorst: I guess we could do two `case` statements instead, I would think.

I'm not sure it is clearer? I suspected myself that:
`case a if cond => x`
`case a => y`
would result in `y` if `cond` was `false`, but I wasn't 100% convinced until I checked to be sure.

I think maybe:
`case a => if cond x else y`
is clearer?

cquiroz: I guess it is just a matter of taste. It is up to you

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -157,8 +157,13 @@ object BagsState {
 
     // Successful AGS lookup while running (and not edited).  Apply the update
     // and move to IdleState.
-    private[ags] override def succeed(results: Option[AgsStrategy.Selection]): StateTransition =
-      (IdleState(k, Some(hash)), BagsManager.applyAction(k, results))
+    private[ags] override def succeed(results: Option[AgsStrategy.Selection]): StateTransition = {
+      val h = results.fold(hash) { r =>
+        if (r.posAngle === ctx.getPositionAngle) hash
```
swalker2m: I worried that an exact comparison would be a bit flakey since BAGS computes the position angle.  Seems like it would be worth it to compare within a small delta?

sraaphorst: Thinking...
Yes, this is probably right. I'm going to add this. It's not too late to make it into the current build.

sraaphorst: Unfortunately, I don't think we have access to the `AlmostEqual` code here 👎 , so we will probably have to implement this ourselves. Seems annoying, as we have similar code in `PositionAnglePanel`, I believe.

Maybe this should live somewhere else?

sraaphorst: Sorry... in the `ParallacticAngleControls.didParallacticAngleChange` method.

sraaphorst: Oh, I see now that `AlmostEqual` is limited to the `edu.gemini.spModel.core` test files.
It seems we should have an operation like this on `Angle`, seeing as there are multiple cases where it is necessary. Does anyone know if anything like this exists? I can't find it, and don't want to reimplement, but it might be worthwhile to have an operation defined in `Angle` that compares to a certain set precision, say `0.005`?

tpolecat: I don't see any reason in principle why we couldn't move `AlmostEqual` out of `test`. @swalker2m?

tpolecat: I guess it's kind of arbitrary. We call all doubles almost equal if they differ by less than `0.00001` which may not always be good enough, depending on scale. I would want to remove that instance and specialize the instances for data types that rely on it.

swalker2m: I'm fine with promoting it if it seems useful.  My hesitation is that it always depends on context when two things are considered close enough.  Not even `Angle` has an `AlmostEqual` instance that works in every use case.  Maybe nothing does really?

At any rate, it's a one-liner  (`(a.toDegrees - b.toDegrees).abs < 0.0001` or something) so we could just type it out here.

sraaphorst: Agreed it's very trivial. It's just annoying to have these `Angle` comparisons in two different places, i.e. `BagsManager` and `ParallacticAngleControls`. If for any reason, we want to adjust the precision, then we will need to change in multiple places.

As this requires more thought, for now, I suppose we can just take the easy way out of doing this one liner, which I'll implement and get out this morning, but if we can think of enough cases where an `AlmostEqual` operation would be useful, this would be a nice thing to have.

PR #1141, opened 2016-12-02
REL-2997: Use small delta when making checks to PA changes.

sraaphorst:
This tiny PR simply addresses @swalker2m's point about doing `Angle` comparisons in BAGS.

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
@@ -156,10 +156,10 @@ object BagsState {
       (RunningEditedState(k), ioUnit)
 
     // Successful AGS lookup while running (and not edited).  Apply the update
-    // and move to IdleState.
+    // and move to IdleState. If the PA changes, update the hash.
     private[ags] override def succeed(results: Option[AgsStrategy.Selection]): StateTransition = {
       val h = results.fold(hash) { r =>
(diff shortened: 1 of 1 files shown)


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -156,10 +156,10 @@ object BagsState {
       (RunningEditedState(k), ioUnit)
 
     // Successful AGS lookup while running (and not edited).  Apply the update
-    // and move to IdleState.
+    // and move to IdleState. If the PA changes, update the hash.
     private[ags] override def succeed(results: Option[AgsStrategy.Selection]): StateTransition = {
       val h = results.fold(hash) { r =>
-        if (r.posAngle === ctx.getPositionAngle) hash
+        if ((r.posAngle.toDegrees - ctx.getPositionAngle.toDegrees).abs < 1e-4) hash
```
swalker2m: We can probably require it to be even closer.  This is the kind of problem we might run into:

````
scala> val x = 35.77
x: Double = 35.77

scala> x + 180.0 - 180.0
res0: Double = 35.77000000000001
````

So even `1e-10` seems fine?

swalker2m:
We discussed this more and I realized I was wrong and that Sebastian's original version with the exact comparison was correct.  We want the hash value to match the observation and that will only happen if it is exactly the same position angle as we get in the AGS results.

PR #1142, opened 2016-12-02
REL-2956: Reserve space for BAGS feedback message.

sraaphorst:
The remaining issue in REL-2956 was a UI issue where the BAGS feedback message would appear and disappear, thus making the other components in the target editor "bounce" up and down to accommodate its UI space requirements.

This simple PR reserves the appropriate amount of room for BAGS feedback so that this is no longer an issue. Nothing is displayed if BAGS is not running.

Right now, the BAGS feedback appears at the bottom of the window, and the guide star feedback above as per the attached picture. (Note the small empty space under the Guide Speed: FAST message.) I think this looks nicer than the alternative and makes sure that it is clear than having empty space between the target editor and the feedback, but that is a matter of personal tastes. If anyone has an argument to reverse these, this can easily be done.

screen shot 2016-12-02 at 9 34 31 am


sraaphorst:
Incidentally, here is the alternative, with space for BAGS feedback reserved above the guide star quality row.
screen shot 2016-12-02 at 10 03 01 am

swalker2m:
Slight preference for version 2.

PR #1145, opened 2017-01-10
REL-2881 WIP

tpolecat:
WIP fixing the sync/permissions issue. There's a race condition that will cause the same service instance to be re-used for simultaneous requests which has the effect of running all of them with the same set of permissions.

```
[info] Secure Service Factory getSecureService (singular) should
[info]   + work fine with normal service registrations
[info]   + pass principals correctly via getSecureServices
[info]   + ignore principals when called normally
[error]   x return a new service instance each time requested
[error]    '(foo,Set(PImpl(Bob), PImpl(Steve)))' is not equal to '(foo,Set(PImpl(Sam), PImpl(Harry)))' (SecureServiceFactorySpec.scala:100)
[error] Actual:   ...Impl([Bob]),...l([Steve])))
[error] Expected: ...Impl([Sam]),...l([Harry])))
```

Not clear how to fix it yet but at least it seems to be isolated.

### DIFF
--- a/bundle/edu.gemini.util.osgi/build.sbt
+++ b/bundle/edu.gemini.util.osgi/build.sbt
@@ -16,6 +16,9 @@ unmanagedJars in Test ++= Seq(
   new File(baseDirectory.value, "../../bundle/edu.gemini.osgi.main/lib/org.apache.felix-4.2.1.jar")
 )
 
+libraryDependencies ++= Seq(
+  "org.scalaz" %% "scalaz-core" % ScalaZVersion)
+
 osgiSettings
 
 ocsBundleSettings
@@ -24,8 +27,7 @@ OsgiKeys.bundleActivator := None
 
 OsgiKeys.bundleSymbolicName := name.value
 
-OsgiKeys.dynamicImportPackage := Seq("")
+OsgiKeys.dynamicImportPackage := Seq("*")
 
 OsgiKeys.exportPackage := Seq(
   "edu.gemini.util.osgi")
-
--- a/bundle/edu.gemini.util.osgi/src/main/scala/edu/gemini/util/osgi/SecureServiceFactory.scala
+++ b/bundle/edu.gemini.util.osgi/src/main/scala/edu/gemini/util/osgi/SecureServiceFactory.scala
@@ -1,52 +1,144 @@
 package edu.gemini.util.osgi
 
 import org.osgi.framework._
+import org.osgi.framework.FrameworkUtil.createFilter
 import java.security.Principal
+import java.util.{ Set => JSet, Hashtable => JHashtable, Dictionary => JDictionary }
 import scala.util.DynamicVariable
 import collection.JavaConverters._
+import scalaz._, Scalaz._
+import scala.reflect.ClassTag
+
+
+trait SecureServiceFactory[A] {
+
+  def getService(ps: Set[Principal]): A
+
+}
+
+abstract class SecureServiceFactoryForJava[A] extends SecureServiceFactory[A] {
+
+  final def getService(ps: Set[Principal]): A =
+    getService(ps.asJava)
+
+  def getService(ps: JSet[Principal]): A
+
+}
 
 object SecureServiceFactory {
 
-  private val principals: DynamicVariable[Set[Principal]] = 
-    new DynamicVariable(Set())
+  /**
+   * When we register a `SecureService[A]` with the container, the registration will have
+   * `ServiceProp` bound to the runtime classname of `A`, which is what we use for lookups.
+   */
+  val ServiceProp = "service"
 
-  /**  Equivalent to getServiceReferences followed by getService. */
-  def getSecureServices(context: BundleContext, clazz: String, filter: String, ps: Set[Principal]): List[(ServiceReference[_], Any)] = 
-    principals.withValue(ps) {
-      Option(context.getServiceReferences(clazz, filter)).map(_.toList.map { ref => 
-        (ref, context.getService(ref))
-      }).getOrElse(Nil)
+  /** Filter is a semigroup under conjunction. */
+  implicit val SemigroupFilter: Semigroup[Filter] =
+    Semigroup.instance((a, b) => createFilter(s"(&$a$b)"))
+
+  /** Construct an equality filter. */
+  def filter(key: String, value: String = "*"): Filter =
+    createFilter(s"($key=$value)")
+
+  /** Construct an filter binding `ServiceProp` to the name of the given class. */
+  def serviceFilter[A](clazz: Class[A]): Filter =
+    filter(ServiceProp, clazz.getName)
+
+  /**
+   * Look up a service factory for the given clazz with any additional filter conditions, create an
+   * instance with the given set of principals, evaluate the given function and return the result,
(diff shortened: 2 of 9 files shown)


cquiroz:
This looks fine to me. Would you make the test pending or ignored to get the build passing?

tpolecat:
> Would you make the test pending or ignored to get the build passing?

I'll add the fix to this PR once I come up with one.

cquiroz:
👍

bundle/edu.gemini.util.osgi/src/main/scala/edu/gemini/util/osgi/SecureServiceFactory.scala
```diff
@@ -1,52 +1,144 @@
...
+    ps:      Set[Principal],
+    ofilter: Option[Filter]
+  )(f: A => B): Option[B] = {
+    val fcName = classOf[SecureServiceFactory[A]].getName
+    val filter = ofilter.foldLeft(serviceFilter(clazz))(_ |+| _).toString
+    val orefs  = Option(context.getServiceReferences(fcName,filter))
+    val oref   = orefs.flatMap(_.headOption)
+    oref.map { ref =>
+      val svc = context.getService(ref).asInstanceOf[SecureServiceFactory[A]]
+      try f(svc.getService(ps)) finally context.ungetService(ref)
```
swalker2m: As long as `getService` returns a new service each time we're good now?

bundle/edu.gemini.util.osgi/src/main/scala/edu/gemini/util/osgi/SecureServiceFactory.scala
```diff
@@ -1,52 +1,144 @@
...
+   * service properties, such that instances of `A` can be constructed via `withSecureService`
+   * above.
+   */
+  def registerSecureService[A](
+    context: BundleContext,
+    factory: SecureServiceFactory[A],
+    clazz:   Class[A],
+    props:   Map[String, String]
+  ): ServiceRegistration[SecureServiceFactory[A]] = {
+    val propsʹ = new JHashtable[String, Object]
```
sraaphorst: You're talking about the prime, right? I was impressed to see that, too.

tpolecat:
Hi, I did't expect you guys to look at this before I had a chance to write it up. Here's the deal:

- For secure services we need a *new* instance constructed with client principals for each request. I had thought a `ServiceFactory` would do that, but I was wrong. A `ServiceFactory` ensures that each *bundle* has no more than one instance, which doesn't work because we're multi-threaded.
- So instead of registering a `ServiceFactory[Foo]` I propose that we now register a `SecureServiceFactory[Foo]`, which by erasure becomes a `SecureServiceFactory[_]` and the service type is specified by an extra service property. The static methods on `SecureServiceFactory` handle registration and lookup.
- The `TrpcServlet` now first tries to look up an non-secure service by class name, and if it's not found it tries to look up a `SecureServiceFactory` whose service attribute matches the given class name.

bundle/edu.gemini.util.osgi/src/main/scala/edu/gemini/util/osgi/SecureServiceFactory.scala
tpolecat: That's the idea, yes.

bundle/edu.gemini.util.osgi/src/main/scala/edu/gemini/util/osgi/SecureServiceFactory.scala
tpolecat: Turns out it has been in Scala forever but everyone had been trying to use the wrong one. The U+2032 PRIME doesn't work but U+02B9 MODIFIER LETTER PRIME does work.

cquiroz: Is this rpm safe? We had an incident with a greek N on the last release

bundle/edu.gemini.util.osgi/build.sbt
```diff
@@ -24,8 +27,7 @@ OsgiKeys.bundleActivator := None
 
 OsgiKeys.bundleSymbolicName := name.value
 
-OsgiKeys.dynamicImportPackage := Seq("")
+OsgiKeys.dynamicImportPackage := Seq("*")
```
tpolecat: We need this because service lookup is now done in a static method on `SecureServiceFactory`.

bundle/jsky.app.ot.testlauncher/src/main/scala/jsky/app/ot/testlauncher/ServiceChecker.scala
```diff
@@ -0,0 +1,42 @@
...
+import edu.gemini.auxfile.server.AuxFileServer
+import edu.gemini.itc.shared.ItcService
+import edu.gemini.pot.spdb.IDBQueryRunner
+import edu.gemini.services.client.TelescopeScheduleService
+import edu.gemini.spModel.core.{Peer, Site}
+import edu.gemini.too.event.api.TooService
+import edu.gemini.util.security.auth.keychain.KeyService
+import edu.gemini.util.trpc.client.TrpcClient
+
+object ServiceChecker extends App {
```
tpolecat: This is just a sanity check that you can run from IDEA to make sure services are responding.

bundle/edu.gemini.util.osgi/src/main/scala/edu/gemini/util/osgi/SecureServiceFactory.scala
tpolecat: I have no idea, what happened with greek N?

swalker2m: Presumably if it doesn't appear as a filename (for a class file) it should be okay.

cquiroz: By accident there was a class that used the greek N in the name instead of the regular N. That is valid scala code and everything worked fine on spdb and dowloaded OT but when building the operational rpms for the OT (In the new build box) something, likely an encoding bug/feature produced an rpm that didn't work because of that.

Now this is a local variable so it should be less of a problem since the issue above exploded when the class was imported via the classloader

cquiroz: The other greek vals in different parts of the code work fine

tpolecat: Ah, ok. Yeah chars in classnames need to make sense to the filesystem. I think this shouldn't be an issue. We have lots of extended chars in source code elsewhere in OCS.

tpolecat:
So, due to this changing a pretty basic piece of machinery we need to be sure testing hits all the services and tests permissions and stuff. Do the scientists have a script for this?

tpolecat:
Ok I was finally able to figure out how to test TOO with help from @swalker2m and it works so I feel ok merging this.

PR #1147, opened 2017-01-16
REL-3034: Added K-blue and K-red filters for F2.

sraaphorst:


cquiroz:
This looks good but there are some "side-effects"

* The name of the filters should be on the spModel side too, at
https://github.com/tpolecat/ocs/blob/e19b95555dab351183a99ae08f3f0a2a31ee231d/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/flamingos2/Flamingos2.java#L128-L128
* This change must be tested in ITAC, could you add  a task to remind us about this
* It would be good to have the 1-1 correspondence of filters tested like ;at
https://github.com/tpolecat/ocs/blob/3c42e02ac47d719a754d99c08576e92a4b114126/bundle/edu.gemini.phase2.skeleton.servlet/src/test/scala/edu/gemini/phase2/skeleton/factory/SpBlueprintFactoryTest.scala#L19-L19

sraaphorst:
The ITAC task has been added as requested.

There seems to be a separate task for the new K filters for the OT, i.e. REL-3031, which is why I didn't initially include those changes here, but some checks have failed and I'm looking into what caused this, and maybe this is why.

I'm guessing these should be done in tandem?

cquiroz:
Right, if they are pending on the OT they cannot be tested. @swalker2m Should we do the OT side of this right away?

sraaphorst:
Whooops... no, the error was caused by me neglecting to make a required change to a spec file. Ooops. That's done now.

I'm going to wait to check this code in since I've started making the OT changes and don't want to finish those until I get confirmation from @swalker2m about completing them.

PR #1149, opened 2017-01-16
REL-3034: Updating F2 blueprint mapping.

sraaphorst:


cquiroz:
Do you think you can add a test?  Blueprints are not very easy to tests but there are examples

PR #1150, opened 2017-01-17
REL-3039: Unhide the F2 MOS option.

sraaphorst:
Previously, the F2 MOS option was always filtered out since it was unavailable.

It is hoping to be made available in 2017B, so this simply undoes the code that filtered out this option from appearing in the resource configuration for F2.

### DIFF
--- a/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/dtree/inst/Flamingos2.scala
+++ b/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/dtree/inst/Flamingos2.scala
@@ -34,7 +34,7 @@ object Flamingos2 {
   class ModeNode extends SingleSelectNode[Unit, F2Mode, Unit](()) {
     val title       = "Select Instrument Mode"
     val description = "Flamingos2 can be used for both imaging and spectroscopy."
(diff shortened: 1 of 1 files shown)


bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/dtree/inst/Flamingos2.scala
```diff
@@ -34,7 +34,7 @@ object Flamingos2 {
   class ModeNode extends SingleSelectNode[Unit, F2Mode, Unit](()) {
     val title       = "Select Instrument Mode"
     val description = "Flamingos2 can be used for both imaging and spectroscopy."
-    def choices     = F2Mode.values.filter(_ != F2Mode.Mos).toList // Hide temporarily the MOS option REL-1355
```
tpolecat: `filterNot`?

swalker2m:
Is everything on the phase2/execution side of this ready or is there more work to do?

sraaphorst:
I believe that everything is done on the phase 2 side of things, as far as I can tell.

It seems that MOS is being treated the same as Imaging and Long-slit Spectroscopy in the OT code, so I suspect that there isn't anything else that is required. I'll ask to be certain on the JIRA task, though.

PR #1148, opened 2017-01-16
OCS3 XML fetch servlet

swalker2m:
This PR adds a servlet named "ocs3" to the ODB.  Its job is to return an XML representation of an observation or program in a format that is suitable for ingestion into the OCS3 model and database.  In particular, it replaces the observation sequence hierarchy with the sequence XML returned by the WDBA and renames the "container" nodes with their types.  You use this servlet like:

    http://hostname:8442/ocs3/GS-2017A-Q-1

or

    http://hostname:8442/ocs3/GS-2017A-Q-1-2

An example of the XML returned is

````xml

<observation name="GS-2017A-Q-1-2">
  <data>
    <param name="title" value="Flamingos2 Observation"/>
    <param name="libraryId" value=""/>
    <param name="priority" value="LOW"/>
    <param name="tooOverrideRapid" value="false"/>
    <param name="phase2Status" value="PI_TO_COMPLETE"/>
    <paramset name="schedulingBlock">
      <param name="start" value="1493703000000"/>
      <paramset name="duration">
        <param name="tag" value="unstated"/>
      </paramset>
    </paramset>
    <param name="qaState" value="UNDEFINED"/>
    <param name="overrideQaState" value="false"/>
  </data>
  <scheduling name="Observing Conditions">
    <data>
      <param name="CloudCover" value="ANY"/>
      <param name="ImageQuality" value="ANY"/>
      <param name="SkyBackground" value="ANY"/>
      <param name="WaterVapor" value="ANY"/>
      <param name="ElevationConstraintType" value="NONE"/>
      <param name="ElevationConstraintMin" value="0.0"/>
      <param name="ElevationConstraintMax" value="0.0"/>
      <paramset name="timing-window-list"/>
    </data>
  </scheduling>
  <telescope name="Targets">
    <data>
      <paramset name="targetEnv">
        <paramset name="base">
          <paramset name="target">
            <param name="name" value="Untitled"/>
            <paramset name="coordinates">
              <param name="ra" value="0.0"/>
              <param name="dec" value="0.0"/>
            </paramset>
            <param name="tag" value="sidereal"/>
          </paramset>
        </paramset>
        <paramset name="guideEnv">
          <param name="primary" value="0"/>
          <paramset name="guideGroup">
            <param name="tag" value="AutoInitialTag"/>
          </paramset>
        </paramset>
      </paramset>
    </data>
  </telescope>
  <instrument name="Flamingos2" type="Flamingos2">
    <data>
[...]

PR #1151, opened 2017-01-19
OCS3 Export Updates

swalker2m:
This PR builds upon #1148:

* Adds an `exportOcs3` OSGi shell command that works just like `exportXML` except that it uses the OCS3 XML format.  This will make it possible to import from files instead of always having to have an ODB running.

* Adds a program listing option to the export servlet.  Since the servlet exports one program at a time, this allows a client to discover the available programs and request them one by one.

* Fixes a few issues with the OCS3 XML generation.


tpolecat:
Looks good. Did you build a PIO parser in Gem or are you just reading the matrix directly?

swalker2m:
> Did you build a PIO parser in Gem or are you just reading the matrix directly?

I'm working on it now, trying to figure out something reasonable.  it mixes in the sequence XML directly (which isn't PIO-based).  I also mapped all the PIO `container` elements to the name of what they actually contain like `program`, `note`, `observation`, etc.  They still wrap `paramset`s and `param`s though.

PR #1154, opened 2017-01-31
REL-2985: PIT calculation of observing and calibration overheads

sraaphorst:
This is a substantial change, as described in parent task REL-2925:
http://jira.gemini.edu:8080/browse/REL-2925

Previously, the user would calculate the program time and enter it into the PIT, and this was included for each observation:
`<observation...><time unit="hrs">1.0</time>...</observation>`
for example.

REL-2985 requires the following:

1. Instead of accepting calculated program time, the PIT must now accept integration time, look up overheads from a table (see the `Overheads` class) based on the instrument configuration, and manually perform the calculations to obtain program time, partner time, and total time.

2. When importing old proposals, the existing time must be assumed to be program time, and the calculations reversed as per the instructions on REL-2985 to calculate integration time, as this is the parameter that determines the other time values along with the instrument configuration.

The instrument configuration parameters can be found here:
http://jira.gemini.edu:8080/secure/attachment/13761/nightpartnercal_overheads.xlsx

This PR includes the majority of the functionality from the PIT side of things (conversion from older proposals, all calculations, schema changes, and PDF generation changes), but probably still needs some work, including tests for the overheads calculations.

Additionally, I am hoping to use this PR as an opportunity to:

1. Begin a discussion with @swalker2m and @cquiroz regarding what need to be done for ITAC and Phase 2 (REL-2927 and REL-2928 respectively), which are currently unassigned, as this will be necessary for the 2017B release, and may require additional handling from the PIT.

2. Get feedback on the current changes.

### DIFF
--- a/bundle/edu.gemini.model.p1.pdf/src/main/resources/edu/gemini/model/p1/pdf/templates/xsl-NOAO.xml
+++ b/bundle/edu.gemini.model.p1.pdf/src/main/resources/edu/gemini/model/p1/pdf/templates/xsl-NOAO.xml
@@ -1179,7 +1179,13 @@
                                    border-bottom-style="solid"
                                    border-bottom-width="0.25pt">
                         <fo:block font-weight="bold">
-                            Obs. Time
+                            Total
+                        </fo:block>
+                        <fo:block font-weight="bold">
+                            Program
+                        </fo:block>
+                        <fo:block font-weight="bold">
+                            Partner
                         </fo:block>
                     </fo:table-cell>
                 </fo:table-row>
@@ -1218,7 +1224,13 @@
                         </fo:table-cell>
                         <fo:table-cell>
                             <fo:block>
-                                <xsl:value-of select="format-number(time, '0.0')"/>&#160;<xsl:value-of select="time/@units"/>
+                                <xsl:value-of select="format-number(totalTime, '0.0')"/>&#160;<xsl:value-of select="totalTime/@units"/>
+                            </fo:block>
+                            <fo:block>
+                                <xsl:value-of select="format-number(progTime, '0.0')"/>&#160;<xsl:value-of select="progTime/@units"/>
+                            </fo:block>
+                            <fo:block>
+                                <xsl:value-of select="format-number(partTime, '0.0')"/>&#160;<xsl:value-of select="partTime/@units"/>
                             </fo:block>
                         </fo:table-cell>
                     </fo:table-row>
--- a/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Observation.scala
+++ b/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Observation.scala
@@ -34,9 +34,11 @@ object Observation {
   val empty = new Observation(None, None, None, None, M.Band.BAND_1_2)
 
   // Convenience method to extract the IntOrProgTime from the mutable observation.
-  def mTimeDisjunction(m: M.Observation): Option[Observation.IntOrProgTime] =
-    Option(m.getIntTime).map(TimeAmount(_).left[TimeAmount]).orElse(Option(m.getProgTime).map(TimeAmount(_).right[TimeAmount]))
-
+  def mTimeDisjunction(m: M.Observation): Option[Observation.IntOrProgTime] = {
+    def toTimeDisj(t: M.TimeAmount, f: TimeAmount => Observation.IntOrProgTime) =
+      Option(t).map(x => f(TimeAmount(x)))
+    toTimeDisj(m.getIntTime, _.left[TimeAmount]).orElse(toTimeDisj(m.getProgTime, _.right[TimeAmount]))
+  }
 }
 
 // REL-2985: It is unfortunate that we have to pass progTime here, but it is necessary for the migration to 2017B,
--- a/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Observation.scala
+++ b/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Observation.scala
@@ -1,36 +1,43 @@
 package edu.gemini.model.p1.immutable
 
+import edu.gemini.model.p1.overheads.Overheads
 import edu.gemini.model.p1.{mutable => M}
+
 import scalaz._
 import Scalaz._
 
 object Observation {
 
-  // Some lenses
+  // Some lenses.
   val blueprint:Lens[Observation, Option[BlueprintBase]] = Lens.lensu((a, b) => a.copy(blueprint = b), _.blueprint)
   val condition:Lens[Observation, Option[Condition]] = Lens.lensu((a, b) => a.copy(condition = b), _.condition)
   val target:Lens[Observation, Option[Target]] = Lens.lensu((a, b) => a.copy(target = b), _.target)
   val meta:Lens[Observation, Option[ObservationMeta]] = Lens.lensu((a, b) => a.copy(meta = b), _.meta)
-  val time:Lens[Observation, Option[TimeAmount]] = Lens.lensu((a, b) => a.copy(time = b), _.time)
+  val intTime:Lens[Observation, Option[TimeAmount]] = Lens.lensu((a, b) => a.copy(intTime = b), _.intTime)
 
   def apply(blueprint:Option[BlueprintBase],
             condition:Option[Condition],
             target:Option[Target],
             band:Band,
-            time:Option[TimeAmount]) = new Observation(blueprint, condition, target, time, band)
+            intTime:Option[TimeAmount],
+            progTime:Option[TimeAmount] = None) = new Observation(blueprint, condition, target, intTime, progTime, band)
 
   def apply(m:M.Observation) = new Observation(m)
 
-  def unapply(o:Observation) = Some((o.blueprint, o.condition, o.target, o.time, o.band))
+  def unapply(o:Observation) = Some((o.blueprint, o.condition, o.target, o.intTime, o.calculatedTimes, o.band))
 
-  val empty = new Observation(None, None, None, None, M.Band.BAND_1_2)
+  val empty = new Observation(None, None, None, None, None, M.Band.BAND_1_2)
 
 }
 
+// REL-2985: It is unfortunate that we have to pass progTime here, but it is necessary for the migration to 2017B,
+// since the old "time" user-specified parameter is now progTime, and we have changed to have progTime now calculated
+// as a function of parameter intTime: thus, if progTime is defined, we must calculate intTime from it.
 class Observation private (val blueprint:Option[BlueprintBase],
                            val condition:Option[Condition],
                            val realTarget:Option[Target],
-                           val time:Option[TimeAmount],
+                           private val intTimeParam:Option[TimeAmount],
+                           private val progTimeParam:Option[TimeAmount],
                            val band:Band,
                            val meta:Option[ObservationMeta] = None,
                            val enabled:Boolean = true) {
@@ -45,23 +52,49 @@ class Observation private (val blueprint:Option[BlueprintBase],
     case None => realTarget
   }
 
+  // REL-2985 requires special handling of the times.
+  // 2017A and earlier: intTimeParam will be None, and calculatedTimesParam will be defined with progTime.
+  // If a blueprint exists - which it should - we can calculate the intTime from the progTime and then the rest
+  // of the observation times from the intTime.
+  val intTime: Option[TimeAmount] = {
+    if (intTimeParam.nonEmpty) intTimeParam
+    else for {
+      b <- blueprint
+      c <- Overheads(b)
+      p <- progTimeParam
+    } yield c.intTimeFromProgTime(p)
+  }
+
+  // 2017B and later: if the blueprint and the intTime are defined, recalculate the calculatedTimes.
+  val calculatedTimes: Option[ObservationTimes] =  for {
+    b <- blueprint
+    c <- Overheads(b)
+    t <- intTime
+  } yield c.calculate(t)
+
+  def progTime:  Option[TimeAmount] = calculatedTimes.map(_.progTime)
+  def partTime:  Option[TimeAmount] = calculatedTimes.map(_.partTime)
+  def totalTime: Option[TimeAmount] = calculatedTimes.map(_.totalTime)
+
   def copy(blueprint:Option[BlueprintBase] = blueprint,
            condition:Option[Condition] = condition,
            target:Option[Target] = target,
-           time:Option[TimeAmount] = time,
+           intTime:Option[TimeAmount] = intTime,
+           calculatedTimes:Option[ObservationTimes] = calculatedTimes,
            band:Band = band,
            meta:Option[ObservationMeta] = None, // metadata resets to None on copy, by default
            enabled:Boolean = enabled) =
-    new Observation(blueprint, condition, target, time, band, meta, enabled)
+    new Observation(blueprint, condition, target, intTime, calculatedTimes.map(_.progTime), band, meta, enabled)
 
   def this(m:M.Observation) = this (
-    Option(m.getBlueprint).map(BlueprintBase(_)),
-    Option(m.getCondition).map(Condition(_)),
-    Option(m.getTarget).map(Target(_)),
-    Option(m.getTime).map(TimeAmount(_)),
-    Option(m.getBand).getOrElse(M.Band.BAND_1_2),
-    Option(m.getMeta).map(ObservationMeta(_)),
-    m.isEnabled
+      Option(m.getBlueprint).map(BlueprintBase(_)),
+      Option(m.getCondition).map(Condition(_)),
+      Option(m.getTarget).map(Target(_)),
+      Option(m.getIntTime).map(TimeAmount(_)),
+      Option(m.getProgTime).map(TimeAmount(_)),
+      Option(m.getBand).getOrElse(M.Band.BAND_1_2),
+      Option(m.getMeta).map(ObservationMeta(_)),
+      m.isEnabled
   )
 
   def mutable(n:Namer) = {
@@ -69,36 +102,39 @@ class Observation private (val blueprint:Option[BlueprintBase],
     m.setBlueprint(blueprint.map(_.mutable(n)).orNull)
     m.setCondition(condition.map(_.mutable(n)).orNull)
     m.setTarget(target.map(_.mutable(n)).orNull)
-    m.setTime(time.map(_.mutable).orNull)
+    m.setIntTime(intTime.map(_.mutable).orNull)
+    m.setProgTime(progTime.map(_.mutable).orNull)
+    m.setPartTime(partTime.map(_.mutable).orNull)
+    m.setTotalTime(totalTime.map(_.mutable).orNull)
     m.setBand(band)
     m.setMeta(meta.map(_.mutable).orNull)
     m.setEnabled(enabled)
     m
   }
 
-  def isEmpty = blueprint.isEmpty && condition.isEmpty && target.isEmpty && time.isEmpty
+  def isEmpty = blueprint.isEmpty && condition.isEmpty && target.isEmpty && intTime.isEmpty
 
   override def equals(a:Any) = a match {
     case o:Observation => kernel == o.kernel
     case _             => false
   }
 
-  private lazy val kernel = (blueprint, condition, target, time, band, meta)
+  private lazy val kernel = (blueprint, condition, target, intTime, band, meta)
 
   override lazy val hashCode = kernel.hashCode()
 
   def isPartialObservationOf(o:Observation) =
-    (time.isEmpty)  && // If I have time defined, I might be incomplete but I'm not partial
+    intTime.isEmpty  && // If I have time defined, I might be incomplete but I'm not partial
     (band == o.band) && // must be the same band
-    (blueprint.isEmpty || target.isEmpty || condition.isEmpty || time.isEmpty) && // Must have *some* empty component
+    (blueprint.isEmpty || target.isEmpty || condition.isEmpty || intTime.isEmpty) && // Must have *some* empty component
     isPartial(blueprint, o.blueprint) &&
     isPartial(target, o.target) &&
     isPartial(condition, o.condition) &&
-    isPartial(time, o.time)
+    isPartial(intTime, o.intTime)
 
   private def isPartial[A](a:Option[A], b:Option[A]) = a.isEmpty || a == b
 
   override def toString =
-    "Observation(%s, %s, %s, %s, %s)".format(blueprint, condition, target, time, band)
+    "Observation(%s, %s, %s, %s, %s)".format(blueprint, condition, target, intTime, band)
 
 }
--- a/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Observation.scala
+++ b/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Observation.scala
@@ -32,6 +32,11 @@ object Observation {
   def unapply(o:Observation) = Some((o.blueprint, o.condition, o.target, o.intTime, o.calculatedTimes, o.band))
 
   val empty = new Observation(None, None, None, None, M.Band.BAND_1_2)
+
+  // Convenience method to extract the IntOrProgTime from the mutable observation.
+  def mTimeDisjunction(m: M.Observation): Option[Observation.IntOrProgTime] =
+    Option(m.getIntTime).map(TimeAmount(_).left[TimeAmount]).orElse(Option(m.getProgTime).map(TimeAmount(_).right[TimeAmount]))
+
 }
 
 // REL-2985: It is unfortunate that we have to pass progTime here, but it is necessary for the migration to 2017B,
@@ -87,16 +92,11 @@ class Observation private (val blueprint:Option[BlueprintBase],
            enabled:Boolean = enabled) =
     new Observation(blueprint, condition, target, intTime.map(_.left), band, meta, enabled)
 
-
-  // Convenience method to extract the IntOrProgTime from the mutable observation.
-  private def mTimeDisjunction(m: M.Observation): Option[Observation.IntOrProgTime] =
-    Option(m.getIntTime).map(TimeAmount(_).left[TimeAmount]).orElse(Option(m.getProgTime).map(TimeAmount(_).right[TimeAmount]))
-
   def this(m:M.Observation) = this (
       Option(m.getBlueprint).map(BlueprintBase(_)),
       Option(m.getCondition).map(Condition(_)),
       Option(m.getTarget).map(Target(_)),
-      mTimeDisjunction(m),
+      Observation.mTimeDisjunction(m),
       Option(m.getBand).getOrElse(M.Band.BAND_1_2),
       Option(m.getMeta).map(ObservationMeta(_)),
       m.isEnabled
--- a/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Observation.scala
+++ b/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Observation.scala
@@ -7,6 +7,7 @@ import scalaz._
 import Scalaz._
 
 object Observation {
+  type IntOrProgTime = TimeAmount \/ TimeAmount
 
   // Some lenses.
   val blueprint:Lens[Observation, Option[BlueprintBase]] = Lens.lensu((a, b) => a.copy(blueprint = b), _.blueprint)
@@ -19,15 +20,13 @@ object Observation {
             condition:Option[Condition],
             target:Option[Target],
             band:Band,
-            intTime:Option[TimeAmount],
-            progTime:Option[TimeAmount] = None) = new Observation(blueprint, condition, target, intTime, progTime, band)
+            time:Option[IntOrProgTime]) = new Observation(blueprint, condition, target, time, band)
 
   def apply(m:M.Observation) = new Observation(m)
 
   def unapply(o:Observation) = Some((o.blueprint, o.condition, o.target, o.intTime, o.calculatedTimes, o.band))
 
-  val empty = new Observation(None, None, None, None, None, M.Band.BAND_1_2)
-
+  val empty = new Observation(None, None, None, None, M.Band.BAND_1_2)
 }
 
 // REL-2985: It is unfortunate that we have to pass progTime here, but it is necessary for the migration to 2017B,
@@ -36,8 +35,7 @@ object Observation {
 class Observation private (val blueprint:Option[BlueprintBase],
                            val condition:Option[Condition],
                            val realTarget:Option[Target],
-                           private val intTimeParam:Option[TimeAmount],
-                           private val progTimeParam:Option[TimeAmount],
+                           private val timeParam:Option[Observation.IntOrProgTime],
                            val band:Band,
                            val meta:Option[ObservationMeta] = None,
                            val enabled:Boolean = true) {
@@ -56,13 +54,12 @@ class Observation private (val blueprint:Option[BlueprintBase],
   // 2017A and earlier: intTimeParam will be None, and calculatedTimesParam will be defined with progTime.
   // If a blueprint exists - which it should - we can calculate the intTime from the progTime and then the rest
   // of the observation times from the intTime.
-  val intTime: Option[TimeAmount] = {
-    if (intTimeParam.nonEmpty) intTimeParam
-    else for {
+  val intTime: Option[TimeAmount] = timeParam.map {
+    case -\/(intTimeParam)  => intTimeParam
+    case \/-(progTimeParam) => (for {
       b <- blueprint
       c <- Overheads(b)
-      p <- progTimeParam
-    } yield c.intTimeFromProgTime(p)
+    } yield c.intTimeFromProgTime(progTimeParam)) | TimeAmount.empty
   }
 
   // 2017B and later: if the blueprint and the intTime are defined, recalculate the calculatedTimes.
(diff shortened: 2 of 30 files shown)


bundle/edu.gemini.model.p1.pdf/src/main/resources/edu/gemini/model/p1/pdf/templates/xsl-NOAO.xml
```diff
@@ -1126,7 +1126,13 @@
...
                                    border-bottom-width="0.25pt">
                         <fo:block font-weight="bold">
-                            Obs. Time
+                            Total
+                        </fo:block>
+                        <fo:block font-weight="bold">
+                            Program
+                        </fo:block>
+                        <fo:block font-weight="bold">
+                            Partner
```
sraaphorst: The instructions for the PDF generation, due to limited width, are to have the former time column label changed to:

Total
Program
Partner

and then have the three times listed on separate rows, hence this change and the change below in terms of PDF generation.

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Observation.scala
```diff
@@ -1,36 +1,43 @@
...
 object Observation {
 
-  // Some lenses
+  // Some lenses.
   val blueprint:Lens[Observation, Option[BlueprintBase]] = Lens.lensu((a, b) => a.copy(blueprint = b), _.blueprint)
   val condition:Lens[Observation, Option[Condition]] = Lens.lensu((a, b) => a.copy(condition = b), _.condition)
   val target:Lens[Observation, Option[Target]] = Lens.lensu((a, b) => a.copy(target = b), _.target)
   val meta:Lens[Observation, Option[ObservationMeta]] = Lens.lensu((a, b) => a.copy(meta = b), _.meta)
-  val time:Lens[Observation, Option[TimeAmount]] = Lens.lensu((a, b) => a.copy(time = b), _.time)
+  val intTime:Lens[Observation, Option[TimeAmount]] = Lens.lensu((a, b) => a.copy(intTime = b), _.intTime)
```
cquiroz: No need for a lens for `progTime`?

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/overheads/Overheads.scala
```diff
@@ -0,0 +1,130 @@
...
+
+import edu.gemini.model.p1.immutable._
+
+import scalaz.Scalaz._
+
+// This is not the ideal package for this to live in, but to avoid circular bundle references, we put it here.
+sealed trait Overheads {
+  def partnerOverheadFraction: Double
+  def acquisitionOverhead: Duration
+  def typicalExpTime: Option[Duration]
```
sraaphorst: This may not be actually necessary, since it is usually used in the calculation (in the spreadsheet listed in the description) for `otherOverheadFraction`, I believe.

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/overheads/Overheads.scala
```diff
@@ -0,0 +1,130 @@
+package edu.gemini.model.p1.overheads
+
+import java.time.Duration
```
sraaphorst: Not sure if I should be using `Duration` here due to the weirdness of `Duration.ofMinutes(59).toHours.getValue()` returning 0, but it seemed like a good idea before I figured that out.

Leaving it as is for now.

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Observation.scala
```diff
@@ -45,60 +52,89 @@ class Observation private (val blueprint:Option[BlueprintBase],
...
+
+  // 2017B and later: if the blueprint and the intTime are defined, recalculate the calculatedTimes.
+  val calculatedTimes: Option[ObservationTimes] =  for {
+    b <- blueprint
+    c <- Overheads(b)
+    t <- intTime
+  } yield c.calculate(t)
+
+  def progTime:  Option[TimeAmount] = calculatedTimes.map(_.progTime)
+  def partTime:  Option[TimeAmount] = calculatedTimes.map(_.partTime)
```
cquiroz: What is this? Partner time?

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/util/Rows.scala
```diff
@@ -1,15 +1,16 @@
...
 
 trait Rows { this: GridBagPanel =>
 
   border = DLU4_BORDER
   private var row = 0
-  def addRow(a: Component, b: Component, f:GridBagPanel.Fill.Value = GridBagPanel.Fill.Horizontal, wy:Int = 0 ) {
+  def addRow(a: Component, b: Component, f:GridBagPanel.Fill.Value = GridBagPanel.Fill.Horizontal, wy: Int = 0, gw: Int = 1) {
     add(a, new Constraints { gridx = 0; gridy = row; ipadx = 10; ipady = 4; anchor = GridBagPanel.Anchor.NorthEast })
-    add(b, new Constraints { gridx = 1; gridy = row; fill = f; weightx = 1; weighty = wy })
+    add(b, new Constraints { gridx = 1; gridy = row; fill = f; weightx = 1; weighty = wy; gridwidth = gw })
```
sraaphorst: Note minor changes to this class to simplify adding rows and minimize UI work.

I will post a shot of the new UI momentarily.

bundle/edu.gemini.shared.gui/src/main/scala/edu/gemini/shared/gui/textComponent/NumberField.scala
```diff
@@ -19,8 +19,7 @@ object NumberField {
 
 class NumberField(d: Option[Double], allowEmpty: Boolean, format: java.text.Format = NumberField.df) extends FormattedTextField(format) with SelectOnFocus {
   d.orElse(Some(0)).foreach { d =>
-    import NumberField.df
-    text = df.format(d)
+    text = format.format(d)
```
sraaphorst: This was almost certainly a bug in the original code. The default formatter was always being used instead of the provided formatter.

cquiroz:
How's that the partner time is calculated out of the integration time? How would it decide to what parter assign what?

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Observation.scala
cquiroz: Or is the time that will be "paid" by the partners?

bundle/edu.gemini.shared.gui/src/main/scala/edu/gemini/shared/gui/textComponent/NumberField.scala
```diff
@@ -31,6 +30,9 @@ class NumberField(d: Option[Double], allowEmpty: Boolean, format: java.text.Form
 
   def valid(d:Double):Boolean = true
 
+  def value: Option[Double] = Option(peer.getValue).map(_.asInstanceOf[Double])
+  def value_=(v: Double): Unit = peer.setValue(v)
+
```
sraaphorst: This is to simplify setting the value on a `FormattedTextField`, since using `text_=` does not have the required effects and this class seems to be lacking in Scala Swing.

sraaphorst:
Super lame proposal, just to demonstrate the changes.

screen shot 2017-01-31 at 12 08 29 pm
[full_proposal.pdf](https://github.com/gemini-hlsw/ocs/files/742365/full_proposal.pdf)

[full_proposal.xml.txt](https://github.com/gemini-hlsw/ocs/files/742367/full_proposal.xml.txt)

[full_proposal_2017A.pdf](https://github.com/gemini-hlsw/ocs/files/742369/full_proposal_2017A.pdf)
[full_proposal_2017A.xml.txt](https://github.com/gemini-hlsw/ocs/files/742370/full_proposal_2017A.xml.txt)

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/overheads/Overheads.scala
```diff
@@ -0,0 +1,130 @@
...
+    // Calculate everything in hours, as per REL-2985 formulae.
+    val progTimeHrs  = {
+      val visTimeHrs = Overheads.visTimeHrs
+      val intTimeHrs = intTime.toHours.value
+      val overheadTimeHrs = (1 + otherOverheadFraction) * intTimeHrs
+      val numAcqs = ((overheadTimeHrs / visTimeHrs).toInt + ((overheadTimeHrs % visTimeHrs > 0) ? 1 | 0)).toDouble
+
+      // Careful here: if we convert to hours directly using java.time API, this will be a Long value representing
+      // the floor of the number of hours.
+      val acqOverheadHrs = acquisitionOverhead.toMinutes.toDouble / 60.0
```
cquiroz: Do you need it twice?

sraaphorst:
The screenshot above was taken from the upconversion of the 2017A program, which is why there is a complaint about the target and ephemeris data. (It does not, nor has it ever been automatically re-looked up when a migration occurs.) Thus, please ignore this. I performed the lookup for the full_proposal.xml.

Also, I had to add .txt extensions since github doesn't allow dragging / attaching .xml files.

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/util/Rows.scala
```diff
@@ -31,7 +40,7 @@ trait Rows { this: GridBagPanel =>
     row = row + 1
   }
   def addSpacer() {
-    addRow(new Label(""), new Label("")) // :-/
+    addRow(new Label(" "), new Label("")) // :-/
```
sraaphorst: This was a bug, or at least did nothing on OS X. Completely empty labels are given a preferred size of (0,0), I believe.

cquiroz:
About the screenshot: Do you only enter the Integation time and the rest is calculated?

bundle/edu.gemini.model.p1/src/main/xsd/Observation.xsd
```diff
@@ -17,9 +17,12 @@
     <xsd:complexType name="Observation">
 
         <xsd:sequence>
-            <xsd:element name="time"  type="TimeAmount"/>
-            <xsd:element name="guide" type="GuideGroup"          minOccurs="0"/>
-            <xsd:element name="meta"  type="ObservationMetaData" minOccurs="0"/>
+            <xsd:element name="intTime"   type="TimeAmount"/>
```
cquiroz: So each of this fields would need to go to ITAC->P2

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/util/Rows.scala
cquiroz: This is a known issue

sraaphorst:
@cquiroz Yes, that's correct. Now, only the integration time is calculated and everything is a function of integration time. Moving from older proposals, we assume the time given is program time and use it to calculate integration time, thus preserving the other times as a function of integration time.

I used disabled NumberFields instead of Labels for the calculated values uniformity and to be able to use the formatter easily, but this could be changed. I really didn't want to have to spend a lot of time playing around with padding and alignment to get everything to line up in the "first draft" of this, as it were.

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/editor/ObservationEditor.scala
```diff
@@ -29,59 +31,115 @@ class ObservationEditor private (obs:Observation, canEdit:Boolean) extends StdMo
...
+    private type Named = {
+      def name: String
+    }
+
+    private class OptionLabel(a:Option[Named], icon0:Icon) extends Label() {
+      horizontalAlignment = Alignment.Left
+      icon = icon0
+      text = a.map(_.name).getOrElse("none")
+      if (a.isEmpty)
+        foreground = Color.GRAY
```
cquiroz: We may have used some Pinkish color for other labels when they are empty, can you check?

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
```diff
@@ -671,7 +671,7 @@ case class TimeProblems(p: Proposal, s: ShellAdvisor) {
   lazy val requested = p.proposalClass.requestedTime
   def obsTimeSum(b: Band) = TimeAmount.sum(for {
     o <- p.observations if o.band == b
-    t <- o.time
+    t <- o.progTime
```
cquiroz: Could you expand on why you selected `progTime` for this check?

cquiroz:
Could you check why travis is complaining?

sraaphorst:
@cquiroz My impression that this "partner time" is not partner time as we generally think, since it is listed in the UI as "Night Basecal Time."

I suspect it refers to the time a partner must give for calibration?

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/overheads/Overheads.scala
sraaphorst: The `acquisitionOverhead` could be type-changed from `Duration` to `Double` (or just `Long`) in mins, which is probably what I will do going forward since the use of `Duration` does seem to be a little unpredictable based on the fact that it uses `Long`s and takes the floor when converting down to a finer grain of measurement.

bundle/edu.gemini.model.p1/src/main/xsd/Observation.xsd
sraaphorst: I believe that Bryan stated that that is what he wanted.

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/editor/ObservationEditor.scala
sraaphorst: You are right: this is indeed the case for some classes (e.g. `SubmitView`) for `foregroundColor`, at least.

I just moved this code for organizational purposes, but it looks like it's just setting the text color to a lighter shade of gray if there is nothing specified. I think it would look odd to set the foreground color in this case?

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
sraaphorst: This, admittedly, I was not entirely sure about, and I have it listed as a question in a list of several questions to be asked. I really more than anything here wanted to get a draft of this out as a "request for comments" and to get the discussion going as to what needs to be done moving forward, so for now I assumed the mapping from the old `time` to the new `progTime` when I wasn't certain and made a note in a text file to revisit this where it wasn't clear.

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Observation.scala
sraaphorst: In the UI and the formulae in REL-2985, it's listed as `partTime` with the instruction to label it in the UI as `Night Basecal Time`.

It seems to be different than the normal idea we have of partner time based on that labeling, but this is a good point that should be clarified.

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Observation.scala
sraaphorst: I wasn't sure about this. The only time `progTime` should change is as a consequence of `intTime`, so I didn't think we needed a `Lens`, since the only instance `progTime` will be directly set is through migration from an older proposal (as opposed to implicitly set by changing `intTime`).

So I'm not entirely certain if a `Lens` is needed, but if there is a reason to use one, I'm happy to add one.

I'm working on some tests, including an upconversion one. I think it's essential since from semester-to-semester, the `Overheads` could change according to my conversation with Bryan, which is also why I recalculate the calculated times from the integration time.

In the final PR (remember that this is not the final PR), I plan to include tests, and will solve the issue of Travis failing 😸

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/overheads/Overheads.scala
```diff
@@ -0,0 +1,130 @@
...
+  // the floor of the number of hours.
+  private def acqOverheadHrs = acquisitionOverhead.toMinutes.toDouble / 60.0
+
+  def calculate(intTime: TimeAmount): ObservationTimes = {
+    // Calculate everything in hours, as per REL-2985 formulae.
+    val progTimeHrs  = {
+      val visTimeHrs = Overheads.visTimeHrs
+      val intTimeHrs = intTime.toHours.value
+      val overheadTimeHrs = (1 + otherOverheadFraction) * intTimeHrs
+      val numAcqs = ((overheadTimeHrs / visTimeHrs).toInt + ((overheadTimeHrs % visTimeHrs > 0) ? 1 | 0)).toDouble
```
tpolecat: I realize that `visTimeHrs` is hardcoded to 2 hours right now, but can it ever conceivably be zero? If so you end up with`numAcqs = 2.147483647E9` which is probably not right.

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
cquiroz: 👍

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/editor/ObservationEditor.scala
cquiroz: Right, it is foreground color

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/overheads/Overheads.scala
cquiroz: What about `Time` in Squants?

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Observation.scala
cquiroz: Ok, it is kind of misleading, I thought it was partner time, which is odd as it depends on how do you apply to time to each partner

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/overheads/Overheads.scala
```diff
@@ -0,0 +1,130 @@
...
+      TimeAmount(partTimeHrs,  TimeUnit.HR),
+      TimeAmount(totalTimeHrs, TimeUnit.HR))
+  }
+
+  // This is needed for 2017A to 2017B migration, as time in former proposals will be migrated to progTime, and
+  // thus intTime - the value that is user modifiable - must be calculated from this.
+  def intTimeFromProgTime(progTime: TimeAmount): TimeAmount = {
+    val visTimeHrs  = Overheads.visTimeHrs
+    val progTimeHrs = progTime.toHours.value
+    val numAcqs     = ((progTimeHrs / visTimeHrs).toInt + ((progTimeHrs % visTimeHrs > 0) ? 1 | 0)).toDouble
```
tpolecat: Same comment as above. There seems to be some duplication.

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/editor/ObservationEditor.scala
```diff
@@ -29,59 +31,115 @@ class ObservationEditor private (obs:Observation, canEdit:Boolean) extends StdMo
...
+  object PartTime extends CalculatedTimeAmountField
+  object PartTimeUnits extends UnitsLabel
+
+  object TotalTime extends CalculatedTimeAmountField
+  object TotalTimeUnits extends UnitsLabel
+
+  def updateTimeLabels(): Unit = {
+    val intTime = TimeAmount(\/.fromTryCatchNonFatal(IntegrationTime.text.toDouble).getOrElse(0.0), Units.selection.item)
+    val obsTimes = calculator.map(_.calculate(intTime))
+    ProgramTime.update(obsTimes.map(_.progTime).getOrElse(TimeAmount.empty))
```
tpolecat: I would be ok defining an additive `Monoid[TimeAmount]` even though it's not associative because it will never affect us at the scales we use. This would allow you to write this as `.update(obsTimes.foldMap(_.progTime))`

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/util/Rows.scala
tpolecat: Can you add a comment explaining this?

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Observation.scala
cquiroz: Yes. in this case an upconversion test is probably important

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Observation.scala
```diff
@@ -1,36 +1,43 @@
...
-  val time:Lens[Observation, Option[TimeAmount]] = Lens.lensu((a, b) => a.copy(time = b), _.time)
+  val intTime:Lens[Observation, Option[TimeAmount]] = Lens.lensu((a, b) => a.copy(intTime = b), _.intTime)
 
   def apply(blueprint:Option[BlueprintBase],
             condition:Option[Condition],
             target:Option[Target],
             band:Band,
-            time:Option[TimeAmount]) = new Observation(blueprint, condition, target, time, band)
+            intTime:Option[TimeAmount],
+            progTime:Option[TimeAmount] = None) = new Observation(blueprint, condition, target, intTime, progTime, band)
```
sraaphorst: The reason is primarily in the migration.

Old proposals will have a `progTime` (and not a fully calculated `ObservationTimes`) and need to calculate the `intTime` from this, and then use this `intTime` to calculate the full `ObservationTimes`.

New 2017B and onward proposals will have an `intTime` and will use this to calculate the full `ObservationTimes`.

My initial thought was actually to pass in an `Option[ObservationTimes]` as you say, but as the only field with a valid value would ever be `progTime`, I thought this might be the wrong approach.

I'm open to alternative suggestions. 😸

sraaphorst: Ideally, we'd take the old `time` parameter in `UpConverter` and calculate the `intTime` from that, but that would be an absolute nightmare, requiring us to parse out the instrument configuration.

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Observation.scala
```diff
@@ -1,36 +1,43 @@
...
+  def unapply(o:Observation) = Some((o.blueprint, o.condition, o.target, o.intTime, o.calculatedTimes, o.band))
 
-  val empty = new Observation(None, None, None, None, M.Band.BAND_1_2)
+  val empty = new Observation(None, None, None, None, None, M.Band.BAND_1_2)
 
 }
 
+// REL-2985: It is unfortunate that we have to pass progTime here, but it is necessary for the migration to 2017B,
+// since the old "time" user-specified parameter is now progTime, and we have changed to have progTime now calculated
+// as a function of parameter intTime: thus, if progTime is defined, we must calculate intTime from it.
```
sraaphorst: `progTime` would be ignored.

I suppose we might be able to somehow separate this into two: one that accepts a `progTime`, calculates the `intTime`, and then creates the object based on this. I'd have to investigate a little more how that would figure into the mutable to immutable conversion when reading.

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/ObservationTimes.scala
```diff
@@ -0,0 +1,4 @@
+package edu.gemini.model.p1.immutable
+
+// All time amounts are in hours, so convert as necessary to nights.
+case class ObservationTimes(progTime: TimeAmount, partTime: TimeAmount, totalTime: TimeAmount)
```
sraaphorst: This is a good point. Yes, you're absolutely right: `totalTime` should not be a parameter.

I got case classes stuck in my head, thinking I was going to match on them at some point, but that never happens, I don't think, so this should just be a `val` in the class.

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Observation.scala
tpolecat: So it sounds like you want `Option[IntTime \/ ProgTime]`?

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/overheads/Overheads.scala
sraaphorst: I'll have to ask on the REL task. My impression is that it will always be fixed to 2, but as it comes up in multiple places in the calculation, I extracted it.

The whole concept of floating point modulo arithmetic seems very strange to me. I understand what it is doing, but coming from a math background, I always associated it with number theory and the ring of integers, but I can see how it works.

sraaphorst:
@tpolecat Regarding changing `new Label("")` to `new Label(" ")`, at least on OS X, a `Label` with no text is given a preferred size of (0,0), so `addSpacer` really, in essence, does nothing, which I assume is not the intent: hence I made one of the `Label`s be nonempty so that a row of space was actually allotted for the `Component`s, which I assume was what was desired by this method from the name.

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/editor/ObservationEditor.scala
sraaphorst: This works for me.

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/overheads/Overheads.scala
sraaphorst: Maybe I'm just being completely blind / stupid here, but I don't immediately see the duplication?

Are you referring to having both `intTime` and `progTime`? Again, this comes from the migration from older proposals. It will result in `progTime` being used to calculate `intTime`, which is then used to (again) calculate `progTime` and `partTime`, which is a bit redundant, but my concern was with the way these were being formatted in the UI and precision. Maybe that isn't an issue and we don't require the roundabout computation.

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/util/Rows.scala
sraaphorst: Will do.

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Observation.scala
sraaphorst: Nice. Yes, that would make much more sense, I think.

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/overheads/Overheads.scala
tpolecat: > I don't immediately see the duplication?

I was referring to lines 26 and 47 which are almost identical:
```scala
val numAcqs = ((overheadTimeHrs / visTimeHrs).toInt + ((overheadTimeHrs % visTimeHrs > 0) ? 1 | 0)).toDouble
val numAcqs = ((progTimeHrs     / visTimeHrs).toInt + ((progTimeHrs     % visTimeHrs > 0) ? 1 | 0)).toDouble
```

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/TimeAmount.scala
```diff
@@ -13,7 +16,9 @@ object TimeAmount {
   // with the difference that we try to keep the time unit from changing if they
   // are all the same.
   def sum(times: Traversable[TimeAmount]): TimeAmount =
-    if (times.isEmpty) empty else (times.head/:times.tail)(_ + _)
+    if (times.isEmpty) empty else (times.head/:times.tail)(_ |+| _)
```
tpolecat: Do we really need `Traversable` here? `TimeAmount.sum(ts)` can normally be written `ts.suml`. In any case this method should just be `foldLeft` starting with `empty`; the `if` is unnecessary.

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Observation.scala
sraaphorst: Now added.

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/editor/ObservationEditor.scala
sraaphorst: @tpolecat  Excuse my Scalaz ignorance: if we add an implicit `Monoid[TimeAmount]`, is there some easy way that we can just get rid of the `TimeAmount.sum` function? This should be `Foldable` and be able to use `FoldableOps` to sum somehow, no? I'm just not sure of the correct syntax or if this is the right approach.

tpolecat: Yeah if you `import scalaz._, Scalaz._` then you can call `.suml` on lists, vectors, sets, things like that. It won't work on mutable things but there shouldn't be many of those.

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/overheads/Overheads.scala
sraaphorst: Much better with Squants. Changed
(And excellent for finally having some exposure to Squants.)

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Observation.scala
```diff
@@ -45,60 +52,89 @@ class Observation private (val blueprint:Option[BlueprintBase],
...
-    Option(m.getTarget).map(Target(_)),
-    Option(m.getTime).map(TimeAmount(_)),
-    Option(m.getBand).getOrElse(M.Band.BAND_1_2),
-    Option(m.getMeta).map(ObservationMeta(_)),
-    m.isEnabled
+      Option(m.getBlueprint).map(BlueprintBase(_)),
+      Option(m.getCondition).map(Condition(_)),
+      Option(m.getTarget).map(Target(_)),
+      Option(m.getIntTime).map(TimeAmount(_)),
+      Option(m.getProgTime).map(TimeAmount(_)),
```
sraaphorst: @tpolecat - hoping you can help me out here.
By your suggestion, I am changing to `Option[TimeAmount \/ TimeAmount]`, so these two preceding lines need to be combined to produce this, but I'm not sure the best way to write this. Right now, I have:

`Option(m.getIntTime).map(TimeAmount(_).right[TimeAmount]).orElse(Option(m.getProgTime).map(TimeAmount(_).left[TimeAmount]))`

but that is truly hideous. There has to be a better way to do this. Any suggestions?

sraaphorst: Carlos was suggesting pattern matching to make it more readable:
`(Option(m.getIntTime), Option(m.getProgTime)) match { ... }`
which is definitely better, I think, but there has to be some scalaz way, I would think, to do this more cleanly?

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/TimeAmount.scala
sraaphorst: Yeah, I was thinking that. In `Submissions.scala`, though, we have:

`  def sumTimes(subs: Traversable[Submission], f: Submission => TimeAmount): TimeAmount =
    subs.map(f).suml`

and even with `import scalaz._, Scalaz._`, I'm getting:

Error:(32, 17) value suml is not a member of Traversable[edu.gemini.model.p1.immutable.TimeAmount]
subs.map(f).suml

tpolecat: Yeah it's not going to work with `Traversable`. But with `Foldable` this operation is just `subs.foldMap(f)` and I'll bet all the callers are `Foldable`. I didn't know this stuff when I wrote the PIT.

Not a big deal if you don't feel like chasing this down.

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Observation.scala
tpolecat: You can't do much better because you have the exceptional case you have to handle. I would probably do it in terms of `align`:

```scala
(op1 align op2) map {
  case This(a)    => a.left
  case That(b)    => b.right
  case Both(a, b) => error condition
}
```

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Observation.scala
```diff
@@ -1,36 +1,51 @@
...
   def apply(m:M.Observation) = new Observation(m)
 
-  def unapply(o:Observation) = Some((o.blueprint, o.condition, o.target, o.time, o.band))
+  def unapply(o:Observation) = Some((o.blueprint, o.condition, o.target, o.intTime, o.calculatedTimes, o.band))
 
   val empty = new Observation(None, None, None, None, M.Band.BAND_1_2)
 
+  // Convenience method to extract the IntOrProgTime from the mutable observation.
+  def mTimeDisjunction(m: M.Observation): Option[Observation.IntOrProgTime] =
+    Option(m.getIntTime).map(TimeAmount(_).left[TimeAmount]).orElse(Option(m.getProgTime).map(TimeAmount(_).right[TimeAmount]))
```
tpolecat: This is a little copy-pastey. How about

```scala
def mTimeDisjunction(m: M.Observation): Option[Observation.IntOrProgTime] = {
  val oi = Option(m.getIntTime)
  val op = Option(m.getProgTime)
  (oi orElse op).map(TimeAmount(_).right[TimeAmount]))
}
```

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/TimeAmount.scala
tpolecat: `times.foldLeft(empty)(_ |+| _)`

tpolecat:
This looks ok to me, at least as far as I can follow in my head. I had a few more stylistic comments but nothing serious.

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Observation.scala
tpolecat: Oh I'm sorry, I didn't see the left vs right difference. Nevermind.

PR #1157, opened 2017-02-03
REL-3069: Queue Visualization error on last day of semester

cquiroz:
This is a very subtle bug where QV will break on the last day of the semester **after sunrise**

The PR brings a test case that reproduced the issue and a tentative fix. An issue with the fix is my lack of knowledge about QV, hence I cannot predict if there are any other side effects of the change.

Also extracting a value from an `Option` was the original symptom. That has been changed to instead return a default but I'm not sure if it is the correct one

I hope some of our users could build this and test QV to check for any problems due to this update

### DIFF
--- a/bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/filter/core/Filter.scala
+++ b/bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/filter/core/Filter.scala
@@ -242,7 +242,7 @@ object Filter {
   // =======================  *** Set time filter ***
   case class SetTime(ctx: QvContext, min: Double = 0, max: Double = 15) extends SimpleRangeFilter {
     def label = "Set Time Hrs"
-    def getter: Obs => Double = { o => TimeUtils.asHours(SolutionProvider(ctx).remainingHours(ctx, o)) }
+    def getter: Obs => Double = { o => TimeUtils.asHours(SolutionProvider(ctx).remainingHours(ctx, o).getOrElse(0L)) }
     def lowest = 0
     def highest = Double.MaxValue
     override def desc = "Filter for targets with number of hours before the target sets tonight in the range min ≤ hours < max."
--- a/bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/filter/core/Filter.scala
+++ b/bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/filter/core/Filter.scala
@@ -68,7 +68,7 @@ case class ConfigurationFilter[A](label: String, instrument: SPComponentType, gr
   def name = selection.map(valueName).mkString(",")
   def predicate(o: Obs) =
     if (!o.getInstrumentComponentType.equals(instrument)) false
-    else selection.intersect(collector(o)).size > 0
+    else selection.intersect(collector(o)).nonEmpty
 
   // implementation of copy method based on the case class copy
   def updated(selection: Set[A]) = copy(selection = selection)
@@ -93,8 +93,8 @@ sealed trait RangeFilter extends Filter {
   def name = toName
 
   override def categoryName = toCategoryName
-  def maxFromString(s: String) = try { s.toDouble } catch { case _: Throwable => highest }
-  def minFromString(s: String) = try { s.toDouble } catch { case _: Throwable => lowest }
+  def maxFromString(s: String): Double = try { s.toDouble } catch { case _: Throwable => highest }
+  def minFromString(s: String): Double = try { s.toDouble } catch { case _: Throwable => lowest }
   def predicate(o: Obs) = {
     val value = getter(o)
     value >= min && value < max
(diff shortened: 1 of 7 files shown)


bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/util/SolutionProvider.scala
```diff
@@ -25,9 +27,12 @@ object SolutionProvider {
     Site.GS -> new SolutionProvider(Site.GS)
   )
 
-  def apply(site: Site) = providers(site)
-  def apply(peer: Peer) = providers(peer.site)
-  def apply(ctx: QvContext) = providers(ctx.site)
+  def currentNight(site: Site, currentTime: Long): Option[Night] =
+    SemesterData.current(site, currentTime).nights.find(n => n.dayEnd > currentTime)
```
cquiroz: This is the core change of this PR it used to be `n.end > currentTime` where `n.end` is the sunrise time

bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/util/SolutionProvider.scala
```diff
@@ -77,17 +82,18 @@ sealed class SolutionProvider(site: Site) extends Publisher {
   def value(valueType: ValueType, night: Night, obs: Obs): Double =
     constraintsCache.value(Seq(night), valueType, obs).head
 
-  def remainingHours(ctx: QvContext, o: Obs): Long = {
-    val n = SemesterData.current(ctx.site).nights.find(n => n.end > System.currentTimeMillis()).get
-    val s = solution(Seq(n), Set[ConstraintType](Elevation), o).restrictTo(n.interval)
-    val set = s.intervals.find(_.end < n.nauticalTwilightEnd).map(_.end).getOrElse(n.nauticalTwilightEnd)
-    set - n.nauticalTwilightStart
+  def remainingHours(ctx: QvContext, o: Obs, currentTime: Long = System.currentTimeMillis()): Option[Long] = {
```
cquiroz: Instead of extracting the `Option` this passes it upstream

bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/filter/core/Filter.scala
```diff
@@ -242,7 +242,7 @@ object Filter {
   // =======================  *** Set time filter ***
   case class SetTime(ctx: QvContext, min: Double = 0, max: Double = 15) extends SimpleRangeFilter {
     def label = "Set Time Hrs"
-    def getter: Obs => Double = { o => TimeUtils.asHours(SolutionProvider(ctx).remainingHours(ctx, o)) }
+    def getter: Obs => Double = { o => TimeUtils.asHours(SolutionProvider(ctx).remainingHours(ctx, o).getOrElse(0L)) }
```
cquiroz: is `0L` a sensible default?

tpolecat: Kind of hard to tell from the code but maybe it should default to `min` since the description claims the value will be in `[min, max)` ... ?

cquiroz: That's a good suggestion

bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/util/SolutionProvider.scala
swalker2m: I know it's not really the point of this bug fix, but I thought I'd point it out nevertheless:  I think this lookup could be done in near constant time if we hash the `Night` objects by start time.  There is code in `edu.gemini.skycalc` for this sort of thing.

bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/util/SemesterData.scala
```diff
@@ -106,13 +105,13 @@ object SemesterData {
...
   private def add(site: Site, semester: Semester) = Cache.add(site, semester)
 
-  def get(site: Site, semester: Semester) = Cache.get(site, semester)
+  def get(site: Site, semester: Semester): SemesterData = Cache.get(site, semester)
 
-  def semesters(site: Site) = Cache.semesters(site)
+  def semesters(site: Site): Seq[SemesterData] = Cache.semesters(site)
 
-  def current(site: Site) = semesters(site, System.currentTimeMillis(), System.currentTimeMillis()).head
+  def current(site: Site, currentTime: Long = System.currentTimeMillis()): SemesterData = semesters(site, currentTime, currentTime).head
```
swalker2m: Thanks for adding the explicit return types.

andrewwstephens:
Set clock to Jan-31 @ 17:12 and see no problem.

PR #1159, opened 2017-02-06
REL-2941: Changes to GSAOI guiding estimates for PIT 2017B

sraaphorst:
These are the modifications requested for the GSAOI guiding estimates for the 2017B PIT.

First off, I clarified the meaning of the progress method, which was extremely unclear, through some renaming. Indeed, there seemed to be some contradictions in the comments / naming.

Second, I now filter out non-Canopus stars. Previously, we returned 0 or 100% depending on whether or not an asterism could be found. As per Bryan's comments on REL-2941, we now return the number of CWFS probes with guide stars divided by three to give the estimate of success.

I also, as per Bryan's recommendation, added a warning if the number of CWFS guide stars is less than three.

### DIFF
--- a/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/mascot/MascotCat.scala
+++ b/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/mascot/MascotCat.scala
@@ -92,17 +92,17 @@ object MascotCat {
    * @param centerRA the base position RA coordinate
    * @param centerDec the base position Dec coordinate
    * @param band determines which magnitudes are used in the calculations: (one of "B", "V", "R", "J", "H", "K")
-   * @param shouldStop called for each asterism as it is calculated, can cancel the calculations by returning true
+   * @param shouldContinue called for each asterism as it is calculated, can cancel the calculations by returning false
    * @return a tuple: (list of stars actually used, list of asterisms found)
    */
   def findBestAsterismInTargetsList(javaList: List[SiderealTarget],
                                     centerRA: Double, centerDec: Double,
                                     band: BandsList, factor: Double,
-                                    shouldStop: (Strehl, Boolean) => Boolean): StrehlResults = {
+                                    shouldContinue: (Strehl, Boolean) => Boolean): StrehlResults = {
     val progress:ProgressFunction = (s: Strehl, count: Int, total: Int) => {
       defaultProgress(s, count, total)
       // TODO: Why is the usable parameter always true?
-      shouldStop(s, true)
+      shouldContinue(s, true)
     }
 
     val (starList, strehlList) = findBestAsterism(javaList, centerRA, centerDec, factor, progress, Mascot.defaultFilter)
--- a/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/mascot/MascotCat.scala
+++ b/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/mascot/MascotCat.scala
@@ -3,7 +3,7 @@ package edu.gemini.ags.gems.mascot
 import java.util.logging.Logger
 
 import edu.gemini.ags.gems.mascot.Mascot.ProgressFunction
-import edu.gemini.spModel.core.{BandsList, MagnitudeBand}
+import edu.gemini.spModel.core.BandsList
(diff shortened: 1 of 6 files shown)


sraaphorst:
Sorry for the take 2 on this PR. I noticed after creating it the first time that I had confused even myself with the naming conventions on the `shouldContinue` function. This should be clear now.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/GemsStrategy.scala
```diff
@@ -143,20 +144,21 @@ trait GemsStrategy extends AgsStrategy {
...
-    def progress(s: Strehl, usable: Boolean): Boolean = {
-      !((usable && s.stars.size >= 2) || (s.stars.size < 2))
-    }
+    // We can terminate the Mascot algorithm immediately if we find a usable 3-star asterism or no asterism is found.
+    def shouldContinue(s: Strehl, usable: Boolean): Boolean = !(
+      // A usable three star asterism.
+      (usable && s.stars.size >= 3) ||
+      // No asterism.
+      (s.stars.size < 2)
+    )
```
swalker2m: I still get a bit confused.  Maybe if we don't flip the result of the comparison at the end as you suggested earlier?  I'm not sure what `usable` means but maybe:

````scala
def shouldContinue(s: Strehl, usable: Boolean): Boolean =
  !usable || s.tars.size == 2
````

Seems like we're saying that only 2 star asterisms require more work?

File: bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/mascot/MascotCat.scala [hunk 2]
```diff
@@ -92,17 +92,17 @@ object MascotCat {
...
                                     centerRA: Double, centerDec: Double,
                                     band: BandsList, factor: Double,
-                                    progressCheck: (Strehl, Boolean) => Boolean): StrehlResults = {
+                                    shouldContinue: (Strehl, Boolean) => Boolean): StrehlResults = {
     val progress:ProgressFunction = (s: Strehl, count: Int, total: Int) => {
       defaultProgress(s, count, total)
-      // Why usable defaults to true?
-      progressCheck(s, true)
+      // TODO: Why is the usable parameter always true?
+      shouldContinue(s, true)
```
swalker2m: I couldn't find where the `Boolean` parameter is ever set to anything but `true`.  Is that the case?  If so,we should just remove it.

sraaphorst: It is never set to anything other than `true`. I really have no idea why it is included.
I have no issue removing it.

sraaphorst: Incidentally, the meaning of `true` and `false` never did change, except in the comments for one method for added confusion, as the method `shouldContinue` was passed to and used promptly reverted back to the intended meaning.

I maintained the original functionality (mostly) and negated.

New commit in a second.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/GemsStrategy.scala
```diff
@@ -143,20 +144,13 @@ trait GemsStrategy extends AgsStrategy {
...
-    // 2. If no asterisms were found.
-    // Returning false will stop the search
-    def progress(s: Strehl, usable: Boolean): Boolean = {
-      !((usable && s.stars.size >= 2) || (s.stars.size < 2))
-    }
-
-    // Iterate over 45 degree position angles if no asterism is found at PA = 0.
-    val gemsCatalogResults = results.map(result => GemsResultsAnalyzer.analyzeGoodEnough(ctx, anglesToTry, result, progress))
+    // Iterate over 45 degree position angles if no 3-star asterism is found at PA = 0.
+    val gemsCatalogResults = results.map(result => GemsResultsAnalyzer.analyzeGoodEnough(ctx, anglesToTry, result, _.stars.size < 3))
```
swalker2m: This is definitely much easier to process.  It does change the meaning from what was installed before though.  If you're sure this is correct, 👍 Either way we need them to do mascot testing with the OT as well since this is used by both just to be sure.

PR #1161, opened 2017-02-07
REL-3059: A couple more required changes for 2017B test for p1-monitor.

sraaphorst:
Sorry... second time doing this, and keep discovering missed details.


swalker2m:
It seems like all of this could be defaulted from the current semester, which can be calculated from the date.  It's work but it might save some effort down the road.  On the other hand since we're actively trying to move to a new build and distribution system anyway maybe it doesn't matter.

At any rate, 👍

sraaphorst:
Agreed. There are just way too many places where these values have to be changed as it currently stands. It's rife with pitfalls and has wasted a lot of time today.

PR #1162, opened 2017-02-10
REL-2973: Added LRU cache for AGS estimation queries to eliminate ODB server load.

sraaphorst:
In the 2017A PIT call for proposals the ODB server was hammered with AGS http queries for guiding estimations.

Investigation into this shows that repeated, identical queries are made.
Thus, to reduce server load, we implement an LRU cache on query string for successful results.

Note that this also seems to occur with the GSA, as per my notes on REL-2973:
http://jira.gemini.edu:8080/browse/REL-2973

Perhaps it would similarly be worthwhile to add a cache for this as well, especially since these queries seem to be fairly time-intensive? I'd like to do so but am wondering if anyone has any reason to object to this.

### DIFF
--- a/bundle/edu.gemini.ags.client.impl/src/main/scala/edu/gemini/ags/client/impl/AgsHttpClient.scala
+++ b/bundle/edu.gemini.ags.client.impl/src/main/scala/edu/gemini/ags/client/impl/AgsHttpClient.scala
@@ -2,7 +2,6 @@ package edu.gemini.ags.client.impl
 
 import edu.gemini.ags.client.api._
 import edu.gemini.model.p1.immutable.Observation
-import edu.gemini.shared.util.LruCache
 import edu.gemini.util.ssl.GemSslSocketFactory
 import java.net.{HttpURLConnection, URL, URLConnection}
 import HttpURLConnection.HTTP_OK
@@ -10,8 +9,11 @@ import HttpURLConnection.HTTP_OK
 import io.Source
 import java.io.{IOException, InputStream}
 import java.util.Collections
+import java.util.{LinkedHashMap, Map}
 import javax.net.ssl.{HostnameVerifier, HttpsURLConnection, SSLSession}
 import java.util.logging.{Level, Logger}
+
+import edu.gemini.ags.client.api.AgsResult.Success
 
 import scalaz._
 import Scalaz._
@@ -26,7 +28,14 @@ object AgsHttpClient {
   }
 
   // LRU cache to minimize lookups.
-  val lRUCache = Collections.synchronizedMap(new LruCache[String, AgsResult.Success](100))
+  val lRUCache = {
+    val CacheLimit = 1000
+    val cache = new LinkedHashMap[String, AgsResult.Success](CacheLimit, 0.75f, true) {
+      override def removeEldestEntry(eldest: Map.Entry[String, Success]): Boolean =
+        size() > CacheLimit
+    }
+    Collections.synchronizedMap(cache)
+  }
 }
 
 import AgsHttpClient.{Log, hostnameVerifier, lRUCache}
@@ -54,10 +63,12 @@ case class AgsHttpClient(host: String, port: Int) extends AgsClient {
         conn.setReadTimeout(timeout)
         Charset.set(conn)
 
-        Response(conn).result match {
-          case s@AgsResult.Success(_) => lRUCache.put(url.toString, s)
-          case other                  => other
+        val result = Response(conn).result
+        result match {
+          case s:AgsResult.Success => lRUCache.put(url.toString, s)
+          case _                   =>
         }
+        result
       }
     } catch {
       case io: IOException =>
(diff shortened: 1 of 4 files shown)


bundle/edu.gemini.ags.client.impl/src/main/scala/edu/gemini/ags/client/impl/AgsHttpClient.scala
```diff
@@ -19,9 +24,12 @@ object AgsHttpClient {
   private val hostnameVerifier: HostnameVerifier = new HostnameVerifier {
      def verify(s: String, sslSession: SSLSession) = true
   }
+
+  // LRU cache to minimize lookups.
+  val lRUCache = Collections.synchronizedMap(new LruCache[String, AgsResult.Success](100))
```
sraaphorst: Unsure about size here. I figured 100 seemed more than large enough, and given the size of the objects being stored in the cache (query `String's as keys and wrapped `Double` guiding estimates as values), this was not problematically large.

bundle/edu.gemini.ags.client.impl/src/main/scala/edu/gemini/ags/client/impl/AgsHttpClient.scala
```diff
@@ -1,16 +1,21 @@
 package edu.gemini.ags.client.impl
 
-import edu.gemini.model.p1.immutable.Observation
 import edu.gemini.ags.client.api._
+import edu.gemini.model.p1.immutable.Observation
+import edu.gemini.shared.util.LruCache
```
sraaphorst: Reusing an existing implementation we have of `LruCache`, which is just a Java `LinkedHashMap`.

bundle/edu.gemini.ags.client.impl/src/main/scala/edu/gemini/ags/client/impl/AgsHttpClient.scala
swalker2m: I might bump this up some more.  It wouldn't take that much space if it were, say, 1000 entries.  There are some pretty long target lists in some proposals.

bundle/edu.gemini.ags.client.impl/src/main/scala/edu/gemini/ags/client/impl/AgsHttpClient.scala
swalker2m: There's so little to configuring an LRU cache that I think I'd be inclined to just instantiate it directly since the alternative is adding another jar file to the distribution.

bundle/edu.gemini.ags.client.impl/src/main/scala/edu/gemini/ags/client/impl/AgsHttpClient.scala
```diff
@@ -39,12 +47,18 @@ case class AgsHttpClient(host: String, port: Int) extends AgsClient {
...
-      Response(conn).result
+      Option(lRUCache.get(url.toString) : AgsResult) | {
+        val conn = url.openConnection().asInstanceOf[HttpsURLConnection]
+        conn.setHostnameVerifier(hostnameVerifier)
+        conn.setSSLSocketFactory(GemSslSocketFactory.get)
+        conn.setReadTimeout(timeout)
+        Charset.set(conn)
+
+        Response(conn).result match {
+          case s@AgsResult.Success(_) => lRUCache.put(url.toString, s)
```
swalker2m: This looks like a bug (?) because a Java `HashMap` returns "the previous value associated with key, or `null` if there was no mapping for key".

Also, there's a side-effect here in updating the cache.  I think it would be a tiny bit clearer to separate the side-effect from the computation.  (Never mind that the computation isn't referentially transparent anyway!). That is, store the result of the computation in a value.  Do the cache updating on success afterwords (even if that means often replacing a success value with the same value you just pulled out of the cache), then just use the value as the result.

sraaphorst: Ugh, you're right. Sorry about that.

Would it make sense to calculate the value, wrap the cache update in an `IO[Unit]` to clarify that it has a side effect, and then `unsafePerformIO` on it, and then return the value, or would that be overkill? Trying to get a better handle on when I should be using `IO`. So something like:

`val result = ...
result match {
    case s: AgsResult.Success => IO { lruCache.put(url.toString, s) }.unsafePerformIO
    case _ =>
}
result`
?

swalker2m: I think overkill given that the computation to get the result is doing network access anyway and that isn't done in IO.  Perhaps if I'd known better I would have written this code differently in the first place but as is, I don't think there is anything to be gained by wrapping that bit in IO.

PR #1163, opened 2017-02-10
GHOST Implementation Ideas

swalker2m:
I've spent some time thinking about how to implement GHOST in the OCS2 model.  No matter what we do, it's likely going to be months of work.  Before estimating effort in greater detail though, I wanted to see if anybody has feedback on the ideas I've had so far.  Or better yet, if anybody has better ideas don't hesitate to share them.

### DIFF
--- a/bundle/edu.gemini.pot/doc/GhostEstimates.md
+++ b/bundle/edu.gemini.pot/doc/GhostEstimates.md
@@ -528,7 +528,11 @@ Search capability.  Needs input from science.
 * Details from science TBD.
 
 
+# ObsLog
 
+### OL - GHOST “segment”
+
+Adding a new instrument to the obslog is fairly straightforward. Currently identified configuration elements consist of: number of red/blue exposures and exposure times, which instrument mode was in use.
 
 
 
--- a/bundle/edu.gemini.pot/doc/GhostEstimates.md
+++ b/bundle/edu.gemini.pot/doc/GhostEstimates.md
@@ -1,10 +1,12 @@
 # GHOST Estimations for High-Level Software
 
-GHOST brings an important new feature to the science program model, the ability to observe two targets simultaneously.  All previous instruments are limited to a single science target per observation.  The purpose of this document is to sketch out ideas for how to implement this feature along with the rest GHOST and then make time estimates for doing the work.
+GHOST brings an important new feature to the science program model, the ability to observe two targets simultaneously.  All previous instruments are limited to a single science target per observation.  The purpose of this document is to sketch out ideas for how to implement this feature along with the rest of GHOST and then make time estimates for doing the work.
 
 The GHOST concept of operations document, section 4.2 “Instrument Operating Modes”, describes a sophisticated set of modes that use the two IFUs with built in guiding capability and dedicated sky fibers in distinct ways. This prescribes a tight level of integration between target definition and instrument features that is unprecedented for Gemini instruments. A few additional facts about dual target support may be relevant to implementation ideas:
 
 * The two target mode applies only to GHOST, and only to GHOST in a particular mode of operations: Dual Target Standard Resolution.  There are other ways to use GHOST that occupy a single target, just as for all other instruments.
+
+* GHOST has two IFUs that offer distinct capabilities and distinct arrangements of sky fibers. Therefore tracking which IFU in particular will be observing a target or sky position is necessary.  That is, it isn’t correct to randomly assign targets to IFUs.
 
 * Because there are two IFUs, in beam-switching and high-resolution modes one can be observing a science target and the other taking sky background at a specific coordinate.
 
--- a/bundle/edu.gemini.pot/doc/GhostEstimates.md
+++ b/bundle/edu.gemini.pot/doc/GhostEstimates.md
@@ -1,4 +1,4 @@
-# GHOST Dual Target Mode
+# GHOST Estimations for High-Level Software
 
 GHOST brings an important new feature to the science program model, the ability to observe two targets simultaneously.  All previous instruments are limited to a single science target per observation.  The purpose of this document is to sketch out ideas for how to implement this feature without refactoring the entire program model so that time estimates for doing the work may be created.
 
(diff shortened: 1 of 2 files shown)


swalker2m:
@arturog8m  I added some guesses for how long the PIT, ITAC, OT, and ObsLog updates might take.  It totals over 4 months of effort. I could easily be convinced that we'll need half as much time or twice as much time though.  I believe the QPT comes out in the wash with all the other changes since it will only care about the base position.

I think Javier or someone with more experience should take a look at the seqexec, TCC, etc.

tpolecat:
I'll look this over again.

bundle/edu.gemini.pot/doc/GhostEstimates.md
```diff
@@ -0,0 +1,610 @@
+# GHOST Estimations for High-Level Software
+
+GHOST brings an important new feature to the science program model, the ability to observe two targets simultaneously.  All previous instruments are limited to a single science target per observation.  The purpose of this document is to sketch out ideas for how to implement this feature along with the rest GHOST and then make time estimates for doing the work.
```
tpolecat:
```
along with the rest GHOST
                   ^-- of ?
```

bundle/edu.gemini.pot/doc/GhostEstimates.md
```diff
@@ -0,0 +1,610 @@
...
+* Because there are two IFUs, in beam-switching and high-resolution modes one can be observing a science target and the other taking sky background at a specific coordinate.
+
+* The two science targets are limited to sidereal targets.  Dual non-sidereal and ToO targets are not supported.
+
+* The concept of a base position to be used for slewing and guiding calculations remains.  The base position in dual target mode is the coordinate that is equidistant from the two targets.  In other words, the point in the middle of a line connecting the two targets.
+
+* There is a minimum (84 arcseconds) and maximum (7 arcminutes) separation of the two targets.
+
+* Future instruments after GHOST are also likely to support multiple targets, each with their own rules for calculating the base position and placement.
+
```
tpolecat: It might be useful to add another point here that the IFUs are distinct and the assignment of targets to one or the other is not arbitrary. This may be obvious in retrospect but wasn't initially and it ended up driving the design.

tpolecat:
I don't see anything obvious that you have missed here, but I do think the estimates are probably optimistic. It might be useful to keep them as a target but double them on the schedule. I also think, based on my experience with the target model, that we should get started sooner rather than later since we have a hard deadline.

swalker2m:
> It might be useful to keep them as a target but double them on the schedule.

Thanks.  I agree, there should be a large contingency allowance at least.  It's very hard to predict how much time large, weakly-typed mutable model updates will take as our experience has shown.

cquiroz:
I skimmed the text and in general seems fine but I agree with @tpolecat that the estimates seems low. In particular, the UI updates generally take a lot longer

ITAC is a concern in general but if we don't need to touch the queue engine that's a big relief. I wonder what to do about the UI where you can edit targets, I guess only display the base?

bundle/edu.gemini.pot/doc/GhostEstimates.md
```diff
@@ -0,0 +1,612 @@
...
+_2 Days_
+
+Existing programs will all have a single target base position.  We’ll need migration code to replace this with a single target asterism.
+
+#### TCC Config
+
+_4 Days_
+
+TODO: It isn’t clear to me what to do in the TCC config.  There may be places where just the base coordinate is needed, and others where all targets have to be listed.  This may require coordination with TCC updates.  Input from Javier on the impact of asterisms would be appreciated.
```
jluhrs: The TCC will need only the base position to point the telescope. It is not clear to me how the guiding will work. If the TCC is expected to set the position of the guide target in the instrument, then it will need all the asterism. It will also need additional controls to enable/disable guiding with each IFU.

bundle/edu.gemini.pot/doc/GhostEstimates.md
```diff
@@ -0,0 +1,612 @@
...
+
+#### Base Target Name
+
+_1 Day_
+
+Several usages of the name of the target at the base position will have to be replaced.  Presumably a comma-separated list of target names will suffice.
+
+* OT target component uses the name of the target by default.  Some combination of multiple target names can be used when necessary presumably.
+
+* The FITS `OBJECT` header is taken from a sequence param derived from the base position target name.  Again, this presumably can be a comma separated list.
```
jluhrs: The space for this keyword is limited to 49 characters (it may be increased if the keyword description is shortened, but only up to 67 characters)

sraaphorst:
Agree with @cquiroz. I have a feeling that this will be a complete nightmare from the target editor perspective, which already required seriously ugly and extremely error-prone hacking to get it working properly with BAGS.

tpolecat:
We do have the benefit of having worked on all this stuff recently. The target and target environment models are sane now, the TPE is better, and the target/target-list editors are better. So I take a small amount of comfort in that. It's still going to be hard though.

bundle/edu.gemini.pot/doc/GhostEstimates.md
jluhrs: According to Steve Margheim, the guiding in GHOST will be controlled manually from a "GHOST Tool". GHOST only needs the RA/Dec of the guide stars, that will most likely be passed by Seqexec as part of the instrument configuration.

swalker2m:
Thinking a bit about GHOST, it seems somewhat difficult to divide up the work since the changes are going to touch so many files.  The only approach that comes to mind:

1. Write a simple quick PR with the `Asterism` trait with no real implementation.  @tpolecat  or I could do that.

2. In parallel
    a. Write the `SingleTarget` implementation and work through replacing all the existing base position references in the target environment with the asterism.  This is a big task.  Since it touches a lot of old awful code I assume I should do this.  If anybody wants to take this cup from me, I'd gladly pass it on though.

    b. Fill in the GHOST-specific asterisms and the algorithm for finding the center default base position.  This involves poring through the GHOST documentation again and defining all the GHOST model in detail.  This is independent, I think, of the `SingleTarget` and all the model updates.  I'm guessing @tpolecat  could do this if we want to make progress in parallel or I could do it after finishing step a.

3. Add the GHOST instrument component.  This should be straightforward.

4. At this point we'd be ready to start updating the OT to handle GHOST asterisms.  I'm guessing here we turn it over to our OT expert @sraaphorst , who in the meantime would be killing off amnesty tasks.

That leaves phase 1 work, templates, ITAC.  In theory someone ( @cquiroz  ?) could be progressing on the Phase 1 side of it while all of this is going on if we wanted to.

What do you guys think?

cquiroz:
Sounds good to me

arturog8m:
Sounds sensible to me. However, the phase 1 work probably needs to be done later or in a separate branch or something, as we aren't offering this in 2018A. The original plan I had was for Phase 1/ITAC mods to be started in October 2017 mostly for that reason.  I think the risks in the PIT are slightly less than in the rest of the OCS (if you think otherwise, please shout!) so I think it would be okay to stick to that schedule. However if we can afford an early start on it, by all means.

swalker2m:
Deleting this old discussion PR.

PR #1164, opened 2017-02-16
REL-3077: Remove Texes from available resource list for 2017B.

sraaphorst:
This tiny PR removed Texes from the available resources for 2017B.

I inquired with Bryan to see what should be done when upconverting proposals from previous semesters that use Texes, and he said they should simply be displayed as-is, i.e. with Texes still listed as the resource: thus, there is nothing else to do on this front other than remove it from the list.


sraaphorst:
And like an idiot, I forgot to update `RootSpec`.
Updated now. Hopefully Travis will give me the green light.

PR #1167, opened 2017-02-21
REL-2941: Fix PIT GSAOI guiding estimates and remove ODGW from search

sraaphorst:
This PR tries to fix the mess that was GSAOI estimates using `GemsStrategy`.

Firstly, it was discovered during the course of this PR that when GSAOI was selected in the PIT, guiding estimates were never being done using `GemsStrategy` due to a shortcoming in `ToContext`, which creates an `ObsContext` from the AGS query string, i.e. the AO component of the resultant `ObsContext` was never created for GSAOI, and only for Altair.

Secondly, the results for `GemsStrategy.estimate` and `GemsStrategy.select` were not consistent. It was found that these two were using different catalog search techniques for no reason Carlos or I could discern, and the `GemsStrategy.estimate` technique was returning no candidates in some cases. I switched `GemsStrategy.estimate` to use the same catalog search technique as `GemsStrategy.select` to make the results consistent with the OT AGS `select`, using the shortcut of terminating if a 3-star Canopus asterism is found.

Two test cases were added: one to ensure the use case found in testing REL-2941 that failed now passes as expected, and one to ensure a 3-star asterism can be found.

Clarifying notes below in code.

### DIFF
--- a/bundle/edu.gemini.ags.servlet/src/main/java/edu/gemini/ags/servlet/ToContext.java
+++ b/bundle/edu.gemini.ags.servlet/src/main/java/edu/gemini/ags/servlet/ToContext.java
@@ -7,10 +7,13 @@ import edu.gemini.shared.util.immutable.Option;
 
 import edu.gemini.shared.util.immutable.Some;
 import edu.gemini.spModel.core.*;
+import edu.gemini.spModel.data.AbstractDataObject;
 import edu.gemini.spModel.gemini.altair.AltairParams;
 import edu.gemini.spModel.gemini.altair.InstAltair;
+import edu.gemini.spModel.gemini.gems.Gems;
 import edu.gemini.spModel.gemini.gmos.InstGmosNorth;
 import edu.gemini.spModel.gemini.gnirs.InstGNIRS;
+import edu.gemini.spModel.gemini.gsaoi.Gsaoi;
 import edu.gemini.spModel.gemini.inst.InstRegistry;
 import edu.gemini.spModel.gemini.michelle.InstMichelle;
 import edu.gemini.spModel.gemini.michelle.MichelleParams;
@@ -183,13 +186,17 @@ public enum ToContext {
             ((PosAngleConstraintAware) inst).setPosAngleConstraint(pac);
         } // otherwise pac is ignored!
 
+        // --- instrument configurable with GeMS?
+        AbstractDataObject aoComp = null;
+        if (inst instanceof Gsaoi) {
+            aoComp = getGeMS(req);
+        }
         // --- instrument configurable with Altair?
-        InstAltair altair = null;
-        if (inst instanceof InstGNIRS ||
-            inst instanceof InstNIRI ||
-            inst instanceof InstNIFS ||
-            inst instanceof InstGmosNorth) {
-            altair = getAltair(req);
+        else if (inst instanceof InstGNIRS ||
+                 inst instanceof InstNIRI ||
+                 inst instanceof InstNIFS ||
+                 inst instanceof InstGmosNorth) {
+            aoComp = getAltair(req);
         } // otherwise altair is ignored
 
         // --- instrument is NIRI?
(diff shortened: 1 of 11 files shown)


bundle/edu.gemini.ags.servlet/src/main/java/edu/gemini/ags/servlet/ToContext.java
```diff
@@ -183,13 +186,17 @@ public ObsContext apply(HttpServletRequest req) throws RequestException {
             ((PosAngleConstraintAware) inst).setPosAngleConstraint(pac);
         } // otherwise pac is ignored!
 
+        // --- instrument configurable with GeMS?
+        AbstractDataObject aoComp = null;
+        if (inst instanceof Gsaoi) {
+            aoComp = getGeMS(req);
+        }
```
sraaphorst: If F2 + GeMS is ever commissioned, we will have to figure out how to handle this, as `estimate` always picks the first available strategy in a prioritized list of strategies for an instrument.

bundle/edu.gemini.ags.servlet/src/main/java/edu/gemini/ags/servlet/ToContext.java
```diff
@@ -210,7 +217,12 @@ public ObsContext apply(HttpServletRequest req) throws RequestException {
 
         final TargetObsComp toc = new TargetObsComp();
         toc.setTargetEnvironment(env);
-        return ObsContext.create(env, inst, site, conds, Collections.emptySet(), altair, new Some<>(sb));
+        return ObsContext.create(env, inst, site, conds, Collections.emptySet(), aoComp, new Some<>(sb));
+    }
+
+    private Gems getGeMS(final HttpServletRequest req) throws RequestException {
+        final Gems gems = new Gems();
+        return gems;
```
sraaphorst: We now create an AO component for GSAOI.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/GemsStrategy.scala
```diff
@@ -42,52 +41,6 @@ trait GemsStrategy extends AgsStrategy {
...
-          angle                                                     <- anglesToTry
-        } yield {
-          val ConeSearchCatalogQuery(_, _, radiusConstraint, mc, _) = result.query
-          val catalogSearchCriterion     = CatalogSearchCriterion("ags", radiusConstraint, mc.head, None, angle.some)
-          val gemsCatalogSearchCriterion = new GemsCatalogSearchCriterion(result.searchKey, catalogSearchCriterion)
-          new GemsCatalogSearchResults(gemsCatalogSearchCriterion, result.catalogResult.targets.rows)
-        }
-    }
-  }
-
```
sraaphorst: All this code comprised the duplicate catalog search method for GeMS and could thus be removed by using the same `search` method for `estimate` and `select`.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/GemsStrategy.scala
```diff
@@ -138,19 +91,19 @@ trait GemsStrategy extends AgsStrategy {
   }
 
   override def estimate(ctx: ObsContext, mt: MagnitudeTable)(ec: ExecutionContext): Future[Estimate] = {
-    // Get the query results and convert them to GeMS-specific ones.
-    val results = toGemsCatalogSearchResults(ctx, catalogResult(ctx, mt)(ec))(ec)
-
     // Create a set of the angles to try.
-    val anglesToTry = (0 until 360 by 45).map(Angle.fromDegrees(_)).toSet
+    val anglesToTry = (0 until 360 by 90).map(Angle.fromDegrees(_)).toSet
```
sraaphorst: Changed to be consistent with `select`, which goes by 90 degree angles.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/GemsStrategy.scala
```diff
@@ -138,19 +91,19 @@ trait GemsStrategy extends AgsStrategy {
...
 
-    // Iterate over 45 degree position angles if no 3-star asterism is found at PA = 0.
+    // Iterate over 90 degree position angles if no 3-star asterism is found at PA = 0.
     val gemsCatalogResults = results.map(result => GemsResultsAnalyzer.analyzeGoodEnough(ctx, anglesToTry, result, _.stars.size < 3))
 
     // We only want Canopus targets, so filter to those and then determine if the asterisms are big enough.
     gemsCatalogResults.map { ggsLst =>
-      val largestAsterism = ggsLst.map(_.guideGroup.grp.toManualGroup.targetMap.keySet.intersection(GemsStrategy.canopusProbes).size).max
-      Estimate(largestAsterism / 3.0)
+      val largestAsterism = ggsLst.map(_.guideGroup.grp.toManualGroup.targetMap.keySet.intersection(GemsStrategy.canopusProbes).size).fold(0)(math.max)
```
sraaphorst: Fixed calling `max` on possibly empty `List` here.

PR #1168, opened 2017-02-23
REL-3079: Altair LGS not available in 2017B.

sraaphorst:
The laser at GN is not currently functional and thus proposals for 2017B must not allow Altair LGS options to be chosen.

This PR does three things:
1. Remove all Altair LGS options from the resource configuration.
2. When upconverting from a former semester, issue a message if an instrument is configured to use Altair LGS.
3. If a resource is found to use Altair LGS, report an error.

### DIFF
--- a/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/transform/UpConverter.scala
+++ b/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/transform/UpConverter.scala
@@ -131,6 +131,12 @@ case class LastStepConverter(semester: Semester) extends SemesterConverter {
  * This converter will upgrade to 2017B
  */
 case object SemesterConverter2017ATo2017B extends SemesterConverter {
+  // TODO: REL-3079 remove this when the GN laser is repaired.
+  val altairLgsNotAvailable: TransformFunction = {
+    case a @ <altair>{ns @ _*}</altair> if (a \\ "lgs").nonEmpty =>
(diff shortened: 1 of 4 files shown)


bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/transform/UpConverter.scala
```diff
@@ -131,6 +131,12 @@ case class LastStepConverter(semester: Semester) extends SemesterConverter {
  * This converter will upgrade to 2017B
  */
 case object SemesterConverter2017ATo2017B extends SemesterConverter {
+  // TODO: REL-3079 remove this when the GN laser is repaired.
+  val altairLgsNotAvailable: TransformFunction = {
+    case a @ <altair>{ns @ _*}</altair> if (a \\ "lgs").nonEmpty =>
+      StepResult(s"Altair Laser Guidestar is currently not offered", a).successNel
```
cquiroz: Will this just remove the node right?

sraaphorst: It doesn't actually remove the node at all: it just issues the message.
We want to keep the node in place for now and put an error in the problems section, preventing submission until the user manually fixes the problem.

cquiroz: Excellent 👍

sraaphorst:
I added REL-3080 to track these changes to ITAC.

PR #1172, opened 2017-02-27
REL-2926: Fixed NIFS to work with Altair in calculating overheads.

sraaphorst:
I had accidentally omitted NIFS + Altair from calculations of overheads.

As these are the same as standard NIFS, this simply switched to doing the overheads lookup using the blueprint superclass, `NifsBlueprintBase`.

### DIFF
--- a/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/overheads/Overheads.scala
+++ b/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/overheads/Overheads.scala
@@ -97,7 +97,7 @@ object Overheads extends (BlueprintBase => Option[Overheads]) {
     case _: GnirsBlueprintImaging       => SimpleOverheads(0.10, 10, 0.193).some
     case _: GnirsBlueprintSpectroscopy  => SimpleOverheads(0.25, 15, 0.080).some
 
-    case _: NifsBlueprint               => SimpleOverheads(0.25, 11, 0.175).some
+    case _: NifsBlueprintBase           => SimpleOverheads(0.25, 11, 0.175).some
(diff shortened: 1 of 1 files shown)


bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/overheads/Overheads.scala
```diff
@@ -97,7 +97,7 @@ object Overheads extends (BlueprintBase => Option[Overheads]) {
     case _: GnirsBlueprintImaging       => SimpleOverheads(0.10, 10, 0.193).some
     case _: GnirsBlueprintSpectroscopy  => SimpleOverheads(0.25, 15, 0.080).some
 
-    case _: NifsBlueprint               => SimpleOverheads(0.25, 11, 0.175).some
+    case _: NifsBlueprintBase           => SimpleOverheads(0.25, 11, 0.175).some
```
cquiroz: So, this would apply to NIFS with or without Altair?

sraaphorst: Yes, that's right. There are two subclasses of this, namely `NifsBlueprint` and `NifsBlueprintAO`. Before, I was just matching on `NifsBlueprint` so `NifsBlueprintAo` was not returning anything in the overhead lookup. Now it will return these `SimpleOverheads` for both.

tpolecat: I wonder if it would be more clear to add two cases here rather than matching the superclass. I don't feel strongly about it though. This match is very brittle, sadly.

PR #1173, opened 2017-02-28
REL-2941: Addition of dedicated AGS ODB server for PIT estimates

sraaphorst:
This adds the configuration for the dedicated AGS ODB server and configures the PIT to use it.

### DIFF
--- a/app/spdb/build.sbt
+++ b/app/spdb/build.sbt
@@ -36,7 +36,8 @@ ocsAppManifest := {
               gnodbtest(v),
             odbproduction(v),
               gsodb(v),
-              gnodb(v)
+              gnodb(v),
+	      gnagsodb(v)
     )
   )
 }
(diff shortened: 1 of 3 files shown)


app/spdb/build.sbt
```diff
@@ -502,6 +503,30 @@ def gnodb(version: Version) = AppConfig(
   )
 ) extending List(odbproduction(version), gnodb_credentials(version))
 
+// GN AGS ODB: hbfauxodb-lv1 / gnauxodb
+def gnagsodb(version: Version) = AppConfig(
```
cquiroz: Is there any risk on having an extra ODB, I guess it makes me nervous to have the actual dirs and servers on the props here. ODB may actively do something on those servers?

sraaphorst: What kind of risk are you thinking?

This is actually necessary because the IP address needs to be set as it is in the configuration or the ODB will not allow connections.

I used the same configuration for most things as gnodbtest, which shouldn't do anything with external effects, and we won't have a crontab file launching any services, so I don't foresee any issues.

cquiroz: Right, no cronjobs, that would do. And I guess datamanager is off too

sraaphorst: Yes, the frameworklevel has to be set to 100 for the datamanager to run, and that must be manually done.

PR #1174, opened 2017-02-28
REL-2941: Remove the GSAOI ODGW search and selection capabilities from the OT and PIT

cquiroz:
SSL dark magic


tpolecat:
I haven't looked at Let's Encrypt … does this let us generate our own app-specific certs?

cquiroz:
ITS uses let's encrypt for the internet facing proxy servers. They generated one for us to use in AGS but java doesn't have it on the root store

On this PR I update the custom keystore used by the ocs to include Let's encrypt intermediate cert and trust it to close the TSL handshake

cquiroz:
In the name of expediency I'll override the travis build and merge

PR #1175, opened 2017-02-28
Feature/rel 3058

andrewwstephens:
This PR simply updates the ITC web form so that the default CCD option for GMOS-N is Hamamatsu.  I  rearranged the options for both GMOS-N and GMOS-S so that the default (Hamamatsu) is first on both for consistency.  I also removed the static read noise and dark current values from both GMOS-N and GMOS-S forms since those are detector-specific.  The correct information is displayed in the ITC output.


cquiroz:
Superseded by #1176

PR #1183, opened 2017-03-14
REL-3061: Rewrote ImEither and fixed null pointer exceptions.

sraaphorst:
This PR consists of two basic elements:

1. Rewriting the `ImEither` class as per @tpolecat's earlier recommendation to change `ImEither` into an interface with implementations `Left` and `Right` instead of the previous way, which included two `Options` in a concrete `ImEither` class. This would run the risk of having an `ImEither` with both `left` and `right` values set to `None`, which violates how this class is supposed to function.

2. There was an issue in `EdCompTargetList` where it was extremely rare, but it was possible that the `ImEither` in this class representing the selection of a target or group could be `null`. I have changed the variable to an `Option<ImEither>` instead to avoid this and eliminate the problem.

Note that the `ImEither` is still right-biased, as is the case with the scalaz disjunction, but for convenience, I have included methods specifically for left access so that we don't have to call `swap` to access left. If the sentiment is that this is not a good idea, I can change it back to be right-biased and drop the left-based operations / qualifiers.

### DIFF
--- a/bundle/edu.gemini.shared.util/src/main/java/edu/gemini/shared/util/immutable/ImEither.java
+++ b/bundle/edu.gemini.shared.util/src/main/java/edu/gemini/shared/util/immutable/ImEither.java
@@ -9,29 +9,45 @@ import java.util.Optional;
 public interface ImEither<L,R> extends Serializable {
     ImEither<R,L> swap();
 
-    // Right biased qualifiers / operations.
+    // === Right biased qualifiers / operations ===
     <T> T foldRight(final T zero,
                     final Function<? super R, ? extends T> rightFunc);
+
     boolean exists(final Predicate<? super R> rightFunc);
+
     boolean forAll(final Predicate<? super R> rightFunc);
+
     void forEach(final Consumer<? super R> rightFunc);
+
     <R2> ImEither<L,R2> map(final Function<? super R, ? extends R2> rightFunc);
+
     <R2> ImEither<L,R2> flatMap(final Function<? super R, ? extends ImEither<L,R2>> rightFunc);
+
     boolean isRight();
+
     Optional<R> toOptional();
 
-    // Left biased qualifiers / operations.
+
+    // === Left biased qualifiers / operations ===
     <T> T foldLeft(final T zero,
                    final Function<? super L, ? extends T> leftFunc);
+
     boolean existsLeft(final Predicate<? super L> leftFunc);
+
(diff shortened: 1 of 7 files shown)


bundle/edu.gemini.shared.util/src/main/java/edu/gemini/shared/util/immutable/ImEither.java
```diff
@@ -1,120 +1,47 @@
...
-    }
-
-    public Option<R> toOption() {
-        return right;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.function.Predicate;
+import java.util.Optional;
+
+public interface ImEither<L,R> extends Serializable {
```
cquiroz: Just a silly formatting note, I guess we normally leave a blank line between method definitions

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
```diff
@@ -104,6 +104,8 @@ public void textBoxKeyPress(final TextBoxWidget tbwe) {
 
         _agsPub.subscribe((obs, oldOptions, newOptions) -> updateGuiding());
 
+        _curSelection = None.instance();
```
cquiroz: I guess this is the actual bug fix

sraaphorst: Yes, this coupled with changing `_curSelection` from an `ImEither` to an `Option<ImEither>` and then using it in that way so that we never have to deal with `null`, which came up as a very rare thing.

sraaphorst:
`ImEither` is actually pretty new... I just wrote it while making the BAGS changes to facilitate, as previously instead we had two variables and just relied on one of them always being `null` and the code being maintained properly in the hopes that that was the case (which, I'll add, it wasn't) 😸

bundle/edu.gemini.shared.util/src/main/java/edu/gemini/shared/util/immutable/ImEither.java
sraaphorst: I'm happy to change it for consistency. I was just trying to keep left-based and right-based qualifiers / operators grouped together for easy visibility. I'll add the spaces.

PR #1184, opened 2017-03-15
REL-3101: General code / editors cleanup

sraaphorst:
There is a ton of cruft in the OT editors, especially from our former use of JFormDesigner.

This cleans up a number of editors and a bit of general code.


cquiroz:
How do we have a commit with author gemini-hlswbuild?

sraaphorst:
This seems to be a configuration problem on my new MacBook... I didn't notice until after I made the PR and have no idea what is misconfigured, but I tried creating new branches and the same thing is happening.

I am going through some online help pages now to try to fix this, but I'm not sure how this is going on, since I don't even have the gemini-hlswbuild credentials stored on this machine as far as I know.

sraaphorst:
There we go. I had to do a:
`sudo git config --system --unset credential.helper`
and now I seem to be committing and pushing properly.
So strange.

PR #1185, opened 2017-03-21
REL-3055: Set HAMAMATSU to default for GMOS-N static component.

sraaphorst:
Previously, E2V was the default CCD manufacturer for GMOS-N.

This was set in `InstGmosCommon` and then overridden in `InstGmosSouth` to set the CCD manufacturer to HAMAMATSU.

Since now both sites will have HAMAMATSU as the default CCD manufacturer, the default in `InstGmosCommon` was changed from E2V to HAMAMATSU and the (no longer necessary) override removed from `InstGmosSouth`.


sraaphorst:
The issue seems to be that the protocol used to access HORIZONS was http, and should have been https. I switched that in `CgiHorizonsConstants` and now everything works as it should.

Tested it locally, pushed a commit for it, and will merge it once Travis says yes.

Possibly should have made a JIRA task specifically for this, but it seems like a lot of time wasting for a one character addition, so I'm not going to bother.

cquiroz:
👍

PR #1188, opened 2017-03-23
REL-1787: Add ability to import timing windows

sraaphorst:
In the OT, previously, timing windows needed to be added in the Observation Conditions (Site Quality) node one at a time using the following dialog:

add

In the case where there were many timing windows, this could become very tedious, so REL-1787 requested the ability to import timing windows from a text file and add them to the existing timing windows.

I added an "Import" button to the UI, as you can see under the cursor in the following screen shot:

import button

This pops up a modal file chooser dialog that allows the user to select a timing window file (.tw):

import dialog

The file is then parsed using Scala parser combinators in the `TimingWindowImporter.TimingWindowParser` object, and the parsed timing windows added to the list. If any lines fail to parse, a dialog containing a table with line numbers and line contents is displayed.

Here is an example of an input file (note that I had to rename it with a `txt` file extension to get it to attach here in github) and the results of the parse, which contains failures:

[timingwindows.txt](https://github.com/gemini-hlsw/ocs/files/864751/timingwindows.txt)

import results

I also took this opportunity to clean up the code considerably:

1. Now that the site quality (of which timing window classes are a subset) have increased in number, I created a new package, namely `jsky.app.ot.gemini.editor.sitequality`, to group these files together.

2. I performed a massive cleanup of the `TimingWindowDialog` to bring it up-to-date with current Java coding practices and make it much more readable and encapsulate members properly. There were also a few small bugs in this (e.g. modality was global instead of simply limited to the OT) that I fixed as well during the cleanup.

3. I moved a couple of classes with reusability potential that were not being exploited to shared packages. Most notably, the PIT contained a very handy `Chooser` class that functions as a file chooser, but additionally has the benefit of using preferences to store / retrieve (for each type of file chooser) the last directory accessed. Additionally, the `HourMinuteFormat` was moved due to its possible utility, but as I ended up performing parsing of date-time using the new Java time API instead of the old Java formatter classes, I did not end up using this.

4. I simplified and cleaned up the classes that I needed to use / modify during this task. I will include notes in specific code changes to elaborate as necessary.

5. Most importantly, I added the `TimingWindowImporter`, which contains the UI elements and parser needed for the parsing, and the `TimingWindowImporterSpec` which tests the timing window parser.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/obscomp/SPSiteQuality.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/obscomp/SPSiteQuality.java
@@ -121,7 +121,7 @@ public class SPSiteQuality extends AbstractDataObject implements PropertyProvide
 
     }
 
-    static class TimingWindowList extends LinkedList<TimingWindow> {
+    private static class TimingWindowList extends LinkedList<TimingWindow> {
 
         private static final long serialVersionUID = 2L;
         private static final String NAME = "timing-window-list";
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/obscomp/SPSiteQuality.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/obscomp/SPSiteQuality.java
@@ -18,6 +18,8 @@ import edu.gemini.spModel.type.SpTypeUtil;
 import java.beans.PropertyDescriptor;
 import java.io.Serializable;
 import java.util.*;
+import java.util.function.Function;
+import java.util.function.Supplier;
 
 /**
  * Site Quality observation component.
@@ -666,24 +668,30 @@ public class SPSiteQuality extends AbstractDataObject implements PropertyProvide
         return Collections.unmodifiableList(_timingWindows);
     }
 
-    public void setTimingWindows(List<TimingWindow> windows) {
-        List<TimingWindow> prev = Collections.unmodifiableList(new ArrayList<>(_timingWindows));
-        _timingWindows.clear();
-        _timingWindows.addAll(windows);
-        firePropertyChange(TIMING_WINDOWS_PROP.getName(), prev, getTimingWindows());
-    }
-
-    public void addTimingWindow(TimingWindow tw) {
-        List<TimingWindow> prev = Collections.unmodifiableList(new ArrayList<>(_timingWindows));
-        _timingWindows.add(tw);
-        firePropertyChange(TIMING_WINDOWS_PROP.getName(), prev, getTimingWindows());
-    }
-
-    public void removeTimingWindow(TimingWindow tw) {
-        List<TimingWindow> prev = Collections.unmodifiableList(new ArrayList<>(_timingWindows));
-        if (_timingWindows.remove(tw)) {
+    // Common method to make changes to the timing window list and then fire a property change if appropriate.
+    private void changeTimingWindows(final Supplier<Boolean> changer) {
+        final List<TimingWindow> prev = Collections.unmodifiableList(new ArrayList<>(_timingWindows));
+        if (changer.get())
             firePropertyChange(TIMING_WINDOWS_PROP.getName(), prev, getTimingWindows());
-        }
+    }
+
+    public void setTimingWindows(final List<TimingWindow> tws) {
+        changeTimingWindows(() -> {
+            _timingWindows.clear();
+            return _timingWindows.addAll(tws);
+        });
+    }
+
+    public void addTimingWindows(final List<TimingWindow> tws) {
+        changeTimingWindows(() -> _timingWindows.addAll(tws));
+    }
+
+    public void addTimingWindow(final TimingWindow tw) {
+        changeTimingWindows(() -> _timingWindows.add(tw));
+    }
+
+    public void removeTimingWindow(final TimingWindow tw) {
+        changeTimingWindows(() -> _timingWindows.remove(tw));
     }
 
     /*
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/obscomp/SPSiteQuality.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/obscomp/SPSiteQuality.java
@@ -14,11 +14,11 @@ import edu.gemini.spModel.type.DisplayableSpType;
 import edu.gemini.spModel.type.ObsoletableSpType;
 import edu.gemini.spModel.type.SequenceableSpType;
 import edu.gemini.spModel.type.SpTypeUtil;
+import jdk.nashorn.internal.runtime.Timing;
 
 import java.beans.PropertyDescriptor;
 import java.io.Serializable;
 import java.util.*;
-import java.util.function.Function;
 import java.util.function.Supplier;
 
 /**
@@ -62,7 +62,7 @@ public class SPSiteQuality extends AbstractDataObject implements PropertyProvide
         private final long start, duration, period;
         private final int repeat;
 
-        public TimingWindow(long start, long duration, int repeat, long period) {
+        public TimingWindow(final long start, final long duration, final int repeat, final long period) {
             this.start = start;
             this.duration = duration;
             this.repeat = repeat;
@@ -74,7 +74,7 @@ public class SPSiteQuality extends AbstractDataObject implements PropertyProvide
             this(System.currentTimeMillis(), 24 * MS_PER_HOUR, 0, 0);
         }
 
-        TimingWindow(ParamSet params) {
+        TimingWindow(final ParamSet params) {
             this(Pio.getLongValue(params, START_PROP, 0),
                  Pio.getLongValue(params, DURATION_PROP, 0),
                  Pio.getIntValue(params, REPEAT_PROP, 0),
@@ -97,8 +97,8 @@ public class SPSiteQuality extends AbstractDataObject implements PropertyProvide
             return start;
         }
 
-        ParamSet getParamSet(PioFactory factory) {
-            ParamSet params = factory.createParamSet(NAME);
+        ParamSet getParamSet(final PioFactory factory) {
+            final ParamSet params = factory.createParamSet(NAME);
             Pio.addLongParam(factory, params, START_PROP, start);
             Pio.addLongParam(factory, params, DURATION_PROP, duration);
             Pio.addIntParam(factory, params, REPEAT_PROP, repeat);
@@ -114,11 +114,21 @@ public class SPSiteQuality extends AbstractDataObject implements PropertyProvide
         public TimingWindow clone() {
             try {
                 return (TimingWindow) super.clone();
-            } catch (CloneNotSupportedException e) {
+            } catch (final CloneNotSupportedException e) {
                 throw new Error("This was supposed to be impossible.");
             }
         }
 
+        @Override
+        public boolean equals(final Object other) {
+            if (other == null || !(other instanceof TimingWindow))
+                return false;
+            final TimingWindow tw = (TimingWindow) other;
+            return  start == tw.start &&
+                    duration == tw.duration &&
+                    period == tw.period &&
+                    repeat == tw.repeat;
+        }
     }
 
     private static class TimingWindowList extends LinkedList<TimingWindow> {
@@ -126,34 +136,30 @@ public class SPSiteQuality extends AbstractDataObject implements PropertyProvide
         private static final long serialVersionUID = 2L;
         private static final String NAME = "timing-window-list";
 
-        ParamSet getParamSet(PioFactory factory) {
-            ParamSet params = factory.createParamSet(NAME);
-            for (TimingWindow tw: this)
-                params.addParamSet(tw.getParamSet(factory));
+        ParamSet getParamSet(final PioFactory factory) {
+            final ParamSet params = factory.createParamSet(NAME);
+            forEach(tw -> params.addParamSet(tw.getParamSet(factory)));
             return params;
         }
 
-        void setParamSet(ParamSet params) {
+        void setParamSet(final ParamSet params) {
             clear();
             if (params != null) {
-                for (ParamSet ps: params.getParamSets())
-                    add(new TimingWindow(ps));
+                params.getParamSets().forEach(ps -> add(new TimingWindow(ps)));
             }
         }
 
         @Override
         public TimingWindowList clone() {
-            TimingWindowList ret = new TimingWindowList();
-            for (TimingWindow tw: this)
-                ret.add(tw.clone());
+            final TimingWindowList ret = new TimingWindowList();
+            forEach(tw -> ret.add(tw.clone()));
             return ret;
         }
-
     }
 
     @Override
     public SPSiteQuality clone() {
-        SPSiteQuality ret = (SPSiteQuality) super.clone();
+        final SPSiteQuality ret = (SPSiteQuality) super.clone();
         ret._timingWindows = _timingWindows.clone();
         return ret;
     }
@@ -162,7 +168,7 @@ public class SPSiteQuality extends AbstractDataObject implements PropertyProvide
         byte getPercentage();
     }
 
-    private static <T extends PercentageContainer> Option<T> read(String strVal, T[] values) {
+    private static <T extends PercentageContainer> Option<T> read(final String strVal, final T[] values) {
         if (strVal.length() < 3) return None.instance();
 
         String tail = strVal.substring(2);
@@ -175,7 +181,7 @@ public class SPSiteQuality extends AbstractDataObject implements PropertyProvide
             return None.instance();
         }
 
-        for (T val : values) {
+        for (final T val : values) {
             if (val.getPercentage() == perc) return new Some<>(val);
         }
         return None.instance();
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/obscomp/SPSiteQuality.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/obscomp/SPSiteQuality.java
@@ -14,7 +14,6 @@ import edu.gemini.spModel.type.DisplayableSpType;
 import edu.gemini.spModel.type.ObsoletableSpType;
 import edu.gemini.spModel.type.SequenceableSpType;
 import edu.gemini.spModel.type.SpTypeUtil;
-import jdk.nashorn.internal.runtime.Timing;
 
 import java.beans.PropertyDescriptor;
 import java.io.Serializable;
--- a/bundle/edu.gemini.shared.util/src/main/java/edu/gemini/shared/util/HourMinuteFormat.java
+++ b/bundle/edu.gemini.shared.util/src/main/java/edu/gemini/shared/util/HourMinuteFormat.java
@@ -1,0 +1,58 @@
+package edu.gemini.shared.util;
+
+import java.text.FieldPosition;
+import java.text.Format;
+import java.text.ParsePosition;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class HourMinuteFormat extends Format {
+    private static final long MS_PER_MINUTE = 1000 * 60;
+    private static final long MS_PER_HOUR = MS_PER_MINUTE * 60;
+    private static final Pattern HH_MM_PAT = Pattern.compile("^(\\d+):(\\d{2})$");
+    private static final Pattern HH_MM_SS_PAT = Pattern.compile("^(\\d+):(\\d{2}):(\\d{2})$");
+
+    public static final HourMinuteFormat HH_MM    = new HourMinuteFormat(false);
+    public static final HourMinuteFormat HH_MM_SS = new HourMinuteFormat(true);
+
+    private final boolean showSeconds;
+
+    private HourMinuteFormat(final boolean showSeconds) {
+        this.showSeconds = showSeconds;
+    }
+
+    @Override
+    public StringBuffer format(final Object obj, final StringBuffer toAppendTo, final FieldPosition pos) {
+
+        final long ms = (Long) obj;
+        final long hh = ms / MS_PER_HOUR;
+        final long mm = (ms % MS_PER_HOUR) / MS_PER_MINUTE;
+
+        final String s;
+        if (showSeconds) {
+            final long ss = (ms % MS_PER_MINUTE) / 1000;
+            s = String.format("%02d:%02d:%02d", hh, mm, ss);
+        } else {
+            s = String.format("%02d:%02d", hh, mm);
+        }
+
+        return toAppendTo.append(s);
+    }
+
+    @Override
+    public Object parseObject(final String source, final ParsePosition pos) {
+        final Pattern p = showSeconds ? HH_MM_SS_PAT : HH_MM_PAT;
+        final Matcher m = p.matcher(source);
+        if (!m.matches()) {
+            pos.setIndex(0);
+            pos.setErrorIndex(0);
+            return null;
+        }
+
+        final long hh = Long.parseLong(m.group(1));
+        final long mm = Long.parseLong(m.group(2));
+        final long ss = showSeconds ? Long.parseLong(m.group(3)) : 0;
+        pos.setIndex(m.end());
+        return hh * MS_PER_HOUR + mm * MS_PER_MINUTE + ss * 1000;
+    }
+}
--- a/bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/sitequality/SiteQualityPanel.java
+++ b/bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/sitequality/SiteQualityPanel.java
@@ -294,21 +294,21 @@ class TimingWindowTableModel extends DefaultTableModel implements PropertyChange
 	private static final long MS_PER_MINUTE = MS_PER_SECOND * 60;
 	private static final long MS_PER_HOUR = MS_PER_MINUTE * 60;
 
-    private static final SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss z");
-    static {
-    	dateFormat.setTimeZone(TimeZone.getTimeZone("UTC"));
-    }
+    private static final SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss z") {{
+    	setTimeZone(TimeZone.getTimeZone("UTC"));
+    }};
 
 	private SPSiteQuality sq;
 
-	void setSiteQuality(SPSiteQuality siteQuality) {
+	void setSiteQuality(final SPSiteQuality siteQuality) {
 		if (sq != null) sq.removePropertyChangeListener(this);
 		sq = siteQuality;
 		if (sq != null) sq.addPropertyChangeListener(this);
 		fireTableDataChanged();
 	}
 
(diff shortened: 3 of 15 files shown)


bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/obscomp/SPSiteQuality.java
```diff
@@ -112,46 +113,52 @@ public String toString() {
...
             try {
                 return (TimingWindow) super.clone();
-            } catch (CloneNotSupportedException e) {
+            } catch (final CloneNotSupportedException e) {
                 throw new Error("This was supposed to be impossible.");
             }
         }
 
+        @Override
+        public boolean equals(final Object other) {
```
sraaphorst: I needed to add an `equals` method to `TimingWindow` for the `TimingWindowImporterSpec` comparisons to work properly.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/obscomp/SPSiteQuality.java
```diff
@@ -666,24 +673,30 @@ public void setElevationConstraintMax(double max) {
         return Collections.unmodifiableList(_timingWindows);
     }
 
-    public void setTimingWindows(List<TimingWindow> windows) {
-        List<TimingWindow> prev = Collections.unmodifiableList(new ArrayList<>(_timingWindows));
-        _timingWindows.clear();
-        _timingWindows.addAll(windows);
-        firePropertyChange(TIMING_WINDOWS_PROP.getName(), prev, getTimingWindows());
+    // Common method to make changes to the timing window list and then fire a property change if appropriate.
+    private void changeTimingWindows(final Supplier<Boolean> changer) {
```
sraaphorst: There was quite a bit of repeated code in the timing window modifiers, so I abstracted this out to minimize and simplify.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/TimingWindowDialog.java
```diff
@@ -1,381 +0,0 @@
-package jsky.app.ot.gemini.editor;
```
sraaphorst: Moved to the new package and heavily modified (although the appearance is the same).

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/sitequality/SiteQualityPanel.java
```diff
@@ -166,43 +164,59 @@
...
+						}));
                     ButtonFlattener.flatten(this);
 				}});
 
+                add(new JButton(Resources.getIcon("eclipse/download.gif")) {{
+                    setToolTipText("Import timing windows from text file");
+                    setFocusable(false);
+                    addActionListener(e -> {
+                        final TimingWindowImporter importer = new TimingWindowImporter(SiteQualityPanel.this);
+						final TimingWindowParser.TimingWindowParseResults results = importer.openImport();
```
sraaphorst: No idea why this line looks so indented: it appears normal in IDEA for me.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/sitequality/TimingWindowDialog.java
```diff
@@ -0,0 +1,285 @@
...
+import java.awt.*;
+import java.awt.event.FocusAdapter;
+import java.awt.event.FocusEvent;
+import java.awt.event.KeyAdapter;
+import java.awt.event.KeyEvent;
+import java.text.*;
+import java.util.Date;
+import java.util.TimeZone;
+
+class TimingWindowDialog extends JDialog {
```
sraaphorst: The moved and heavily modified `TimingWindowDialog` class.
I think it is much cleaner now, and modality works as one would expect.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/sitequality/TimingWindowDialog.java
```diff
@@ -0,0 +1,285 @@
...
+            field.addKeyListener(new KeyAdapter() {
+                @Override
+                public void keyReleased(KeyEvent e) {
+                    updateEnabledState();
+                }
+            });
+
+            // Whenever one of the fields loses focus, its contents are automatically set to the last valid edit,
+            // so the OK button should be enabled.
+            field.addFocusListener(new FocusAdapter() {
```
sraaphorst: This is new: the enabled status of the OK button relies on the contents of all `JFormattedTextFields` being acceptable, but the method to trigger an update on this only occurred when a `JFormattedTextField` received a `KeyListener.keyReleased` event.

This was not sufficient, as if a `JFormattedTextField` loses focus when it contains illegal contents, the contents are reset to their previous legal value; thus, to make the OK button enabled properly, we also need to enable it whenever a `JFormattedTextField` loses focus.

bundle/jsky.app.ot/src/main/resources/resources/conf/UIInfo.xml
```diff
@@ -105,7 +105,7 @@
             type="siteQuality"
             imageKey="component.gif"
             shortDescription="This component describes the conditions under which this observation can be observed."
-            uiClassName="jsky.app.ot.gemini.editor.EdCompSiteQuality"/>
+            uiClassName="jsky.app.ot.gemini.editor.sitequality.EdCompSiteQuality"/>
```
cquiroz: Do we still use this crap? What happens if you delete the xml?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/sitequality/TimingWindowImporter.scala
```diff
@@ -0,0 +1,213 @@
...
+
+  private def hhmmParser: Parser[Long] = (arbitraryDigits <~ ":") ~ timeDigits ^^ {
+    case hh ~ mm => hhmmssToLong(hh, mm)
+  }
+
+  // Parse number of repetitions for a timing window:
+  // -1  = infinite repetitions.
+  //  0  = no repetitions.
+  // x>0 = x repetitions.
+  private def repetitionsParser: Parser[Int] = """(-1)|0|([1-9]\d*)""".r ^^ { _.toInt }
```
cquiroz: Should me support constants on the file format?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/sitequality/TimingWindowImporter.scala
```diff
@@ -0,0 +1,213 @@
...
+import scala.collection.JavaConverters._
+import scala.util.parsing.combinator.RegexParsers
+import scalaz._
+import Scalaz._
+
+
+// Prompt the user for a file and then parse it and return a list of TimingWindows.
+// Since the calling class is Java and uses Java components, to get the modality right, we have to use Java Swing / AWT.
+// If we try to wrap the AWT component in a Scala Swing component, we are globally modal, which is not what we want.
+class TimingWindowImporter(owner: Component) {
```
cquiroz: Very nice

bundle/jsky.app.ot/src/test/scala/jsky/app/ot/gemini/editor/sitequality/TimingWindowImporterSpec.scala
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    "ignore empty lines and comments" in {
+      val lines = List("", "  ", "#", "# Comment", " # abc ", "    ## nothing here")
+      val results = TimingWindowParser.parseTimingWindows(lines)
+      (results.successes.isEmpty must beTrue) and (results.failures.isEmpty must beTrue)
+    }
+
+    "create timing windows from complete data, ignoring whitespace" in {
+      val t1Tup = ("2018-01-01", "01:02:03", "04:05", 1.some, "01:02:03".some)
```
cquiroz: does it make sense to filter e.g. dates on the past? Would that be an error?

bundle/jsky.app.ot/src/test/scala/jsky/app/ot/gemini/editor/sitequality/TimingWindowImporterSpec.scala
```diff
@@ -0,0 +1,110 @@
...
+      val t2Tup = ("   2018-01-02  ", "  02:03:04   ", "   06:07  ", 2.some, "   03:04:05    ".some)
+      val (successes, _, results) = runParsing(t1Tup.right, t2Tup.right)
+      (successes must === (results.successes)) and (results.failures.isEmpty must beTrue)
+    }
+
+    "identify a success and a failure" in {
+      val t1Tup = ("2017-06-10", "23:59:59", "48:00", 3.some, "123:45:00".some)
+      val fail1 = "Ceci n'est pas un timing window."
```
cquiroz: Why the Québécois?

cquiroz:
Wow great job, this is probably going to be very useful for our users. Do we have to do this in other parts, like the PIT or QPT?

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/sitequality/SiteQualityPanel.java
sraaphorst: It is convention for us to use four spaces and not tabs, right?
I was prompted with some of these older files when modifying them, and I always chose four spaces. I'll see if I can figure it out.

sraaphorst:
This shouldn't affect any other parts of the software, and I think this is the only place where timing windows are entered, as they are defined in `jsky.app.ot`. At least, according to the JIRA task, the change was only requested in the OT.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/sitequality/SiteQualityPanel.java
cquiroz: I strongly prefer four spaces 😄

bundle/edu.gemini.shared.util/src/main/java/edu/gemini/shared/util/HourMinuteFormat.java
```diff
@@ -0,0 +1,58 @@
+package edu.gemini.shared.util;
+
+import java.text.FieldPosition;
+import java.text.Format;
+import java.text.ParsePosition;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class HourMinuteFormat extends Format {
```
sraaphorst: Bryan and Andy stated how they wanted the file format to look, so I just went along with that. 😸 
Keep in mind that `HourMinuteFormat` is some seriously old code... I just moved it and didn't end up using it. But I think having it in `edu.gemini.shared.util` makes a lot more sense than having it in `jsky.app.ot.gemini.editor` and I think when possible, things like this should be shared. Part of the reason we probably have so many implementations of certain things are because they aren't shared and nobody can find them 😺

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/sitequality/SiteQualityPanel.java
sraaphorst: It will block until the dialog is closed, and the dialog is modal, so the OT will require you to dismiss the dialog before you can do anything else in the OT.

This is pretty standard for any `Dialog` or `JDialog`, and blocks when you call `setVisible(true)`, which is all that I do in `openImport` to show the dialog.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/sitequality/SiteQualityPanel.java
```diff
@@ -166,43 +164,59 @@
...
+
+                        final int newIndex = table.getModel().getRowCount();
+                        owner.getDataObject().addTimingWindows(results.successesAsJava());
```
sraaphorst: Thanks! I figured it was important to indicate failures, although haha delving back into the world of creating `JTable`s instead of just modifying existing implemented ones may result in a few nights of nightmares!

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/sitequality/SiteQualityPanel.java
```diff
@@ -166,43 +164,59 @@
...
+                            new TimingWindowParseFailureDialog(sqpFrame, results.failures()).setVisible(true);
+                    });
+                    ButtonFlattener.flatten(this);
+                }});
+
                 add(new JButton(Resources.getIcon("eclipse/remove.gif")) {{
                     setToolTipText("Remove timing window");
                     setEnabled(false);
                     setFocusable(false);
 					table.getSelectionModel().addListSelectionListener(e -> setEnabled(table.getSelectedRowCount() > 0));
```
sraaphorst: I hadn't noticed this case, but good eye. (I tried but missed catching this exception, I guess 😺 .)
I'll see if I can go back through the files and force space indenting instead of tab indenting, since this appears normal in my IDEA but looks hokey here and will probably end up annoying at some point to someone.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/sitequality/SiteQualityPanel.java
```diff
@@ -250,38 +256,59 @@ public void updateEnabledState(boolean enabled) {
         elevMin.setEnabled(e);
         elevMax.setEnabled(e);
     }
-}
 
+	private class GBC extends GridBagConstraints {
```
sraaphorst: There are SO many of these GBCs. I was thinking about trying to dig them out and see how similar they are. We really should have a common implementation of this given that I think a lot of it is copy-paste.

I don't mind doing that and it would be a good improvement, but I think I'd rather do it in another PR.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/sitequality/SiteQualityPanel.java
```diff
@@ -250,38 +256,59 @@ public void updateEnabledState(boolean enabled) {
...
 
 	private static final long MS_PER_SECOND = 1000;
 	private static final long MS_PER_MINUTE = MS_PER_SECOND * 60;
 	private static final long MS_PER_HOUR = MS_PER_MINUTE * 60;
 
-    private static final SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss z");
-    static {
-    	dateFormat.setTimeZone(TimeZone.getTimeZone("UTC"));
-    }
+    private static final SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss z") {{
```
sraaphorst: I'm not sure how `DefaultTableModel`s work in terms of thread safety, but I assume they would be thread-safe. No reason why this can't be changed to a `DateFormatter` instead, though. I'll go ahead and do that: it should be straightforward.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/sitequality/SiteQualityPanel.java
cquiroz: Yes, that's fine, And each is slightly different so abstracting them out can be a paint

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/sitequality/SiteQualityPanel.java
```diff
@@ -250,38 +256,59 @@ public void updateEnabledState(boolean enabled) {
...
+    	setTimeZone(TimeZone.getTimeZone("UTC"));
```
sraaphorst: `TimingWindow`s have always been entered and displayed in UTC in the `TimingWindowDialog`, and that looks like how Bryan and Andy - from my understanding - want things to continue through the imports. (Note that I didn't change this... I just eliminated the `static` block by moving it here instead. Not sure if one way is considered preferable to another.)

I don't

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/sitequality/SiteQualityPanel.java
```diff
@@ -338,53 +365,27 @@ private static String formatPeriod(final TimingWindow tw) {
...
 
-	private static String formatRepeat(TimingWindow tw) {
+	private static String formatRepeat(final TimingWindow tw) {
 		if (tw.getDuration() == TimingWindow.WINDOW_REMAINS_OPEN_FOREVER) return null;
 		switch (tw.getRepeat()) {
-		case -1: return "forever";
-		case 0: return "never";
-		case 1: return "1 time";
-		default: return tw.getRepeat() + " times";
+			case -1: return "forever";
```
sraaphorst: You're right. This is kind of stupid, since there are constants already defined in the `TimingWindow` class precisely for this. I'll change -1 and 0 to use these.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/sitequality/SiteQualityPanel.java
cquiroz: That's correct. they have to be UTC, my mistake

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/sitequality/TimingWindowDialog.java
```diff
@@ -0,0 +1,285 @@
...
+class TimingWindowDialog extends JDialog {
+    private static final long MS_PER_MINUTE = 1000 * 60;
+    private static final long MS_PER_HOUR   = MS_PER_MINUTE * 60;
+
+    private static final String INSTRUCTIONS =
+        "Specify a timing window. Starting time should be in the form YYYY-MM-DD hh:mm:ss " +
+        "and will be interpreted in UTC. Window duration and repeat period are specified in " +
+        "hours and minutes.";
+
+    private static final SimpleDateFormat UTC   = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss") {{
```
sraaphorst: This one, I'm not sure how to handle, as it's used by a `JFormattedTextField`.
Calling the `DateTimeFormatter.getFormat()` method to get an appropriate object causes an exception.

There is some information here but I'm not sure I understand the comment:
https://bugs.openjdk.java.net/browse/JDK-8034818

The "customer submitted workarounds" also seem complex. It appears this is just a failure to update `JFormattedTextField` (probably because nobody is 😸 ).

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/sitequality/TimingWindowDialog.java
```diff
@@ -0,0 +1,285 @@
...
+    private final JRadioButton repeatTypeForever;
+    private final JRadioButton repeatTypeFixed;
+
+    private final ImList<JLabel> repeatLabels;
+
+    private final JButton okButton;
+    private final JButton cancelButton;
+
+    // Indicates if this dialog was cancelled.
+    private boolean cancelled;
```
sraaphorst: Yes, that should be the case. It should only ever be written to when the object is constructed and in `ActionListener`s for the OK and Cancel buttons.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/sitequality/TimingWindowDialog.java
```diff
@@ -0,0 +1,285 @@
...
+
+
+    TimingWindowDialog(final Frame owner) throws HeadlessException {
```
sraaphorst: LOL your guess is as good as mine.

From `HeadlessException`:

"Thrown when code that is dependent on a keyboard, display, or mouse
 * is called in an environment that does not support a keyboard, display,
 * or mouse."

Seems completely unnecessary. I'm going to remove it.

bundle/jsky.app.ot/src/main/resources/resources/conf/UIInfo.xml
sraaphorst: I'm pretty sure we do, and I'm kind of scared to try it, but I will 😹 .

This updated automatically during the refactor, but I'm going to commit the other PR comment changes and then give it a go even though I feel fairly certain it will fail. I'd love to be surprised, though!

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/sitequality/TimingWindowImporter.scala
sraaphorst: Thanks! I appreciate that!

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/sitequality/TimingWindowImporter.scala
sraaphorst: Yes, Bryan said this was absolutely essential and listed -1 and 0 in his file format.

bundle/jsky.app.ot/src/test/scala/jsky/app/ot/gemini/editor/sitequality/TimingWindowImporterSpec.scala
sraaphorst: Due to repetitions, I think, dates in the past should be acceptable, as even if the date is in the past, with repetitions, the periods defined by the info could extend (possibly infinitely) into the future.

bundle/jsky.app.ot/src/test/scala/jsky/app/ot/gemini/editor/sitequality/TimingWindowImporterSpec.scala
sraaphorst: I was just trying to appear cultured since I so rarely have the opportunity to do so! Haha.
300px-magrittepipe

cquiroz: 👍

bundle/jsky.app.ot/src/main/resources/resources/conf/UIInfo.xml
sraaphorst: Yes, I just checked, and it is definitely used. `OT.initUI` calls `UIInfoXML.init`, which parses this file and uses it to create a table of editors.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/sitequality/SiteQualityPanel.java
sraaphorst: I switched to that, and I think it seems to have solved the problem!

bundle/jsky.app.ot/src/main/resources/resources/conf/UIInfo.xml
cquiroz: :hurtrealbad: I'm going to jump in front of a train

sraaphorst: I think we need an OCS emotional support group.
Or Pisco Sours with the cake on Fridays. 😆

andrewwstephens:
Very nice!!!

In the timingwindows.txt file, should the first failure case should say "A timing window missing the *duration* should be rejected?

# FAIL: A timing window missing repetitions should be rejected.
2018-01-01 00:00

bundle/edu.gemini.shared.util/src/main/java/edu/gemini/shared/util/HourMinuteFormat.java
jluhrs: @cquiroz We do have FITS support for this.

bundle/jsky.app.ot/src/main/resources/resources/conf/UIInfo.xml
jluhrs: @cquiroz But, trains haven't reached Quillota in a long time...

sraaphorst:
@andrewwstephens You are absolutely right! My bad.
There's no reason to update it, though... the file is just attached to this PR as an example and won't appear in the code base.

Thanks for the kind words 👍

bundle/jsky.app.ot/src/main/resources/resources/conf/UIInfo.xml
cquiroz: @jluhrs a "carreta" will do 😄

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/sitequality/TimingWindowDialog.java
cquiroz: Would this be the only thing pending?

cquiroz: Ok, I read the lineked document and no can do. Let's use `SimpleDateFormat` but lets encapsulate it, I think if it is private, non static and access are synchronized it should be safe to use. Or is it only passed to `JFormattedTextField`?

sraaphorst: This is the only thing pending. I'll finish this off so we can merge it and hopefully be done with it!

sraaphorst: Yes, it is only passed to the `JFormattedTextField` and not used anywhere else, so I'm going to just move it to where the field is constructed and not have it as a static member.

sraaphorst: There we go. Let me know if you think this is better, and we can merge this beast, I hope.

sraaphorst: Huh? My latest commit doesn't seem to be showing up. Not sure what is going on.

sraaphorst: Oh, stupid me... I pushed to central instead of origin. Fixed.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/sitequality/TimingWindowDialog.java
```diff
@@ -59,6 +53,8 @@
         cancelled = true;
 
         // Initialize components.
+        final SimpleDateFormat UTC = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss");
+        UTC.setTimeZone(TimeZone.getTimeZone("UTC"));
```
cquiroz: Yes. I like this a lot more

PR #1189, opened 2017-03-23
REL-3100: Updating HORIZONS protocol to https.

sraaphorst:
See description. HORIZONS access via http no longer seems to work.


sraaphorst:
Trying to figure out what made Travis unhappy, as all test cases passed locally with the change.
Pushed another commit with the changes required for the PIT for comet / asteroid HORIZONS lookup.

PR #1194, opened 2017-03-27
REL-3100: HORIZONS lookups broken in OCS

cquiroz:
Horizons lookups are broken since jpl move to https. Their certificate by DigiCert isn't compatible with the OT/SPDB/PIT because the root cert is newer than the keystore in our JVM  (8u73)

A solution is to upgrade the JVM to anything newer than 8u91, however we can't do that easily due to conflicts with the obslog jsp page generation

This PR uses an alternative, it replaces the ancient apache http client used by Horizons with the built in JVM url connection and it provides an explicit keystore to complete de TLS handshake.

The keystore contains DigiCert intermediate and root certificate compatible with the OT. Note that this keystore will conflict with the one develop since we added another keystore for AGS.

This solves the problem in the OT and SPDB but the PIT will be handled in another PR


bundle/edu.gemini.horizons.api/src/main/scala/edu/gemini/horizons/server/backend/HS2.scala
```diff
@@ -1,29 +1,55 @@
...
+  private val hostnameVerifier: HostnameVerifier = new HostnameVerifier {
+     def verify(s: String, sslSession: SSLSession) = true
+  }
+
+  private val timeout = 3 * 60000  // 1 min ?
+
+  private object ConnectionCharset {
+    val default = Charset.forName("UTF-8")
+
+    def set(conn: URLConnection) {
```
tpolecat: Since this is only used in one place could we either inline it or mark it as unsafe since it's a side-effect?

bundle/edu.gemini.horizons.api/src/main/scala/edu/gemini/horizons/server/backend/HS2.scala
```diff
@@ -1,29 +1,55 @@
...
 // work-in-progress; this will get a better name and will probably move elsewhere
 object HorizonsService2 {
+  private val LOG = Logger.getLogger(this.getName)
+
+  // Lets any remote endpoint be connectable
+  private val hostnameVerifier: HostnameVerifier = new HostnameVerifier {
+     def verify(s: String, sslSession: SSLSession) = true
+  }
+
+  private val timeout = 3 * 60000  // 1 min ?
```
arturog8m: what are the units of this? I can't figure out how this is one minute. Or is the comment wrong?

bundle/edu.gemini.horizons.api/src/main/scala/edu/gemini/horizons/server/backend/HS2.scala
```diff
@@ -1,29 +1,55 @@
...
+
+  private object ConnectionCharset {
+    val default = Charset.forName("UTF-8")
+
+    def set(conn: URLConnection) {
+      conn.setRequestProperty("Accept-Charset", default.displayName())
+    }
+
+    def get(conn: URLConnection): Charset =
+      Option(conn.getContentType).getOrElse("").replace(" ", "").split(';').find {
```
tpolecat: You can use `.orZero` here instead of `.getOrElse("")`.

bundle/edu.gemini.horizons.api/src/main/scala/edu/gemini/horizons/server/backend/HS2.scala
```diff
@@ -325,41 +350,44 @@ object HorizonsService2 {
...
-    new Resource[GetMethod] {
-      def close(m: GetMethod): IO[Unit] =
-        IO(m.releaseConnection)
+  implicit def ResponseStreamResource: Resource[ResponseStream] =
+    new Resource[ResponseStream] {
+      def close(m: ResponseStream): IO[Unit] =
+        IO(m.in.close)
+    }
+
+  private def readRemote(baseURL: String, params: Map[String, String]): ResponseStream = {
```
tpolecat: Can we rename this `unsafeReadRemote` or something since it's a side-effect?

sraaphorst:
Looks good, but Travis is complaining, unfortunately.

We will have to fix this silly obslog problem in the near future, as evidenced by this. It is kind of ridiculous that we have to stick with an old JVM because the obslog is unhappy.

We should also look at merging the PIT HORIZONS class with the OT / ODB HORIZONS class.

sraaphorst:
And thanks for figuring this out... I tried a number of things with no success, so I am glad that you were able to find a solution.

sraaphorst:
Also, it is always nice to see the code simplified and cleaned in edited files up as you did!

tpolecat:
Approved with a few suggestions. I don't understand the CI failure.

jdnavarro:
Where did you take the keystore included with this PR? From a newer JVM release? Does it mean that the certificates have to be updated by checking in a new keystore?

Nothing wrong with that, I'm just trying to understand the problem.

bundle/edu.gemini.horizons.api/src/main/scala/edu/gemini/horizons/server/backend/HS2.scala
sraaphorst: Good to know. I had no idea that `orZero` existed.

bundle/edu.gemini.horizons.api/src/main/scala/edu/gemini/horizons/server/backend/HS2.scala
tpolecat: Ha good catch. Yeah let's fix this.

bundle/edu.gemini.horizons.api/src/main/scala/edu/gemini/horizons/server/backend/HS2.scala
sraaphorst: So `orZero` just relies on an underlying `Monoid` and retrieves the operation identity?
Sensible.

cquiroz:
@tpolecat I don't see your suggestions?

tpolecat:
@cquiroz I mean, the code comments above. You replied to one.

cquiroz:
@jdnavarro I got the certificates from DigiCert and added them to the certificate loaded `GemSslSocketFactory`

bundle/edu.gemini.horizons.api/src/main/scala/edu/gemini/horizons/server/backend/HS2.scala
sraaphorst: Good point. Seems like it should be:

`import scala.concurrent.duration._`

and `1.min.toMillis`
?

bundle/edu.gemini.horizons.api/src/main/scala/edu/gemini/horizons/server/backend/HS2.scala
cquiroz: Yeah, it should be inside IO

cquiroz:
@tpolecat Yes, the code was originally side-effecty but I didn't want to change it too much. The charset one was quite clear a problem and I tried to fix it but the synchronization started to be "funny" and tests failed

BTW, a bunch of the new code is taken from the AGS Https client

tpolecat:
Good to merge assuming CI issue can get fixed.

cquiroz:
I spent some time thinking about this and my final conclusion (probably incorrect 😄 ) is that using `IO` and and having  a global synchronized lock is not correct

I'd rather use `Task` instead and force the execution to be sequential defining a single threaded executor service

I don't have enough time to do this at the moment so I'll settle for just naming the method as `unsafe`

tpolecat:
@cquiroz yeah a single-thread executor might be nicer. It shouldn't be the concern of the implementation. But I also agree that it's not worth introducing right now.

PR #1197, opened 2017-03-28
REL-3100: HORIZONS lookups broken in OCS

cquiroz:
This is a replay of PR #1194 
It however contains an updated keystore with the certificates for both branches


tpolecat:
👍

PR #1198, opened 2017-03-29
REL-3100: Fixed Horizons lookups to work with the PIT.

sraaphorst:
This is a simple fix for Horizons lookups for the PIT that mirrors @cquiroz's fix for Horizons lookups for the OT and ODB.

It is a temporary measure until we switch the PIT to use the same code as the OT / ODB.

I cleaned up the code a bit as well. The object type, for example, was never used in the query, but only in the response.


bundle/edu.gemini.horizons.api/src/main/java/edu/gemini/horizons/api/HorizonsQuery.java
```diff
@@ -90,11 +69,6 @@ private StepUnits(int minValue, int maxValue, String suffix) {
      */
     private Site _site = Site.GN;
 
-    /**
-     * The object type to look.
-     */
-    private ObjectType _type;
```
cquiroz: Was this used only on the PIT?

bundle/edu.gemini.horizons.api/src/main/java/edu/gemini/horizons/server/backend/CgiQueryExecutor.java
```diff
@@ -60,80 +53,68 @@ private static synchronized String _formatDate(Date date) {
...
+            for (final Map.Entry<String, String> e: queryParams.entrySet()) {
+                if (first) first = false;
+                else queryParamString.append("&");
+                queryParamString.append(e.getKey()).append("=").append(URLEncoder.encode(e.getValue(), charSet));
+            }
+            final URL url = new URL(CgiHorizonsConstants.HORIZONS_URL + "?" + queryParamString);
+            LOG.info("Horizons request " + url);
+
+            final HttpsURLConnection conn = (HttpsURLConnection) url.openConnection();
+            conn.setHostnameVerifier((v1,v2) -> true);
```
cquiroz: Pity we can't do this in scala

tpolecat:
There's nothing obviously wrong here but it would be nice to have a few tests, so at least we would have some coverage if things go bad again.

bundle/edu.gemini.horizons.api/src/main/java/edu/gemini/horizons/server/backend/CgiQueryExecutor.java
sraaphorst: There's no reason why I couldn't rewrite `CgiQueryExecutor` in scala, but at that point, it feels like it starts to pseudo-move to the solution for REL-3111.

cquiroz: Right, there is no point on that

cquiroz:
@tpolecat HS2 has some very nice tests. Since we aim to replace this code with HS2's maybe we don't need to implement tests at the moment? WDYT?

sraaphorst:
Agree with @cquiroz: if something changes on JPL's end, the existing `HS2Spec` tests should catch it, so it seems redundant to me to spend time implementing test cases for this when it is just a temporary measure for five months.

bundle/edu.gemini.horizons.api/src/main/java/edu/gemini/horizons/api/HorizonsQuery.java
sraaphorst: Yes, `HorizonsQuery` and related classes are only used in the PIT, and the `ObjectType` member here was never used except in the `main` method of this class, so I stripped that all out.

The `ObjectType` is used in the `ReplyType`, though, in the `HorizonsResponse`, so I moved it there.

tpolecat:
I lack your confidence, but if you're ok without a test then that's your call. It needs a bunch of monkey testing for sure.

cquiroz:
The HS2 tests failed on the bug we are dealing with at least. Unfortunately it didn't show in travis because the tests requires an oldish JVM

tpolecat:
> Unfortunately it didn't show in travis because the tests requires an oldish JVM

(╯°□°）╯︵ ┻━┻

PR #1204, opened 2017-04-04
REL-1101: Fix GNIRS description in e-obslog

sraaphorst:
GNIRS was incorrectly reporting MIR as the prism in the e-obslog even when cross-dispersed.

I found that this was due to the incorrect handling of the data in the `ConfigMap` `CROSSDISPERSED_KEY` value. It was assumed in `GNIRSLogSegment` that this would be `Yes` or `No`, but in actuality, it is always `No`, `SXD`, or `LXD`.

This fixes that by removing checks to see if the value is `Yes` since this is never the case, and setting other values appropriately based on this.

Note that this is (eep) from Q3 amnesty list, but I had started it awhile back and was so close to finishing it when the final amnesty list was released that I just wanted to finish it up. I was just waiting for Shane to show me how to generate the e-obslog.

I also cleaned up a bit as I went through the code to try to understand what was going on.


bundle/edu.gemini.obslog/src/main/java/edu/gemini/obslog/instruments/GNIRSLogSegment.java
```diff
@@ -35,40 +30,34 @@
...
-        if (typeValue == null) typeValue = EMPTY_STRING;
-
         // This should work if filterValue == null - the check is added to keep
         // oldValueOf from complaining
         if (filterValue != null) {
-            GNIRSParams.Filter f = GNIRSParams.Filter.getFilter(filterValue, null);
+            final GNIRSParams.Filter f = GNIRSParams.Filter.getFilter(filterValue, null);
             if (f != null) {
+                final String typeValue = ImOption.apply(map.sget(TYPE_KEY)).getOrElse("");
                 map.put(FILTER_KEY, !typeValue.equals("DARK") ? f.logValue() : "DARK");
```
tpolecat: Are these constants not defined anywhere? Is the whole app like this?

sraaphorst: It truly is horrible, isn't it?
A lot of these constants are actually defined in `GNIRSParams`.
I'll go and make some substitutions to clean this up.

`GpiLogSegment` seems to be a lot better written, and the others are shorter, so I'm not going to touch them for now (may take a peek later tonight), but I'll tidy up this little corner of the OCS code.

swalker2m: There should be a real enumerated type for this, but I don't think there is one.  It gets stuck in the sequence as a `String`.  The constants are defined in `edu.gemini.spModel.obscomp.InstConstants`, for what it's worth.

tpolecat: image

sraaphorst: Actually, I think we can improve on this a lot, but the problem is that there were some different values pre 2010, so we do need to do some ugly hackery in a couple cases, from what I can tell, and my fix may break things for old obslogs re: cross-dispersed.
Yuck.

sraaphorst: Okay, I rewrote the class in Scala and think it's much improved, but it requires testing.

bundle/edu.gemini.obslog/src/main/scala/edu/gemini/obslog/instruments/GNIRSLogSegment.scala
```diff
@@ -0,0 +1,154 @@
...
+
+import GNIRSParams._
+import GNIRSLogSegment._
+
+import scalaz._
+import Scalaz._
+
+class GNIRSLogSegment(logItems: java.util.List[OlLogItem], obsLogOptions: OlLogOptions) extends InstrumentLogSegment(SegmentType, logItems, obsLogOptions) {
+
+  override def decorateObservationData(map: ConfigMap): Unit = Option(map).foreach(m => {
```
swalker2m: Obviously we have to leave this method with the given `ConfigMap` modified to work with the Obslog framework as is.  If we're rewriting this though, one idea would be to build up the changes we want to make to it and apply them all at the end instead of mutating as we go.

bundle/edu.gemini.obslog/src/main/scala/edu/gemini/obslog/instruments/GNIRSLogSegment.scala
```diff
@@ -0,0 +1,154 @@
...
+        ps match {
+          case PixelScale.PS_015 if lambda > 2.5 => Camera.SHORT_RED
+          case PixelScale.PS_015                 => Camera.SHORT_BLUE
+          case _                 if lambda > 2.5 => Camera.LONG_RED
+          case _                                 => Camera.LONG_BLUE
+        }
+      }.getOrElse("?")
+
+      // Prism
+      val prismName = yesOrNotNo(c) ? (ps match {
```
swalker2m: This is a difference from the previous version, and also the bug fix we're trying to do?  Now it is "if not no" instead of "if yes".

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/acqcam/AcqCamGUI.java
```diff
@@ -1,230 +0,0 @@
...
-package jsky.app.ot.gemini.acqcam;
-
-import java.awt.*;
-import javax.swing.*;
-import javax.swing.border.*;
-
-import jsky.util.gui.DropDownListBoxWidget;
-import jsky.util.gui.NumberBoxWidget;
-
-public class AcqCamGUI extends JPanel {
```
swalker2m: Good to git rid of stuff we aren't using.  Since this is unrelated to this PR and there are so many deletions, it might be best (in the future) to relegate this type of thing to a separate PR or at least a separate commit.

swalker2m:
There are obvious improvements here in your new version.  I definitely appreciate the initiative.  Thinking about it a bit more, I would be inclined to fix the bug making as few changes as possible or else rewrite it as you've done and do the comparison test cases.

One thing to consider is that the Obslog as a whole needs to be rewritten anyway but that should wait for the new program model.  When it is rewritten, I hope it looks much different than it does today.  In other words, even if you do go for the full replacement option, which is fine with me, it might be short-lived.  (If your definition of "short" is "years" I guess :-) )

arturog8m:
As this is legacy code and we are aiming to address all these amnesty items so we can actually focus on working on the new OCS, I agree we should try to make these fixes with as little changes as possible in the current OCS. 

While it is nice to tidy up this part of the code base, the effort necessary to ensure it doesn't have unintended consequences is not worth it, so I would prefer we go ahead with the original fix.

cquiroz:
This is slightly related. Should we try to fix the obslog issue that doesn't let us use newer JVMs? While not very sexy work it could be a showstopper for other changes as we have seen lately

arturog8m:
Yes - using the same principle - we need to update it to "keep it working" and that issue has caused problems in operations, so to me that means we have to at least get that resolved. That will be a separate issue/PR though.

cquiroz:
Yes, I don't recall that issue on the amnesty list though
Nice avatar!

arturog8m:
Right, it's not - let's add it and figure out when we can do it.

bundle/edu.gemini.obslog/src/main/scala/edu/gemini/obslog/instruments/GNIRSLogSegment.scala
```diff
@@ -0,0 +1,154 @@
...
+
+import scalaz._
+import Scalaz._
+
+class GNIRSLogSegment(logItems: java.util.List[OlLogItem], obsLogOptions: OlLogOptions) extends InstrumentLogSegment(SegmentType, logItems, obsLogOptions) {
+
+  override def decorateObservationData(map: ConfigMap): Unit = Option(map).foreach(m => {
+    decorateFilter(m)
+    decorateSlitWidth(m)
+    decorateDisperser(m)
```
sraaphorst: Ooops. I wrote the code for it, but neglected to add it to the list. Good catch.
As I said, this is untested: I finished it late Tuesday night and just wanted to throw it up - working or not - to see what people thought.

bundle/edu.gemini.obslog/src/main/scala/edu/gemini/obslog/instruments/GNIRSLogSegment.scala
```diff
@@ -0,0 +1,154 @@
...
+    GNIRSParams.Order.FIVE,
+    GNIRSParams.Order.FOUR,
+    GNIRSParams.Order.THREE,
+    GNIRSParams.Order.TWO
+  )
+
+  // This is to handle the hideousness of cross dispersers.
+  // Yes used to be a supported option, and now it is not: it is divided into SXD and LXD.
+  // Thus, for backward compatibility, we determine if the cross disperser is yes (legacy) or not no.
+  def yesOrNotNo(s: String): Boolean = s.equalsIgnoreCase("yes") || !s.equalsIgnoreCase("no")
```
sraaphorst: Yes, you are right.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/acqcam/AcqCamGUI.java
sraaphorst: Sure. I agree that is preferable.

sraaphorst:
I am going to cancel this PR, and revert, then, to the simplest fix possible for now (leaving the hideously bad code intact), as was originally the case. We can attempt to clean up this code at a later date.

PR #1205, opened 2017-04-06
REL-1101: Fix for GNIRS e-obslog data

sraaphorst:
This is the minimum possible fix for `GNIRSLogSegment`.

As per my previous PR, at one point, the cross dispersed property was either `No` or `Yes`.

Now it is `No`, `SXD`, or `LXD`.

Thus, I replaced the `Yes` checks, since this value never shows up, with ones that will work. 

Instead of `if YES then A else B`, we now do `if NO then B else A`.


sraaphorst:
Yes, we know this code is terribly written but at this point, we just need the problem fixed without spending too much time on it as per my previous much more detailed PR, which risks introducing errors.

PR #1207, opened 2017-04-06
REL-3117: Initial Asterism trait

swalker2m:
This simple PR just presents the proposed `Asterism` interface.  Having an `Asterism` trait will unlock progress on the `SingleTargetAsterism` implementation, `TargetEnvironment` update, and GHOST-specific asterism development.

One potentially controversial question is whether the targets should be `SPTarget` (which is mutable) or `Target` (which is immutable).  If this were new code here would be no question.  In `TargetEnvironment` though, the base position is an `SPTarget` so it seemed like the path of least resistance to make the asterism targets `SPTarget`s.  We drag targets around, for example, in the TPE so making these immutable implies replacing a lot of existing code. It seems like it would be prohibitively difficult.  I'm certainly happy to entertain the idea though if someone thinks I'm overestimating the complexity.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/env/Asterism.scala
+++ b/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/env/Asterism.scala
@@ -1,0 +1,63 @@
+package edu.gemini.spModel.target.env
+
+import edu.gemini.spModel.core.Coordinates
+import edu.gemini.spModel.target.SPTarget
+import edu.gemini.shared.util.immutable.{Option => GOption}
+import edu.gemini.shared.util.immutable.ScalaConverters._
+import edu.gemini.skycalc.{Coordinates => SCoordinates}
+
+
+import java.time.Instant
+
+import scalaz.NonEmptyList
+
+/** Collection of stars that make up the science target(s) for an observation,
+  * along with any configuration details unique to the instrument in use.
+  */
+trait Asterism {
+
+  /** All targets that comprise the asterism.  There must be at least one target.
+    */
+  def targets: NonEmptyList[SPTarget]
+
+  /** Slew coordinates and AGS calculation base position.  By default this
+    * will be a computation that finds the coordinate minimizing the maximum
+    * distance to any target in the asterism.
+    */
+  def basePosition(time: Instant): Option[Coordinates] =
+    ???  // we should supply the "smallest-circle" implementation here
+
+  //
+  // "Base position" convenience methods already in use extensively throughout
+  // the codebase.  Defined in terms of the base position, these are methods
+  // on SPTarget that are used directly in calls like
+  // targets.getBase.getRaDegrees(when)
+  //
+
+  type GOLong   = GOption[java.lang.Long]
+  type GODouble = GOption[java.lang.Double]
+
+  private def gcoords[A](time: GOLong)(f: Coordinates => A): GOption[A] =
+    (for {
+      t <- time.asScalaOpt
+      c <- basePosition(Instant.ofEpochMilli(t))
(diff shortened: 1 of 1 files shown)


bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/env/Asterism.scala
```diff
@@ -0,0 +1,63 @@
...
+  /** All targets that comprise the asterism.  There must be at least one target.
+    */
+  def targets: NonEmptyList[SPTarget]
+
+  /** Slew coordinates and AGS calculation base position.  By default this
+    * will be a computation that finds the coordinate minimizing the maximum
+    * distance to any target in the asterism.
+    */
+  def basePosition(time: Instant): Option[Coordinates] =
+    ???  // we should supply the "smallest-circle" implementation here
```
cquiroz: These always worry me as we sometimes forget them and they endup in production

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/env/Asterism.scala
```diff
@@ -0,0 +1,63 @@
...
+import scalaz.NonEmptyList
+
+/** Collection of stars that make up the science target(s) for an observation,
+  * along with any configuration details unique to the instrument in use.
+  */
+trait Asterism {
+
+  /** All targets that comprise the asterism.  There must be at least one target.
+    */
+  def targets: NonEmptyList[SPTarget]
```
cquiroz: In general I'd prefer to use the new `Target` class but I certainly see the point that the consequences downstream could be very big. Is it so that `SPTarget` has the listeners to inform about changes?

tpolecat: Yeah, if it turns out that there are callers who don't depend on it being an `SPTarget` then we can change the interface. I think this is ok for now.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/env/Asterism.scala
tpolecat: This is used all over the place so it will be obvious if it remains unimplemented. Implementing it is one of the next tasks.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/env/Asterism.scala
```diff
@@ -0,0 +1,63 @@
...
+  // "Base position" convenience methods already in use extensively throughout
+  // the codebase.  Defined in terms of the base position, these are methods
+  // on SPTarget that are used directly in calls like
+  // targets.getBase.getRaDegrees(when)
+  //
+
+  type GOLong   = GOption[java.lang.Long]
+  type GODouble = GOption[java.lang.Double]
+
+  private def gcoords[A](time: GOLong)(f: Coordinates => A): GOption[A] =
```
sraaphorst: Just out of curiosity, why are the `time` parameters `Option`s?
When would we call these with `None` and why?

swalker2m: These methods are intended to make it possible to use an `Asterism` directly in many places in the codebase where today an `SPTarget` base position is used.  For example, if we replace the "base" of a `TargetEnvironment` with an `Asterism` then a call such as `env.getBase.getRaDegrees(when)` would not need to be modified.  For that reason, the methods match exactly those available in the `SPTarget` API.

To answer the question more directly though, the scheduling block from which the time of interest comes is an `Option` in `SPObservation`.

PR #1208, opened 2017-04-07
REL-3119: Initial GHOST asterism model

swalker2m:
Based on a careful reading of the GHOST ConOps document, this PR represents an initial attempt at defining the GHOST asterism data model.  GHOST offers a fairly sophisticated set of modes and configuration options.  Not all options are applicable to all modes.  Rather than creating a generic `Asterism` implementation that contains mutually exclusive options, I'm attempting to constrain the choices to match those available in the various operating modes.  This helps guide user interface development and limits the possibility of ending up in an inconsistent state.

I expect there will be a number of iterations required to get the model in good shape.  When the changes die down I anticipate adding `Lens`es to make it easier to manipulate.

A few outstanding questions for science-staff include:

1) __Explicit Guide Fiber State__.  The ConOps document details the need to allow the user to explicitly enable or disable guide fibers.  Bryan asked a reasonable question that I didn't see an answer to yet.  Namely, do we really need this?  Usually we can default the state based on target magnitude.  Would the user know enough to go against the default settings or can the ability to turn guiding on or off explicitly be relegated to lower-level control screens that only a telescope operator would see?  As you can see in the model, making it overridable introduces a degree of additional complexity. 

2) __Cloud Cover__.  The rules for defaulting guide fiber state don't take into account cloud cover.  I assume that will be necessary.  Does anybody have more details there?

3) __Beam-Switching Modes__.  In my effort estimation document Bryan notes that we won't add explicit support for beam-switching modes.  This confuses me since it has a prominent place in the "Instrument Operating Modes" section.  If we are doing beam-switching as described in ConOps, I'd prefer to make it explicit and constrain the configuration as described rather than infer beam-switching from the asterism plus other targets in the target environment.  Does anybody have any further input about this?

4) __Explicit Binning__. This question is similar to the one about guiding state, but the model as it stands took the opposite approach.  Namely, binning is always defaulted here based on mode.  The ConOps document does mention a desire for the user to be able to explicitly set the value to anything but this is not (yet) implemented.  Is this really something that is needed in the user interface?

5) __Default Binning__.  Rather than specifying bright/faint/very faint can we infer this from the target magnitude (+ cloud cover?).  Table 9 doesn't make it clear how we would select between faint and very faint so I didn't do this.

6) __Two Target Base Position__.  I believe Bryan mentioned a desire to allow the definition of a base position at some point other than the default position between the two targets.  I don't see this mentioned in the ConOps document.  I'd prefer to avoid this complication unless it is really needed of course.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostAsterism.scala
+++ b/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostAsterism.scala
@@ -26,22 +26,22 @@ object GhostAsterism {
   /** The GHOST guide fibers can be enabled or disabled for each science target.
     * Typically they are enabled for bright targets (< B mag 18 for standard
     * resolution, < B mag 17 for high resolution) but disabled for faint targets.
-    * OIWFS state can also be explicitly disabled, say, in a crowded field where
-    * OIWFS is less effective.
+    * Guide state can also be explicitly disabled, say, in a crowded field where
+    * guide fibers are less effective.
     */
-  sealed trait OiwfsState extends Product with Serializable
+  sealed trait GuideFiberState extends Product with Serializable
 
-  object OiwfsState {
-    case object Enabled  extends OiwfsState
-    case object Disabled extends OiwfsState
+  object GuideFiberState {
+    case object Enabled  extends GuideFiberState
+    case object Disabled extends GuideFiberState
 
-    val enabled: OiwfsState  = Enabled
-    val disabled: OiwfsState = Disabled
+    val enabled: GuideFiberState  = Enabled
+    val disabled: GuideFiberState = Disabled
 
     val All = NonEmptyList(enabled, disabled)
 
-    implicit val EqualOiwfsState: Equal[OiwfsState] =
-      Equal.equalA[OiwfsState]
+    implicit val EqualGuideFiberState: Equal[GuideFiberState] =
+      Equal.equalA[GuideFiberState]
   }
 
 
@@ -52,7 +52,8 @@ object GhostAsterism {
     * There is a default guiding state based on magnitude, but this can be
     * explicitly overridden.
     */
-  final case class GhostTarget(target: SPTarget, explicitOiwfsState: Option[OiwfsState]) {
+  final case class GhostTarget(target: SPTarget,
+                               explicitGuideFiberState: Option[GuideFiberState]) {
 
     def coordinates(when: Option[Instant]): Option[Coordinates] =
       target.getCoordinates(when.map(_.toEpochMilli))
@@ -60,31 +61,31 @@ object GhostAsterism {
 
   object GhostTarget {
 
-    /** The magnitude at which OIWFS state is disabled by default at standard
+    /** The magnitude at which guide state is disabled by default at standard
       * resolution.
       */
     val StandardResCutoff: Magnitude =
       Magnitude(18.0, MagnitudeBand.B, None, MagnitudeSystem.Vega)
 
-    /** The magnitude at which OIWFS state is disabled by default at high
+    /** The magnitude at which guide state is disabled by default at high
       * resolution.
       */
     val HighResCutoff: Magnitude =
       Magnitude(17.0, MagnitudeBand.B, None, MagnitudeSystem.Vega)
 
-    private def defaultOiwfsState(t: GhostTarget, cutoff: Magnitude): OiwfsState =
-      t.target.getMagnitude(MagnitudeBand.B).forall(_.value < cutoff.value) ? OiwfsState.enabled | OiwfsState.disabled
+    private def defaultGuideFiberState(t: GhostTarget, cutoff: Magnitude): GuideFiberState =
+      t.target.getMagnitude(MagnitudeBand.B).forall(_.value < cutoff.value) ? GuideFiberState.enabled | GuideFiberState.disabled
 
-    private def oiwfsState(t: GhostTarget, cutoff: Magnitude): OiwfsState =
-      t.explicitOiwfsState | defaultOiwfsState(t, cutoff)
+    private def guideFiberState(t: GhostTarget, cutoff: Magnitude): GuideFiberState =
+      t.explicitGuideFiberState | defaultGuideFiberState(t, cutoff)
 
-    /** Computes the OiwfsState for the given target in standard resolution mode. */
-    def standardResOiwfsState(t: GhostTarget): OiwfsState =
-      oiwfsState(t, StandardResCutoff)
+    /** Computes the GuideFiberState for the given target in standard resolution mode. */
+    def standardResGuideFiberState(t: GhostTarget): GuideFiberState =
+      guideFiberState(t, StandardResCutoff)
 
-    /** Computes the OiwfsState for the given target in high resolution mode. */
-    def highResOiwfsState(t: GhostTarget): OiwfsState =
-      oiwfsState(t, HighResCutoff)
+    /** Computes the GuideFiberState for the given target in high resolution mode. */
+    def highResGuideFiberState(t: GhostTarget): GuideFiberState =
+      guideFiberState(t, HighResCutoff)
   }
 
 
@@ -200,11 +201,11 @@ object GhostAsterism {
         case Faint | VeryFaint => YBinning.Four
       }
 
-    def ifu1OiwfsState: OiwfsState =
-      GhostTarget.standardResOiwfsState(ifu1)
+    def ifu1GuideFiberState: GuideFiberState =
+      GhostTarget.standardResGuideFiberState(ifu1)
 
-    def ifu2OiwfsState: OiwfsState =
-      GhostTarget.standardResOiwfsState(ifu2)
+    def ifu2GuideFiberState: GuideFiberState =
+      GhostTarget.standardResGuideFiberState(ifu2)
   }
 
 
@@ -249,11 +250,12 @@ object GhostAsterism {
         case Faint | VeryFaint => YBinning.Eight
       }
 
-    /** Deterimines the OIWFS state for the IFU observing the science object.
-      * For the IFU observing a sky position, OiwfsState is always disabled.
+    /** Deterimines the guide fiber state for the IFU observing the science
+      * object. For the IFU observing a sky position, GuideFiberState is always
+      * disabled.
       */
-    def oiwfsState: OiwfsState =
-      GhostTarget.standardResOiwfsState(target)
+    def guideFiberState: GuideFiberState =
+      GhostTarget.standardResGuideFiberState(target)
 
     // Calculate the diametrically opposed position, assuming we know where
     // the target is.
@@ -376,10 +378,10 @@ object GhostAsterism {
     override def yBinning: YBinning =
       mode.yBinning
 
-    /** Deterimines the OIWFS state for the HRIFU1.  Typically this will be
-      * enabled since the target is bright but may be explicitly turned off.
+    /** Deterimines the guide fiber state for the HRIFU1.  Typically this will
+      * be enabled since the target is bright but may be explicitly turned off.
       */
-    def oiwfsState: OiwfsState =
-      GhostTarget.highResOiwfsState(target)
+    def guideFiberState: GuideFiberState =
+      GhostTarget.highResGuideFiberState(target)
   }
 }
--- a/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostAsterism.scala
+++ b/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostAsterism.scala
@@ -1,6 +1,6 @@
 package edu.gemini.spModel.gemini.ghost
 
-import edu.gemini.spModel.core.{Coordinates, Magnitude, MagnitudeBand, MagnitudeSystem, Offset}
+import edu.gemini.spModel.core._
 import edu.gemini.spModel.gemini.obscomp.SPSiteQuality.CloudCover
 import CloudCover.{PERCENT_70 => CC70, PERCENT_80 => CC80, PERCENT_90 => CC90, ANY => CCAny}
 import edu.gemini.spModel.target.SPTarget
@@ -50,13 +50,16 @@ object GhostAsterism {
     * There is a default guiding state based on magnitude, but this can be
     * explicitly overridden.
     */
-  final case class GhostTarget(target: SPTarget,
+  final case class GhostTarget(spTarget: SPTarget,
                                xBinning: XBinning,
                                yBinning: YBinning,
                                explicitGuideFiberState: Option[GuideFiberState]) {
 
     def coordinates(when: Option[Instant]): Option[Coordinates] =
-      target.getCoordinates(when.map(_.toEpochMilli))
+      spTarget.getCoordinates(when.map(_.toEpochMilli))
+
+    def copyWithClonedTarget: GhostTarget =
+      copy(spTarget = spTarget.clone)
   }
 
   object GhostTarget {
@@ -74,7 +77,7 @@ object GhostAsterism {
       Magnitude(17.0, MagnitudeBand.B, None, MagnitudeSystem.Vega)
 
     private def defaultGuideFiberState(t: GhostTarget, cutoff: Magnitude): GuideFiberState =
-      t.target.getMagnitude(MagnitudeBand.B).forall(_.value < cutoff.value) ? GuideFiberState.enabled | GuideFiberState.disabled
+      t.spTarget.getMagnitude(MagnitudeBand.B).forall(_.value < cutoff.value) ? GuideFiberState.enabled | GuideFiberState.disabled
 
     private def guideFiberState(t: GhostTarget, cutoff: Magnitude): GuideFiberState =
       t.explicitGuideFiberState | defaultGuideFiberState(t, cutoff)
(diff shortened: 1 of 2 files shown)


bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostAsterism.scala
```diff
@@ -0,0 +1,368 @@
...
+  }
+
+  object XBinning {
+    case object One extends XBinning(1)
+    case object Two extends XBinning(2)
+
+    val one: XBinning = One
+    val two: XBinning = Two
+
+    val All = NonEmptyList(one, two)
```
cquiroz: Are you referring via the `one` and `two` variable to get the correct type inferred?

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostAsterism.scala
```diff
@@ -0,0 +1,368 @@
...
+
+    implicit val EqualXBinning = Equal.equalA[XBinning]
+  }
+
+
+  /** Y-binning options. */
+  sealed abstract class YBinning(val intValue: Int) extends Product with Serializable {
```
cquiroz: Any pros/cons of making this a trait?

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostAsterism.scala
swalker2m: Yeah, here it could just as easily be `NonEmptyList(One, Two)` but in code like

````scala
  val binOpt: Option[XBinning] = ...
  binOpt.fold(XBinning.One) { b => ??? }
````

You have to explicitly tag `XBinning.One` as `XBinning` because it cannot infer that.  If you use `XBinning.one` though it picks up the right type.

File: bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostAsterism.scala (hunk 2 above)
Comment (swalker2m) [inline bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostAsterism.scala]: I did it that way to make it trivial to specify the `intValue`.  Otherwise I guess you would have to provide a definition for each `case object` or else provide a switch in the trait like:

````scala
sealed trait YBinning extends Product with Serializable {

  def intValue: Int = this match {
    case One   => 1
    case Two   => 2
    case Four  => 4
    case Eight => 8
  }
}
````

That seems like it would be fine too.

File: bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostAsterism.scala [hunk 3]
```diff
@@ -0,0 +1,368 @@
...
+/** GHOST asterism model.
+  */
+object GhostAsterism {
+
+  /** The GHOST guide fibers can be enabled or disabled for each science target.
+    * Typically they are enabled for bright targets (< B mag 18 for standard
+    * resolution, < B mag 17 for high resolution) but disabled for faint targets.
+    * OIWFS state can also be explicitly disabled, say, in a crowded field where
+    * OIWFS is less effective.
+    */
```
arturog8m: This is probably my lack of understanding about the GHOST specifics, but this comment was somewhat confusing to me. It starts talking about the GHOST guide fibers, but then it talks about the OIWFS state - I wonder if we can define first what an OIWFS state is, so it is clear that this state is about the guide fibers? I might be understanding all this wrong, so apologies for the noise in that case.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostAsterism.scala
```diff
@@ -0,0 +1,368 @@
...
+    val All = NonEmptyList(enabled, disabled)
+
+    implicit val EqualOiwfsState = Equal.equalA[OiwfsState]
+  }
+
+
+  /** GHOST targets are associated with a guiding state (enabled or disabled),
+    * referring to whether the dedicated guide fibers surrounding the science
+    * target should be used.
+    *
```
arturog8m: A guiding state is an OIWFS state, I presume

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostAsterism.scala
swalker2m: You're right.  I'm not sure this counts as an OIWFS and would probably be better named `GuideState` or `GuideFiberState`.  I'll change it.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostAsterism.scala
swalker2m: The "GemOps v3" document refers to these as OIWFS so I must have picked that up from there.  I'll make it more consistent though.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostAsterism.scala
```diff
@@ -0,0 +1,368 @@
...
+  object XBinning {
+    case object One extends XBinning(1)
+    case object Two extends XBinning(2)
+
+    val one: XBinning = One
+    val two: XBinning = Two
+
+    val All = NonEmptyList(one, two)
+
+    implicit val EqualXBinning = Equal.equalA[XBinning]
```
tpolecat: It's good practice to always have a return type annotation on any implicit definition because you can run into subtle problems with the inferred type being too precise. It's ok here but just as a matter of habit I would write it as

```scala
implicit val EqualXBinning: Equal[XBinning] = Equal.equalA
```

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostAsterism.scala
```diff
@@ -0,0 +1,368 @@
...
+
+    implicit val EqualFiberAgitatorState = Equal.equalA[FiberAgitatorState]
+  }
+
+  /** The high resolution asterism comes in two flavors, normal high resolution
+    * mode and precision radial velocity (PRV) mode.  Normal high resolution
+    * mode can use bright or faint detector binning, but PRV is always bright.
+    * PRV can be set up with fiber agitation while normal high resolution mode
+    * cannot.
+    */
```
arturog8m: What does it mean that PRV can be set up with fiber agitation?

arturog8m: Oh, I see that in the HighResMode object below - pardon the noise.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostAsterism.scala
tpolecat: Sealed abstract class seems fine to me. But note that this value doesn't participate in equality, hashing, etc. (This is ok since all instances are singletons).

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostAsterism.scala
```diff
@@ -0,0 +1,368 @@
...
+    case object Eight extends YBinning(8)
+
+    val one: YBinning   = One
+    val two: YBinning   = Two
+    val four: YBinning  = Four
+    val eight: YBinning = Eight
+
+    val All = NonEmptyList(one, two, four, eight)
+
+    implicit val EqualYBinning = Equal.equalA[YBinning]
```
tpolecat: Would it make sense to define `Order` here and elsewhere when there's a natural ordering? I don't have a strong opinion but the UI might want to sort them rather than relying on the arbitrary order of `All`.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostAsterism.scala
swalker2m: From the ConOps description:

> This mode is for science targets requiring the maximum possible wavelength calibration precision. This mode is equivalent to the High-Resolution Mode, but employs two additional system components to increase the radial velocity precision performance of GHOST.
>
> The first is a fiber agitator (REQ-5231), which increases wavelength calibration accuracy by reducing modal noise induced by imperfections, stresses, or strain within the fiber bundle, or tracking errors within the telescope guide system. This element is also required for any high S/N observations (e.g. very low equivalent width lines). The agitator states (‘on’ or ‘off’) are controlled by the user via the OT.

arturog8m:
Looks pretty nice!

swalker2m:
__Comments from Steve__

> 1. Explicit Guide Fiber State. The ConOps document details the need to allow the user to explicitly enable or disable guide fibers. Bryan asked a reasonable question that I didn't see an answer to yet. Namely, do we really need this? Usually we can default the state based on target magnitude. Would the user know enough to go against the default settings or can the ability to turn guiding on or off explicitly be relegated to lower-level control screens that only a telescope operator would see? As you can see in the model, making it overridable introduces a degree of additional complexity.

I'm OK with making this not available to the user and only settable via the GHOST app by the operator.  

> 2. Cloud Cover. The rules for defaulting guide fiber state don't take into account cloud cover. I assume that will be necessary. Does anybody have more details there?

The effect of cloud cover on default guide state should probably use the same method of CC impact on determining the 'goodness' of P2WFS or OIWFS guide star selection.

> 3. Beam-Switching Modes. In my effort estimation document Bryan notes that we won't add explicit support for beam-switching modes. This confuses me since it has a prominent place in the "Instrument Operating Modes" section. If we are doing beam-switching as described in ConOps, I'd prefer to make it explicit and constrain the configuration as described rather than infer beam-switching from the asterism plus other targets in the target environment. Does anybody have any further input about this?

Supporting Beam-Switching explicitly within the OT has always been assumed to be difficult to implement properly.  I'm happy to reconsider.  Under the current requirement set, beam-switching is supporting by having two separate GHOST sequences with the target states switched and the observers would manually swap between the two as necessary.  This is more operationally expensive but the use of this mode is expected to be low.  Again, happy to reconsider if there is a way to support explicitly

> 4. Explicit Binning. This question is similar to the one about guiding state, but the model as it stands took the opposite approach. Namely, binning is always defaulted here based on mode. The ConOps document does mention a desire for the user to be able to explicitly set the value to anything but this is not (yet) implemented. Is this really something that is needed in the user interface?

I would prefer that the OT allows for explicit setting of the binning mode for each detector.  Some science may benefit from the non-default values. 

> 5. Default Binning. Rather than specifying bright/faint/very faint can we infer this from the target magnitude (+ cloud cover?). Table 9 doesn't make it clear how we would select between faint and very faint so I didn't do this.

At this point, I'd recommend going with default binning being determined by resolution mode only.  1x2 for standard, and 1x1 for high.  I think the boundary between this and using other binning modes is fuzzy enough not to be useful. 

> 6. Two Target Base Position. I believe Bryan mentioned a desire to allow the definition of a base position at some point other than the default position between the two targets. I don't see this mentioned in the ConOps document. I'd prefer to avoid this complication unless it is really needed of course.

The situation I can imagine is that the base position may need to be at some arbitrary point, rather than the default, to reach a suitable P2WFS guide star.

swalker2m:
> I would prefer that the OT allows for explicit setting of the binning mode for each detector. Some science may benefit from the non-default values.

> At this point, I'd recommend going with default binning being determined by resolution mode only. 1x2 for standard, and 1x1 for high. I think the boundary between this and using other binning modes is fuzzy enough not to be useful.

I'm trying to sort out the binning mode vs. x/y binning settings.  Unfortunately even after all these years I'm amazingly ignorant about these things.  I guess I'd thought of the binning mode (bright / faint / very faint) as just a convenient way to specify the desired binning without dipping down to explicit x/y binning values.  Are the two actually independent settings?

If I understand what you're saying we'd need a binning mode (bright / faint / very faint) for each detector. Presumably the value that the user would select would ultimately depend on the magnitude of target in the desired conditions.  I assume then that the x/y binning should also differ per detector, or is there just one value for the observation as a whole for some reason?

To be clear, we can discard all the logic for default x/y binning that takes into account observation mode along with binning mode (bright / faint / very faint) and use 1x2 for standard resolution and 1x1 for high resolution?

swalker2m:
As you can see by my follow-up questions for science, there is still work to do.  I added a few more commits to address some of the feedback received so far.  I realize that over in another branch @tpolecat has refined the `Asterism` trait as well and so the GHOST asterisms will need to be adjusted to match when that is merged.

swalker2m:
From Steve:

> My worries in beam-switching are having target changes during an observing sequence, plus the difficulties in avoiding positioner collisions during the movement of the IFUs. But perhaps these are not as difficult as I envision or may be required anyway (prevent collisions, etc).
>
> If we can do it, great.

swalker2m:
From Steve:

> Yes, I think we can discard the logic for setting up default x/y binning and simply use 1x1 for high and 1x2 for standard resolution.
>
> While we could try to get a bit fancier with target magnitude logic, the real decision to bin beyond the defaults will be highly dependent on the specific science case and type of target, not simply just the magnitude.  All this means that we just as likely get it wrong as right and so we might as well make it easier on ourselves and easier to describe to our users.  
>
> While I think that the binning will usually be the same for both detectors, we should allow for independent settings to accommodate both possibilities.

swalker2m:
I guess there's a difference between being able to represent the request in the OT and actually executing it.  I don't know what problems it would introduce in the seqexec or below.

Maybe @jluhrs  or Cristian can help fill me in.

You say "having target changes" but I understood it to be the same target, just switched from one detector to the other.  Perhaps it will be easier to discuss this in person sometime when you're back from your trip.

swalker2m:
To be clear do we need both binning mode (bright / faint / very faint) and x/y binning values or just the x/y binning values?

swalker2m:
> To be clear do we need both binning mode (bright / faint / very faint) and x/y binning values or just the x/y binning values?

From Steve: just the x/y binning values

swalker2m:
I've updated the GHOST asterisms to match updates to the `Asterism` trait by @tpolecat.  I know this will need additional work, but we could merge this now to have an instance of a multi-target asterism.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/env/Asterism.scala
```diff
@@ -45,7 +45,8 @@ trait Asterism {
   def basePositionProperMotion: Option[ProperMotion]
 
   /** Return a display name for this asterism. */
-  def name: String
+  def name: String =
+    allTargets.map(_.name).toList.mkString
```
tpolecat: Can we `.mkString(", ")` or something so they don't get mashed together?

tpolecat: I think `.map(_.name).intercalate(", ")` will do it actually.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/env/Asterism.scala
```diff
@@ -36,7 +36,7 @@ trait Asterism {
     * only possibilities.
     * TODO: should we Church encode this instead?
     */
-  def targets: Target \/ (SiderealTarget, SiderealTarget)
+  def targets: Target \/ (Target, Target)
```
tpolecat: This defeats the purpose of the `target` method … things like `.isSidereal` depend on knowing that if there's more than one target they have to be sidereal.

swalker2m: I understood we would be removing this altogether?  At any rate, since the asterisms will need to hold mutable `SPTarget`s, I'm not sure how to guarantee that they are and always will be sidereal.

swalker2m: I'm not sure `isSidereal` makes sense for an asterism anyway.  Perhaps if it is implemented as "are all targets sidereal"?

tpolecat: There's nothing preventing us from constructing a single asterism that contains a mixture of TOO, sidereal, and non-sidereal targets; but we have a lot of code that asks "is this a sidereal observation?" and I don't know how to answer that question in the general case. We may need to examine every call site and try to figure out why they're asking.

tpolecat: But *in practice* every asterism will be `Target \/ (SiderealTarget, SiderealTarget)` … anything else is an error condition that the UI should have prevented. It's irritating that we can't count on this in a stronger way. I don't know how to solve it other than getting rid of `SPTarget`.

swalker2m:
Same comment from before applies again:

I've updated the GHOST asterisms to match updates to the Asterism trait by @tpolecat. I know this will need additional work, but we could merge this now to have an instance of a multi-target asterism.

I'll merge this if the tests pass.

tpolecat:
> I'll merge this if the tests pass.

👍

PR #1209, opened 2017-04-08
REL-724: OT tree search updates

swalker2m:
This PR addresses a task request to allow the OT's observation search panel to also work with group nodes.  Since it wasn't setup to handle anything other than observations, I just replaced it with (I think) a simpler implementation.  The changes include:

1) Searches the title of all nodes in the tree, not just observations.
2) As before a search for a number assumes an observation number.  The user can now quote a number with " characters though to search for numbers in node titles.


swalker2m:
I wrote an implementation of this with a `Zipper` but in the end I think it wasn't a good idea since the program is mutable.  It would allow stepping through filtered program nodes, but the nodes can be modified at any time after the `Zipper` is created.  You have to constantly refresh the zipper just in case something changes so I decided then to just repeat the search on demand each time as before.

We could easily expand upon this in various ways.  For example, we could search within note text as well or search for dataset file names, etc.  We could also easily implement going backwards through the tree nodes.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/viewer/SearchPanel.scala
```diff
@@ -0,0 +1,120 @@
...
+
+/** Defines a label and text box to use for searching for program nodes.
+  */
+class SearchPanel(statusFilter: StatusFilter, tree: SPTree) extends GridBagPanel {
+  import SearchPanel._
+
+  private var viewer = Option.empty[SPViewer]
+
+  def searchString: Option[String] =
+    Option(textBox.text).filterNot(_.isEmpty)
```
tpolecat: Do we want to trim whitespace, i.e., `Option(textBox.text).map(_.trim).filterNot(_.isEmpty)`?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/viewer/SearchPanel.scala
```diff
@@ -0,0 +1,120 @@
...
+  object nextAction extends Action("Next") {
+
+    def apply: Unit = {
+      val focus = for {
+        v     <- viewer
+        s     <- searchString
+        t     <- Option(v.getTree)
+        n     <- Option(t.getCurrentNode)
+        p     <- Option(n.getProgram)
+        (l, r) = p.toStream.span(_.getNodeKey =/= n.getNodeKey)
```
tpolecat: Very clever.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/viewer/SearchPanel.scala
```diff
@@ -0,0 +1,120 @@
...
+object SearchPanel {
+  // A search for a number is assumed to be an observation number.  If you want
+  // to search all nodes for one with a title containing the number you can
+  // include it in quotes.
+  private val ObsNumber    = """(\d+)""".r
+  private val QuotedString = """"([^"]*)"""".r
+
+  private def dataObjectMatcher(f: ISPDataObject => Boolean): ISPNode => Boolean =
+    (n: ISPNode) =>
+      n.getDataObject match {
```
tpolecat: This could be shortened to `_.getDataObject match {`

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/viewer/SearchPanel.scala
```diff
@@ -0,0 +1,120 @@
...
+        case d: ISPDataObject => f(d)
+        case _                => false
+      }
+
+  private def titleMatcher(text: String): ISPNode => Boolean = {
+    val textʹ = text.toLowerCase.trim
```
tpolecat: N.B. if you trim above you won't need to do so here. this could also go inside the lambda below which would let you get rid of a set of braces.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/viewer/SearchPanel.scala
```diff
@@ -0,0 +1,120 @@
...
+      Option(obj.getTitle).map(_.toLowerCase).exists(_.contains(textʹ))
+    }
+  }
+
+  private def obsNumberMatcher(num: Int): ISPNode => Boolean =
+    (n: ISPNode) =>
+      n match {
+        case o: ISPObservation => o.getObservationNumber === num
+        case _                 => false
+      }
```
tpolecat: This could just be 

```scala
private def obsNumberMatcher(num: Int): ISPNode => Boolean = {
  case o: ISPObservation => o.getObservationNumber === num
  case _                 => false
}
```

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/viewer/SearchPanel.scala
swalker2m: I think I was under the impression that `getDataObject` returned an `Object` but in fact it is already an `ISPDataObject` so this can even be a one liner

````scala
n => Option(n.getDataObject).exists(f)
````

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/viewer/SearchPanel.scala
swalker2m: Thanks, I looked at this again and decided to move the `trim` down to where the matcher is created so I could make quoted text match case exactly while unquoted search text ignores case.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/viewer/SearchPanel.scala
sraaphorst: I'm a bit confused. Isn't `p` an `ISPProgram`? I'm not sure what the `toStream` method is acting on or where it is coming from.

This looks good to me. As @tpolecat said, very nice, elegant solution for the search.

swalker2m:
> I'm a bit confused. Isn't p an ISPProgram? I'm not sure what the toStream method is acting on or where it is coming from.

It's added via `edu.gemini.spModel.rich.pot.sp.RichNode`.  I agree though that implicits can be confusing.

sraaphorst:
Ahhh, yes. I see now. Thanks for clarifying that for me!

PR #1210, opened 2017-04-10
REL-2984: Add CFH exchange participant to the OT

sraaphorst:
This simply follows through on the work done on the PIT and adds CFH as an exchange partner in the OT. By adding it to the `TimeAcctCategory`, since the UI and reports for time keeping iterate over the enum, this propagates automatically through as it should.


bundle/edu.gemini.spdb.shell/src/main/java/edu/gemini/spdb/shell/misc/TimeAcctSetterFunctor.java
```diff
@@ -1,136 +0,0 @@
-//
```
sraaphorst: This class was unused.

sraaphorst:
Note that I also did some minor cleanup of the code as I went. Sorry for the added confusion.

PR #1211, opened 2017-04-10
WIP - asterism suffering

tpolecat:
This adds the initial easy bits in the `SPTarget` to `Asterism` migration in `TargetEnvironment`. For now `getAsterism` is an *additional* method that simply wraps the existing `base` in a single-element asterism. Eventually `getBase` will be removed and the `TargetEnvironment` itself will be modified to actually deal with asterisms internally.

Much more to come. I will ask for feedback when I hit things with non-obvious answers.

### DIFF
--- a/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsHash.scala
+++ b/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsHash.scala
@@ -59,13 +59,13 @@ object AgsHash {
     // Base Position
     Option(ctx.getTargets).foreach { t =>
       val time = Some(new java.lang.Long(when)).asGeminiOpt
-      val base = t.getBase
+      val asterism = t.getAsterism
 
       def toData(coord: GemOption[java.lang.Double]): Int =
         coord.asScalaOpt.map(_.doubleValue).##
 
-      val ra  = toData(base.getRaDegrees(time))
-      val dec = toData(base.getDecDegrees(time))
+      val ra  = toData(asterism.getRaDegrees(time))
+      val dec = toData(asterism.getDecDegrees(time))
 
       ra +=: dec +=: buf
     }
--- a/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsHash.scala
+++ b/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsHash.scala
@@ -7,12 +7,12 @@ import edu.gemini.spModel.gemini.altair.InstAltair
 import edu.gemini.spModel.gemini.flamingos2.Flamingos2
 import edu.gemini.spModel.gemini.gmos.{InstGmosNorth, InstGmosSouth}
 import edu.gemini.spModel.obs.context.ObsContext
-
 import java.time.Instant
+
+import edu.gemini.spModel.target.SPTarget
 
 import scala.collection.mutable.ListBuffer
 import scala.util.hashing.{MurmurHash3 => M3}
-
 import scala.collection.JavaConverters._
 
 /** Computes an `Int` hash value that corresponds to the set of inputs to the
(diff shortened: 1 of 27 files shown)


bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsHash.scala
```diff
@@ -59,13 +59,13 @@ object AgsHash {
...
       val time = Some(new java.lang.Long(when)).asGeminiOpt
-      val base = t.getBase
+      val asterism = t.getAsterism
 
       def toData(coord: GemOption[java.lang.Double]): Int =
         coord.asScalaOpt.map(_.doubleValue).##
 
-      val ra  = toData(base.getRaDegrees(time))
-      val dec = toData(base.getDecDegrees(time))
+      val ra  = toData(asterism.getRaDegrees(time))
```
cquiroz: If an asterism has many targets what does it mean to get RaDec? Which target is selected?

tpolecat: The base position is defined as the point equidistant from the targets in the asterism, so if that's all you need then an `Asterism` is as good as an `SPTarget`.

swalker2m: If there are just two targets, as for GHOST dual-target mode, the base position is equidistant from the two targets.  In general by default though it is the center of the smallest circle that encloses all of the targets.  That minimizes the maximum distance to any target in the asterism.

cquiroz: Nice, that will make interoperability much simpler with all the code that assume there is only one target

cquiroz: @swalker2m Do you do the calculation with the smallest circle to account for the curvature of the sky? In a plane it seems the middle point would be the same as the center of the smallest circle

bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/nifs/NifsRule.java
```diff
@@ -184,17 +185,17 @@ public IConfigMatcher getMatcher() {
         P2Problems prob = new P2Problems();
         public IP2Problems check(ObservationElements elements)  {
             for (ObsContext ctx : elements.getObsContext()) {
-                final SPTarget baseTarget       = ctx.getTargets().getBase();
+                final Asterism asterism       = ctx.getTargets().getAsterism();
                 final Option<SPTarget> oiTarget = getOITarget(ctx);
                 final Option<SPTarget> aoTarget = getAOTarget(ctx);
 
-                if (baseTarget == null || oiTarget.isEmpty() || aoTarget.isEmpty()) return null;
+                if (asterism == null || oiTarget.isEmpty() || aoTarget.isEmpty()) return null;
```
swalker2m: There was always supposed to be a base position, and there should similarly always be an asterism now.  I guess it's not worth going through and removing the `null` checks though?

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsHash.scala
swalker2m: I haven't done it yet, and I may not because we don't have an asterism with more than two targets yet.  The GHOST dual target base position is calculated by interpolating the two targets with `Coordinates.interpolate`.

My thought though was that I would pick a random target as a base position (0,0) and then calculate the angular separation between it and the other targets as an `Offset` with `Coordinates.difference(c0, c1).offset`.  In other words, the curvature comes into account when calculating the offset and when converting back to RA/Dec but the smallest circle calculation would work with the projection of the points on a plane.  It's likely I'm confused though, which is another good reason to wait until we have a 3+ target asterism.

sraaphorst: 👍 
This was an enlightening discussion, and this change seems very straightforward.

cquiroz: I guess the circle would work fine given that our FOV is normally small but I don't know enough about the scales to tell

bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/nifs/NifsRule.java
tpolecat: My inclination is to rock the boat as little as possible in this PR because it's going to get very big before I'm done. But removing these checks would be a useful thing to do at some point.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsHash.scala
```diff
@@ -56,18 +56,17 @@ object AgsHash {
...
-      ra +=: dec +=: buf
+      def hashTarget(t: SPTarget): Unit = {
+        def toData(coord: GemOption[java.lang.Double]): Int =
+          coord.asScalaOpt.map(_.doubleValue).##
+        val ra  = toData(t.getRaDegrees(time))
+        val dec = toData(t.getDecDegrees(time))
+        ra +=: dec +=: buf
+      }
+      env.getAsterism.targets.foreach(hashTarget)
     }
```
tpolecat: I'm kind of assuming we want to redo AGS if any component of the asterism moves (as opposed to only looking at movement of the center point). I think in every real-life case these will be equivalent, but I'm trying to anticipate corner cases. I guess it depends on what the GHOST strategy is sensitive to.

swalker2m: 👍  GHOST will use a combination of guide fibers (surrounding the science targets) and P2.  The guide fiber state won't be controlled by AGS, I assume, so it boils down to finding a P2 star.  It seems to me that the positions of the individual stars will matter there since we'll want to not only find a P2 star that is reachable from the base position but also one that doesn't vignette the science targets.

tpolecat:
Closed in favor of #1222

PR #1212, opened 2017-04-11
REL-246: reorder baseline cals to place flats before arcs

swalker2m:
This update would rearrange any flats in a baseline calibration to happen before any arcs (instead of the other way around as it is currently done).

Since we're modifying the sequence calculation, the question "what about existing executed sequences" naturally comes to mind.  Smart GCal was written to handle this type of change and simple monkey tests and a couple of unit tests confirm that it works, at least to some degree.  At any rate any problems with this feature are already present in the codebase.

### DIFF
--- a/bundle/edu.gemini.pot/src/test/scala/edu/gemini/spModel/gemini/seqcomp/BaselineReorderTest.scala
+++ b/bundle/edu.gemini.pot/src/test/scala/edu/gemini/spModel/gemini/seqcomp/BaselineReorderTest.scala
@@ -1,0 +1,79 @@
+package edu.gemini.spModel.gemini.seqcomp
+
+import edu.gemini.spModel.config.ConfigBridge
+import edu.gemini.spModel.config.map.ConfigValMapInstances.IDENTITY_MAP
+import edu.gemini.spModel.config2.ItemKey
+import edu.gemini.spModel.gemini.calunit.CalUnitParams._
+import edu.gemini.spModel.gemini.flamingos2.Flamingos2
+import org.junit.Assert._
+import org.junit.{Before, Test}
+
+final class BaselineReorderTest extends GcalExecTest {
+
+  val observeTypeItem  = new ItemKey("observe:observeType")
+
+  @Before override def setUp(): Unit = {
+    super.setUp(progId)
+
+    addObsComponent(Flamingos2.SP_TYPE)
+
+    // 2 baseline night observations (4 datasets)
+    (1 to 2).foreach { _ =>
+      addSeqComponent(getObs.getSeqComponent, SeqRepeatSmartGcalObs.BaselineNight.SP_TYPE)
+    }
+  }
+
+  val flat = CalImpl(Set(Lamp.IR_GREY_BODY_HIGH), Shutter.OPEN,   Filter.ND_20, Diffuser.IR, 1,  4.0, 1, arc = false)
+  val arc  = CalImpl(Set(Lamp.AR_ARC),            Shutter.CLOSED, Filter.NIR,   Diffuser.IR, 1, 15.0, 1, arc = true )
+
+  @Test def testOrderChange(): Unit = {
+
+    // Originally, arcs came before flats.
+    setupGcal(arc, flat)
+    exec(1)
+    exec(2)
+
+    // Now flats come before arcs
+    setupGcal(flat, arc)
+
+    // Generate the sequence.
+    val cs = ConfigBridge.extractSequence(getObs, null, IDENTITY_MAP)
+
+    assertEquals(4, cs.getAllSteps.size)
+
+    // Even though flats come before arcs, the first two steps are already
+    // executed so they don't flip to flat, arc.
+    val actual   = cs.getItemValueAtEachStep(observeTypeItem)
+    val expected = List("ARC", "FLAT", "FLAT", "ARC")  // yes, Strings
+
+    actual.zip(expected).foreach { case (a,e) => assertEquals(a, e) }
+  }
+
+  @Test def testPartialExecution(): Unit = {
+
+    // Originally, arcs came before flats.  Do just one of the two baseline
+    // calibration steps.
+    setupGcal(arc, flat)
+    exec(1)
+
+    // Now flats come before arcs
+    setupGcal(flat, arc)
+
+    // Generate the sequence.
+    val cs = ConfigBridge.extractSequence(getObs, null, IDENTITY_MAP)
+
+    assertEquals(4, cs.getAllSteps.size)
+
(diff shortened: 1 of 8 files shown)


bundle/edu.gemini.spModel.smartgcal/src/main/java/edu/gemini/spModel/smartgcal/provider/CalibrationProviderImpl.java
```diff
@@ -33,18 +33,30 @@ public CalibrationProviderImpl(CalibrationRepository repository) {
...
-        ConfigurationKey configKey = key.getConfig();
+        final ConfigurationKey configKey = key.getConfig();
+        final List<Calibration> cals;
         if (key instanceof CalibrationKeyImpl.WithWavelength) {
-            Double wavelength = ((CalibrationKeyImpl.WithWavelength) key).getWavelength();
-            return getCalibrations(
-                    configKey.getInstrumentName(),
-                    configKey,
-                    wavelength);
+            final Double wavelength = ((CalibrationKeyImpl.WithWavelength) key).getWavelength();
```
sraaphorst: Not that it's really important, but I suppose there's no reason to use `Double` instead of `double` here, especially since this is unboxed in the called method?

I'm not even sure why `CalibrationKeyImpl.WithWavelength` has a `Double` instead of a `double`, since a quick glance at the code that instantiates these objects in the instrument specifically seems to make sure `null` will never be passed into the constructor for the wavelength value.

bundle/edu.gemini.pot/src/test/scala/edu/gemini/spModel/gemini/seqcomp/BaselineReorderTest.scala
```diff
@@ -0,0 +1,79 @@
...
+
+import edu.gemini.spModel.config.ConfigBridge
+import edu.gemini.spModel.config.map.ConfigValMapInstances.IDENTITY_MAP
+import edu.gemini.spModel.config2.ItemKey
+import edu.gemini.spModel.gemini.calunit.CalUnitParams._
+import edu.gemini.spModel.gemini.flamingos2.Flamingos2
+import org.junit.Assert._
+import org.junit.{Before, Test}
+
+final class BaselineReorderTest extends GcalExecTest {
```
cquiroz: Any reason to prefer Junit over scalatest? Is it due to `GcalExecTest`?

bundle/edu.gemini.pot/src/test/scala/edu/gemini/spModel/gemini/seqcomp/CalImpl.scala
```diff
@@ -0,0 +1,31 @@
...
+import edu.gemini.spModel.gemini.calunit.CalUnitParams.{Diffuser, Filter, Shutter, Lamp}
+import edu.gemini.spModel.gemini.calunit.smartgcal.Calibration
+
+import java.lang.{Boolean => JBoolean, Double => JDouble, Integer => JInteger}
+
+import scala.beans.BeanProperty
+import scala.collection.JavaConverters._
+
+final case class CalImpl(lamps: Set[Lamp],
+           @BeanProperty shutter: Shutter,
```
cquiroz: Will the annotation create the getter/setter for you?

bundle/edu.gemini.spModel.smartgcal/src/main/java/edu/gemini/spModel/smartgcal/provider/CalibrationProviderImpl.java
```diff
@@ -33,18 +33,30 @@ public CalibrationProviderImpl(CalibrationRepository repository) {
...
+                flats.add(cal);
+            } else {
+                arcs.add(cal);
+            }
+        });
+
+        // Return the flats before the arcs.
+        final List<Calibration> result = new ArrayList<>(flats);
+        result.addAll(arcs);
+        return result;
```
swalker2m: This is the heart of the thing.  There are more Java-8ish ways to do this but I find them equally awkward.  Dunno.

This one may be clearer.  Makes several passes through the calibrations.
````java
        final Stream<Calibration> calStream  = cals.stream();
        final Stream<Calibration> flatStream = calStream.filter(c -> c.isFlat());
        final Stream<Calibration> arcStream  = calStream.filter(c -> c.isArc());
        return Stream.concat(flatStream, arcStream).collect(Collectors.toList());
````

Then there's the awkward `Collectors`:

````java
        final Map<Boolean, List<Calibration>> flatsArcs =
                cals.stream()
                    .collect(Collectors.partitioningBy(c -> c.isFlat()));
        
        final List<Calibration> result = new ArrayList(flatsArcs.get(true));
        result.addAll(flatsArcs.get(false));
        return result;
````

I'm inclined to leave it as is unless there's a better way?

bundle/edu.gemini.pot/src/test/scala/edu/gemini/spModel/gemini/seqcomp/BaselineReorderTest.scala
swalker2m: There was an existing test using JUnit that was very similar so I just went with the flow.

bundle/edu.gemini.pot/src/test/scala/edu/gemini/spModel/gemini/seqcomp/CalImpl.scala
swalker2m: Yes.  This is just an implementation for the test cases, by the way.

bundle/edu.gemini.spModel.smartgcal/src/main/java/edu/gemini/spModel/smartgcal/provider/CalibrationProviderImpl.java
swalker2m: Good question.  I'm very hesitant to refactor Smart GCal though.  I just made cosmetic changes to these lines (added a `final` and removed the `return`s because I need to manipulate the result before handing it back.

The implementation in `ocs3` is straightforward, thanks to the new sequence model and the database, so I'll just leave further improvements until `ocs3`.

bundle/edu.gemini.spModel.smartgcal/src/main/java/edu/gemini/spModel/smartgcal/provider/CalibrationProviderImpl.java
sraaphorst: I was looking at precisely this.... using `Collectors.paritioningBy` to recommend, but it got so horrible and messy that the original solution you had seemed so much better and more readable.

PR #1213, opened 2017-04-12
REL-246: work around an issue with Smart GCal testing

swalker2m:
While working on an unrelated feature, I noticed a sporadic failure with the Smart GCal test cases.  I believe the issue lies in an implementation class called `CalibrationProviderHolder`.  This class maintains a shared mutable reference to another class, the `CalibrationProvider`.  I'm not too interested in refactoring Smart GCal to fix this kind of thing because we have reimplemented it in a much less convoluted way already for ocs3.  Design questions aside, this creates a problem for the test cases which need to swap out the `CalibrationProvider` for testing and can't work in an environment in which the provider is changing out from under them.

There is some way to get limited serialization of sbt tests, but this needs to work across all test cases that are doing Smart Gcal.  I don't know of another way to handle this besides just using a shared lock, which is what this PR proposes.

__By the way, I'm looking forward to some kind of animated GIF indicating disgust with this whole thing, so please don't disappoint!__  Alternative suggestions are also welcome.

### DIFF
--- a/bundle/edu.gemini.pot/src/test/scala/edu/gemini/spModel/gemini/seqcomp/BaselineReorderTest.scala
+++ b/bundle/edu.gemini.pot/src/test/scala/edu/gemini/spModel/gemini/seqcomp/BaselineReorderTest.scala
@@ -27,53 +27,57 @@ final class BaselineReorderTest extends GcalExecTest {
   val arc  = CalImpl(Set(Lamp.AR_ARC),            Shutter.CLOSED, Filter.NIR,   Diffuser.IR, 1, 15.0, 1, arc = true )
 
   @Test def testOrderChange(): Unit = {
+    locking {
 
-    // Originally, arcs came before flats.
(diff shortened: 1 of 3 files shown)


bundle/edu.gemini.pot/src/test/scala/edu/gemini/spModel/gemini/seqcomp/BaselineReorderTest.scala
```diff
@@ -27,53 +27,57 @@ final class BaselineReorderTest extends GcalExecTest {
   val arc  = CalImpl(Set(Lamp.AR_ARC),            Shutter.CLOSED, Filter.NIR,   Diffuser.IR, 1, 15.0, 1, arc = true )
 
   @Test def testOrderChange(): Unit = {
+    locking {
```
cquiroz: Given that this is in tests and as you say this will be soon deprecated I'm quite ok with this

PR #1214, opened 2017-04-13
REL-684: Observation Status Mail Updates

swalker2m:
Two changes to observation status mails:

1) Now all contacts are emailed when there is a status change.  That means any PI, NGO, and Gemini contacts are included.  The different mails will differ in who the message is to vs who is cc'ed though.

2) A new status email is sent when a ToO observation is set to ON_HOLD.

The test cases for this code had been disabled for some reason, so much of the effort involved was getting them running again.


bundle/edu.gemini.spdb.reports.collection/src/main/java/edu/gemini/dbTools/mail/PrepareMessageAction.java
```diff
@@ -95,244 +100,102 @@ private Properties getProperties() {
...
-        if ((adrs == null) || (adrs.length == 0)) {
-            if (logError) {
-                _log.warning(action + ": no Gemini contact addresses. Prog id = " +
-                        _mail.getProgramId());
-            }
-            return null;
-        }
-        return adrs;
-    }
+            final Message.RecipientType ccType = (to.length > 0) ? CC : TO;
```
cquiroz: So if there is on `to` addresses the `cc` ones are used as `to`?

bundle/edu.gemini.spdb.reports.collection/src/test/java/edu/gemini/dbTools/mail/test/TestProgram.java
```diff
@@ -23,7 +26,7 @@ protected String getGeminiAddressesStr() {
     }
 
     protected OdbMailConfig getMailConfig() {
-        throw new Error("Not implemented, sorry."); // TODO
+        return config;
```
cquiroz: 👍

tpolecat:
Ow my eyes. I need to stop trying to read the deleted code. Much better. 👍

bundle/edu.gemini.spdb.reports.collection/src/main/java/edu/gemini/dbTools/mail/PrepareMessageAction.java
swalker2m: Correct.

PR #1215, opened 2017-04-23
Feature/REL-3032: Add ITC support for new Flamingos2 K-red & K-blue filters

andrewwstephens:


cquiroz:
It seems you didn't need to update the baseline for this change?

andrewwstephens:
@cquiroz - These are new filters so they did not affect any test results.

PR #1218, opened 2017-04-27
REL-3128: Changing from des to queryString as des is not necessarily …

sraaphorst:
As per Rob's comments on previous REL-3128 PR.


bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
```diff
@@ -81,7 +81,7 @@ class LchQueryFunctor(queryType: LchQueryFunctor.QueryType,
             n <- target.getNonSiderealTarget
             h <- n.horizonsDesignation
           } {
-            setHorizonsObjectId(h.des)
+            setHorizonsObjectId(h.queryString)
```
tpolecat: Assuming it doesn't need to be human-readable, 👍

sraaphorst: No, no human readability required 😸

PR #1220, opened 2017-04-28
REL-3128: Updating ODB browser bundle.

sraaphorst:
Updating the ODB browser bundle to be at the same version as the LTTS.


swalker2m:
No client side changes needed?  From the version number difference it looks like the API might have changed.

sraaphorst:
Honestly, nothing changed in the actual XML, so I'm not even sure we need to update this jar. There was a previous PR that I merged under this REL task that now uses a previously unused but already existing field, but that's not used by anything other than the LTTS.

Since the LTTS version number was bumped and thus the bundle, I thought better safe than sorry.

PR #1222, opened 2017-05-01
Asterism Updates part A

tpolecat:
This is the first set of updates to support asterisms in the target environment. It updates all the low-hanging fruit but leaves UI work for later PRs.


bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/ScienceTargetStrategy.scala
```diff
@@ -15,25 +15,29 @@ case class ScienceTargetStrategy(key: AgsStrategyKey, guideProbe: ValidatableGui
...
-    val siderealTarget = ctx.getTargets.getBase.toSiderealTarget(ctx.getSchedulingBlockStart)
-    val posAngle       = ctx.getPositionAngle
-    val assignment     = AgsStrategy.Assignment(guideProbe, siderealTarget)
-    val selection      = AgsStrategy.Selection(posAngle, List(assignment))
+    val siderealTargets = toSiderealTargets(ctx)
+    val assignments     = siderealTargets.map(AgsStrategy.Assignment(guideProbe, _))
+    val posAngle        = ctx.getPositionAngle
+    val selection       = AgsStrategy.Selection(posAngle, assignments)
     Future.successful(Some(selection))
   }
```
tpolecat: This algorithm used to assign the probe to the [one and only] base position, coerced into being a sidereal target. This coercion operation is ridiculously sketchy (it turns non-sidereal targets into fixed targets and falls back to a zenith target when coordinates are unknown) but I'm not touching it. The change simply returns an assignment for all science targets rather than assuming one. The selection takes a `List` so I assume this is ok.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gems/Canopus.java
```diff
@@ -329,13 +330,13 @@ public boolean inRange(ObsContext ctx, Offset offset){
...
-        SPTarget base   = ctx.getTargets().getBase();
-        SPTarget target = spTargetOpt.getValue();
+        Asterism asterism = ctx.getTargets().getAsterism();
+        SPTarget target   = spTargetOpt.getValue();
 
         final Option<Long> when = ctx.getSchedulingBlockStart();
 
         return
-            base.getSkycalcCoordinates(when).flatMap(bc ->
+            asterism.getSkycalcCoordinates(when).flatMap(bc ->
```
tpolecat: It would be really nice to search and destroy the remaining uses of Skycalc coordinates.

bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/data/ObservationProvider.scala
```diff
@@ -181,7 +184,15 @@ case class PositionProvider(ctx: QvContext, base: ObservationProvider) extends O
...
-        val newTarget = o.getTargetEnvironment.getBase.clone()
+
+        // The check above ensures that this is a nonsidereal observation, which means it must have
+        // a single-target asterism. If this is not the case it means the model has changed.
+        val newTarget = o
+          .getTargetEnvironment
+          .getAsterism
+          .getNonSiderealSpTarget
+          .getOrElse(sys.error("The asterism is not a single nonsidereal target."))
+
```
tpolecat: I worry about checks like this, but I don't know how we can avoid them. We have code that's down in holes where we can only arrive based on some outside knowledge that we have a single-target environment (all the instrument-specific editors, for instance).

bundle/edu.gemini.spdb.rollover.servlet/src/main/scala/edu/gemini/rollover/servlet/RolloverServlet.scala
```diff
@@ -55,10 +55,12 @@ final class RolloverServlet(site: Site, odb: IDBDatabaseService, user: java.util
...
-      <target>
-        <ra>{ ro.target.coords.map(_.ra).getOrElse(0.0) }</ra>
-        <dec>{ ro.target.coords.map(_.dec).getOrElse(0.0) }</dec>
-      </target>
+      {ro.targets.map { t =>
+        <target>
+          <ra>{  t.coords.map(_.ra).getOrElse(0.0) }</ra>
+          <dec>{ t.coords.map(_.dec).getOrElse(0.0) }</dec>
+        </target>
+      }}
```
tpolecat: The rollover servlet output may contain more than one target now. I don't know who consumes this. The "smash to zero" logic for the coordinates is ridiculous.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcPanel.scala
```diff
@@ -47,6 +47,12 @@ sealed trait ItcPanel extends GridBagPanel {
   def display: Component
   def visibleFor(t: SPComponentType): Boolean
 
+  /** TODO: we need to display a list of asterism members and let the user select one. For now we
+    * just select the first target in the asterism.
+    */
+  def selectedTarget: Option[Target] =
+    Option(owner.getContextTargetEnv).map(_.getAsterism.allTargets.head)
+
```
tpolecat: The `ItcPanel` needs to add a new UI element to let you select the asterism component. Right now it just uses the first target. This was easy to stub out but other UI elements won't be so easy.

bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/data/ObservationProvider.scala
cquiroz: I don't think there is a much better way do this without bubbling up the `Option` which as you said will bring a bunch of other changes

bundle/edu.gemini.spdb.rollover.servlet/src/main/scala/edu/gemini/rollover/servlet/RolloverServlet.scala
cquiroz: Rollover servlet is used by ITAC. We should be careful here to avoid changes that impact the output especially when we are 2 weeks away from the next ITAC meeting

tpolecat: Ok, that's helpful. Initially I guess ITAC can just look at the first target (that's probably what will happen anyway if it's just selecting the element by path). But I will check.

Re: merging this isn't going to happen any time soon so you don't need to worry about upcoming ITAC.

cquiroz: I think for rollovers targets aren't that important since the time for the rollover just get discounted. Two things to watch for:

* Are we feeding the rollovers back to SPDB and thus using another target could be problematic? (Probably we are not doing that)
* Confirm that the the rollover time just get discounted and not that the proposals are fed to the queue engine where visibility restrictions could come into play

tpolecat: Yes, verified that it will continue to work with multiple targets. ITAC just [grabs the first one](https://github.com/gemini-hlsw/itac/blob/7ce7fcd341c026c6ac9a37a5ce472e79acc267f4/phase1-data/src/main/java/edu/gemini/tac/persistence/rollover/RolloverObservation.java#L58).

tpolecat: I don't understand either of the two points above. These are ITAC questions?

cquiroz: Yes they are internal to ITAC. We should create a ticket to verify them

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/env/Asterism.scala
```diff
@@ -1,31 +1,66 @@
...
+  /** All Targets that comprise the asterism. */
+  def allTargets: NonEmptyList[Target] =
+    targets.fold(NonEmptyList(_), p => NonEmptyList(p._1, p._2))
+
+  /** An asterism is a single generic target, or a pair of sidereal targets. These are currently the
+    * only possibilities.
+    * TODO: should we Church encode this instead?
     */
-  def targets: NonEmptyList[SPTarget]
+  def targets: Target \/ (SiderealTarget, SiderealTarget)
```
swalker2m: I assumed in situations where it matters we'll need to match on the specific asterism type.  As soon as a new instrument comes along with three targets this is no longer valid.

bundle/edu.gemini.spdb.rollover.servlet/src/main/scala/edu/gemini/rollover/servlet/RolloverServlet.scala
swalker2m: For rollover, probably just the base position is what we want.  It's about reserving time for the observation during the semester so we don't schedule too many new observations for the corresponding part of the sky.  This may just work as is, but we'd need to check the consumer of this to make sure it can handle multiple targets if we leave it this way.  I guess it *might* just read the first target which would be fine.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcPanel.scala
swalker2m: It's a good question how to handle this.  For GHOST dual-target I think the two targets need to be nearly the same magnitude since they are being observed by the same detector.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/tpe/TpeContext.scala
```diff
@@ -77,7 +77,8 @@ final class TargetContext(obs: Option[ISPObservation]) extends TpeSubContext[ISP
 
   def envOrDefault: TargetEnvironment = env.getOrElse(TargetEnvironment.create(baseOrDefault))
 
-  def base: Option[SPTarget] = env.map(_.getBase)
+  // TODO: deal with multi-target asterisms
+  def base: Option[SPTarget] = env.map(_.getAsterism.allSpTargets.head)
```
swalker2m: This would be another good place to mark with a "TODO ASTERISM" or some distinct label that we can easily search for later.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/ScienceTargetStrategy.scala
tpolecat: Hm ok, that's my fault for trusting the types.

bundle/edu.gemini.spdb.rollover.servlet/src/main/scala/edu/gemini/rollover/servlet/RolloverServlet.scala
tpolecat: It only reads the first target, but if the base position is what we want I can change it.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcPanel.scala
tpolecat: Yeah there are other places in code that I have avoided so far that look at the magnitude table. I don't know how to deal with these yet. It may be the case that it only happens in a context where we "know" there's only one target.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsHash.scala
```diff
@@ -59,13 +59,13 @@ object AgsHash {
...
-      val base = t.getBase
+      val asterism = t.getAsterism
 
       def toData(coord: GemOption[java.lang.Double]): Int =
         coord.asScalaOpt.map(_.doubleValue).##
 
-      val ra  = toData(base.getRaDegrees(time))
-      val dec = toData(base.getDecDegrees(time))
+      val ra  = toData(asterism.getRaDegrees(time))
+      val dec = toData(asterism.getDecDegrees(time))
```
tpolecat: Ah you're right, thanks. This didn't make it over from the previous PR.

bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/general/GeneralRule.java
```diff
@@ -574,7 +565,8 @@ private SPTarget getObsTarget(final ObservationElements elements)  {
             if (targetEnv != null) {
                 final ObsClass obsClass = ObsClassService.lookupObsClass(elements.getObservationNode());
                 if (obsClass == ObsClass.SCIENCE) {
-                    final SPTarget target = targetEnv.getBase();
+                    // TODO: this needs to handle multiple targets
+                    final SPTarget target = targetEnv.getAsterism().allSpTargets().head();
```
tpolecat: Good idea re: `TODO`

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gpi/GpiCB.java
```diff
@@ -88,17 +88,6 @@ public void apply(Magnitude magnitude) {
...
-        if (targetcomp != null) {
-            final TargetObsComp toc = (TargetObsComp) targetcomp.getDataObject();
-            if (toc != null) {
-                if (toc.getTargetEnvironment().getBase() != null) {
-                    ImList<Magnitude> magnitudes = toc.getTargetEnvironment().getBase().getMagnitudesJava();
-                    magnitudes.filter(new MagnitudeFilter(MagnitudeBand.H$.MODULE$)).headOption().foreach(new MagnitudeSetter(Gpi.MAG_H_PROP));
-                    magnitudes.filter(new MagnitudeFilter(MagnitudeBand.I$.MODULE$)).headOption().foreach(new MagnitudeSetter(Gpi.MAG_I_PROP));
-                }
-            }
-        }
```
tpolecat: Yeah I talked to Carlos and he thought it was probably left over and could be deleted. It doesn't do anything and it's been there since we imported OCS into GitHub.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/TargetEnvironment.java
```diff
@@ -235,8 +241,9 @@ public boolean containsTarget(SPTarget target) {
      * Returns <code>true</code> if the given target is the base position in
      * this environment.
      */
+    @Deprecated
     public boolean isBasePosition(SPTarget target) {
-        return target == base;
+        return target == getBase();
```
tpolecat: Yeah I haven't figured out a fix for that one yet. It's not clear whether it's sufficient to say `allSpTargets.exists(_ == target)` or not. All the call sites come from code I don't understand.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/env/Asterism.scala
tpolecat: Yeah I coded it this way specifically so it would break when we add a new target type, since exhaustive matching isn't super reliable and we have so many build warnings it could easily be missed.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/obsComp/TargetObsComp.java
```diff
@@ -73,25 +74,27 @@ public Object clone() {
...
         return targetEnv;
     }
 
     /**
+     * Get the target environment (never null).
+     */
+    public Asterism getAsterism() {
+      return getTargetEnvironment().getAsterism();
+    }
+
```
sraaphorst: Just a note that the comment header for this method is incorrect.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/env/Asterism.scala
tpolecat: So, the important observation here is that the only new case we need to handle is a pair of *sidereal* targets, which is much easier than dealing with an arbitrary number of generic targets. Answering questions about the ephemeris, TOO status, etc., are much easier because we don't need to consider them in the multi-target case.

swalker2m: I suppose it makes sense to worry about the three target case, or whatever else is coming, when it happens.  This will make it easier to make all the other changes that need to be made to support switching out the base position with an asterism without.  It seems to me at some point though we'll have to punt and match on the asterism.

tpolecat: I guess I'm hoping that we'll be off of OCS before this gets any weirder.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsHash.scala
```diff
@@ -56,18 +56,20 @@ object AgsHash {
...
 
       def toData(coord: GemOption[java.lang.Double]): Int =
         coord.asScalaOpt.map(_.doubleValue).##
 
-      val ra  = toData(asterism.getRaDegrees(time))
-      val dec = toData(asterism.getDecDegrees(time))
+      asterism.allSpTargets.map { sp =>
+        val ra  = toData(sp.getRaDegrees(time))
+        val dec = toData(sp.getDecDegrees(time))
+        ra +=: dec +=: buf
```
cquiroz: What's `+=:`?

swalker2m: It appends an element to mutable `ListBuffer`.  The individual hash values for the various fields of interest are added to a list which is then hashed to a single value by `MurmurHash3`.

tpolecat:
rebased

bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/ConfigExtractor.scala
```diff
@@ -264,7 +266,7 @@ object ConfigExtractor {
         fieldLens <- extract[FieldLens]     (c, AoFieldLensKey)
         wfsMode   <- extract[GuideStarType] (c, AoGuideStarTypeKey)
       } yield {
-        val separation = distance(targetEnv.getBase, guideStar, when)
+        val separation = distance(targetEnv.getAsterism, guideStar.getTarget, when)
```
tpolecat: Is this correct? We're computing the distance from the base position to the guide star.

swalker2m: @andrewwstephens can you provide some insight?  This is sort of a theoretical question for the time being since GN / Altair isn't used with an instrument that has multiple targets.  If it were used with a multi-target instrument would it make sense to consider the distance between the center point of the asterism and the guide star for strehl calculations?  Perhaps AO and multiple targets wouldn't mix in this way in practice?

I found this calculation which seems relevant:

````java
    // Calculates the strehl noise from the guide star distance
    //  the return values is used to calculate the strehl
    public double getStrehlNoiseDist() {
        double dgs0, dgspow;

        if (fieldLensIsIn()) {
            dgs0 = 30d; //if FL in (FL always in for LGS)
            dgspow = 2d; //if FL in (FL always in for LGS)

        } else {
            dgs0 = 5d; //if FL out
            dgspow = 1d; //if FL out
        }
        //double strehlNoiseDistance = Math.exp(-1*Math.pow(guideStarDistance/12.5,2)*Math.pow(1650/wavelength,2));  //Old caclulation

        return Math.exp(-1 * Math.pow(altair.guideStarSeparation() / dgs0, dgspow) * Math.pow(1650 / wavelength, 2));
    }
````

**Comment by tpolecat** [comment]
I have renamed `TargetEnvironment::getBase` to `getArbitraryTargetFromAsterism` to make it stand out a bit more in existing code, and added a delegate of the same name to `TargetObsComp`. Both are deprecated and will be removed for 18A.

As things stand I think we're probably close to being able to merge this. Everything works as before, but the target environment contains an `Asterism` instead of a base `SPTarget`, with PIO code and a model updater in place. Most UI code doesn't yet use the asterism; most back-end code does. I don't think it's necessarily beneficial to try to do all of this in a single PR.

Remaining work can be identified by finding callers of `getArbitraryTargetFromAsterism` and comments tagged with `TODO:ASTERISM:`.

File: bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/ConfigExtractor.scala (hunk 14 above)
Comment (andrewwstephens) [inline bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/ConfigExtractor.scala]: Good question.  While we need the distance from the base to know if a particular guide star is viable, what matters for the calculation of the AO correction quality (Strehl) is the distance between each science target and the guide star.  The Strehl will be the highest for the target closest to the guide star and lowest for the target farthest from the guide star.

File: bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/env/Asterism.scala (hunk 6 above)
Comment (tpolecat) [inline bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/env/Asterism.scala]: This turned out to be unnecessary. We can remove it after the GHOST asterism is integrated.

File: bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/ConfigExtractor.scala (hunk 14 above)
Comment (tpolecat) [inline bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/ConfigExtractor.scala]: I'm inclined to leave this as-is for now. I will add a comment with a link to this conversation so we can review later.

File: bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/general/GeneralRule.java [hunk 15]
```diff
@@ -310,17 +312,14 @@ public IP2Problems check(final ObservationElements elements)  {
...
 
-            if (scienceTarget == null) { //really unlikely
-                problems.addError(PREFIX+"NO_SCIENCE_TARGET_MSG", NO_SCIENCE_TARGET_MSG, elements.getTargetObsComponentNode().getValue());
-            } else {
-                if ("".equals(scienceTarget.getName().trim())) {
-                    problems.addError(PREFIX+"EMPTY_TARGET_NAME_MSG", EMPTY_TARGET_NAME_MSG, elements.getTargetObsComponentNode().getValue());
-                }
+            final Asterism asterism = elements.getTargetObsComp().getValue().getAsterism();
+            if ("".equals(asterism.name().trim())) {
+                problems.addError(PREFIX+"EMPTY_TARGET_NAME_MSG", EMPTY_TARGET_NAME_MSG, elements.getTargetObsComponentNode().getValue());
```
swalker2m: I think the idea is to check that all science targets have a name.  This would only flag an error if none of them have a name?

bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/general/GeneralRule.java
```diff
@@ -347,25 +346,18 @@ private static boolean hasScienceObserves(final ConfigSequence sequence) {
...
-                        final double pm_dec = pm.deltaDec().velocity();
-                        final double total = pm_ra * pm_ra + pm_dec * pm_dec;
-
-                        if (total > MAX_PM * MAX_PM) { //to avoid sqrt call
-                            final P2Problems problems = new P2Problems();
-                            problems.addWarning(PREFIX + "TARGET_PM_RULE", MESSAGE, elements.getTargetObsComponentNode().getValue());
-                            return problems;
-                        }
-                    }
+            final scala.Option<ProperMotion> opm = elements.getTargetObsComp().getValue().getAsterism().basePositionProperMotion();
```
swalker2m: Haven't seen this definition yet, but I think perhaps the check should be applied to all targets in the asterism.  If any has a big PM then flag an error?

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/obsComp/TargetObsComp.java
swalker2m: @sraaphorst would you expand on this?  The idea is that by design a target component always has a target environment and a target environment always has an asterism (which in turn always has at least one start).

bundle/edu.gemini.spModel.io/src/main/java/edu/gemini/spModel/io/impl/PioSpXmlParser.java
```diff
@@ -231,6 +232,9 @@ private ISPRootNode _parseDocument(Document doc) throws Exception {
         // Update pre-2017A programs
         To2017A.updateProgram(doc);
 
+        // Update pre-2017B programs
+        To2018A.updateProgram(doc);
```
swalker2m: I guess you mean pre-2018A in the comment.  There is a 2017B migration too, but I guess that will be revealed when you merge.

bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/TargetGroupConfig.java
```diff
@@ -23,8 +21,8 @@
     public static final String TYPE_VALUE="targetgroup";
 
     public static TargetGroupConfig createBaseGroup(final TargetEnvironment env) {
-        final ImList<SPTarget> targets = env.getUserTargets().cons(env.getBase());
-        return new TargetGroupConfig(TccNames.BASE, targets, ImOption.apply(env.getBase()));
+        final ImList<SPTarget> targets = env.getUserTargets().cons(env.getArbitraryTargetFromAsterism());
+        return new TargetGroupConfig(TccNames.BASE, targets, ImOption.apply(env.getArbitraryTargetFromAsterism()));
```
swalker2m: Might be worth a TODO:ASTERISM label?  I'm not sure what needs to happen here but somehow the TCC is going to need to know about all the targets, I guess.

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
```diff
@@ -61,7 +61,7 @@ class LchQueryFunctor(queryType: LchQueryFunctor.QueryType,
 
     def makeTargetNode(target: SPTarget, env: TargetEnvironment): Option[Serializable] = {
       def targetType: Option[String] = {
-        if (env.isBasePosition(target)) Some("Base")
+        if (env.getArbitraryTargetFromAsterism eq target) Some("Base")
```
sraaphorst: Sure. The LTTS uses this information, it appears, but only for output purposes and to compare observation targets. It also uses coordinates, though, so this shouldn't be an issue, I wouldn't think.

I think we can label them all base, but I also don't see a good reason to not label them asterism to clarify the distinction. I think it boils down to a matter of preference over functionality.

bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/general/GeneralRule.java
sraaphorst: Yes, that's my impression. Shouldn't we make sure all targets IN the asterism have a name, and otherwise produce an error message?

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/obsComp/TargetObsComp.java
sraaphorst: This seems fine to me. I suspect my comment was on a previous commit (that I can't find now) about a small fiddly thing.

I can't recall what I meant by it now.

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
tpolecat: All uses of `getArbitraryTargetFromAsterism` will be crushed so we will have a chance to revisit this.

bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/TargetGroupConfig.java
tpolecat: `getArbitraryTargetFromAsterism` is equivalent to a `TODO` marker because all the uses need to be removed.

bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/general/GeneralRule.java
tpolecat: Yep I'll change.

bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/general/GeneralRule.java
tpolecat: Yep I'll change.

tpolecat:
Ok I addressed your issues @swalker2m. Merging with CI turns green.

swalker2m:
> Ok I addressed your issues @swalker2m. Merging with CI turns green.

👍

tpolecat:
Hell. I need to rebase.

tpolecat:
This was a kind of violent rebase because both branches introduced `To2017B` (mistakenly on my branch … I renamed it as `To2018A` but git isn't smart enough to figure that out). So I had to re-add `To2017B` from `develop` after merging. Anyway I will review the whole thing again to be sure the conversion code is intact.

PR #1224, opened 2017-05-04
REL-1808: GMOS ITC central wavelength check

leashma:
I modified gmos.java to throw an error if the user tries to submit a central wavelength that is out of the range for GS or GN. I closely followed the snytax for GNIRS for consistency, and compiled/successfully ran all of the ITC tests.


cquiroz:
There are some troubles on how this PR was made. After discussions with @andrewwstephens it will be resubmitted

PR #1225, opened 2017-05-04
REL-1808: GMOS ITC central wavelength check

leashma:
…I added a block to gmos.java to send an error if the central wavelength isn't between specified limits for GS and GN.

### DIFF
--- a/bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gmos/Gmos.java
+++ b/bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gmos/Gmos.java
@@ -310,7 +310,7 @@ public abstract class Gmos extends Instrument implements BinningProvider, Spectr
                         throw new RuntimeException("Central wavelength must be between 300 nm and 1000 nm.");
                     break;
                 default:
-                    throw new Error("invalid site");
+                    throw new RuntimeException("invalid site");
             }
 
         }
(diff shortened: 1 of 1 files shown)


cquiroz:
Thanks @leashma for you first PR. This looks good to me but travis is having a tantrum so we may need to wait a bit to merge it

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gmos/Gmos.java
```diff
@@ -296,6 +296,23 @@ private void validate() {
...
             }
 
+            // central wavelength, site dependent
+            double _centralWavelength = getCentralWavelength();
+            switch (gp.site()) {
+                // User-input central wavelength for GN
+                case GN:
+                    if (_centralWavelength < 360 || _centralWavelength > 1000)
+                        throw new RuntimeException("Central wavelength must be between 360 nm and 1000 nm.");
+                    break;
```
arturog8m: I'm curious why we are throwing a `RuntimeException` here - is it expected that the users need to know about this? If so, how do they know?

cquiroz: I'm not that happy about that either but it certainly is the style of ITC. You throw an exception and the outer layers deal with it

I'd say it is outside the scope of this PR though

arturog8m: Ah. Okay - I see now this is the way ITC deal with these things in other places. Good. Sorry for the noise. Travis is still failing though.

cquiroz: Yes, I requested @leashma to rebase on develop to fix the travis build

sraaphorst: Seems like the issue is that `sbt` isn't found. I would imagine re-running with @cquiroz's earlier fixes might solve the problem?

cquiroz:
@leashma did a rebase and the tests are passing

PR #1227, opened 2017-05-07
REL-2928: (Part A)

swalker2m:
This PR changes the time accounting model to split the time award across program and partner.  Before the model was essentially a map from partner to double (hours) and now it is a map from partner to `TimeAcctAward` which in turn contains a `Duration` for the program award and a `Duration` for the partner award.

The data migration simply converts the old hours amount to an explicit program award in milliseconds and a 0 ms partner award.  The program admin editor is updated to allow editing both values.  (Bryan was opposed to making this split too visible, but this is only available to staff members and there's no good way to allow editing total hours without offering an editor for both parts.)

### DIFF
--- a/bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/SpProgramFactory.scala
+++ b/bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/SpProgramFactory.scala
@@ -27,6 +27,8 @@ import Scalaz._
  */
 object SpProgramFactory {
 
+  private val MsPerHour = Duration.ofHours(1).toMillis
+
   private val NGO_TIME_ACCT = Map(
     AR -> TimeAcctCategory.AR,
     AU -> TimeAcctCategory.AU,
@@ -202,7 +204,7 @@ object SpProgramFactory {
           val jmap = ratios.map { case (cat, rat) =>
             // REL-2928 TODO: handle splitting program and partner award.  Here
             // we just keep everything as the program award for now.
-            val ms    = ((hrs * rat) * 3600000).round // ms for this category
+            val ms    = ((hrs * rat) * MsPerHour).round // ms for this category
             val award = new TimeAcctAward(Duration.ofMillis(ms), Duration.ZERO)
             (cat, award)
           }.toMap.asJava
--- a/bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/SpProgramFactory.scala
+++ b/bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/SpProgramFactory.scala
@@ -1,5 +1,6 @@
 package edu.gemini.phase2.skeleton.factory
 
+import edu.gemini.spModel.core.Affiliate
 import edu.gemini.spModel.gemini.obscomp.SPProgram.ProgramMode._
 import edu.gemini.spModel.too.TooType
 import edu.gemini.model.p1.immutable._
@@ -12,13 +13,14 @@ import edu.gemini.spModel.gemini.obscomp.SPProgram
 import edu.gemini.spModel.gemini.obscomp.SPProgram.{PIInfo, ProgramMode}
 import edu.gemini.shared.util.TimeValue
 
-import edu.gemini.spModel.timeacct.{TimeAcctAllocation, TimeAcctCategory}
+import edu.gemini.spModel.timeacct.{TimeAcctAllocation, TimeAcctAward, TimeAcctCategory}
 import edu.gemini.spModel.gemini.phase1.{GsaPhase1Data => Gsa}
+
+import java.time.Duration
 
 import scala.collection.JavaConverters._
 import scalaz._
 import Scalaz._
-import edu.gemini.spModel.core.Affiliate
 
 /**
  * Factory for creating an SPProgram from a Phase 1 Proposal document.
@@ -197,8 +199,14 @@ object SpProgramFactory {
       timeAccountingRatios(proposal) match {
         case Nil    => None
         case ratios =>
-          val jmap  = ratios.toMap.mapValues(d => new java.lang.Double(d)).asJava
-          Some(new TimeAcctAllocation(hrs, jmap))
+          val jmap = ratios.map { case (cat, rat) =>
+            // REL-2928 TODO: handle splitting program and partner award.  Here
+            // we just keep everything as the program award for now.
+            val ms    = ((hrs * rat) * 3600000).round // ms for this category
+            val award = new TimeAcctAward(Duration.ofMillis(ms), Duration.ZERO)
+            (cat, award)
+          }.toMap.asJava
+          Some(new TimeAcctAllocation(jmap))
       }
     }
 
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/obscomp/SPProgram.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/obscomp/SPProgram.java
@@ -968,13 +968,6 @@ public class SPProgram extends AbstractDataObject implements ISPStaffOnlyFieldPr
             }
         }
 
-//        TimeValue awardedTime = getAwardedTime();
-//        if (awardedTime != null) {
-//            Pio.addParam(factory, paramSet, AWARDED_TIME_PROP,
-//                    String.valueOf(awardedTime.getTimeAmount()),
-//                    awardedTime.getTimeUnits().name());
-//        }
-
         // Write the time accounting information.
         if (_timeAllocation != null) {
             ParamSet timeAcctPset;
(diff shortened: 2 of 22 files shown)


swalker2m:
A screen shot of the updated program admin dialog.

progadmin

swalker2m:
Remaining steps include:

1. Update the `SpProgramFactory` to add the correct partner time allocation on Phase 1 import.
2. Cron job task that updates the partner time allocation with executed partner time on a daily basis for old programs.  This will need to run until we no longer have pre-2017B programs in the queue.
3. Update remaining time calculation to subtract partner time.

Anything else?  I'm not sure what reports / labels need to be adjusted, if any.

cquiroz:
In general it seems fine to me. There is a test that didn't pass though

sraaphorst:
Looks like this is what didn't pass, which I suppose is unsurprising given the changes:

TimeValueTest.testNormalConstruction expected: 1.0 but was 0.001

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/timeacct/TimeAcctAllocation.java
```diff
@@ -165,21 +115,15 @@ public String toString() {
...
         if (this == o) return true;
         if (o == null || getClass() != o.getClass()) return false;
-
         final TimeAcctAllocation that = (TimeAcctAllocation) o;
-
-        if (Double.compare(that.totalTime, totalTime) != 0) return false;
-        if (!allocMap.equals(that.allocMap)) return false;
-
-        return true;
+        return Objects.equals(allocMap, that.allocMap);
```
sraaphorst: I did not know about the `Objects` static methods. Nice.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/timeacct/TimeAcctPio.java
```diff
@@ -15,49 +16,53 @@
...
-            if (hours == 0) continue;
-            
-            ParamSet ratioParamSet = factory.createParamSet(ALLOC_PARAM_SET);
-            Pio.addParam(factory, ratioParamSet, CATEGORY_PARAM, cat.name());
-            Pio.addDoubleParam(factory, ratioParamSet, HOURS_PARAM, hours);
-            pset.addParamSet(ratioParamSet);
+            final TimeAcctAward award = allocation.getAward(cat);
+            if (award.isZero()) continue;
+
+            final ParamSet allocParamSet = factory.createParamSet(ALLOC_PARAM_SET);
```
sraaphorst: Much clearer here. I'm not sure why this was called `ratioParamSet` in the old code, since it didn't store ratios, from what I can tell, but just the allocated hours per category.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/progadmin/TimeAcctUI.java
```diff
@@ -96,12 +128,27 @@ private JPanel getRatiosPanel() {
...
+        gbc.gridx     = 1;
+        pan.add(new JLabel("Program"), gbc);
+        gbc.gridx     = 2;
+        pan.add(new JLabel("Partner"), gbc);
+        gbc.gridx     = 5;
+        pan.add(new JLabel("Program"), gbc);
+        gbc.gridx     = 6;
+        pan.add(new JLabel("Partner"), gbc);
+
+        // Time Accountig Allocation
```
sraaphorst: Stupid note: just a typo here in `Accounting`.

sraaphorst:
Ahhh... I see what is breaking the test case now.

Previously:
`t2 = new TimeValue(1, TimeValue.Units.seconds);`
was invoking the old (unique) public constructor in `TimeValue` that takes a `double`, but now since the `long` constructor has been made public, it invokes this instead, which never does the conversion to milliseconds.

If we change this to:
`t2 = new TimeValue(1d, TimeValue.Units.seconds);`
to force the `double` constructor to be called, then everything passes.

We should probably be careful about this because there may be other places in the code base that use non-double values in `TimeValue` creations, which would now generate unexpected results.

sraaphorst:
The only other place I see this happening in a search is in `InstantiationFunctorTest`:

`def newParameters(targetName: String, cc: CloudCover): TemplateParameters =
    TemplateParameters.newInstance(newTarget(targetName), newSiteQuality(cc), new TimeValue(1, TimeValue.Units.hours))`

However, I think the `long` constructor is fundamentally flawed, given that previously it was inaccessible and everything went through the `double` constructor.

I can't see an easy way to avoid this completely apart from making one of the constructors a `static` creation method or something to that effect, e.g. `TimeValue.fromMillis(long milliseconds, Units timeUnits)` or something to that effect.

swalker2m:
> I can't see an easy way to avoid this completely apart from making one of the constructors a static creation method or something to that effect, 

Thanks yes, I think that's a good idea.

bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/SpProgramFactory.scala
```diff
@@ -197,8 +199,14 @@ object SpProgramFactory {
       timeAccountingRatios(proposal) match {
         case Nil    => None
         case ratios =>
-          val jmap  = ratios.toMap.mapValues(d => new java.lang.Double(d)).asJava
-          Some(new TimeAcctAllocation(hrs, jmap))
+          val jmap = ratios.map { case (cat, rat) =>
+            // REL-2928 TODO: handle splitting program and partner award.  Here
+            // we just keep everything as the program award for now.
+            val ms    = ((hrs * rat) * 3600000).round // ms for this category
```
tpolecat: While we're at it we could avoid the magic number with something like this.

```scala
Duration.ofMillis(Duration.ofHours(hrs).toMillis * ratio)
```

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/obscomp/SPProgram.java
```diff
@@ -1005,8 +1005,8 @@ public ParamSet getParamSet(PioFactory factory) {
 
             // Write the awarded time (for GSA only -- will be ignored on
             // import)
-            Pio.addParam(factory, paramSet, AWARDED_TIME_PROP,
-                    String.valueOf(_timeAllocation.getTotalTime()), "hours");
+            final double hrs = _timeAllocation.getSum().getTotalHours();
+            Pio.addParam(factory, paramSet, AWARDED_TIME_PROP, String.valueOf(hrs), "hours");
```
tpolecat: Sorry, super minor but can we change `String.valueOf` to `Double.toString`? This has a very Java 1 feel to it.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/timeacct/TimeAcctAward.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    public Duration getTotalAward() {
+        return programAward.plus(partnerAward);
+    }
+
+    // Convenience method to provide total award converted to hours.
+    public double getTotalHours() {
+        final long ms = getTotalAward().toMillis();
+        return ((double) ms) / 3600000.0;
```
tpolecat: `Duration.ofHours(1).toMillis` maybe?

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/timeacct/TimeAcctAward.java
```diff
@@ -0,0 +1,82 @@
...
+
+import java.io.Serializable;
+import java.time.Duration;
+import java.util.Objects;
+
+
+/**
+ * A tuple combining program and partner time awards.
+ */
+public final class TimeAcctAward implements Serializable {
```
tpolecat: Is there a good reason not to implement this in Scala?

bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/SpProgramFactory.scala
tpolecat: Actually nevermind, `hrs` is a double isn't it?

swalker2m: Right, that would round down the number of hours.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/timeacct/TimeAcctAward.java
swalker2m: Not really, I guess.  It's siblings `TimeAcctAllocation` and `TimeAcctCategory`, and most of the client code that uses it as well, were written pre-Scala so it just seemed natural to do it in Java as well.

tpolecat: Ok. That makes sense.

swalker2m:
After a brief meeting this morning, it was collectively decided to collect and store the program vs. partner time award split, but keep the actual math and values displayed in reports and labels the same as before (pure program award).  I've added a few commits to make that happen.

swalker2m:
I left the "Program Admin" editor showing both program and partner time as in the screenshot.  I'm not sure Bryan will want to keep it, but it seems like it is inevitable that we'll need to be able to display and edit the partner time award so I left it alone for now.  Only staff members have access to this editor.

PR #1228, opened 2017-05-09
REL-3031: Make F2 dark filter obsolete.

sraaphorst:
As per @andrewwstephens' note on REL-3031, the F2 dark filter is now obsolete.


sraaphorst:
Sorry... for some reason, the commit came from gemini-hlswbuild.
Argh. Will re-create and reopen.

PR #1229, opened 2017-05-09
REL-3031: Obsolete F2 Dark filter.

sraaphorst:
As per @andrewwstephens' comment, the F2 Dark filter is now obsolete.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/flamingos2/Flamingos2.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/flamingos2/Flamingos2.java
@@ -125,7 +125,7 @@ public final class Flamingos2 extends ParallacticAngleSupportInst
     /**
      * Filters
      */
-    public enum Filter implements DisplayableSpType, SequenceableSpType, LoggableSpType {
+    public enum Filter implements DisplayableSpType, SequenceableSpType, LoggableSpType, ObsoletableSpType {
 
         OPEN("Open", "Open", new Some<>(1.6)),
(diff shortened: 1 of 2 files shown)


swalker2m:
Does it ultimately use the `SpTypeComboBoxRenderer` which handles obsolete items specially?  It marks an obsolete selection with an asterisk when selected and only removes it from the list of options if it wasn't already selected.

sraaphorst:
Ooops. I only saw your comment, @swalker2m, after I merged.

I followed the same pattern here that was used for the Lyot wheel, which has obsolete items.

I'll go back and give it another look.

sraaphorst:
I just created a 2017A F2 program with a Dark filter and imported into this, and it looks to preserve the Dark filter, although it doesn't appear in the drop down combo box anymore. No asterisk, though. I'll look for the pattern you mentioned, as it sounds preferable.

swalker2m:
This is probably not a big deal but I guess if it isn't too bad it is better to be consistent with the other obsolete items, though it sounds like the lyot wheel is already done differently.

sraaphorst:
Yeah, the Lyot wheel is handled without the asterisk.

I think this would involve migrating from `ComboPropertyCtrl` to `JComboBox` from what I see in `EdCompInstGmos` and `GmosForm`, and since all the other combo boxes (including Lyot wheel) are using `ComboPropertyCtrl` in `Flamingos2Editor` (which clumps view and controller together instead of separating them out as in most of the other editors), I'm hesitant to make this change and introduce the inconsistency in UI classes.

If you think it's worthwhile, I'll go ahead.

swalker2m:
Okay that makes sense to me. I think we can probably just leave it as is then.

PR #1231, opened 2017-05-09
REL-3136: Update QPT icons.

sraaphorst:
See subject.


cquiroz:
I guess the `ico` comes in a single resolution right?

sraaphorst:
Yes, that's right. The `icns` file contains multiple resolutions, but the `ico` is just a single resolution.

Ain't Windows grand?

PR #1232, opened 2017-05-09
REL-2928-B: partner time cron job

swalker2m:
This PR would add a temporary cron job that sets the partner award equal to the executed partner time for the program for pre-2017B programs.  The idea is that this will keep old programs from seeing a difference in remaining time, even as new partner calibrations are added and executed.

This is not ready to merge as is (:-1:).  Only the last commit is different from the previous REL-2928 PR.  Assuming we go ahead with this approach, I'd merge that one first and then rebase this one so that only the last commit is included.

There are no test cases.  I did monkey testing but I'm not sure unit testing is going to be worth the significant effort that it will require.  On top of that, we still don't have 100% commitment to this approach.  I just wanted to put this out there because this task is time critical.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/obs/PartnerTimeAwardUtil.scala
+++ b/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/obs/PartnerTimeAwardUtil.scala
@@ -24,20 +24,24 @@ import scala.collection.JavaConverters._
   * pre-2017B program in the queue this code can be deleted.
   */
 object PartnerTimeAwardUtil {
-  private val Log = Logger.getLogger(PartnerTimeAwardUtil.getClass.getName)
 
   /** Sets the partner time award to the executed partner time amount in the
     * given program.
     */
-  def setPartnerAwardToExecuted(prog: ISPProgram): Unit = {
+  def setPartnerAwardToExecuted(prog: ISPProgram, log: Logger): Unit = {
 
     // Distributes amt over all the map keys as close to the given ratios as
     // possible, but ensuring that the sum of all the values is amt.
     def distribute[A](amt: Long, ratios: Map[A, Double]): Map[A, Long] = {
-      val proportions = ratios.map { case (a, r) =>
-        val f = amt * r
-        (a, f.floor.toLong, f - f.floor)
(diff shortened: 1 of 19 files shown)


bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/obs/PartnerTimeAwardUtil.scala
```diff
@@ -0,0 +1,97 @@
...
+
+  /** Sets the partner time award to the executed partner time amount in the
+    * given program.
+    */
+  def setPartnerAwardToExecuted(prog: ISPProgram): Unit = {
+
+    // Distributes amt over all the map keys as close to the given ratios as
+    // possible, but ensuring that the sum of all the values is amt.
+    def distribute[A](amt: Long, ratios: Map[A, Double]): Map[A, Long] = {
+      val proportions = ratios.map { case (a, r) =>
```
tpolecat: Can you add a comment and annotate the return type here?

bundle/edu.gemini.spdb.reports.collection/src/main/scala/edu/gemini/dbTools/timeAcct/PartnerTimeAwardCron.scala
```diff
@@ -0,0 +1,26 @@
...
+  */
+object PartnerTimeAwardCron {
+
+  /** Cron job entry point.  See edu.gemini.spdb.cron.osgi.Activator. */
+  def run(ctx: BundleContext)(tmpDir: File, logger: Logger, env: java.util.Map[String, String], user: java.util.Set[Principal]): Unit = {
+    val odbRef = ctx.getServiceReference(classOf[IDBDatabaseService])
+    val odb    = ctx.getService(odbRef)
+
+    logger.log(Level.INFO, "Start PartnerTimeAwardCron")
+    PartnerTimeAwardFunctor.query(odb, user)
```
tpolecat: Would it make sense to pass `logger` through to `PartnerTimeAwardUtil` so we can tie its messages back to the cron job?

swalker2m:
In the end we decided to go with a simpler approach so this won't be needed.

PR #1234, opened 2017-05-12
REL-2942: Change program support labels

sraaphorst:
The REL task for this PR is confusing, as the requirements have changed several times.
In the end, the description should be ignored, and only the later comments read.

Essentially, this primarily affects the program details (`ProgramForm`), and generalizes the concept of the former `NGO Contact Email` label to `Principal Contact Email`, and changes the `NGO Check?` label in the table to `Check?` to follow suit.

I tried to make this change as transparent as possible by relabeling variables and methods to reflect the change. This will, however, introduce discrepancies into the OCS code base in a few places, such as the ODB Browser, where the principal contact email will still be exported as the NGO contact email due to the XML format, which we do not wish to change as the scope would be too large and affect too many existing things, such as GSA information.

Several discussions were had and several iterations of the requirements proposed, and it was decided that living with the few discrepancies / inconsistencies was acceptable and the preferred solution.

Since this task has been stretched across quite some time, it resulted in me investigating a good chunk of the code, and I performed some minor cleanup along the way, getting rid of old commented out code and cleaning up some warnings as well.

### DIFF
--- a/bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
+++ b/bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
@@ -145,7 +145,7 @@ class LchQueryFunctor(queryType: LchQueryFunctor.QueryType,
         setSemester(prog.scienceSemester.orNull)
         setTitle(spProg.getTitle)
         setContactScientistEmail(spProg.getContactPerson)
-        setNgoEmail(spProg.getNGOContactEmail)
+        setNgoEmail(spProg.getPrimaryContactEmail)
         setNotifyPi(spProg.getNotifyPi.displayValue)
         setRollover(prog.rolloverStatus.displayValue)
 
--- a/bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/auxfile/AttachmentTableModel.java
+++ b/bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/auxfile/AttachmentTableModel.java
@@ -11,28 +11,28 @@ import javax.swing.table.AbstractTableModel;
 public final class AttachmentTableModel extends AbstractTableModel {
     public enum Col {
         NAME("Name", String.class) {
-            @Override public String getValue(AuxFile file) {
+            @Override public String getValue(final AuxFile file) {
                 return file.getName();
             }
         },
         SIZE("Size", Integer.class) {
-            @Override public Object getValue(AuxFile file) {
+            @Override public Object getValue(final AuxFile file) {
                 return file.getSize();
             }
         },
         LAST_MOD("Last Modified (UTC)", Date.class) {
-            @Override public Object getValue(AuxFile file) {
-                long l = file.getLastModified();
+            @Override public Object getValue(final AuxFile file) {
+                final long l = file.getLastModified();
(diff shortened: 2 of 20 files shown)


cquiroz:
It seems your commits have multiple authors, perhaps geminihlsw?

cquiroz:
I don't see anything obviously wrong with this though I don't really understand the context of the change

sraaphorst:
=sigh=
That is going to be truly annoying to fix. Unsure of how to address it.
I deleted the `geminihlsw-build` account since it was unused (I created it for the build machine but never used it) to avoid this problem coming up again.

sraaphorst:
Actually, I think that ended up easier than expected. I added the email address associated with that account to my GitHub account and I think it fixed things and now associates the commits with me.

sraaphorst:
(Still lists them as coming from me multiple times, under different email addresses, I guess, due to `sraaphor` versus `sraaphorst`, but at least they are all associated with me.)

cquiroz:
Maybe you need to force push to get the new addresses in place

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
```diff
@@ -145,7 +145,7 @@ class LchQueryFunctor(queryType: LchQueryFunctor.QueryType,
         setSemester(prog.scienceSemester.orNull)
         setTitle(spProg.getTitle)
         setContactScientistEmail(spProg.getContactPerson)
-        setNgoEmail(spProg.getNGOContactEmail)
+        setNgoEmail(spProg.getPrimaryContactEmail)
```
sraaphorst: Here is where you can see the biggest inconsistency:
The OT will use Primary Contact Email, whereas the ODB Browser will continue to report this as NGO Contact Email.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/auxfile/AttachmentTableModel.java
```diff
@@ -66,24 +66,16 @@ public int getColumnCount() {
...
         return Col.values()[col].displayValue();
     }
 
-    public Class getColumnClass(int col) {
+    public Class getColumnClass(final int col) {
         return Col.values()[col].getColumnClass();
     }
-
-    @Override public boolean isCellEditable(int rowIndex, int columnIndex) {
-        return false; // columnIndex == Col.CHECKED.ordinal();
```
sraaphorst: Always returning `false` is the default implementation, so there was no need to override now that the old code was commented out.

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
swalker2m: It also seems confusing that we now have a "contact person" and a "primary contact email".  Both are email addresses.  It seems like it should be primary contact email and secondary contact email, or maybe primary contact email and gemini contact email.

sraaphorst:
I agree with your assessment, @swalker2m. Originally, the requirements were to have a "primary contact email" and "secondary contact email", but the requirements changed to have this instead.

I'm going to merge this for today's test release, but I'm going to copy your comment to the JIRA task to see what science thinks.

PR #1235, opened 2017-05-12
Asterism Updates part B

tpolecat:
This is a follow-on to #1222 that updates the TPE to handle asterism. It does the following:
- The TPE base position is now defined as the base position of the Asterism. It is updated when any target in the asterism moves.
- The base position feature now draws all targets, and has been renamed the asterism feature.
- The base position itself is not drawn. A new feature is needed for this.

Further work here probably needs to wait until it's possible to construct and manipulate multi-target asterisms in the target list editor. There will certainly be usability issues to deal with. But this PR gets us to a sensible starting point.

Note that only commits after 1bbc418 (May 10 and later) are relevant.


bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/TargetEnvironment.java
```diff
@@ -36,12 +36,23 @@
...
     public static TargetEnvironment create(SPTarget base) {
-        ImList<SPTarget> user = ImCollections.emptyList();
-        return new TargetEnvironment(new Asterism.Single(base), GuideEnvironment$.MODULE$.Initial(), user);
+        return create(new Asterism.Single(base));
     }
 
-    /**
+  /**
+   * Creates a TargetEnvironment with a single-target asterism but no other
+   * associated stars.
```
swalker2m: I suppose it isn't necessarily a single-target asterism.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/tpe/TpeContext.scala
```diff
@@ -74,7 +74,7 @@ final class TargetContext(obs: Option[ISPObservation]) extends TpeSubContext[ISP
 
   def envOrNull: TargetEnvironment = env.orNull
 
-  def envOrDefault: TargetEnvironment = env.getOrElse(TargetEnvironment.create(new SPTarget))
+  def envOrDefault: TargetEnvironment = env.getOrElse(TargetEnvironment.create(new SPTarget));
```
swalker2m: `;`?

tpolecat:
Will re-open on another branch.

PR #1238, opened 2017-05-16
REL-1787: Fixed arithmetic overflow in converting from hh:mm:ss to Long in ms.

sraaphorst:
In `TimingWindowImporter`, in converting from an `hh` and `mm` (and optional `ss`) specified as `Int` to a `Long` in ms, I did not convert the `hh`, `mm`, and `ss` to `Long`s first, which resulted in 32-bit integer arithmetic overflow for large values of `hh` (e.g. 720 as stated in @andrewwstephens's comment in the REL task). 

Now I convert the time elements first to `Long`s so that the arithmetic is performed in 64 bits, and the computation acts as expected instead of resulting in a negative time value.

I also added a test case for this.

In addition, as per Andy's recommendation, I changed the icon for timing window imports for clarity.

### DIFF
--- a/bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/sitequality/SiteQualityPanel.java
+++ b/bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/sitequality/SiteQualityPanel.java
@@ -179,7 +179,7 @@ final class SiteQualityPanel extends JPanel {
                     ButtonFlattener.flatten(this);
                 }});
 
-                add(new JButton(Resources.getIcon("eclipse/download.gif")) {{
+                add(new JButton(Resources.getIcon("eclipse/openbrwsr.gif")) {{
                     setToolTipText("Import timing windows from text file");
                     setFocusable(false);
                     addActionListener(e -> {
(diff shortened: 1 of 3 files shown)


cquiroz:
How can you overflow in hours? I guess is not hours of a day

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/sitequality/SiteQualityPanel.java
```diff
@@ -179,7 +179,7 @@
                     ButtonFlattener.flatten(this);
                 }});
 
-                add(new JButton(Resources.getIcon("eclipse/download.gif")) {{
+                add(new JButton(Resources.getIcon("eclipse/openbrwsr.gif")) {{
```
cquiroz: Is this related to the change?

sraaphorst: No, it's just a number of hours and not hours in a day, so, for example, in what Andy showed, he was using 720:00, and the calculation became:

`1000 * (60 * (60 * hh + mm) + ss) = 2592000000`, which is greater than `2^31`, so we were overflowing.

sraaphorst: No, it's just a number of hours and not hours in a day, so, for example, in what Andy showed, he was using 720:00, and the calculation became:

`1000 * (60 * (60 * hh + mm) + ss) = 2592000000`, which is greater than `2^31`, so we were overflowing.

PR #1242, opened 2017-05-17
REL-2928: Populate TimeAcctAward properly from times in proposal observations.

sraaphorst:
This PR simply modifies the `SpProgramFactory` class used in importing P1 XMLs accepted by ITAC to properly populate the program and partner time fields in the new `TimeAcctAward` class.

We will not be using the new P1 calculated partner time in the OT as of yet, but we will store this information for a future release as per @swalker2m's previous PR for this REL task.

### DIFF
--- a/bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/SpProgramFactory.scala
+++ b/bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/SpProgramFactory.scala
@@ -198,19 +198,17 @@ object SpProgramFactory {
 
   def timeAcctAllocation(proposal: Proposal): Option[TimeAcctAllocation] =
     awardedHours(proposal).filter(_ > 0.0) flatMap { hrs =>
-      timeAccountingRatios(proposal) match {
-        case Nil    => None
-        case ratios =>
-          val progTime = hoursFromObservations(proposal, _.progTime)
-          val partTime = hoursFromObservations(proposal, _.partTime)
-          require((hrs - (progTime |+| partTime).hours).abs < 1e-1,
-            "Awarded time must be equal to the sum of program time and partner calibration time")
+      val progTime = hoursFromObservations(proposal, _.progTime)
+      val partTime = hoursFromObservations(proposal, _.partTime)
 
+      timeAccountingRatios(proposal) match {
+        case Nil => None
+        case ratios =>
           val jmap = ratios.map { case (cat, rat) =>
-            def duration(v: TimeAmount): Duration =
+            def durationRatio(v: TimeAmount): Duration =
               Duration.ofMillis(((v.hours * rat) * MsPerHour).round)
 
-            val award = new TimeAcctAward(duration(progTime), duration(partTime))
+            val award = new TimeAcctAward(durationRatio(progTime), durationRatio(partTime))
             (cat, award)
           }.toMap.asJava
           Some(new TimeAcctAllocation(jmap))
--- a/bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/SpProgramFactory.scala
+++ b/bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/SpProgramFactory.scala
@@ -201,11 +201,16 @@ object SpProgramFactory {
       timeAccountingRatios(proposal) match {
         case Nil    => None
         case ratios =>
+          val progTime = hoursFromObservations(proposal, _.progTime)
+          val partTime = hoursFromObservations(proposal, _.partTime)
+          require((hrs - (progTime |+| partTime).hours).abs < 1e-1,
+            "Awarded time must be equal to the sum of program time and partner calibration time")
+
           val jmap = ratios.map { case (cat, rat) =>
-            // REL-2928 TODO: handle splitting program and partner award.  Here
-            // we just keep everything as the program award for now.
-            val ms    = ((hrs * rat) * MsPerHour).round // ms for this category
-            val award = new TimeAcctAward(Duration.ofMillis(ms), Duration.ZERO)
+            def duration(v: TimeAmount): Duration =
+              Duration.ofMillis(((v.hours * rat) * MsPerHour).round)
+
+            val award = new TimeAcctAward(duration(progTime), duration(partTime))
             (cat, award)
           }.toMap.asJava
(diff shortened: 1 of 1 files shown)


bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/SpProgramFactory.scala
```diff
@@ -215,6 +220,9 @@ object SpProgramFactory {
   def awardedHours(proposal: Proposal): Option[Double] =
     itacAcceptance(proposal) map { a => a.award.toHours.value }
 
+  private def hoursFromObservations(proposal: Proposal, sf: Observation => Option[TimeAmount]): TimeAmount =
+    proposal.observations.foldLeft(TimeAmount.empty)(_ |+| sf(_).getOrElse(TimeAmount.empty))
```
sraaphorst: There may be a nicer way to do this. I was first doing it with `map` and `flatten` to get rid of `None` in the `List[Option[TimeAmount]]`, but the compiler did not like `flatMap`.

bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/SpProgramFactory.scala
```diff
@@ -201,11 +201,16 @@ object SpProgramFactory {
       timeAccountingRatios(proposal) match {
         case Nil    => None
         case ratios =>
+          val progTime = hoursFromObservations(proposal, _.progTime)
+          val partTime = hoursFromObservations(proposal, _.partTime)
+          require((hrs - (progTime |+| partTime).hours).abs < 1e-1,
+            "Awarded time must be equal to the sum of program time and partner calibration time")
```
cquiroz: this feels a bit risky. I guess the PIT enforces this right? and in principle  ITAC doesn't change it but still it could be changed by hand

bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/SpProgramFactory.scala
```diff
@@ -201,11 +201,16 @@ object SpProgramFactory {
...
+
           val jmap = ratios.map { case (cat, rat) =>
-            // REL-2928 TODO: handle splitting program and partner award.  Here
-            // we just keep everything as the program award for now.
-            val ms    = ((hrs * rat) * MsPerHour).round // ms for this category
-            val award = new TimeAcctAward(Duration.ofMillis(ms), Duration.ZERO)
+            def duration(v: TimeAmount): Duration =
+              Duration.ofMillis(((v.hours * rat) * MsPerHour).round)
```
cquiroz: Can you do the conversions with the `java.time` package?

bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/SpProgramFactory.scala
sraaphorst: Agree that this feels a bit risky. This should be enforced somehow, and if it currently fails, the ODB will log an exception and the skeleton script will return it.

My concern is accumulated floating point arithmetic errors across many observations. I'm not sure if this will ever happen.

We should somehow, though, ensure that the ITAC awarded time is equal to the P1 required time here.

sraaphorst: (And yes, the PIT enforces this.)

bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/SpProgramFactory.scala
sraaphorst: I'm not sure of a way offhand, as the `java.time.Duration.ofXXX` always requires a `long` and we are working with `double` values in `TimeAmount`.

cquiroz: Ok, this is probably fine

bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/SpProgramFactory.scala
cquiroz: Ok, this is at the end of a servlet call right? What would you get if requires fails? A 500 or a 400?

cquiroz: OTOH ITAC should enforce this too

cquiroz: But you can't really trust ITAC to do so. We are not really checking this. An extra problem is that ITAC stores this in the DB whose floating point representation doesn't necessarily match Java's

swalker2m: We have the ratios across partners based on the partner awarded times.  We can sum the program and partner times in the observations and figure out the ratio of program vs. partner.  Then we could apply those two sets of ratios to the sum of the awards to come up with values that by definition match the actual awards?  Is this the concern we're trying to address?  It should work out the same but if for whatever reason the awards don't match the sum of observation times we wouldn't be assigning more or less than the actual awards.

cquiroz: I'm ok with validating, I'm just slightly worried that `require` is a bit heavy handed

sraaphorst: After discussing with @swalker2m, we have decided to remove this `require` check.

In the past, we never checked to ensure that the ITAC awarded time was equal to the sum of the observation times, so there isn't really a good reason to change that behaviour now.

PR #1246, opened 2017-05-17
REL-566: ignore acquisition and cal for ToO triggers

swalker2m:
This PR would change ToO triggers to filter out anything but science observations.

All the work was in creating a simple test case.  Turns out aligning the starts to make a ToO observation trigger is tricky.

### DIFF
--- a/bundle/edu.gemini.too.event/src/main/scala/edu/gemini/too/event/service/TooCondition.scala
+++ b/bundle/edu.gemini.too.event/src/main/scala/edu/gemini/too/event/service/TooCondition.scala
@@ -2,9 +2,10 @@ package edu.gemini.too.event.service
 
 import edu.gemini.pot.spdb.IDBTriggerCondition
 import edu.gemini.pot.sp.{ISPObservation, SPUtil, SPCompositeChange}
-import edu.gemini.spModel.obs.{ObservationStatus, SPObservation}
+import edu.gemini.spModel.obs.{ObsClassService, ObservationStatus, SPObservation}
 import edu.gemini.spModel.obs.ObsPhase2Status.ON_HOLD
 import edu.gemini.spModel.obs.ObservationStatus.READY
+import edu.gemini.spModel.obsclass.ObsClass.SCIENCE
 import edu.gemini.spModel.too.Too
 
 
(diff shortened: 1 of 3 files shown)


swalker2m:
I'm quite open to the opinion that it is too late to do this for 17B.  I could just repurpose this PR to the development branch.

bundle/edu.gemini.too.event/src/test/scala/edu/gemini/too/event/service/TooServiceTest.scala
```diff
@@ -0,0 +1,85 @@
...
+
+  private def setOnHold(): Unit =
+    setPhase2Status(ObsPhase2Status.ON_HOLD)
+
+  private def setReady(): Unit =
+    setPhase2Status(ObsPhase2Status.PHASE_2_COMPLETE)
+
+  def testValidTooTrigger(): Unit = {
+    setOnHold()
+    Thread.sleep(500)
```
swalker2m: Could be problematic.  The triggering is done by adding a task to a thread pool so I don't really know when it will actually happen.

bundle/edu.gemini.too.event/src/main/scala/edu/gemini/too/event/service/TooCondition.scala
```diff
@@ -13,26 +14,36 @@ import edu.gemini.spModel.too.Too
...
+ * (REL-566: skip acquisition and calibration observations as well.)
+ *
  * This condition is registered with the database such that when it occurs, the
  * [[edu.gemini.too.event.service.TooService]] is executed to record the event.
  */
 object TooCondition extends IDBTriggerCondition {
-  private def isDataObjectUpdate(change: SPCompositeChange): Boolean =
-    change.getPropertyName == SPUtil.getDataObjectPropertyName
+  def triggeredObservation(change: SPCompositeChange): Option[ISPObservation] = {
+    def castIf[T:Manifest](o: Object): Option[T] =
```
tpolecat: Would you mind adding a note that this technique is unsound in general, and only works for non-parameterized class types? I would almost rather see `Option(a) collect { case a: Whatever => a }` in the two cases where this is called, just to discourage people from thinking it's ok.

swalker2m: Why don't we just cast it without checking?  The API lets you store whatever you want as a data object of any program node, but if we do that everything will break long before getting to ToO triggers anyway.

tpolecat: I guess if we got a `None` there nothing would happen, silently, which might be less helpful than an exception.

bundle/edu.gemini.too.event/src/test/scala/edu/gemini/too/event/service/TooServiceTest.scala
tpolecat: You could use a `Promise` here and have the test thingy complete it, then await the resulting `Future` with a long timeout. This would avoid sleeping longer than necessary.

project/OcsBundle.scala
```diff
@@ -425,7 +425,7 @@ trait OcsBundle {
   lazy val bundle_edu_gemini_too_event = 
     project.in(file("bundle/edu.gemini.too.event")).dependsOn(
       bundle_edu_gemini_shared_skyobject,
-      bundle_edu_gemini_pot,
+      bundle_edu_gemini_pot % "test->test;compile->compile",
```
sraaphorst: What does this do, out of curiosity?

bundle/edu.gemini.too.event/src/test/scala/edu/gemini/too/event/service/TooServiceTest.scala
swalker2m: Most of the waiting, as in this case, is making sure that nothing happens -- that there is no ToO event.  I'm not sure how to either speed that up or guarantee that I've waited long enough without changing the ToO API and implementation.

project/OcsBundle.scala
swalker2m: I wanted to use a class in the test code of the `pot` bundle.  This bizarre looking appendage makes the `pot` test classes available to the tests in the `too_event` bundle.  I'm sure there's a better, more complete, explanation but having obtained the result I wanted I didn't look deeper.

swalker2m:
Turns out `TooCondition` isn't even used, in practice.  So this is all wrong anyway.  I'll straighten it out in a new PR.

PR #1247, opened 2017-05-18
REL-2942: Program form contact email label changes.

sraaphorst:
Science has requested (take three or four) that we change the labels in the `ProgramForm` for contact emails to reflect the way that the process now works as opposed to how it used to work with rigidly defined contacts.

I have updated the labels and the names of the widgets accordingly.

Unless someone comments on this PR insisting that I do so, at this point, I am not going to update the `SPProgram` accessors and member variables because this far into the process, it seems risky.

I made notes at the top of `ProgramForm` and `EdProgram` to indicate that there will be an inconsistency in labels / `SPProgram` data member names. This also translates to the ODB Browser, but that was - a week or two ago - agreed to be acceptable.

### DIFF
--- a/bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/EdProgram.java
+++ b/bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/EdProgram.java
@@ -31,6 +31,7 @@ import java.util.logging.Logger;
 
 /**
  * This is the editor for the Science Program component.
+ * NOTE: program labels are no longer compatible with data objects because of REL-2942.
  */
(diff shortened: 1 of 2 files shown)


arturog8m:
Is this the same as #1248 ?

sraaphorst:
More Travis issues that have nothing to do with this PR (which passes all tests on my machine):
unresolved dependency: org.scala-sbt#collections;0.13.11: not found
download failed: org.scala-sbt#logging;0.13.11!logging.jar
download failed: org.scala-sbt#relation;0.13.11!relation.jar
Error during sbt execution: Error retrieving required libraries
  (see /home/travis/.sbt/boot/update.log for complete log)
Error: Could not retrieve sbt 0.13.11

tpolecat:
There was a hiccup at bintray, where the sbt stuff lives. Seems to be working now.

PR #1252, opened 2017-05-22
WIP: remove asterism predicates

tpolecat:
This removes `isSidereal`, etc., from `Asterism` and as a side-effect removes some other cruft. As a side-effect it removes special handling of non-sidereal targets in QV.

After merging this we need to add a new JIRA issue for QV, to warn the user about non-sidereal targets that don't have defined targets at the reference date.


tpolecat:
This totally broke QV but I don't know how yet.

tpolecat:
Got it working again, but the filters need to be parameterized on time which means basically everything in QV has to be parameterized. Tried 4 different approaches and all failed. Will resume tomorrow.

tpolecat:
In the end I started over 13 times but finally got it working. When you change the reference date in QV it re-juggles all the filters and graphs and stuff correctly now.

tpolecat:
I'm going to go over this first. I'll let you know when it's ready for review.

tpolecat:
I regret to inform you that this is ready for review.

bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/QvContext.scala
```diff
@@ -45,11 +46,11 @@ case class QvContext(peer: Peer, dataSource: DataSource, source: ObservationProv
...
   // ===== Data providers / filters used in this context ====
-  private val _nonSidProvider = new PositionProvider(this, source)
-  private val _mainFilter = FilterProvider(_nonSidProvider)
-  private val _tableFilter = FilterProvider(_mainFilter)
-  private val _selection = SelectionProvider(_tableFilter)
-  
+  private lazy val _nonSidProvider = new PositionProvider(this, source)
+  private lazy val _mainFilter = FilterProvider(this, _nonSidProvider)
+  private lazy val _tableFilter = FilterProvider(this, _mainFilter)
+  private lazy val _selection = SelectionProvider(this, _tableFilter)
```
swalker2m: I've shied away from `lazy` values because of the complicated semantics and because usually it isn't clear that you're gaining that much, if anything.  Are these things expensive and usually not needed, or do you think we should just embrace `lazy`?

bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/filter/core/Filter.scala
```diff
@@ -213,10 +214,10 @@ object Filter {
     def label = "RA"
     def lowest = RA.MinValue
     def highest = RA.MaxValue
-    def getter = {o: Obs => o.getRa/15.0 }
+    def getter = (o: Obs, ctx: QvContext) => o.raDeg(ctx) / 15.0
```
swalker2m: This is the main change, I think.  I couldn't find the actual definition of `raDeg` but I'm guessing it does what I expect it does.

bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/QvContext.scala
sraaphorst: @swalker2m, can you explain your aversion to `lazy val` as a curiosity? I'm not sure why you would want to avoid it.

swalker2m: I admit that I don't have a well founded argument against it but I think [this SIP entry](http://docs.scala-lang.org/sips/pending/improved-lazy-val-initialization.html) does a good job of indicating the complexity and potential for deadlock.  Locking and checking the initialization state is involved as well with every access to a `lazy val`, so if it is heavily used then it isn't clear you've gained anything.  Unless there's something that is really expensive to initialize and yet not always needed, or impossible to initialize on construction it just seems like a premature optimization.

sraaphorst: Thanks for that clarification. I appreciate the link.

tpolecat: `lazy` is necessary here because the constructors actually do things with the `QvContext` which is not yet initialized (since we're passing `this`). It NPEs if we don't do this.

bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/filter/core/Filter.scala
tpolecat: Yes, well spotted. This is what that provoked all the other QV changes. The method `.raDeg` is available by an implicit conversion in the `edu.gemini.qv.plugin` package object [here](https://github.com/tpolecat/ocs/blob/4b5cff17929aab7255599e5d095cb1570554a974/bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/package.scala#L14-L31).

tpolecat:
rebased just to be sure

PR #1253, opened 2017-05-25
REL-2928: Minimum time now adjusted to only be partner time.

sraaphorst:
As per our discussion on Tuesday, minimum time is now scaled to only contain program time.

### DIFF
--- a/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Proposal.scala
+++ b/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Proposal.scala
@@ -116,11 +116,15 @@ case class Proposal(meta:Meta,
   def resetObservationMeta:Proposal = copy(observations = observations.map(_.copy(meta = None)))
 
   // Calculate the program time as a ratio of total time.
+  private def timeSum(extract: Observation => Option[TimeAmount]): TimeAmount =
+    TimeAmount.sum(observations.map(extract).flatten)
+
+  def programTimeHours: TimeAmount = timeSum(_.progTime)
+  def partnerTimeHours: TimeAmount = timeSum(_.partTime)
+
   def programTimeRatio: Double = {
-    def timeSum(extract: Observation => Option[TimeAmount]): TimeAmount =
-      TimeAmount.sum(observations.map(extract).flatten)
-    val progTime  = timeSum(_.progTime)
-    val partTime  = timeSum(_.partTime)
+    val progTime  = programTimeHours
+    val partTime  = partnerTimeHours
     val totalTime = progTime |+| partTime
     totalTime.isEmpty ? 1.0 | (progTime.hours / totalTime.hours)
   }
--- a/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Proposal.scala
+++ b/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Proposal.scala
@@ -115,6 +115,16 @@ case class Proposal(meta:Meta,
 
   def resetObservationMeta:Proposal = copy(observations = observations.map(_.copy(meta = None)))
 
+  // Calculate the program time as a ratio of total time.
+  def programTimeRatio: Double = {
+    def timeSum(extract: Observation => Option[TimeAmount]): TimeAmount =
+      TimeAmount.sum(observations.map(extract).flatten)
+    val progTime  = timeSum(_.progTime)
+    val partTime  = timeSum(_.partTime)
+    val totalTime = progTime |+| partTime
+    totalTime.isEmpty ? 1.0 | (progTime.hours / totalTime.hours)
+  }
+
   private def this(m:M.Proposal) = this(
     Meta(m.getMeta),
     Semester(m.getSemester),
--- a/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Proposal.scala
+++ b/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Proposal.scala
@@ -119,10 +119,10 @@ case class Proposal(meta:Meta,
   private def timeSum(extract: Observation => Option[TimeAmount]): TimeAmount =
     TimeAmount.sum(observations.map(extract).flatten)
 
-  lazy val programTime: TimeAmount = timeSum(_.progTime)
-  lazy val partnerTime: TimeAmount = timeSum(_.partTime)
+  val programTime: TimeAmount = timeSum(_.progTime)
+  val partnerTime: TimeAmount = timeSum(_.partTime)
 
-  lazy val programTimeRatio: Double = {
+  val programTimeRatio: Double = {
     val totalTime = programTime |+| partnerTime
     totalTime.isEmpty ? 1.0 | (programTime.hours / totalTime.hours)
   }
--- a/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Proposal.scala
+++ b/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Proposal.scala
@@ -119,14 +119,12 @@ case class Proposal(meta:Meta,
   private def timeSum(extract: Observation => Option[TimeAmount]): TimeAmount =
     TimeAmount.sum(observations.map(extract).flatten)
 
-  def programTimeHours: TimeAmount = timeSum(_.progTime)
-  def partnerTimeHours: TimeAmount = timeSum(_.partTime)
+  lazy val programTime: TimeAmount = timeSum(_.progTime)
+  lazy val partnerTime: TimeAmount = timeSum(_.partTime)
 
-  def programTimeRatio: Double = {
-    val progTime  = programTimeHours
-    val partTime  = partnerTimeHours
-    val totalTime = progTime |+| partTime
-    totalTime.isEmpty ? 1.0 | (progTime.hours / totalTime.hours)
+  lazy val programTimeRatio: Double = {
+    val totalTime = programTime |+| partnerTime
+    totalTime.isEmpty ? 1.0 | (programTime.hours / totalTime.hours)
   }
 
   private def this(m:M.Proposal) = this(
--- a/bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/SpProgramFactory.scala
+++ b/bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/SpProgramFactory.scala
@@ -70,13 +70,13 @@ object SpProgramFactory {
     prog.setThesis(isThesis(proposal))
 
     prog.setPIInfo(piInfo(proposal))
-    hostNgoEmail(proposal) foreach prog.setPrimaryContactEmail
+    hostNgoEmail(proposal).foreach(prog.setPrimaryContactEmail)
 
     // Note: not a typo -- "contact person" is an email
-    gemEmail(proposal) foreach { e => prog.setContactPerson(e) }
+    gemEmail(proposal).foreach(prog.setContactPerson)
 
-    minBand3Time(proposal) foreach prog.setMinimumTime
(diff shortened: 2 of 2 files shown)


bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Proposal.scala
```diff
@@ -115,6 +115,16 @@ case class Proposal(meta:Meta,
 
   def resetObservationMeta:Proposal = copy(observations = observations.map(_.copy(meta = None)))
 
+  // Calculate the program time as a ratio of total time.
+  def programTimeRatio: Double = {
+    def timeSum(extract: Observation => Option[TimeAmount]): TimeAmount =
+      TimeAmount.sum(observations.map(extract).flatten)
```
sraaphorst: Note that IDEA tells me I should use `flatMap` here, but as `Option` is not `GenTraversableOnce`, I can't.

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Proposal.scala
```diff
@@ -115,6 +115,16 @@ case class Proposal(meta:Meta,
...
+    val progTime  = timeSum(_.progTime)
+    val partTime  = timeSum(_.partTime)
+    val totalTime = progTime |+| partTime
+    totalTime.isEmpty ? 1.0 | (progTime.hours / totalTime.hours)
```
cquiroz: Why would `totalTime` ever be empty?

cquiroz: And if it is empty why would it default to 1?

cquiroz: Right, it is 1 because it is a ratio

bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/SpProgramFactory.scala
```diff
@@ -70,13 +70,13 @@ object SpProgramFactory {
...
-    hostNgoEmail(proposal) foreach { e => prog.setPrimaryContactEmail(e) }
+    hostNgoEmail(proposal) foreach prog.setPrimaryContactEmail
 
     // Note: not a typo -- "contact person" is an email
     gemEmail(proposal) foreach { e => prog.setContactPerson(e) }
 
-    minBand3Time(proposal) foreach { tv => prog.setMinimumTime(tv) }
-    timeAcctAllocation(proposal) foreach { alloc => prog.setTimeAcctAllocation(alloc) }
+    minBand3Time(proposal) foreach prog.setMinimumTime
+    timeAcctAllocation(proposal) foreach prog.setTimeAcctAllocation
```
cquiroz: Were we going to have a rule that `minTime` couldn't be bigger than `totalTime`. is that enforced?

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Proposal.scala
sraaphorst: It could be empty in theory if no time was assigned to the observations. That being said, this method should never be called if that happens, but I wanted to eliminate any risk of dividing by zero.

bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/SpProgramFactory.scala
sraaphorst: That should always be the case, but I don't think we discussed enforcing it; at least, it was never enforced before.

If we want to enforce it, we could do so by throwing an exception so that the skeleton script fails and reports the failure, I guess? This would also log an exception in the ODB, I think.

cquiroz: Right, I'm not sure if we should enforce it, but my preference if we do is to do validations early when the program is parsed so we can return a proper HTTP error code but maybe that would be too complicated

sraaphorst: My hesitation here is that I don't see any validation done in `SpProgramFactory` as it stands, so I'm not sure this is the place to introduce it, especially so close to release.

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Proposal.scala
sraaphorst: I guess we could just propagate `NaN` through if we really wanted to, but that seems like the wrong approach to me.

bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/SpProgramFactory.scala
cquiroz: I guess we can live with the exception, @swalker2m what do you think?

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Proposal.scala
cquiroz: I agree, I guess it is not possible in practice to have an empty totalTime so the fallback seems correct

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Proposal.scala
```diff
@@ -115,6 +115,20 @@ case class Proposal(meta:Meta,
 
   def resetObservationMeta:Proposal = copy(observations = observations.map(_.copy(meta = None)))
 
+  // Calculate the program time as a ratio of total time.
+  private def timeSum(extract: Observation => Option[TimeAmount]): TimeAmount =
+    TimeAmount.sum(observations.map(extract).flatten)
+
+  def programTimeHours: TimeAmount = timeSum(_.progTime)
+  def partnerTimeHours: TimeAmount = timeSum(_.partTime)
```
swalker2m: Is there something mutable here that makes these be `def`s and not `val`s or do we just not want to store them?  If they were `val`s you wouldn't need temporary variables in the `programTimeRatio` method, which could be a `val` too come to think of it.

sraaphorst: I was actually just thinking about that myself. This class should be fully immutable, so there's no reason why we can't make them all `lazy val`s.

bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/SpProgramFactory.scala
```diff
@@ -70,13 +70,13 @@ object SpProgramFactory {
     prog.setThesis(isThesis(proposal))
 
     prog.setPIInfo(piInfo(proposal))
-    hostNgoEmail(proposal) foreach { e => prog.setPrimaryContactEmail(e) }
+    hostNgoEmail(proposal) foreach prog.setPrimaryContactEmail
```
swalker2m: On a trivial note, we've never really documented Scala standards and we're a bit all over the place, but for some reason I'm inclined to prefer `hostNgoEmail(proposal).foreach(prog.setPrimaryContactEmail)` in this case.  Seems to be the ["official" recommendation](http://docs.scala-lang.org/style/method-invocation) too.

👍 for the eta-reduction, btw!

sraaphorst: I'll make this change and the changes to `Proposal` from `def` to `lazy val` and push another commit momentarily. Thanks for the tips!

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Proposal.scala
swalker2m: I think I'd just go with the simpler `val`, if you're just trying to optimize construction.

sraaphorst:
Alright, done! I also noticed some other places in `SpProgramFactory` that weren't following the style recommendation Shane pointed out and some others where eta-reduction could be done, so I cleaned up the class to be consistent.

bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/SpProgramFactory.scala
swalker2m: I think for now we should just scale the minimum time and store it, since before we were just storing it without any validation?   In theory this validation should be done upstream but if we want to check the proposal again before ingesting it we might want to do so with a bit more time to think it all through.

sraaphorst: @swalker2m Agree. 👍 
If we are going to start validating here, we should be thorough and not just validate one aspect.

swalker2m:
In the end then, we decided to keep using the sum of observation times rather than scaling the award by the program and parter ratios?

sraaphorst:
@swalker2m Yes, that's right. We determined that it should always be the case that awarded time = sum over observations (progTime + partTime), so we could maintain things as they were.

sraaphorst:
Sure. I'd like to get this merged for today's test release, even though it probably won't be tested until we can coordinate with ITAC. I'll hold off on merging it with develop until we get confirmation from Rodrigo, though.

So if nobody has any objections, I'll go ahead and merge and then see if it's a good time to start the test release?

bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/SpProgramFactory.scala
```diff
@@ -179,47 +179,42 @@ object SpProgramFactory {
...
+    awardedHours(proposal).filter(_ > 0.0) flatMap { _ =>
       timeAccountingRatios(proposal) match {
         case Nil => None
         case ratios =>
           val jmap = ratios.map { case (cat, rat) =>
             def durationRatio(v: TimeAmount): Duration =
               Duration.ofMillis(((v.hours * rat) * MsPerHour).round)
 
-            val award = new TimeAcctAward(durationRatio(progTime), durationRatio(partTime))
+            val award = new TimeAcctAward(durationRatio(proposal.programTime), durationRatio(proposal.partnerTime))
```
swalker2m: > Yes, that's right. We determined that it should always be the case that awarded time = sum over observations (progTime + partTime), so we could maintain things as they were.

I'm probably being overly pessimistic but I'd feel better if the skeleton creation and the itac email were based on the same total award value taken from the same place in the model.  Then there's zero chance that we come up with different answers.  Something like:

```scala
awardedHours(proposal).filter(_ > 0.0).flatMap { hrs =>
...
  case ratios =>
    val jmap = ratios.map { case (cat, partnerShare) =>
      def duration(rat: Double): Duration =
        Duration.ofMillis((hrs.hours * rat * partnerShare) * MsPerHour).round)

      val award = new TimeAcctAward(
                    duration(proposal.programTimeRatio),
                    duration(1.0 - proposal.programTimeRatio))
      (cat, award)
    }.toMap.asJava
...
}
```

sraaphorst: Ooops. I saw this after I merged. I'm fine either way, as the values should be the same, but I'll change this.

PR #1254, opened 2017-05-25
REL-2928: Changes as per PR comments from previous PR to use awarded time.

sraaphorst:
We now use awarded time instead of summing across observations.


swalker2m:
Seems good. 👍 Maybe @cquiroz should take a quick look before we close the book on it though.

cquiroz:
close the book? i like your optimism

PR #1255, opened 2017-05-25
REL-2784 Make date formatting order consistent

jdnavarro:
This commit focuses on dates which don't follow the year, month and day order.
Where changed the format followed is yyyy-MMM-dd. Dates with different
format (like slashes as separator instead of dashes) but with the right order
get a pass for now.

Please watch out for any date format change that would break something that's not validated with a test. I tried not touch anything that looked like it was part of a path or filename though.

Not all dates are consistent for now because I couldn't find a large enough set of changes that could pass the tests. It was unfeasible to run the whole test suite for every date format change but from this commit I think I can check the tests for every date format change. I hope to at least fix the ones with different separators.

### DIFF
--- a/bundle/edu.gemini.obslog/src/main/java/edu/gemini/obslog/obslog/DBNightlyPlanInfo.java
+++ b/bundle/edu.gemini.obslog/src/main/java/edu/gemini/obslog/obslog/DBNightlyPlanInfo.java
@@ -1,8 +1,8 @@
 package edu.gemini.obslog.obslog;
 
 import java.io.Serializable;
-import java.text.SimpleDateFormat;
-import java.util.Date;
+import java.time.Instant;
+import java.time.format.DateTimeFormatter;
 
 //
 // Gemini Observatory/AURA
@@ -22,7 +22,7 @@ public class DBNightlyPlanInfo implements Serializable, Comparable<DBNightlyPlan
     // last modified timestamp
     private long _timestamp;
 
-    private static SimpleDateFormat _formatter = new SimpleDateFormat("yyyy-MMM-dd HH:hh:ss");
+    private static DateTimeFormatter dateFormat = DateTimeFormatter.ofPattern("yyyy-MMM-dd HH:hh:ss");
 
     private String _lastModified;
 
@@ -37,7 +37,7 @@ public class DBNightlyPlanInfo implements Serializable, Comparable<DBNightlyPlan
         _title = title;
         _planID = planID;
         _timestamp = timestamp;
-        _lastModified = _formatter.format(new Date(timestamp));
+        _lastModified = dateFormat.format(Instant.ofEpochMilli(_timestamp));
     }
 
     public String getTitle() {
--- a/bundle/edu.gemini.obslog/src/main/java/edu/gemini/obslog/obslog/DBNightlyPlanInfo.java
+++ b/bundle/edu.gemini.obslog/src/main/java/edu/gemini/obslog/obslog/DBNightlyPlanInfo.java
@@ -22,7 +22,7 @@ public class DBNightlyPlanInfo implements Serializable, Comparable<DBNightlyPlan
     // last modified timestamp
     private long _timestamp;
 
-    private static SimpleDateFormat _formatter = new SimpleDateFormat("MM/dd/yy HH:hh:ss");
+    private static SimpleDateFormat _formatter = new SimpleDateFormat("yyyy-MMM-dd HH:hh:ss");
 
     private String _lastModified;
 
--- a/bundle/edu.gemini.spdb.reports.collection/src/main/java/edu/gemini/dbTools/odbState/ObservationState.java
+++ b/bundle/edu.gemini.spdb.reports.collection/src/main/java/edu/gemini/dbTools/odbState/ObservationState.java
@@ -15,10 +15,10 @@ import org.dom4j.DocumentFactory;
 import org.dom4j.Element;
 
 import java.io.Serializable;
-import java.text.DateFormat;
-import java.text.SimpleDateFormat;
 import java.util.Calendar;
 import java.util.GregorianCalendar;
+import java.text.DateFormat;
+import java.text.SimpleDateFormat;
 
 public final class ObservationState implements Serializable {
     private static final long serialVersionUID = 2;
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/auxfile/AuxFileEditor.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/auxfile/AuxFileEditor.scala
@@ -4,8 +4,9 @@ import edu.gemini.auxfile.api.AuxFile
 import edu.gemini.spModel.core.SPProgramID
 import jsky.app.ot.gemini.editor.ProgramForm
 import jsky.app.ot.vcs.VcsOtClient
-import java.text.SimpleDateFormat
-import java.util.{Collections, Date, TimeZone}
+import java.time.{Instant, ZoneId}
+import java.time.format.DateTimeFormatter
+import java.util.{Collections, Date}
 
 import jsky.util.gui.Resources
 
@@ -27,13 +28,11 @@ object AuxFileEditor {
       lab.text = size.toString
   })
 
-  private val Format = new SimpleDateFormat("yyyy-MMM-dd HH:mm:ss") {
-    setTimeZone(TimeZone.getTimeZone("UTC"))
-  }
+  private val DateFormat = DateTimeFormatter.ofPattern("yyyy-MMM-dd HH:mm:ss").withZone(ZoneId.of("UTC"))
 
   private val DateCellRenderer = new LabelCellRenderer[Date]((lab, date) => {
     lab.horizontalAlignment = Alignment.Center
-    lab.text = Format.format(date)
+    lab.text = DateFormat.format(date.toInstant)
   })
 
   private val CheckIcon = Resources.getIcon("eclipse/check.gif")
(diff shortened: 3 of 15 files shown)


bundle/edu.gemini.obslog/src/main/java/edu/gemini/obslog/obslog/DBNightlyPlanInfo.java
```diff
@@ -22,7 +22,7 @@
     // last modified timestamp
     private long _timestamp;
 
-    private static SimpleDateFormat _formatter = new SimpleDateFormat("MM/dd/yy HH:hh:ss");
+    private static SimpleDateFormat _formatter = new SimpleDateFormat("yyyy-MMM-dd HH:hh:ss");
```
swalker2m: I know it isn't part of this PR, but a `static SimpleDateFormat` is always a danger sign.  The issue is that it isn't thread-safe.  I think we might as well switch these to the newer thread-safe `java.time.format.DateTimeFormatter` while we're at it.

cquiroz: That's correct, I'll see to change it

cquiroz: I agree we should try to use the new API where possible

bundle/jsky.util/src/main/java/jsky/util/DateUtil.java
```diff
@@ -17,12 +17,12 @@
 public class DateUtil {
 
     /**
-     * Return a string with the UTC time in the format /mm/dd/yy hh:mm:ss,
+     * Return a string with the UTC time in the format yyyy-MMM-dd hh:mm:ss,
      * given the current UTC time in ms.
      */
     public static String formatUTC(long time) {
```
cquiroz: Did you check how many places use this function? we maybe changing something important upstream

cquiroz:
i'm a bit surprised this didn't affect test. So you avoided any changes that would me a test fail?

bundle/edu.gemini.spdb.reports.collection/src/main/java/edu/gemini/dbTools/odbState/ObservationState.java
```diff
@@ -178,7 +178,7 @@ private static String _getNight(final long time) {
         final StringBuilder buf = new StringBuilder();
         buf.append(startDay).append("/").append(endDay).append(" ");
 
-        final DateFormat f = new SimpleDateFormat("MMM yyyy");
+        final DateFormat f = new SimpleDateFormat("yyyy-MMM-dd");
```
swalker2m: This formatted date doesn't seem to escape an XML file into which it is written.  I don't know if anything parses that `String` somewhere else.  I think maybe we could just leave this one.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/auxfile/AuxFileEditor.scala
```diff
@@ -27,7 +27,7 @@ object AuxFileEditor {
       lab.text = size.toString
   })
 
-  private val Format = new SimpleDateFormat("MM/dd/yy HH:mm:ss") {
+  private val Format = new SimpleDateFormat("yyyy-MMM-dd HH:mm:ss") {
```
swalker2m: Did you check these dialog boxes, etc after the format change.  I think this one goes in a table so it would be worth checking that the formatted dates aren't too big for the column.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/tpe/feat/TpeEphemerisFeature.scala
```diff
@@ -188,7 +188,7 @@ object TpeEphemerisFeature {
   }
 
   val formatDate: Long => String = {
-    val df = new SimpleDateFormat("dd-MMM HH:mm")
+    val df = new SimpleDateFormat("yyyy-MMM-dd HH:mm")
```
swalker2m: This one writes on the TPE display on top of everything else so it might be better to keep it short, even if technically inconsistent?  That is maybe just switch the `dd-MMM` as you've done but leave off the year?

jdnavarro:
Sorry guys for not replying earlier, for some reason I'm not getting the emails from this thread.

bundle/edu.gemini.obslog/src/main/java/edu/gemini/obslog/obslog/DBNightlyPlanInfo.java
jdnavarro: Good to know, I didn't know about it. I'll take care of it.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/auxfile/AuxFileEditor.scala
jdnavarro: Not sure how to get to that dialog. Let me try getting there on my own first and if can't I'll ask you.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/tpe/feat/TpeEphemerisFeature.scala
jdnavarro: I'll change it to `MMM-dd` then.

jdnavarro:
> i'm a bit surprised this didn't affect test. So you avoided any changes that would me a test fail?

Actually I first changed all the dates that didn't look like a filename or URL path but even after trying, I couldn't get several tests to pass. I started taking out dates by different criteria (one date by one would be infeasible) and this last change was the first one I found where all the tests passed.

bundle/jsky.util/src/main/java/jsky/util/DateUtil.java
jdnavarro: No usages found with IDEA in ocs. There is also another formatting method below that didn't change but also couldn't find any usages. ¿Do you think those methods are used by ocs API consumers? ¿Can we remove them?

jdnavarro: When grepping for formatUTC I could find one usage of `formatUTC` in [`EdProgramHelper`](https://github.com/gemini-hlsw/ocs/blob/aa6c1b9131fb59aad82c067ffff2b906af523d38/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/EdProgramHelper.scala#L39-L40).

Do you think that will be problematic?

swalker2m: > Do you think that will be problematic?

That should be fine.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/auxfile/AuxFileEditor.scala
swalker2m: I probably mislead you talking about "dialog boxes".  It's just the "File Attachment" tab on the top-level program node.  Those table columns are evenly distributed based on available space.  We can always explicitly shrink the "Size" and "NGO Check?" column if necessary.

jdnavarro: The columns in the "File Attachment" tab look fine:

file_attachment

However it doesn't look good in the "Sync History" tab:

sync_history

Do you think it's worth trying to increase the column width for *start* and *end*? I have no idea how difficult or easy that would be.

cquiroz: It would be nice to increase the width but Column's widths is not that trivial in Swing, There is a utility trait https://github.com/gemini-hlsw/ocs/blob/develop/bundle/jsky.app.ot/src/main/scala/edu/gemini/catalog/ui/TableColumnsAdjuster.scala but i don't know how applicable is for your particular case

swalker2m: It looks like it should be straightforward.  You can add a bit to the "Start" and "End" columns [here](https://github.com/gemini-hlsw/ocs/blob/develop/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/EdProgramHelper.scala#L25-L29) and take away a corresponding amount from "Keys Used".  I think it is worth doing.

jdnavarro: It was unexpectedly easy 🍰 

sync_history

jdnavarro:
This is now ready for a second review.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/auxfile/AuxFileEditor.scala
cquiroz: unexpectedly easy and Swing? Sounds suspicious :smile:

jdnavarro:
> I guess I expected this to steal space from the "Keys Used" column but maybe it doesn't matter. "Keys Used" might just grab whatever is left-over anyway.

Hmmm, where is it taking the space from? I'm resizing manually the "Start" and "End" columns but I can only see the "Keys Used" column shrinking...

swalker2m:
> Hmmm, where is it taking the space from? I'm resizing manually the "Start" and "End" columns but I can only see the "Keys Used" column shrinking...

I think this is fine.  The last column is set, I believe, to grow or shrink to consume whatever space is left.

PR #1258, opened 2017-05-26
REL-2928: Use only enabled observations of the appropriate band in time scaling

sraaphorst:
Beforehand, in order to calculate the ratio of program time to total time (for scaling minimum time and awarded time during the skeleton import), we simply summed over all the observations in a proposal.

@cquiroz and @swalker2m pointed out that we should only be doing so for enabled observations in the appropriate band. I have extracted the common shared code out of `Phase1FolderFactory` into the package module that will be used by both, and since this is no longer really a property of a `Proposal` that we should be considering in the P1 model, I moved code from the `Proposal` class that I added here as well.

### DIFF
--- a/bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/Phase1FolderFactory.scala
+++ b/bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/Phase1FolderFactory.scala
@@ -89,23 +89,13 @@ object Phase1FolderFactory {
   }
 
 
-  // If there is an itac acceptance, then use its band assignment.  Otherwise
-  // just figure we will use the "normal" band 1/2 observations.
-  private def band(proposal: Proposal): Band =
-    (for {
-      itac   <- proposal.proposalClass.itac
-      accept <- itac.decision.right.toOption
-    } yield accept.band).getOrElse(1) match {
-      case 3 => Band.BAND_3
-      case _ => Band.BAND_1_2
-    }
-
   def create(site: core.Site, proposal: Proposal): Either[String, Phase1Folder] = {
     val empty: Either[String, Folder] = Right(Folder.empty(site))
 
-    val b       = band(proposal)
+    val b       = extractBand(proposal)
     val time    = proposal.semester.midPoint
-    val efolder = (empty/:proposal.observations.filter(obs => obs.band == b && obs.enabled)) { (e, obs) =>
+    val obs     = enabledObs(proposal)
+    val efolder = (empty/:obs) { (e, obs) =>
       e.right flatMap { _.add(obs, time) }
     }
 
--- a/bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/Phase1FolderFactory.scala
+++ b/bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/Phase1FolderFactory.scala
@@ -92,7 +92,6 @@ object Phase1FolderFactory {
   def create(site: core.Site, proposal: Proposal): Either[String, Phase1Folder] = {
     val empty: Either[String, Folder] = Right(Folder.empty(site))
 
-    val b       = extractBand(proposal)
     val time    = proposal.semester.midPoint
     val obs     = enabledObs(proposal)
     val efolder = (empty/:obs) { (e, obs) =>
--- a/bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/Phase1FolderFactory.scala
+++ b/bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/Phase1FolderFactory.scala
@@ -93,12 +93,9 @@ object Phase1FolderFactory {
     val empty: Either[String, Folder] = Right(Folder.empty(site))
 
     val time    = proposal.semester.midPoint
-    val obs     = enabledObs(proposal)
-    val efolder = (empty/:obs) { (e, obs) =>
-      e.right flatMap { _.add(obs, time) }
-    }
-
-    efolder.right map { _.toPhase1Folder }
+    val obs     = proposal.usableObservations
+    val efolder = obs.foldLeft(empty){ (e, obs) => e.right.flatMap( _.add(obs, time)) }
+    efolder.right.map(_.toPhase1Folder)
   }
 }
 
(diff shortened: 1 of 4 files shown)


bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/package.scala
```diff
@@ -1,4 +1,38 @@
...
+import edu.gemini.model.p1.immutable.{Band, Observation, Proposal, TimeAmount}
+
+import scalaz._
+import Scalaz._
+
 package object factory {
+  // Filter to only the enabled observations by appropriate band.
+  def enabledObs(proposal: Proposal): List[Observation] = {
+    val band = extractBand(proposal)
+    proposal.observations.filter(obs => obs.band == band && obs.enabled)
```
cquiroz: We should double check that what's exported from ITAC as enabled corresponds here

bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/package.scala
```diff
@@ -1,4 +1,38 @@
...
+
+  // If there is an itac acceptance, then use its band assignment.  Otherwise
+  // just figure we will use the "normal" band 1/2 observations.
+  def extractBand(proposal: Proposal): Band =
+    (for {
+      itac   <- proposal.proposalClass.itac
+      accept <- itac.decision.right.toOption
+    } yield accept.band).getOrElse(1) match {
+      case 3 => Band.BAND_3
+      case _ => Band.BAND_1_2
```
cquiroz: After some discussions this seems correct. The "band" in itac is actually from 1 to 7 😄

sraaphorst: Glad to hear it 😸, since I extracted this directly from `Phase1FolderFactory`, assuming that that code was doing the correct thing.

bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/package.scala
sraaphorst: This is also from `Phase1FolderFactory`, which is why I assumed the correspondence was correct, but yes, we should confirm that.

bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/Phase1FolderFactory.scala
```diff
@@ -89,23 +89,13 @@ object Phase1FolderFactory {
...
-    }
-
   def create(site: core.Site, proposal: Proposal): Either[String, Phase1Folder] = {
     val empty: Either[String, Folder] = Right(Folder.empty(site))
 
-    val b       = band(proposal)
+    val b       = extractBand(proposal)
     val time    = proposal.semester.midPoint
-    val efolder = (empty/:proposal.observations.filter(obs => obs.band == b && obs.enabled)) { (e, obs) =>
+    val obs     = enabledObs(proposal)
```
swalker2m: We don't use `b` from line 95 since it is handled in `enabledObs`.

sraaphorst:
I just realized that I could clean this up a bit more, too, since the band was used only in `Phase1FolderFactory` for filtering observations. Since I moved this code to the package object, I didn't need to have a band value in `Phase1FolderFactory` and could embed the band code in the observation filtering code in the package object.

bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/package.scala
```diff
@@ -1,4 +1,38 @@
 package edu.gemini.phase2.skeleton
 
+import edu.gemini.model.p1.immutable.{Band, Observation, Proposal, TimeAmount}
+
+import scalaz._
+import Scalaz._
+
 package object factory {
+  // Filter to only the enabled observations by appropriate band.
+  def enabledObs(proposal: Proposal): List[Observation] = {
```
swalker2m: Since "enabled" is a thing in `Observation` and since this method does more than filter on `enabled`, perhaps we should name this something else like `usableObservations` or something?

bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/package.scala
```diff
@@ -1,4 +1,38 @@
 package edu.gemini.phase2.skeleton
 
+import edu.gemini.model.p1.immutable.{Band, Observation, Proposal, TimeAmount}
+
+import scalaz._
+import Scalaz._
+
 package object factory {
```
swalker2m: All of the methods you add here take `Proposal` as a single argument.  You could also bedazzle them into existence with an `implicit class` that takes a `Proposal` in its constructor.  No big deal but it gives you nicer syntax I think.

bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/Phase1FolderFactory.scala
```diff
@@ -89,23 +89,13 @@ object Phase1FolderFactory {
...
+    val efolder = (empty/:obs) { (e, obs) =>
```
tpolecat: This syntax is probably going away in 2.13 so we should probably prefer `.foldLeft` and `.foldRight`.

sraaphorst: I should have realized that, since @swalker2m had just sent me the link yesterday to the style guide that said exactly that these operators should be avoided. I'll change that.

sraaphorst: I should have realized that, since @swalker2m had just sent me the link yesterday to the style guide that said exactly that these operators should be avoided. I'll change that.

sraaphorst:
Agreed with @swalker2m that the pimped (bedazzled?) `Proposal` class is nicer, so I went ahead and made that change.

Also agree with @tpolecat's comment, which I should have caught (ha) since @swalker2m pointed me to the page on the scala style guide yesterday that says exactly this!

bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/Phase1FolderFactory.scala
cquiroz: I'm glad, I never liked the operator syntax 🕺

swalker2m:
Looks good to me. 👍 For what it's worth, I always preferred `/:`and `:\' too but I guess I'll have to switch.

PR #1259, opened 2017-05-31
REL-3153: Obsolete GMOS North R150_G5306

swalker2m:
Makes R150_G5306 obsolete to make way for R150_G5308, which should be selected when the PI chooses an R150 filter in the phase 1 proposal.

### DIFF
--- a/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/package.scala
+++ b/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/package.scala
@@ -108,7 +108,7 @@ package object immutable {
     val B600  = M.GmosNDisperser.B600_G5307
     val R600  = M.GmosNDisperser.R600_G5304
     val R400  = M.GmosNDisperser.R400_G5305
-    val R150  = M.GmosNDisperser.R150_G5306
+    val R150  = M.GmosNDisperser.R150_G5308
   }
 
   type GmosNFilter = M.GmosNFilter
@@ -256,7 +256,7 @@ package object immutable {
     val HStar   = M.GpiObservingMode.H_STAR
     val HLiwa   = M.GpiObservingMode.H_LIWA
     val HDirect = M.GpiObservingMode.DIRECT_H_BAND
-    
+
     def isCoronographMode(mode: GpiObservingMode):Boolean = mode.value.startsWith("Coronograph")
     def isDirectMode(mode: GpiObservingMode):Boolean = mode.value.endsWith("direct")
     def scienceBand(mode: GpiObservingMode):Option[String] = {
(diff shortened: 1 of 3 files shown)


cquiroz:
Is there some conversion code for older proposals from 5306 to 5308

sraaphorst:
This looks right to me, but as @cquiroz mentioned, do we want to do any conversion for old proposals in `UpConverter`?

swalker2m:
In the Phase 1 XML, the disperser is the same either way:

```
                <disperser>R150</disperser>
```

I thought this was just about picking the correct Phase 2 disperser to match `R150`. There are two options, the now obsolete `R150_G5306` and `R150_G5308`.  The update to the `.xjb` file was done so that it would find the correct one instead of the obsolete version.  It's been a while since I touched this code though so I may be missing something, but I don't think there is any up conversion to be done in this case?

swalker2m:
Here is where the lookup is done (in the skeleton servlet):

```scala
  private def spEnum[E1 <: Enum[E1], E2 <: Enum[E2]](e1: E1, c2: Class[E2]): Either[String, E2] = {
    val e2Opt = c2.getEnumConstants find { _.name() == e1.name() }
    e2Opt.toRight("Could not find matching enum value for %s.%s".format(e1.getClass.getName, e1.name()))
  }
```
and it is called like this:

```scala
    private def disperser(p1: GmosNDisperser): Either[String, GmosNorthType.DisperserNorth] =
     spEnum(p1, classOf[GmosNorthType.DisperserNorth])
```

and the Phase 1 `GmosNDisperser` constants are controlled by the `.xjb` file:

```
        <!-- GmosNDisperser => GmosNorthType.DisperserNorth -->
        <jxb:bindings node="./xsd:simpleType[@name='GmosNDisperser']/xsd:restriction">
            ...
            <jxb:bindings node="./xsd:enumeration[@value='R150']">
                <jxb:typesafeEnumMember name="R150_G5308"/>
            </jxb:bindings>
        </jxb:bindings>
```

which is used in generating the Java-based mutable Phase 1 model.

So while the underlying enum name will differ in the Phase 1 model behind the latest PIT, we pass around the proposal XML.  Perhaps the Phase 1 model version should be bumped up though?

cquiroz:
@swalker2m I see the change is only on the `.xjb`. then you'r right it doesn't really affect the phase1 model

swalker2m:
@cquiroz do you think that the p1 model version should be updated?  I think that implies updating the `pitVersion` in the top-level build file?  Presumably the last digit, "serial compatibility", would be moved up.  So, it would be

```
pitVersion in ThisBuild := OcsVersion("2017B", false, 2, 1, 1)  // was 2, 1, 0
```

Would that cause any issues at skeleton import time?  I think not since the `schemaVersion` remains the same?

cquiroz:
In this case it seems we don't need to update the p1 model, it is only on the mapping to p2, sorry for misleading you

PR #1263, opened 2017-06-01
REL-3162: SPDB build file changes.

sraaphorst:
@swalker2m was attempting to connect to the ODB with VisualVM but found the startup scripts were missing several defines and parameters: this fixes the SPDB `build.sbt` to now add these to the startup scripts.

### DIFF
--- a/app/spdb/build.sbt
+++ b/app/spdb/build.sbt
@@ -51,6 +51,7 @@ def common(version: Version) = AppConfig(
     "-Dcom.sun.management.jmxremote.port=2407",
     "-Dcom.sun.management.jmxremote.ssl=false",
     "-Djava.awt.headless=true",
+    "-Djava.net.preferIPv4Stack=true",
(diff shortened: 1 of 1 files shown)


app/spdb/build.sbt
```diff
@@ -512,6 +516,7 @@ def gnagsodb(version: Version) = AppConfig(
     "-Dedu.gemini.site=north",
     "-Dcron.archive.edu.gemini.dbTools.html.ftpHost=gsagn.hi.gemini.edu",
     "-Dcron.odbMail.SITE_SMTP_SERVER=smtp.hi.gemini.edu",
+    "-Djava.rmi.server.hostname=gnauxodb",
```
swalker2m: Does `gnauxodb` still exist?

sraaphorst: Yes... this is still the empty ODB we use for PIT AGS estimates. We haven't yet simplified this down, but will aim to do so for the PIT 2018A release.

PR #1265, opened 2017-06-01
update spModel core libs

tpolecat:
This updates the `spModel.core` dependencies to versions that have 2.12 artifacts available, which we need in order to move compile it to 2.12 for Gem. There were a handful of compile errors that are fixed here. After this change you can say

    sbt ++2.12.2 bundle_edu_gemini_spModel_core/packageBin

to create a jarfile for core compiled with 2.12

### DIFF
--- a/bundle/edu.gemini.spModel.core/build.sbt
+++ b/bundle/edu.gemini.spModel.core/build.sbt
@@ -11,7 +11,7 @@ libraryDependencies ++= Seq(
   "org.scala-lang.modules" %% "scala-parser-combinators" % "1.0.4",
   "org.scalaz" %% "scalaz-core" % ScalaZVersion,
   "org.scalaz" %% "scalaz-effect" % ScalaZVersion,
-  "com.squants"  %% "squants"  % "0.6.2"
+  "com.squants"  %% "squants"  % "0.6.3"
   )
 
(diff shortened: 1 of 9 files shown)


cquiroz:
Nice, I wonder if we could use this technique to move ocs3 to 2.12 as well. We depend on many more packages though

tpolecat:
@cquiroz yeah it may be harder for you. Need to figure out a way to do it though.

tpolecat:
Any reason not to approve this?

cquiroz:
Just forgot

bundle/edu.gemini.spModel.core/build.sbt
```diff
@@ -11,7 +11,7 @@ libraryDependencies ++= Seq(
   "org.scala-lang.modules" %% "scala-parser-combinators" % "1.0.4",
   "org.scalaz" %% "scalaz-core" % ScalaZVersion,
   "org.scalaz" %% "scalaz-effect" % ScalaZVersion,
-  "com.squants"  %% "squants"  % "0.6.2"
+  "com.squants"  %% "squants"  % "0.6.3"
```
swalker2m: Didn't we want to get to 1.3.0?

PR #1266, opened 2017-06-02
quiet down the build

tpolecat:
This sets up the Travis build to run `sbt -q` with a custom logging config that only reports errors. (I verified that you do in fact see compile/test failures). I turned most of the `println`s that were in test output (I couldn't find all of them) into logger messages, so you still see them in a normal build and when running, but not in the CI log.

The CI log for this build should demonstrate silence. We'll see.

I also turned on caching which should speed things up a tiny bit. We'll see.

### DIFF
--- a/logging.properties
+++ b/logging.properties
@@ -1,0 +1,59 @@
+############################################################
+#  	Default Logging Configuration File
+#
+# You can use a different file by specifying a filename
+# with the java.util.logging.config.file system property.
+# For example java -Djava.util.logging.config.file=myfile
+############################################################
+
+############################################################
+#  	Global properties
+############################################################
+
+# "handlers" specifies a comma separated list of log Handler
+# classes.  These handlers will be installed during VM startup.
+# Note that these classes must be on the system classpath.
+# By default we only configure a ConsoleHandler, which will only
+# show messages at the INFO and above levels.
+handlers= java.util.logging.ConsoleHandler
+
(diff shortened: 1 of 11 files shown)


logging.properties
```diff
@@ -0,0 +1,59 @@
...
+# To also add the FileHandler, use the following line instead.
+#handlers= java.util.logging.FileHandler, java.util.logging.ConsoleHandler
+
+# Default global logging level.
+# This specifies which kinds of events are logged across
+# all loggers.  For any given facility this global level
+# can be overriden by a facility specific level
+# Note that the ConsoleHandler also has a separate level
+# setting to limit messages printed to the console.
+.level= SEVERE
```
tpolecat: This is the only thing that is changed from the default config.

tpolecat:
Oh wait now I know why we didn't do this before … it was timing out because there was no output. Probably need to add some hacking.

tpolecat:
This seems to work. See the build transcript.

cquiroz:
I guess it would blurp something if there is an actual error, right?

cquiroz:
I wonder if we should include in travis building the specific modules supporting 2.12

tpolecat:
> I guess it would blurp something if there is an actual error, right?

yes

> I wonder if we should include in travis building the specific modules supporting 2.12

good idea, i will add that

PR #1267, opened 2017-06-02
REL-1787: Changing icon for timing window import again.

sraaphorst:
Subject line says it all.


swalker2m:
From the REL task:

> Maybe use "import targets" icon from the PIT?

Is that what this is?

sraaphorst:
Yes, that is exactly what this is.

Hopefully with regards to this icon, this third time will be the charm 😸

PR #1268, opened 2017-06-02
REL-1787: Changing icon for timing window import again.

sraaphorst:
This tracks the icon change to the develop branch.


sraaphorst:
Strange. I think I've come across this before, though:

[info] CoordinatesSpec
[info] 
[info] Coordinates Offsetting should
[info]   + be consistent with diff  + flip the RA when Dec overflowsCoordinates Diff should
[info]   + have identity (1)  + have identity (2)Coordinates Serialization should
[info]   + Support Java BinaryCoordinates Angular Separation should
[info]   + be in [0, 180]  + be 0 for the same Coordinates  + be consistent with RA offsetting at the equator  + be consistent with Dec offsetting (includes polar discontinuity)Coordinates Interpolation should
[error]   x be consistent with fractional angular separation
[error]    Falsified after 57 passed tests.
[error]    > ARG_0: Coordinates(RA(Angle(283.4000000000001°)),Dec(Angle(300.0°)))
[error]    > ARG_1: Coordinates(RA(Angle(323.1999999999998°)),Dec(Angle(90.0°)))

swalker2m:
This is the item that is failing:

```scala
    "be consistent with fractional angular separation" ! 
      forAll { (c1: Coordinates, c2: Coordinates) =>
        val sep = c1.angularDistance(c2)
        (-1.0 to 2.0 by 0.1).forall { f =>
          val stepSep = c1.interpolate(c2, f).angularDistance(c1)
          stepSep.toDegrees ~= (sep * f).toSignedDegrees.abs
        }
      }
```
If you look at the coordinates that fail

```
[error] > ARG_0: Coordinates(RA(Angle(283.4000000000001°)),Dec(Angle(300.0°)))
[error] > ARG_1: Coordinates(RA(Angle(323.1999999999998°)),Dec(Angle(90.0°)))
```

You can see 150° of separation in declination.  I think that the multiplication in the last line just doesn't take into account that we're dealing with circles and it wraps around.

PR #1272, opened 2017-06-07
REL-3171: fix TigraTable creation

swalker2m:
This PR updates the way that `TigraTableCreator` obtains a `VcsService`, taking advantage of the new `SecureServiceFactory` utility.  The previous version was failing to find a `VcsService` and therefore failing to update the TigraTable data.


tpolecat:
TigraTable … hoo boy. That takes me back.

PR #1275, opened 2017-06-15
REL-3175: TigraTable Redux

swalker2m:
#1272 introduced a solution for a bug with Tigra table creation.  The problem there was that the `VcsService` upon which it relies couldn't be found in the OSGi service registration.

Once that problem was fixed, we encountered new issues while testing:

1. After scheduling a Tigra table creation, there would be a 15 minute delay before it would actually begin
2. Once started, the Tigra table functor would run for about 45 minutes before finally producing the requested table

There are two commits in this PR, each addresses a separate issue.  First, the 15 minute delay was caused by the Google Services registration code which was done in the global execution context and would spend most of those 15 minutes doing a `Thread.sleep`.  It also suffered from a race condition assigning the service registration.  To fix the problem, like a caveman I just create a new `Thread` to do the registration and use `synchronized`.

Second, the 45 minute run time of the `TigraTableFunctor` was caused by doing repeated VCS log searches to find the last PI sync time.  Skipping the last sync time lookup the table could be created in one minute.  I modified the code to perform just one query, loading up the last sync time for all programs into memory.  I think given the size of the data (226M in testing, 63M in production) that should be fine.  I'm open to suggestions though, particularly if I'm not using doobie to its fullest potential (@tpolecat).  With this change, the exact same data is produced but in 74 seconds instead of 45 minutes.


swalker2m:
I should point out that the 15 minute delay was only an issue at ODB startup.  Once the Google services thing eventually gives up it was no longer an issue.  Nevertheless I thought it should be fixed.

Also, during the 45 minute `TigraTableFunctor` run the CPU on the machine was 100% busy.  I believe there are other unrelated issues causing occasional periods of high CPU load, but this will shorten one of them to just over a minute instead of 45 minutes.

sraaphorst:
This looks great, and I am impressed that you found the problem and were able to explain it so clearly.

PR #1277, opened 2017-06-15
REL-3182: Removed leading 1 from negative HMS.

sraaphorst:
I accidentally inserted a 1 when displaying negative HMS values.
This removes that.

### DIFF
--- a/bundle/jsky.coords/src/main/java/jsky/coords/HMS.java
+++ b/bundle/jsky.coords/src/main/java/jsky/coords/HMS.java
@@ -187,7 +187,7 @@ public class HMS implements Serializable {
 
     public String toString(boolean showSeconds, boolean showFractionalSeconds) {
         final NumberFormat nf = showFractionalSeconds ? nf_frac : nf_noFrac;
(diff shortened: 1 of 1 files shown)


swalker2m:
I agree, this looks fine but let's wait and see if this bug appears in the OT before merging it into the patch release? If we can't find a way to make it show up then we could just put it into `develop`.

sraaphorst:
Sure. I've been digging through the code for some time now and I haven't yet found a place where I think it's likely to crop up: everything seems like the `HMS` will have non-negative values guaranteed, and the only possible instances where I think it may be a problem is in `SkycatCatalog:_getQueryUrl` L722, or `SexagesimalTableCellRenderer` / `SexagesimalTableCellEditor`, but that is only because I'm not certain how these classes are used.

I think we will be in the clear. I think, though, instead of making any of the changes related to `HMS` to the `develop` branch, we should untangle the idea of `HMS` being used to format both times and RA.

All in all, `HMS` is a mess: there is so much unused cruft in this class.

PR #1279, opened 2017-06-21
qpt test launcher

tpolecat:
This adds a standalone launcher for QPT. When it comes up the keychain (which contains the peer mapping) will treat localhost as if it were Gemini South.

### DIFF
--- a/bundle/edu.gemini.qpt.client/src/main/scala/edu/gemini/qpt/TestLauncher.scala
+++ b/bundle/edu.gemini.qpt.client/src/main/scala/edu/gemini/qpt/TestLauncher.scala
@@ -1,0 +1,51 @@
+package edu.gemini.qpt
+
+import java.io.File
+
+import edu.gemini.ags.conf.ProbeLimitsTable
+import edu.gemini.qpt.ui.ShellAdvisor
+import edu.gemini.qpt.ui.action.PublishAction
+import edu.gemini.spModel.core.{Peer, Site, Version}
+import edu.gemini.ui.workspace.impl.Workspace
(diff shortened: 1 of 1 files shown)


bundle/edu.gemini.qpt.client/src/main/scala/edu/gemini/qpt/TestLauncher.scala
```diff
@@ -0,0 +1,51 @@
...
+import edu.gemini.qpt.ui.ShellAdvisor
+import edu.gemini.qpt.ui.action.PublishAction
+import edu.gemini.spModel.core.{Peer, Site, Version}
+import edu.gemini.ui.workspace.impl.Workspace
+import edu.gemini.util.trpc.auth.TrpcKeyChain
+import edu.gemini.util.security.auth.keychain.Action._
+import edu.gemini.util.security.auth.ui.AuthDialog
+
+/** Standalone launcher for running QPT inside an IDE. */
+object QptTestLauncher {
```
cquiroz: Shouldn't this extend `App`? Would scala take any object with `main` as an app?

jdnavarro:
I keep having problems with my JDK/IDEA setup, so I think it's going to take a while for me to actually run this Launcher. I'll let you know when I do, meanwhile, it's fine with me if you merge it.

bundle/edu.gemini.qpt.client/src/main/scala/edu/gemini/qpt/TestLauncher.scala
tpolecat: No, in general you don't want to use `App` because it has weird initialization semantics. Just use an object than implements `main`.

PR #1280, opened 2017-06-22
REL-1485: OT elevation plot background

jdnavarro:
This is an early PR to check I'm going in the right direction. I have to still figure out how to add a legend and some thin lines.

Now the background colors of the elevation plots of the OT go like this:

Light grey: Sunset -> Start Nautical twilight (12°)
Dark grey: Start Nautical twilight (12°) -> End Nautical twilight (12°)
Light grey: End Nautical twilight (12°) -> Sunrise

elevation_plots


swalker2m:
Looks like it is correct.  Perhaps the colors, including the lavender background the QPT uses, should match exactly as well?

jdnavarro:
I was also trying to rearrange the legend and draw lines in the boundaries between areas but I couldn't come up with a way to display it properly.

This is how is looking now, it's not exactly the same colors as in the QPT but hope is good enough.

lilac_elevation_plot

PR #1282, opened 2017-06-27
SEQNG-271: Port to scala 2.12

cquiroz:
This PR enables to compile on scala 2.12 all the dependencies that are needed for ocs3. For the most part this is straightforward but the bundle `edu.gemini.util.security` breaks due to a compiler bug against the java mail API

Instead of chasing a solution I divided the bundle into two, one containing the minimal to make trpc work and another (`edu.gemini.util.security.ext`) containing the UI and the mailing parts that would only be compiled with 2.11

Also doobie had to be updated to support 2.12, thanks @tpolecat 

All the applications were tested but this is a bit delicate change, same breaks could happen


bundle/edu.gemini.util.security/build.sbt
```diff
@@ -11,24 +11,18 @@ unmanagedJars in Compile ++= Seq(
...
-  "org.scala-lang.modules" %% "scala-swing" % "2.0.0-M2",
-  "org.scalaz.stream"      %% "scalaz-stream" % "0.8a" intransitive(),
   "org.scalaz"             %% "scalaz-core" % ScalaZVersion,
   "org.scalaz"             %% "scalaz-effect" % ScalaZVersion,
   "org.scalaz"             %% "scalaz-concurrent" % ScalaZVersion,
-  "org.tpolecat"           %% "doobie-core" % "0.3.0-M1",
-  "com.chuusai"            %% "shapeless" % "2.3.0",
-  "org.scodec"             %% "scodec-bits" % "1.0.9",
-  "org.scala-lang"         %  "scala-reflect" % "2.11.7",
-  "org.scala-lang"         %  "scala-compiler" % "2.11.7"
```
swalker2m: The swing stuff moved to `ext` but `scalaz-stream`, `shapeless`, and `scodec-bits` weren't actually needed?

cquiroz: They come as dependencies from doobie, thus we don't need to keep the versions in sync

tpolecat:
👍

cquiroz:
As requested I confirmed that with this change the OT can communicate with spdb

PR #1283, opened 2017-06-29
REL-3189: Transitional Observation Export for New Seqexec

swalker2m:
This PR offers a drop-in replacement for the existing `SeqExecService` used by the new seqexec.  Instead of fetching a `ConfigSequence` from the ODB via `trpc`, it makes use of a modified `Ocs3ExportServlet` to obtain an XML `String` that is then parsed into a `ConfigSequence` on the client side.

Note that dependencies for the `edu.gemini.seqexec.odb` bundle have changed, since `trpc` is no longer necessary but `edu.gemini.spModel.io` is needed.

We decided to try this approach when trying to update ocs3 to Scala 2.12 and finding that it couldn't work with the Scala 2.11 ocs over `trpc`.  This PR removes the need to use `trpc` from the new seqexec.  Keep in mind that this is a transitional service though.  Ultimately we will move to importing observations into a postgres database and using the model being developed in `gem` for the sequence.


cquiroz:
This is great, thanks a lot

bundle/edu.gemini.seqexec.odb/src/main/scala/edu/gemini/seqexec/odb/SeqExecService.scala
```diff
@@ -22,13 +32,77 @@ trait SeqExecService {
...
 
-    override def sequence(oid: SPObservationID): TrySeq[ConfigSequence] =
-      call(_.sequence(oid))
-  }
+  /** Constructs a SeqExecService that works with a servlet running in the ODB.
+    * When a sequence is requested, it contacts the servlet and requests the
+    * observation XML (wrapped in a program shell suitable for importing). The
+    * XML is then parsed and the sequence is extracted.
+    */
+  def client(peer: Peer): SeqExecService =
```
cquiroz: I guess this will be the main client facing API right?
Could we add one that returns `ISPObservation` so we can read more data like the observation name? Can we get `ConfigSequence` from `ISPObservation`?

swalker2m: Right, I wanted to avoid changing the API so that we can try this without changing seqexec code.  If you want to return something in addition to the `ConfigSequence` though you could just change the API.  To get the `ConfigSequence` from an `ISPObservation`:

```scala
import edu.gemini.spModel.config.ConfigBridge
import edu.gemini.spModel.config.map.ConfigValMapInstances.IDENTITY_MAP

```

PR #1284, opened 2017-07-06
REL-472: Altair + GNIRS spectroscopy support for ITC

olesjasmirnova:
Enabled Altair+GNIRS spectroscopy support for the ITC after testing and making sure the calculations are OK.


app/spdb/build.sbt
```diff
@@ -250,12 +250,12 @@ def dnavarro(version: Version) = AppConfig(
     "edu.gemini.auxfile.fits.dest"               -> "/gemsoft/var/data/ictd/test/GS@SEMESTER@/@PROG_ID@",
     "edu.gemini.auxfile.fits.host"               -> "gsconfig.gemini.edu",
     "edu.gemini.auxfile.other.dest"              -> "/gemsoft/var/data/finder/GSqueue/Finders-Test/@SEMESTER@/@PROG_ID@",
-    "edu.gemini.auxfile.root"                    -> "/Users/jnavarro/.auxfile",
+    "edu.gemini.auxfile.root"                    -> "/Users/dnavarro/.auxfile",
```
cquiroz: Why this change, @jdnavarro Would this affect you?

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gnirs/GnirsRecipe.java
```diff
@@ -220,8 +220,8 @@ public SpectroscopyResult calculateSpectroscopy() {
 
             specS2N.setSourceSpectrum(sed);
             specS2N.setBackgroundSpectrum(sky);
-            if (haloThroughput.nonEmpty()) {
-                specS2N.setHaloSpectrum(sed, haloThroughput.get(), IQcalc.getImageQuality());
+            if (altair.isDefined()) {
```
cquiroz: It maybe a good idea to also check the other options like
```
if (altair.isDefined() && halo.isDefined() && haloTroughtput.isDefined()) {
```

olesjasmirnova: Good point, I'll add more checks

app/spdb/build.sbt
olesjasmirnova: I was getting a credentials error after pulling from the central (haven't pulled for A WHILE), so I relinked the credentials file, which resulted in this change from "dnavarro" to "jnavarro".

cquiroz: It would probably be best to return them to its original state. Do you use jnavarro's spdb for testing? You can create your own as well

olesjasmirnova: The change from jnavarro to dnavarro happened in commit 44e80f592b178f294458116ad75bd19c883b6069 back in 2016, made not by me.

I was using my own credentials, but they don't seem to be there any longer. I was reverting some test code before pushing the changes and I may have removed them by accident. I'll introduce them again.

olesjasmirnova: Sorry, I checked this a bit more in detail. The "dnavarro" from my code comes indeed from that commit, but it was changed at some point and I accidentally reverted it during the merge. It should be change back to "jnavarro"

cquiroz: Thanks, if you prefer we can handle this ourselves and i could also create credentials for you

Are you actually using spdb in development or would it be enough to use the itc app. That one is simpler to deploy and test

jdnavarro: Oh, now I understand why I had to create the `.auxfile` dir manually. My global username is *jnavarro* but in svn, where the credentials are taken from, it was created as *dnavarro*. I don't know if that has changed recently.

@olesjasmirnova it's better if you get an account on your own because if I had to setup `ocs` again I may run into trouble without knowing why.

olesjasmirnova: @cquiroz I have used the spdb in the past when working on a task for the OT, and I guess I may need to use it at some point in the future. I can add my credentials myself, as you created them for me a while ago. I imagine that it would be better not to do it in a commit in this branch. What would be the right one?

@jdnavarro, sorry for the mess, I'll fix the build file and send a new pull request.

cquiroz: Don't worry, we could merge this and make another PR for your credentials

cquiroz:
@olesjasmirnova Thanks for the update, I think we are ready to merge

PR #1286, opened 2017-07-07
REL-3819: Prepare the PIT for 2018A

cquiroz:
This PR updates the version numbers and upconverts to start building the PIT 2018A

### DIFF
--- a/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/transform/UpConverter.scala
+++ b/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/transform/UpConverter.scala
@@ -34,16 +34,17 @@ object UpConverter {
   }
 
   // Sequence of conversions for proposals from a given semester. Cleaned up for 2017A since older conversion multiples to increment version no longer necessary.
-  val from2017A:List[SemesterConverter] = List(SemesterConverterToCurrent, SemesterConverter2017ATo2017B, LastStepConverter(Semester(2017, SemesterOption.A)))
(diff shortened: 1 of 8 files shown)


bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/transform/UpConverter.scala
```diff
@@ -34,16 +34,17 @@ object UpConverter {
...
+  val from2013A:List[SemesterConverter] = List(SemesterConverterToCurrent, SemesterConverter2017BTo2018A, SemesterConverter2016BTo2017A, SemesterConverter2016ATo2016B, SemesterConverter2015ATo2015B, SemesterConverter2014BTo2015A, SemesterConverter2014BTo2014BSV, SemesterConverter2013ATo2013B, SemesterConverter2013BTo2014A, SemesterConverter2014ATo2014B, SemesterConverter2017ATo2017B, LastStepConverter(Semester(2013, SemesterOption.A)))
+  val from2012B:List[SemesterConverter] = List(SemesterConverterToCurrent, SemesterConverter2017BTo2018A, SemesterConverter2016BTo2017A, SemesterConverter2016ATo2016B, SemesterConverter2015ATo2015B, SemesterConverter2014BTo2015A, SemesterConverter2014BTo2014BSV, SemesterConverter2012BTo2013A, SemesterConverter2013ATo2013B, SemesterConverter2013BTo2014A, SemesterConverter2014ATo2014B, SemesterConverter2017ATo2017B, LastStepConverter(Semester(2012, SemesterOption.B)))
```
swalker2m: Seems like there must be a way to simplify this?

cquiroz: Not much really. I guess we could remove some duplicated code but that would be limited to the first 2 columns. This lets us move old proposals to current. It will get unwidely with time but by then we'd have probably moved away from this xml based format

PR #1287, opened 2017-07-07
REL-3114: Remove constraint on email address suffixes in PIT

cquiroz:
This PR relaxes the restriction that emails TLDs must be between 2 and 4 characters. The exported PDF now also will behave better for very long email addresses

I added a couple of changes to the build to simplify launching the PIT from sbt rather than IntelliJ

### DIFF
--- a/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/package.scala
+++ b/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/package.scala
@@ -17,8 +17,8 @@ package object immutable {
 
     // Trim lines, join adjacent lines if non-empty
     def unwrapLines: String = s.lines.map(_.trim).dropWhile(_.isEmpty).map {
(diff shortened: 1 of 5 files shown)


bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/package.scala
```diff
@@ -18,7 +18,7 @@ package object immutable {
     // Trim lines, join adjacent lines if non-empty
     def unwrapLines: String = s.lines.map(_.trim).dropWhile(_.isEmpty).map {
       case s0 if s0.isEmpty => "\n"
-      case s0 => s0 + " "
+      case s0               => s0 + " "
```
swalker2m: I guess not part of this change but it seems like we could just match the empty string directly:

```scala
case "" => "\n"
```

cquiroz: true, I'll patch that

PR #1290, opened 2017-07-11
Add Smart Gcal Updating to Seqexec Client

swalker2m:
In the previous REL-3189 (#1283), I updated the seqexec client to retrieve observation XML in order to relieve the dependency on binary serialization and make it possible to compile the seqexec with Scala 2.12. The seqexec client now expands the sequence in its own process instead of directly receiving it from the database.

What I failed to take into account though is that sequence expansion requires smart gcal lookup tables. This PR addresses that oversight.  I wasn't sure how this will be incorporated into the seqexec though, so feel free to update this as necessary.  As it stands there is a `SmartGcal.initialize()` function that has to be called to get a `CalibrationProvider`, which is needed by the `SeqExecService` client.  You pass the Gemini South ODB peer (always GS) and a directory where it can store smart gcal configuration .csv files.  Reading the config files can take a few seconds so you might want to do this in a separate thread.

I also took advantage of the opportunity to update the pre-packaged .csv files which are used until the first update from the repository takes place.  That's probably a bad idea but that's how smart gcal works and I didn't want to dive into it, especially since smart gcal has been completely redone in `gem`.

### DIFF
--- a/bundle/edu.gemini.seqexec.odb/src/main/scala/edu/gemini/seqexec/odb/SmartGcal.scala
+++ b/bundle/edu.gemini.seqexec.odb/src/main/scala/edu/gemini/seqexec/odb/SmartGcal.scala
@@ -1,7 +1,7 @@
 package edu.gemini.seqexec.odb
 
 import edu.gemini.spModel.core.Peer
-import edu.gemini.spModel.gemini.calunit.smartgcal.{CalibrationProvider, CalibrationRepository}
+import edu.gemini.spModel.gemini.calunit.smartgcal.{CalibrationProviderHolder, CalibrationProvider, CalibrationRepository}
 import edu.gemini.spModel.smartgcal.UpdatableCalibrationRepository
 import edu.gemini.spModel.smartgcal.provider.CalibrationProviderImpl
 import edu.gemini.spModel.smartgcal.repository.{CalibrationUpdater, CalibrationFileCache, CalibrationRemoteRepository}
@@ -10,13 +10,13 @@ import java.nio.file.Path
 import java.time.Duration
 
 object SmartGcal {
-  def initialize(peer: Peer, dir: Path): TrySeq[CalibrationProvider] =
+  def initialize(peer: Peer, dir: Path): TrySeq[Unit] =
     catchingAll {
       val service    = new CalibrationRemoteRepository(peer.host, peer.port)
       val cachedRepo = new CalibrationFileCache(dir.toFile)
       val provider   = new CalibrationProviderImpl(cachedRepo)
+      CalibrationProviderHolder.setProvider(provider)
       CalibrationUpdater.instance.addListener(provider)
       CalibrationUpdater.instance.start(cachedRepo, service, Duration.ofHours(1).toMillis)
-      provider
     }
 }
--- a/bundle/edu.gemini.seqexec.odb/src/main/scala/edu/gemini/seqexec/odb/SmartGcal.scala
+++ b/bundle/edu.gemini.seqexec.odb/src/main/scala/edu/gemini/seqexec/odb/SmartGcal.scala
@@ -1,0 +1,22 @@
+package edu.gemini.seqexec.odb
+
+import edu.gemini.spModel.core.Peer
+import edu.gemini.spModel.gemini.calunit.smartgcal.{CalibrationProvider, CalibrationRepository}
+import edu.gemini.spModel.smartgcal.UpdatableCalibrationRepository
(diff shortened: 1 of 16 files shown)


swalker2m:
Don't be (too) discouraged by the size of the change.  The `.csv` file updates are the bulk of it and those can be ignored.

swalker2m:
If you want to simplify this mess, we can ignore the auto-updating bit and just specify a directory where the `.csv` files have been retrieved from SVN.  If they should change in the future, we'd just need to svn up them manually.  That would remove the need to ping the GS ODB for updates.

cquiroz:
Thanks @swalker2m for this. I'm ok with whatever option you think is best. auto-updating is probably better since we may forget to update but I don't have a strong preference

cquiroz:
just to be sure, all this is relevant only on the client side, right? I'm asking since we plan to deploy the odb soon and I want to make sure we don't have to include these changes

bundle/edu.gemini.seqexec.odb/src/main/scala/edu/gemini/seqexec/odb/SmartGcal.scala
```diff
@@ -0,0 +1,22 @@
...
+import edu.gemini.spModel.gemini.calunit.smartgcal.{CalibrationProvider, CalibrationRepository}
+import edu.gemini.spModel.smartgcal.UpdatableCalibrationRepository
+import edu.gemini.spModel.smartgcal.provider.CalibrationProviderImpl
+import edu.gemini.spModel.smartgcal.repository.{CalibrationUpdater, CalibrationFileCache, CalibrationRemoteRepository}
+
+import java.nio.file.Path
+import java.time.Duration
+
+object SmartGcal {
+  def initialize(peer: Peer, dir: Path): TrySeq[CalibrationProvider] =
```
cquiroz: So this need to be called only once during the application?

bundle/edu.gemini.seqexec.odb/src/main/scala/edu/gemini/seqexec/odb/SmartGcal.scala
```diff
@@ -0,0 +1,22 @@
...
+    catchingAll {
+      val service    = new CalibrationRemoteRepository(peer.host, peer.port)
+      val cachedRepo = new CalibrationFileCache(dir.toFile)
+      val provider   = new CalibrationProviderImpl(cachedRepo)
+      CalibrationUpdater.instance.addListener(provider)
+      CalibrationUpdater.instance.start(cachedRepo, service, Duration.ofHours(1).toMillis)
```
cquiroz: Could there be troubles if we do the update at the same exact time that we need to unloop a sequence?

project/OcsBundle.scala
```diff
@@ -242,6 +242,7 @@ trait OcsBundle {
       bundle_edu_gemini_spModel_core,
       bundle_edu_gemini_spModel_io,
       bundle_edu_gemini_spModel_pio,
+      bundle_edu_gemini_spModel_smartgcal,
```
cquiroz: We need to make sure this runs on 2.12. Could we add that bundle to the travis check for 2.12?

swalker2m:
> just to be sure, all this is relevant only on the client side, right?

Correct, the service it talks to just provides access to the observation XML and that doesn't change. We only update the client used by the seqexec for this.

bundle/edu.gemini.seqexec.odb/src/main/scala/edu/gemini/seqexec/odb/SmartGcal.scala
swalker2m: Yeah, assuming it doesn't blow up.

bundle/edu.gemini.seqexec.odb/src/main/scala/edu/gemini/seqexec/odb/SmartGcal.scala
swalker2m: It was designed to handle that, but I don't feel comfortable making guarantees about anything in the Smart Gcal implementation in ocs2.  At any rate, this is the exact same thing done by the ODB and OT.

swalker2m:
I added a(nother) dubious commit when, tracing through the code to see how the updating mechanism works, I saw dependencies on having the `CalibrationProvider` set in something called the `CalibrationProviderHolder`.  Sorry, I'd hoped to avoid it.

cquiroz:
I tested this on the seqexec and it seems to work nicely. Now I get all the steps for smartgcal sequences

Great work!

PR #1292, opened 2017-07-13
REL-3144

olesjasmirnova:
Added GMOS-specific saturation limit warning message. Also updated the well depth values for GMOS-N and S.


bundle/edu.gemini.itc/src/main/scala/edu/gemini/itc/gmos/GmosSaturLimitRule.scala
```diff
@@ -0,0 +1,22 @@
+package edu.gemini.itc.gmos
+
+import edu.gemini.itc.base.{LimitRule, Result}
+
+/** GMOS specific warning rule. */
+final case class GmosSaturLimitRule(adLimit: Double, wellDepth: Double, xBin: Int, yBin: Int,
```
cquiroz: Very nice, moving to scala very well 👍

cquiroz:
I added a commit to get travis to build including your credentials and it seems fine now

PR #1294, opened 2017-07-14
REL-3193: Support Subaru Intensive Programs in PIT

cquiroz:
This PR adds a new proposal class to the PIT: `Subaru Intensive Proposal` class. The PR updates the model, adds UI controls to select the proposal properties and includes a new backend for submissions.

Here's a screenshot

screen shot 2017-07-14 at 12 23 41 pm

### DIFF
--- a/bundle/edu.gemini.model.p1/src/main/xsd/ProposalClass.xsd
+++ b/bundle/edu.gemini.model.p1/src/main/xsd/ProposalClass.xsd
@@ -136,6 +136,7 @@
         <xsd:complexContent>
             <xsd:extension base="ProposalClass">
                 <xsd:sequence>
+                    <xsd:element name="telescope" type="ExchangeTelescope"/>
                     <xsd:element name="submission" type="SubaruIntensiveProgramSubmission"/>
                 </xsd:sequence>
             </xsd:extension>
--- a/bundle/edu.gemini.model.p1/src/main/xsd/ProposalClass.xsd
+++ b/bundle/edu.gemini.model.p1/src/main/xsd/ProposalClass.xsd
@@ -17,6 +17,7 @@
             <xsd:element name="special"         type="SpecialProposalClass"/>
             <xsd:element name="exchange"        type="ExchangeProposalClass"/>
             <xsd:element name="large"           type="LargeProgramClass"/>
+            <xsd:element name="sip"             type="SubaruIntensiveProgramClass"/>
             <xsd:element name="fastTurnaround"  type="FastTurnaroundProgramClass"/>
         </xsd:choice>
     </xsd:complexType>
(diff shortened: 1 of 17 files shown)


bundle/edu.gemini.model.p1/src/main/xsd/ProposalClass.xsd
```diff
@@ -129,6 +130,21 @@
...
+        Subaru Intensive Program
+    -->
+    <xsd:complexType name="SubaruIntensiveProgramClass">
+        <xsd:complexContent>
+            <xsd:extension base="ProposalClass">
+                <xsd:sequence>
+                    <xsd:element name="telescope" type="ExchangeTelescope"/>
+                    <xsd:element name="submission" type="SubaruIntensiveProgramSubmission"/>
+                </xsd:sequence>
+                <xsd:attribute name="tooOption" type="TooOption" use="optional"/>
```
swalker2m: Subaru supports ToO?

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/view/partner/PartnerView.scala
```diff
@@ -267,26 +310,30 @@ class PartnerView extends BorderPanel with BoundView[Proposal] {view =>
...
-          case l: LargeProgramClass          =>
+          case l: LargeProgramClass           =>
             selection.item = l.tooOption
             visible = true
-          case f: FastTurnaroundProgramClass =>
+          case f: FastTurnaroundProgramClass  =>
             selection.item = f.tooOption
             visible = true
-          case _                             =>
+          case SubaruIntensiveProgramClass(_, _, _, ExchangeTelescope.GEMINI, tooOption, _) =>
```
swalker2m: What's the `ExchangeTelescope` and why is it `GEMINI`?

bundle/edu.gemini.model.p1/src/main/xsd/ProposalClass.xsd
cquiroz: No, for this type of proposal you have to select a telescope, which I call an `ExchangeTelescope` for lack of a better word. If your telescope is `Gemini` you have the ToO choice

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/view/partner/PartnerView.scala
cquiroz: I probably answered this above

cquiroz:
There is no work on ITAC for this. we will not import them into ITAC like FT

cquiroz:
BTW, I want to get this in to produce a prototype, lt is likely that there will be changes after user testing

bundle/edu.gemini.model.p1/src/main/xsd/ProposalClass.xsd
swalker2m: So it is "subaru intensive" but has some Gemini observations?

cquiroz: The requirement is that for this type of proposal you select the observatory and that defines the selection of the instruments either Gemini or Subaru

cquiroz:
You have a point on questioning the impact on phase II, this should be considered. I'll post a note on the jira ticket

bundle/edu.gemini.model.p1/src/main/xsd/ProposalClass.xsd
swalker2m: I read the information in REL-3193 but I'm still not sure what this is all about.  Are these LP programs for Gemini or are the observations at Subaru or maybe both?

cquiroz:
Updated according to the latest specifications

PR #1299, opened 2017-07-18
REL-3197: Homogenize SmartGcal Configuration Files

swalker2m:
This PR provides an `exportSmartGcal` shell command that writes all SmartGcal lookup table information in a format that is easily ingested into OCS3.  Namely, unlike the existing `.csv`s the files produced by `exportSmartGcal`:

1) Always use the sequence value for each enumerated type (this conveniently matches the existing `String` parsers for OCS2 types in OCS3).

2) Expands wildcards and regular expressions.

3) Is case-sensitive.

4) Exports wavelengths consistently as integral angstroms for all instruments.

5) Exports exposure times consistently as integral milliseconds.

5) Removes arbitrary missing values separating the instrument configuration from the matching calibration data.

The implementation is suboptimal in that it adds `export` methods in the various SmartGcal classes rather than, say, adding this capability via a typeclass.  It seemed like it would be challenging to get at the data to be exported since they are stored in private mutable fields in the various calibration maps.  At any rate, the point of the work being done in OCS3 is to replace all of this SmartGcal infrastructure.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/calunit/smartgcal/Calibration.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/calunit/smartgcal/Calibration.java
@@ -7,6 +7,7 @@
 
 package edu.gemini.spModel.gemini.calunit.smartgcal;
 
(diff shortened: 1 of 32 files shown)


tpolecat:
👍

PR #1300, opened 2017-07-19
REL-3193: Support SIP programs on the PDF templates

cquiroz:
Small detail forgotten when adding SIP proposals was to include them on the templates. This PR fixes that


cquiroz:
You can do some level of unit testing but it they are very involved

PR #1303, opened 2017-07-24
REL-3193: Updated description of SIP programs

cquiroz:


bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
```diff
@@ -540,7 +540,7 @@ class ProblemRobot(s: ShellAdvisor) extends Robot {
             case e: ExchangeSubmission               => Partners.name(e.partner)
             case s: SpecialSubmission                => s.specialType.value
             case l: LargeProgramSubmission           => "large program"
-            case i: SubaruIntensiveProgramSubmission => "Subaru intensive program"
```
swalker2m: Trivial, but I'm wondering why intensive is capitalized if the other cases aren't.

cquiroz: It probably should be, I followed the convention of the live above that uses `large program` non capitalized

swalker2m: The comment seemed to go with the deletion instead of the addition, where it would have made more sense.  Here it is capitalized.
```
case i: SubaruIntensiveProgramSubmission => "Intensive program observing at Subaru"
```

PR #1305, opened 2017-07-25
REL-3086: Configure the P1 monitor to handle CFHT proposals

cquiroz:
For historical reasons the PIT has mixed the usage of CFHT and CFH to refer to the Canada-France-Hawaii Telescope

This PR fixes this unifying it to use always CFH and it also enables the p1-monitor to handle CFH properly


swalker2m:
This will need a corresponding change in `SpProgramFactory` if not elsewhere.

cquiroz:
@swalker2m Great catch, updated on the last commit. This omission broke the build

PR #1309, opened 2017-07-31
REL-2835 Make QPT use the planned date for non-sidereal targets

jdnavarro:
This is still not ready to be merged since the build is broken but I'm sending it already in order to get early feedback. I'm quite uneasy the *functional* Java code I'm writing 😰.

What's left, aside of making sure it works as it's supposed once I'm able to compile, is to update every instance where `getCoords` is being called, restore the cache and handle the case when no coordinates are passed.

**TODO**

- [ ] What to do when there are no coordinates for the date.
- [x] Restore Cache.


jdnavarro:
I think it's at a state where we can start checking if everything is coming alright in the UI.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/AzimuthListener.java
```diff
@@ -35,7 +35,7 @@ public void propertyChange(PropertyChangeEvent evt) {
 		ApproximateAngle windConstraint = variant.getWindConstraint();
 		if (windConstraint != null) {
 			for (Alloc a: variant.getAllocs()) {
-				AzimuthSolver solver = new AzimuthSolver(schedule.getSite(), a.getObs().getCoords(), windConstraint);
+				AzimuthSolver solver = new AzimuthSolver(schedule.getSite(), a.getObs().getCoords(schedule.getMiddlePoint()), windConstraint);
```
swalker2m: Why doesn't `AzimuthSolver` get the same treatment as `ElevationConstraintSolver`s?  This is, why not construct it with a function that takes a time and returns coordinates instead of taking coordinates at a fixed time?

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/Variant.java
```diff
@@ -703,7 +703,7 @@ private void updateObsFlags() {
 			// can not be scheduled at all tonight.
 			Union<Interval> darkUnion = darkUnionCache.get(obs);
 			if (darkUnion == null) {
-				WorldCoords coords = obs.getCoords();
+				WorldCoords coords = obs.getCoords(getSchedule().getMiddlePoint());
```
swalker2m: Same question here for the `BackgroundSolver`.  It seems like they should all work the same way.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/Alloc.java
```diff
@@ -174,7 +175,7 @@ private synchronized void initCircumstances() {
 		if (circV == null) {
 
 			// Calculate visit circumstances, which includes setup time.
-			WorldCoords coords = new WorldCoords(getObs().getRa(), getObs().getDec());
+			WorldCoords coords = getObs().getCoords(getMiddlePoint());
```
swalker2m: Here again we get coords at the middle point and then use them later on (line 188) at a point where we have a better idea of the precise time for which we wanted the coordinates (and worse, it is changing over a loop).  We should use the most accurate time possible for getting the coordinates in all cases.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/SetupListener.java
```diff
@@ -56,7 +56,7 @@ private boolean compatible(Alloc pred, Alloc succ) {
 		return 
 			pred != null &&
 			pred.getObs().getInstrumentString().equals(succ.getObs().getInstrumentString()) &&
-			pred.getObs().getCoords().equals(succ.getObs().getCoords());
+			pred.getObs().getCoords(pred.getMiddlePoint()).equals(succ.getMiddlePoint());
```
swalker2m: For a non-sidereal target in theory this will never be equals, but maybe that just means you always need setup for two non-sidereal observations even if they are the "same" target?  I don't know the answer to this one.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/util/Interval.java
```diff
@@ -201,8 +201,13 @@ public long getStart() {
 	}
 
 	/**
-	 * Returns the length of this Interval, as <code>end - start</code>.
+	 * Returns the middle point for this Interval.
 	 */
+	public long getMiddlePoint() { return start + end / 2; }
```
swalker2m: Oops, `(start + end) / 2`

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/ui/html/ScheduleDocument.java
```diff
@@ -287,7 +287,7 @@ public String getCond(byte value) {
     }
 
     public RiseTransitSet getRTS(Alloc alloc) {
-        return new RiseTransitSet(schedule.getSite(), alloc.getObs().getCoords(), alloc.getStart());
+        return new RiseTransitSet(schedule.getSite(), alloc.getObs().getCoords(alloc.getVariant().getSchedule().getMiddlePoint()), alloc.getStart());
```
swalker2m: Here we should be consistent and pass in the `Long -> Coordinates` function to `RiseTransitSet`.  In that class there is an `airmass` function that takes a time which you can apply to the function to get the precise coordinates.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/ui/view/candidate/CandidateObsComparator.java
```diff
@@ -34,7 +34,7 @@ public int compare(Obs o1, Obs o2) {
 			return direction * ((ret != 0) ? ret : -o1.compareTo(o2));
 		}
 		
-		return direction * attr.getComparator().compare(o1, o2);
+		return direction * attr.getComparator(variant.getSchedule().getMiddlePoint()).compare(o1, o2);
```
swalker2m: Let's come back to this one.  We need to be sure every time we're using the middle point that there isn't a more accurate option.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/util/Interval.java
jdnavarro: image

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/ui/view/candidate/CandidateObsComparator.java
jdnavarro: I've looked into `table` and `viewer` but couldn't find a more precise timestamp. I'll try passing the timestamp as a parameter to the constructor as we did with `CandidateObsComparator`. It's only instantiated once so it shouldn't be that much work.

jdnavarro: I'm retiring what I proposed above. I can't find a *better* date where `CandidateObsComparator` is [instantiated](https://github.com/gemini-hlsw/ocs/blob/4fd9f5910d33c82a9793232476503edd63dc46b9/bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/ui/view/candidate/CandidateObsViewAdvisor.java#L70).

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/SetupListener.java
swalker2m: Looking at this closely, it appears that we're comparing `WorldCoords` to a `Long` which is never going to be equals.

It's counterintuitive perhaps but maybe the time to use for both the `pred` and `succ` observation would be the `pred` interval end or the `succ` interval start.  The idea is that you can continue observing without a new setup if the two observations are in the same place, which that would presumably tell you.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/ui/view/candidate/CandidateObsComparator.java
swalker2m: I'm not sure, but the midpoint probably makes sense here. We're sorting observations based on some attribute which might vary based on time.  But to compare one observation to another we have to fix the time to be used.  The midpoint of the night seems as good as any time in this case?  @tpolecat do you have any input?  I'm not sure how this is used.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/ui/view/property/adapter/ObsAdapter.java
```diff
@@ -37,7 +37,7 @@ static void addObsProperties(Variant variant, Obs target, PropertyTable table) {
 		table.put(PROP_TITLE, target.getTitle());
 		table.put(PROP_INSTRUMENT, target.getInstrumentStringWithConfig());
 		table.put(PROP_CONSTRAINTS, target.getConstraintsString());
-		table.put(PROP_COORDINATES, target.getCoords());
+		table.put(PROP_COORDINATES, target.getCoords(variant.getSchedule().getMiddlePoint()));
```
swalker2m: The plan is for this one to be updated to use the midpoint of its allocation, correct?

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/SetupListener.java
```diff
@@ -56,7 +56,7 @@ private boolean compatible(Alloc pred, Alloc succ) {
 		return 
 			pred != null &&
 			pred.getObs().getInstrumentString().equals(succ.getObs().getInstrumentString()) &&
-			pred.getObs().getCoords(pred.getMiddlePoint()).equals(succ.getMiddlePoint());
```
swalker2m: I think maybe what we want here is:

```java
pred.getObs().getCoords(pred.getEnd()).equals(succ.getObs().getCoords(pred.getEnd()))
```

That fixes the bug where we are comparing two different types (`WorldCoords` vs. `Long`) and uses the same time, in this case `pred.getEnd()`, for the two observations.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/ui/view/property/adapter/ObsAdapter.java
jdnavarro: Correct. We also want the coordinates to be updated when dragging the observation as @andrewwstephens mentioned. I think I already know how to do it.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/SetupListener.java
swalker2m: I think we need either `pred.getEnd` or `succ.getStart` (i.e., the same time) for both observations.  We're trying to see whether setup is needed when switching from one observation to another so we want to know if they are in different places at the time of the switchover from one to another.

I'm guessing here, it would be good to get verification from @andrewwstephens.  It may be that we want `pred.getEnd` for the first observation and `succ.getStart` for the second and if they are not the same time and it is a fast moving non-sidereal then we need setup anyway even if the target is the same.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/ui/view/candidate/CandidateObsComparator.java
tpolecat: This is used to sort candidate observations in the list, so yes I think middle of the night is the best we can do (unless we want a slider somewhere that lets us pick a reference time).

jdnavarro:
This is ready for another review.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/SetupListener.java
```diff
@@ -56,7 +56,7 @@ private boolean compatible(Alloc pred, Alloc succ) {
 		return 
 			pred != null &&
 			pred.getObs().getInstrumentString().equals(succ.getObs().getInstrumentString()) &&
-			pred.getObs().getCoords().equals(succ.getObs().getCoords());
+			pred.getObs().getCoords(pred.getEnd()).equals(succ.getObs().getCoords((succ.getEnd())));
```
swalker2m: For a non-sidereal target, it seems unlikely that this will ever return `true`.  I think what we're trying to get at here is whether we're still observing the same target.  Since we have no unique ids for targets, we're doing this by checking whether the previous and next observation coordinates are in the same place.  Since targets are moving, to make this check meaningful I believe we need to compare them at the same time.  It's technically possible that two distinct targets could be in exactly the same place at one time and then move apart, but exceedingly unlikely.  So I think pick a time, either `pred.getEnd` or `succ.getStart` and use it for both coordinate lookups.  Alternatively, we could do two comparisons, one of both coordinates at `pred.getEnd` and another of both coordinates at `succ.getStart` but I don't think it's necessary and at any rate still doesn't guarantee correctness.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/ui/view/visualizer/Visualizer.java
```diff
@@ -680,57 +681,57 @@ private void paintGroupBars(Graphics2D g2d) {
...
+	private Shape getElevationCurve(Function<Long, WorldCoords> coords, long start, long end, boolean close, Map<WorldCoords, CachedShape> cache) {
 		synchronized (cacheLock) {
 
 			// Return the cached shape, if any.
-			CachedShape cs = cache.get(coords);
-			if (cs != null && cs.matches(start, end))
-				return cs.shape;
+//			CachedShape cs = cache.get(coords);
+//			if (cs != null && cs.matches(start, end))
+//				return cs.shape;
```
swalker2m: In the end, is the cache just not necessary?  Though again not technically guaranteed to be correct we could probably just make the hash key bigger and add coordinates at a few times (start, middle, end) say.

jdnavarro: If the plot without caching is quick enough we could get away without caching. In my machine it feels fine but haven't used the QPT in anger.

Since functions can't be cached, we'll have to pick the coordinates at certain times as you mention. But can we draw the same exact non-cached plot with just those few coordinates? Is it alright to have an approximation for the elevation plot when cached? Maybe we could cache every step used to draw the plot but that may make the cache too large.

In any case I think we should reassess first if the performance is good enough without caching.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/SetupListener.java
jdnavarro: Ok, I think now I'm getting what you mean. This will do, right? `pred.getObs().getCoords(pred.getEnd()).equals(succ.getObs().getCoords((pred.getEnd())));`

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/ui/view/visualizer/Visualizer.java
swalker2m: The cache is a `Map<WorldCoords, CachedShape>` so the value stored is the whole shape of the plot.  We would "just" need to change the map lookup key to a small collection of, say, 3 coords.  You're probably right though that we shouldn't bother if performance is good enough anyway.  The only way to know is to see how well it works for a full night's plan on moderately old hardware.  If we're not going to use the cache though, we should remove them.

swalker2m: Perhaps we could provide a test version as is and mention this issue as something that needs a close look.  If testers say it is good enough without caching, then we remove the cache and merge that version.

jdnavarro: 😔 Maybe it's not that much work in the end to restore the cache with the 3 coords. Let me give it a try.

swalker2m: Another idea would be to use the observation id as the cache key.  It's true that two or more observations could have the same target and therefore the cache might not be as efficient as it otherwise would be, but at least it would be guaranteed correct.

jdnavarro: > Another idea would be to use the observation id as the cache key.

What happens with the cache if the planned date in the QPT changes?

swalker2m: If the planned date is updated then the model is also updated and [all the caches are cleared](https://github.com/gemini-hlsw/ocs/blob/develop/bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/ui/view/visualizer/Visualizer.java#L128-L132).

jdnavarro:
Now the cache for the plots is restored. It seems to be working fine although I didn't notice much difference in performance.

There is also the remaining point of what to do when there are no coordinates available, is it alright to leave them set to 0 as it is now?

tpolecat:
I was thinking these observations would be excluded altogether … there's a mechanism that lets you explain why things aren't showing up. But zero as a first approximation seems ok to me, then maybe someone could follow up with something friendlier.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/ui/view/visualizer/Visualizer.java
```diff
@@ -709,22 +721,22 @@ private Shape getElevationCurve(Function<Long, WorldCoords> coords, long start,
...
 		synchronized (cacheLock) {
 
 			// Return the cached shape, if any.
-//			Map<WorldCoords, CachedShape> cache = skyBackgroundCurveCache;
-//			CachedShape cs = cache.get(coords);
-//			if (cs != null && cs.matches(start, end))
-//				return cs.shape;
+			Map<String, CachedShapeId> cache = skyBackgroundCurveCache;
+			CachedShapeId cs = cache.get(obsId);
+			if (cs != null && cs.matches(obsId))
```
swalker2m: I think this should be set back the way it was with `start` and `end` times.  The fact that you obtained the `CachedShape` using the `obsId` as key is enough to know that it corresponds to that observation id.  However, if the shape doesn't match the same start/end time period then it shouldn't be used.  I don't think there is any need for a `CachedShapeId` object.

Otherwise this update looks good.

jdnavarro:
> We should update the "setup needed?" calculation at least though before we call it good.

@swalker2m what did you mean by that? Is it some value in the UI? I can't find it...

swalker2m:
> @swalker2m what did you mean by that? Is it some value in the UI? I can't find it...

You [did this already](https://github.com/gemini-hlsw/ocs/pull/1309/commits/8498c0687d9c0c4f35071c2db26eeefadaff6242).

PR #1311, opened 2017-08-04
REL-3220: NPE on p1-monitor

cquiroz:
In certain cases p1-monitor cannot read the list of files being monitored (most likely due to a permissions failure) and it produces an NPE inside the timer. This make is look like the monitor is running but it doesn't process files

This is a simple fix wrapping the `listFiles` call and if it fails it fatally exits


cquiroz:
That's a great suggestion

PR #1315, opened 2017-08-21
Make getCoords always recalculate the coordinates

jdnavarro:
This was causing the issue reported by @andrewwstephens. Now the OT and QPT elevation plot are identical:

coords_fix

(The screenshot from Andy was with Gemini North, but I don't know how to make the QPT switch to the North. In any case that target at that date was also wrong in the QPT for the South before)


cquiroz:
Great!

tpolecat:
Hm, the elevation lines look pixellated. Did you change the rendering hints for those routines?

tpolecat:
Maybe it's always been that way. The hints haven't changed. Are you on a mac?

swalker2m:
@tpolecat I assumed it was because it is a non-sidereal target.

jdnavarro:
> the elevation lines look pixellated. Did you change the rendering hints for those routines?
> Are you on a mac?

Yes I'm on a Mac but I don't remember changing anything related to hinting or scaling. I'm noticing that Swing apps in general look more jagged than they should be. I guess it may have something to do with MacOS autoscaling of apps that don't support HiDPI natively.

I also widened the windows much more than what you'd expect in normal use in order to detect differences. I don't know that can be related.

PR #1317, opened 2017-08-21
REL-1438: user target types

swalker2m:
The target environment currently has a list of `SPTarget` "user" targets.  This update creates a `UserTarget` tuple that pairs a `UserTarget.Type` with an `SPTarget` to give it a meaning.  We don't do anything with this meaning yet but it is still useful because it clearly indicates the purpose of each user target.  The new types are "Blind Offset", "Off Axis", "Tuning Star" and "Other" (which renders as the familiar old "User").  We need a catchall like "Other" if for no other reason to convert existing programs.

A large part of this update went into reformatting the tool buttons that appear to the left of the image in the TPE.  Since there are now more options for creating targets, they needed to be rearranged into two columns.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/UserTarget.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/UserTarget.java
@@ -1,0 +1,70 @@
+package edu.gemini.spModel.target.env;
+
+import edu.gemini.shared.util.immutable.ImOption;
+import edu.gemini.shared.util.immutable.Option;
+import edu.gemini.spModel.target.SPTarget;
+
+import java.io.Serializable;
+import java.util.Objects;
+import java.util.stream.Stream;
+
+/**
+ * Tuple of a user SPTarget and its type.
+ */
+public final class UserTarget implements Serializable {
+
+    /**
+     * Enumeration of the usages of additional non-guide star targets in a
+     * target environment.
+     */
+    public enum Type {
+        blindOffset("Blind-offset"),
+        offAxis("Off-axis"),
+        tuning("Tuning Star"),
+        other("User");
+
+        public final String displayName;
+
+        private Type(String displayName) {
+            this.displayName = displayName;
+        }
+
+        public static Option<Type> fromString(String s) {
+            return ImOption.fromOptional(
+                    Stream.of(values())
+                            .filter(t -> t.name().equals(s))
+                            .findFirst()
+            );
+        }
+    }
+
+    public final Type type;
+    public final SPTarget target;
+
+    public UserTarget(Type type, SPTarget target) {
+        if (type == null) throw new NullPointerException("type cannot be null");
+        if (target == null) throw new NullPointerException("target cannot be null");
+
+        this.type   = type;
+        this.target = target;
+    }
+
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        UserTarget that = (UserTarget) o;
+        return Objects.equals(type, that.type) &&
+                Objects.equals(target, that.target);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(type, target);
+    }
+
+    public UserTarget cloneTarget() {
+        return new UserTarget(type, target.clone());
+    }
+}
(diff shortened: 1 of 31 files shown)


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/tpe/ImageCatalogPanel.scala
```diff
@@ -111,19 +112,28 @@ final class ImageCatalogPanel(imageDisplay: CatalogImageDisplay) {
...
+          contents ++= catalogRows.map { r => row(r.button, r.feedback) }
+          catalogRows.foreach { r =>
+            buttonGroup.add(r.button.peer)
+            r.button.reactions += {
+              case ButtonClicked(_) => requestImage(r.feedback.catalog)
+            }
+          }
+        },
+        Center
+      )
```
swalker2m: This should probably be done with the Mig layout library since the rest of it was done that way.  Suggestions for how to do that or PRs editing this are welcome.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/UserTarget.java
```diff
@@ -0,0 +1,70 @@
...
+import edu.gemini.spModel.target.SPTarget;
+
+import java.io.Serializable;
+import java.util.Objects;
+import java.util.stream.Stream;
+
+/**
+ * Tuple of a user SPTarget and its type.
+ */
+public final class UserTarget implements Serializable {
```
cquiroz: Should we add a `serialVersionUID`?

bundle/jsky.app.ot/src/main/java/jsky/app/ot/tpe/feat/TpeTargetPosFeature.java
```diff
@@ -26,11 +27,59 @@
...
+            final TargetObsComp obsComp = getTargetObsComp();
+            if (obsComp == null) return;
+
+            final Option<SiderealTarget> targetOpt = tme.target;
+            final SPTarget userPos = targetOpt.map(SPTarget::new).getOrElse(() -> {
+                final double ra  = tme.pos.ra().toDegrees();
+                final double dec = tme.pos.dec().toDegrees();
+                // No SkyObject info is present so we use the old way of creating
+                // a target from a mouse event using only the coordinates.
+                return new SPTarget(ra, dec);
```
cquiroz: Will this return get out of `create` or only the inner function?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/tpe/ImageCatalogPanel.scala
cquiroz: I can try in another PR. It is hard to see it with just a diff

swalker2m:
Here's a before (left) and after (right).  I wouldn't try to claim there's an improvement other than being able to fit more buttons in less vertical space.

tpe-buttons

cquiroz:
looks very good te me

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/UserTarget.java
swalker2m: I'm not sure it matters since we don't try to maintain binary compatibility across releases.  If you do add a `serialVersionUID` but forget to update it after making an incompatible change the error is less clear.

cquiroz: oh, I thought we did set the version on the other classes. `Serialization` scares me

bundle/jsky.app.ot/src/main/java/jsky/app/ot/tpe/feat/TpeTargetPosFeature.java
swalker2m: Only out of the `Supplier` implementation.  It needs a `return` if I understand it correctly because it is multiple lines.

I didn't pay much attention to this code, which I moved up from lower down in the original source.  I did add the `UserTarget.Type` parameter to the constructor  and have it make appropriately typed user targets.  Now that I'm looking at it, I see that it is deferring the creation of the new `SPTarget` to when/if it is necessary by using a `java.util.function.Supplier`.

PR #1319, opened 2017-08-23
P1 backend rewrite proposal

jdnavarro:
This PR is not intended for merging but to review the steps I was considering taking for the p1 backend servers rewrite.

The main short term goals of the rewrite are:

- Replace cgi-bin Perl scripts with something easier to maintain.
- Get rid of the p1-monitor. The backend should take care of sending the emails.

I won't start right away because I have other tasks with more priority. The TODO list is written to help me context switch easily to this project whenever I can. So the steps are not fully explained but I think it's still possible to guess what's the intention.

### DIFF
--- a/bundle/edu.gemini.p1backend/TODO.md
+++ b/bundle/edu.gemini.p1backend/TODO.md
@@ -1,0 +1,50 @@
+# Short term
+
+Something that improves the situation now. Nothing for the future. A dead-end by
+design. Something to keep chugging away until we do it right. It shouldn't
+take more work than 2 weeks.
+
+1. Code lives in ocs.
+1. URLs:
+   - Receiver of P1 submissions: phase1.gemini.edu/receive/.../...
+   - Viewer of P1 submissions: phase1.gemini.edu/fetch/.../...
+1. Each backend a different path like now.
+1. Configuration file for each backend (that file is read with every new
+   submission). Used for prepending semester to Ids, close deadline, etc.
+1. Id counter from previous file names (hashes discussed but discarded).
+1. Different backend for joints? If not possible use hard links.
+1. Generate PDFs.
+1. Stick together the XML and PDFs in tarballs.
+1. Send notifications *after* everything is successful. Consult configuration
+   file for emails.
+1. Multiple p1 versions:
+   a. When an XML is sent with an older PIT version, try to migrate to the latest
+      p1 model while parsing (as PIT does) and serialize it using the latest
+      version.
+   b. If there is a problem with the above, just change the Ids and add the
+      required tags by parsing directly the XML with an XML parsing library
+      (scala-xml + parsing combinators).
+1. Proposal viewer. XML stored in different directories with its own set of
+   HTTP basic authentication credentials. SSL should be possible.
+   a. Use Jetty to serve both the proposal receiver (the http4s WAR) and the
+      static files.
+   b. Use Apache/NGINX as reverse proxy like now and run the proposal receiver
+      on its on its own (http4s blaze) bound to 127.0.0.1.
+1. Deployment: Somehow try to attach it to the the release of PIT. Manually:
+   create WAR in development machine, copy WAR to phase1.gemini.edu, reload
+   jetty. But still not clear that's the easiest and safest release method.
+
+
+# Long term
+
+1. Code lives in ocs3.
+1. Store the proposals in db, not XML. We'd still create XML dumps for 3rd party
+   utilities.
+1. Create doobie DAOs for current p1 model or make gem support p1 features.
+1. Proper authentication: Use Gemini's Active directory.
+1. Proper authorization: Proposals with permissions, NGOs/Gemini personnel will
+   have their own set granular permissions (roles?).
+1. Web forms: configuration stored in db, close deadline by NGO, proposal
+   filters, etc.
+1. Docker for deployment.
+1. Foothold for ITAC backend rewrite.
--- a/bundle/edu.gemini.p1backend/TODO.md
+++ b/bundle/edu.gemini.p1backend/TODO.md
@@ -25,14 +25,14 @@ take more work than 2 weeks.
       required tags by parsing directly the XML with an XML parsing library
       (scala-xml + parsing combinators).
 1. Proposal viewer. XML stored in different directories with its own set of
-   HTTP basic authentication credentials. SSL should be possible.
-   a. Use Jetty to serve both the proposal receiver (the http4s WAR) and the
-      static files.
+   HTTP basic authentication credentials. SSL should still be possible.
+   a. Use Jetty to serve both the proposal receiver (the http4s WAR, osgi?) and
+      the static files.
    b. Use Apache/NGINX as reverse proxy like now and run the proposal receiver
-      on its on its own (http4s blaze) bound to 127.0.0.1.
-1. Deployment: Somehow try to attach it to the the release of PIT. Manually:
-   create WAR in development machine, copy WAR to phase1.gemini.edu, reload
-   jetty. But still not clear that's the easiest and safest release method.
+      on its own (http4s blaze) bound to 127.0.0.1.
+1. Deployment: Somehow try to attach it to the release of PIT. Manually: create
+   WAR in development machine, copy WAR to phase1.gemini.edu, reload jetty. But
+   still not clear that's the easiest or safest release method.
 
 
 # Long term
@@ -48,3 +48,66 @@ take more work than 2 weeks.
    filters, etc.
 1. Docker for deployment.
 1. Foothold for ITAC backend rewrite.
+
+
+# To Find out
+
+1. PIT multipart submissions and support in http4s.
+1. Will Gem ids be used for phase I in the future?
+
+
+# Ids
+
+We are open to change the id scheming, but whatever scheme we choose, it should
+*look* the same in the long term: i.e ordered vs unordered, alphanumeric vs
+numeric only, etc.
+
+
+## Counter
+
+### Pros
+
+- Similar to what we have now. Prevents unforeseeable issues.
+- The id has some meaning: proposal 2 arrived right after proposal 1.
+
+
+### Cons
+
+- Have to deal with storage persistency.
+- Have to deal with shared data in what otherwise would be totally independent
+  concurrent proposal submission requests.
+
+## Hashes
+
+ - Non cryptographic hash like Murmurhash (available in all major languages and
+   easy to implement if not available). 32 bits length should be good enough to
+   avoid collisions (TODO: calculate exact probability to be certain).
+ - Encoding 32 bit hashes:
+   - Base58: only numbers and printed characters that can't be mistakenly
+     mistyped. Examples:
+     - 3fq91a
+     - 3fmnCW
+     - 2SJWB1
+   - Base10, only numbers. Examples:
+     - 4294967296
+     - 2147483647
+
+
+### Pros
+
+- Deterministic. Id intrinsic to the data coming with the proposal. It can be
+  calculated cheaply anywhere without depending in anything else. No need to
+  persist anything to disk or share data between 2 simultaneous requests.
+- True identity. If 2 proposals have the same id they are the same proposal. No
+  need to care about accidental misassignments of ids due to HTTP resubmissions
+  overwrites or whatnot. The proposal receiver endpoint would be truly
+  [idempotent](http://www.restapitutorial.com/lessons/idempotency.html) which
+  allows to forget about having to deal with the inherent annoyances of
+  non-idempotent HTTP methods.
+
+
+### Cons
+
+- Id will change if content of proposal is modified.
+- Characters in the id may create problems in 3rd party scripts.
+- If only numbers are used the ids may be too long.
--- a/bundle/edu.gemini.p1backend/build.sbt
+++ b/bundle/edu.gemini.p1backend/build.sbt
@@ -1,0 +1,26 @@
+import OcsKeys._
+
+// note: inter-project dependencies are declared in projects/OcsBundle.scala
+
+name := "edu.gemini.p1backend"
+
+// version set in ThisBuild
+
+libraryDependencies ++= Seq(
+  "org.typelevel" %% "cats-core" % "1.0.0-MF"
+  // The first one to use `cats-effects`. If not released yet, point to the
+  // git repo/branch in `dependsOn` in OcsBundle.scala.
+  // "org.http4s" %% "http4s-dsl" % "0.18.0-SNAPSHOT"
+  )
+
+osgiSettings
+
+ocsBundleSettings
+
+OsgiKeys.bundleActivator := None
+
+OsgiKeys.bundleSymbolicName := name.value
+
+OsgiKeys.dynamicImportPackage := Seq("")
+
+OsgiKeys.exportPackage := Seq("edu.gemini.p1backend")
--- a/project/OcsBundle.scala
+++ b/project/OcsBundle.scala
@@ -734,5 +734,11 @@ trait OcsBundle {
   lazy val bundle_edu_gemini_ui_miglayout =
     project.in(file("bundle/edu.gemini.ui.miglayout"))
 
+  lazy val bundle_edu_gemini_p1backend =
+    project.in(file("bundle/edu.gemini.p1backend")).dependsOn(
+      bundle_edu_gemini_model_p1,
+      bundle_edu_gemini_model_p1_pdf,
+      bundle_edu_gemini_model_p1_submit
+    )
 }
 


bundle/edu.gemini.p1backend/TODO.md
```diff
@@ -0,0 +1,50 @@
...
+Something that improves the situation now. Nothing for the future. A dead-end by
+design. Something to keep chugging away until we do it right. It shouldn't
+take more work than 2 weeks.
+
+1. Code lives in ocs.
+1. URLs:
+   - Receiver of P1 submissions: phase1.gemini.edu/receive/.../...
+   - Viewer of P1 submissions: phase1.gemini.edu/fetch/.../...
+1. Each backend a different path like now.
+1. Configuration file for each backend (that file is read with every new
```
cquiroz: any reason to prefer files than to a db. We probably need a transactional db for the counters regardless

bundle/edu.gemini.p1backend/TODO.md
```diff
@@ -0,0 +1,50 @@
+# Short term
+
+Something that improves the situation now. Nothing for the future. A dead-end by
+design. Something to keep chugging away until we do it right. It shouldn't
+take more work than 2 weeks.
+
+1. Code lives in ocs.
+1. URLs:
+   - Receiver of P1 submissions: phase1.gemini.edu/receive/.../...
```
cquiroz: note that this url isn't user facing but still it is a good idea to make them restish

bundle/edu.gemini.p1backend/TODO.md
```diff
@@ -0,0 +1,50 @@
...
+   - Viewer of P1 submissions: phase1.gemini.edu/fetch/.../...
+1. Each backend a different path like now.
+1. Configuration file for each backend (that file is read with every new
+   submission). Used for prepending semester to Ids, close deadline, etc.
+1. Id counter from previous file names (hashes discussed but discarded).
```
cquiroz: The counter should increase monotonically with submission per partner. We should rollback the counter on failed submissions

bundle/edu.gemini.p1backend/TODO.md
```diff
@@ -0,0 +1,50 @@
...
+1. Different backend for joints? If not possible use hard links.
```
cquiroz: i'm not sure what you mean. i think we could try to do only one submission for a joint and get an id per partner instead

bundle/edu.gemini.p1backend/TODO.md
```diff
@@ -0,0 +1,50 @@
+# Short term
+
+Something that improves the situation now. Nothing for the future. A dead-end by
+design. Something to keep chugging away until we do it right. It shouldn't
+take more work than 2 weeks.
+
+1. Code lives in ocs.
```
cquiroz: ocs3

bundle/edu.gemini.p1backend/TODO.md
```diff
@@ -0,0 +1,50 @@
...
+   - Receiver of P1 submissions: phase1.gemini.edu/receive/.../...
+   - Viewer of P1 submissions: phase1.gemini.edu/fetch/.../...
+1. Each backend a different path like now.
+1. Configuration file for each backend (that file is read with every new
+   submission). Used for prepending semester to Ids, close deadline, etc.
+1. Id counter from previous file names (hashes discussed but discarded).
+1. Different backend for joints? If not possible use hard links.
+1. Generate PDFs.
+1. Stick together the XML and PDFs in tarballs.
+1. Send notifications *after* everything is successful. Consult configuration
```
cquiroz: Note that there are several emails to send. One to the NGO and there are request to send them to the PIs too

bundle/edu.gemini.p1backend/TODO.md
```diff
@@ -0,0 +1,50 @@
...
+1. URLs:
+   - Receiver of P1 submissions: phase1.gemini.edu/receive/.../...
+   - Viewer of P1 submissions: phase1.gemini.edu/fetch/.../...
+1. Each backend a different path like now.
+1. Configuration file for each backend (that file is read with every new
+   submission). Used for prepending semester to Ids, close deadline, etc.
+1. Id counter from previous file names (hashes discussed but discarded).
+1. Different backend for joints? If not possible use hard links.
+1. Generate PDFs.
+1. Stick together the XML and PDFs in tarballs.
```
cquiroz: The tarball should be downloadable for NGOs too

bundle/edu.gemini.p1backend/TODO.md
```diff
@@ -0,0 +1,50 @@
...
+1. Send notifications *after* everything is successful. Consult configuration
+   file for emails.
+1. Multiple p1 versions:
```
cquiroz: This is not correct. We cannot upgrade proposals from older PITs on the backend. each backend needs a semester/version and will only accept that version. The tricky part here is that we should accept multiple versions. Given we can't have multiple versions of model.p1 probably we need to have multiple backends with a front end.

bundle/edu.gemini.p1backend/TODO.md
```diff
@@ -0,0 +1,50 @@
...
+   a. When an XML is sent with an older PIT version, try to migrate to the latest
+      p1 model while parsing (as PIT does) and serialize it using the latest
+      version.
+   b. If there is a problem with the above, just change the Ids and add the
+      required tags by parsing directly the XML with an XML parsing library
+      (scala-xml + parsing combinators).
```
cquiroz: it is simpler to modify directly the p1 proposal in code rather than going to the XML

bundle/edu.gemini.p1backend/TODO.md
```diff
@@ -0,0 +1,50 @@
...
+1. Proposal viewer. XML stored in different directories with its own set of
+   HTTP basic authentication credentials. SSL should be possible.
```
cquiroz: i'd rather not use HTTP Basic and instead use the same auth infrastructure than the Seqexec

bundle/edu.gemini.p1backend/TODO.md
```diff
@@ -0,0 +1,50 @@
...
+   a. Use Jetty to serve both the proposal receiver (the http4s WAR) and the
+      static files.
+   b. Use Apache/NGINX as reverse proxy like now and run the proposal receiver
```
cquiroz: It should be ok to do this especially with multiple backends. Ideally we should be able to have some tools to enable/disable routes depending on what's open

cquiroz: Note that we have a request to let NGOs control opening/closing the backend, thus we may need to figure out how to let users change this with a GUI

bundle/edu.gemini.p1backend/TODO.md
```diff
@@ -0,0 +1,50 @@
...
+      on its on its own (http4s blaze) bound to 127.0.0.1.
+1. Deployment: Somehow try to attach it to the the release of PIT. Manually:
```
cquiroz: As you say this is TBD, the standalone/war route is ok though we may want to go the docker route

bundle/edu.gemini.p1backend/TODO.md
```diff
@@ -0,0 +1,50 @@
...
+   create WAR in development machine, copy WAR to phase1.gemini.edu, reload
+   jetty. But still not clear that's the easiest and safest release method.
+
+
+# Long term
+
+1. Code lives in ocs3.
+1. Store the proposals in db, not XML. We'd still create XML dumps for 3rd party
```
cquiroz: I don't think this is necessary, we only need to track the proposal id and the file location. XML in db is the reason ITAC has had so many troubles

jdnavarro:
> Thank you for putting forward this proposal/plan. I've always envisioned the backends to have a UI for NGOs to use. It should be fairly simple but of course that adds effort. Have you considered that, or is the plan to only do the bare minimum and do admin directly via config/db?

The UI could go in the long-term rewrite. The short-term rewrite should be in functionality identical to what we have now plus some fixes. But they are not mutually exclusive.

> I think we should have a little more details about supporting multiple versions of the backends at the same time. That is likely the most difficult part of this job so we should do some level design on how would that work

I have a vague idea of what the problem is but I'm not exactly sure what happens in detail and how the PIT deals with different versions now. So my idea was to naively try to fix it by upgrading to the latest version (it's not that much work anyway) and experience first hand the problem so that I can learn more from it.

bundle/edu.gemini.p1backend/build.sbt
```diff
@@ -0,0 +1,26 @@
...
+
+name := "edu.gemini.p1backend"
+
+// version set in ThisBuild
+
+libraryDependencies ++= Seq(
+  "org.typelevel" %% "cats-core" % "1.0.0-MF"
+  // The first one to use `cats-effects`. If not released yet, point to the
+  // git repo/branch in `dependsOn` in OcsBundle.scala.
+  // "org.http4s" %% "http4s-dsl" % "0.18.0-SNAPSHOT"
```
jdnavarro: The problem is that 0.16, 0.17 and 0.18 have all major changes in the exposed API. I'm positive 0.18 is going to be more definitive but the problem is not that much in http4s itself but in aligning the dependency versions in our code with http4s (cats, cats-effect, fs2).

By using the 0.18 dependencies we can have some certainty we won't have to rewrite the whole app in 6 months, I hope 🤞.

But even if we finish it very soon, which I don't think it'll happen because it's not that urgent, we won't put it in production until 2018B, minimum. Hopefully by then all the 3rd party APIs on which we depend will be settled.

jdnavarro: What other projects are in ocs depend on http4s?

bundle/edu.gemini.p1backend/TODO.md
jdnavarro: I meant ocs for the short term rewrite. The more ambitious rewrite could go into ocs3.

bundle/edu.gemini.p1backend/TODO.md
jdnavarro: You are right, I didn't consider race conditions. We could probably have the counter in a locked file that is read when starting the app or something but it may be not worth the trouble. I'll think more about this but a full blown database for just for a counter seems overkill to me.

bundle/edu.gemini.p1backend/TODO.md
jdnavarro: The more we think about the counter the more problems it brings and I'm still not convinced it's a must have feature.

bundle/edu.gemini.p1backend/TODO.md
jdnavarro: I didn't explain it well enough. Apparently having the same exact proposal duplicated with different Ids creates problems down the line, so in order to deal with the permissions we could either have different directories for the joint proposals with special permissions or just link them at the file system level in each directory backend. Not pretty, but should do it for the time being.

cquiroz:
I was a bit out of context. i thought we wanted to do this in ocs3 that's why some of my comments refer e.g. to conflicts with the seqexec

bundle/edu.gemini.p1backend/TODO.md
cquiroz: Right

bundle/edu.gemini.p1backend/TODO.md
cquiroz: Well it is not just a counter, it is a transactional counter, it should rollback if we have an error. That's why i was leaning to a db

bundle/edu.gemini.p1backend/TODO.md
cquiroz: What feature? having a counter? We need to have a unique id per proposal/partner ideally increasing. Any other ideas so that NGOs could identify uniquely a proposal?

bundle/edu.gemini.p1backend/TODO.md
cquiroz: if we are not in ocs3 perhaps Basic auth will d

bundle/edu.gemini.p1backend/TODO.md
jdnavarro: It was my understanding that the PIT is backwards compatible with previous proposals created with older PIT versions. That is, it could open proposal from older p1 versions and automatically upgrades it to the newest one. The plan in the backend was to use the same serializer/deserializer that the PIT uses.

Of course, I'm guessing it won't work all that smoothly but this is what I meant when I said that I wanted to see first hand what's exactly is the problem when trying it.

bundle/edu.gemini.p1backend/TODO.md
cquiroz: same as before if in ocs we should just do a standalone app like itc

bundle/edu.gemini.p1backend/TODO.md
cquiroz: yes the PIT will open older proposals and move them to today's version. But a PIT version A can only submit to a backend version A. However we could have multiple versions of the PIT live at the same time, this is because during a semester you can still receive proposals for special programs from the last semester

bundle/edu.gemini.p1backend/TODO.md
jdnavarro: This was plan B if dealing with multiple versions is not straightforward. The changes in the XML are not that much and I'd rather deal with it than having to maintain, configure and deploy multiple backend servers.

In the end maybe there is no other way than to have multiple real backend servers running but I'd rather exhaust first all the easy possibilities first.

bundle/edu.gemini.p1backend/TODO.md
jdnavarro: I'd rather not do it too, but I'm just reusing the HTTP auth that is already in place and we want something quick. We'd have to register the NGOs and all that, right?

Of course, in the long term we should use what we have in the Seqexec.

bundle/edu.gemini.p1backend/TODO.md
jdnavarro: Apparently that's no longer so urgent or straightforward. Gemini personnel would like to also have control over when to to close the deadline for each backend.

But again, once we have more than 1-2 week to rewrite it, we could carefully design a web form to handle all those scenarios.

bundle/edu.gemini.p1backend/TODO.md
jdnavarro: That WAR thing is just the default after asking around around in the office, but yes, even for a short project I think we should have something (even RPMs may be too much work...) that lets us forget about having to deploy the backend server and restart all manually once in a while.

bundle/edu.gemini.p1backend/TODO.md
jdnavarro: The scientist have several scripts and utilities that rely on those XML files. Not offering the XMLs anymore is not an option unless we replace all the utilities they have ourselves.

cquiroz: i meant that we store the xml as file rather than in the db. But yes we certainly need to keep the xml available

bundle/edu.gemini.p1backend/TODO.md
cquiroz: In the ocs we don't produce WARs really. What we do is to produce standalone applications with osgi that contain a web server. The closest case would be the itc app

bundle/edu.gemini.p1backend/TODO.md
cquiroz: Makes sense

bundle/edu.gemini.p1backend/TODO.md
jdnavarro: The idea was to increment the counter only after making sure everything was right, the file is stored to disk with the right Id, the PDFs have been created, the emails have been sent, etc.

But now that you mention it there may be edge cases where we would still need a rollback regardless. Which makes me wonder even more why the counter is so necessary.

bundle/edu.gemini.p1backend/TODO.md
jdnavarro: Yes, I'm referring to the counter. Is a counter the only way to assign ids? They could be random, like is done almost everywhere that I know of where ids are generated.

Why do they need to be in order really? I mean, it's nice that you can get an idea of which proposal came first but of all the people I asked about the backend servers you are the only one who has told me that's an important feature.

There is a ticket where they mention gaps in the ids but that's only confusing because an order is expected but if you don't expect an order the problem is solved.

For me no order is less work but if you tell me you've been told by the users that they need that ordering of ids, I'll take your word for it, and work something out.

bundle/edu.gemini.p1backend/TODO.md
jdnavarro: I understood that having multiple ids for the same proposal is a problem now. What are the multiple ids used for now?

bundle/edu.gemini.p1backend/TODO.md
jdnavarro: Alright, so please help me understand then. A is the newest version, and B is an older version What's the problem on making the backend accept B versions from PIT B and store them as A once accepted?

Do the accepted proposals need to be opened at some point by an older PIT version?

jdnavarro:
Now I'm realizing I didn't explain anywhere that I was thinking in 2 rewrites, one short of one week, just to quickly replace what we have now, and the other more permanent using what we have in `ocs3` where we do it for real.

I'm all up for the replacement for real, but I'd need more than 2 weeks to do it.

bundle/edu.gemini.p1backend/TODO.md
cquiroz: That has been my understanding so far. If users don't need a meaningful id that's fine and then we could remove the need for a db

jdnavarro: I've been thinking more about this id counting and the thing is that even we are promised that it doesn't matter for the users, we won't be certain until the new ids are actually being used. So in theory if we keep counting as now we could avoid any unforeseeable problems with the new ids.

I'll give it some thought and try to come up with a simple way to keep a counter before ruling it out.

cquiroz:
after some discussions I now realize that I was thinking in terms of the full rewrite that we had planned with several enhancements and in the context of ocs3

Instead, this proposal is for a strict replacement of the current backend/p1-monitor and in that sense some comments are out of context

It maybe worth to start by a one-to-one replacement without changing other assumptions like id format or how to handle joints but without going into too much complications about it

jdnavarro:
It's still not clear whether is worth making a short or a long term rewrite of the backend servers.

PR #1322, opened 2017-08-28
REL-2194: Create Notice Severity in QPT for messages that are neither warning or info

jdnavarro:
Initially we thought of downgrading some of the messages from `Warning` to `Info`, but it turns out that `Info` was already taken by QC markers and those need different treatment; they shouldn't show up in the HTML reports.

Now when an Allocation does have `Notice` messages it's still shown in the elevation plots in green like an `Info` message, however in the web reports a `Notice` is highlighted in *blue*.

`Notice` messages have a similar, but still slightly different, icon to `Warning` icons, which makes it more noticeable than `Info` but not as important as real `Warning`s which are shown in *yellow* in the elevation plot.

I included these messages as `Notice`, but please let me know of any other that should go into this new category:

- Parallactic angle
- Blind offsetting
- Non-sidereal


tpolecat:
Looks good. `bundle/edu.gemini.qpt.client/src/main/resources/edu/gemini/qpt/ui/util/warn_tsk_blue.gif` appears to be yellow ... ?

jdnavarro:
Looks good. `bundle/edu.gemini.qpt.client/src/main/resources/edu/gemini/qpt/ui/util/warn_tsk_blue.gif` appears to be yellow ... ?

Well... the color refers to the symbol inside it. I tried first coloring the whole but it didn't look that good, it seems editing icons is not my strong suit.

I also looked in the Eclipse icons repo but couldn't find anything not too awkward for a `Notice` message and that aligned properly in the row where the icon is shown.

tpolecat:
Oh that's very subtle in the diff ;-)

👍

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
```diff
@@ -189,25 +189,25 @@ public void propertyChange(PropertyChangeEvent evt) {
...
 					break;
 
-				case IQ_UQUAL:					
+				case IQ_UQUAL:
 					mm.addMarker(false, this, Severity.Error, "Variant IQ is under-qualified for this observation.", v, a);
 					break;
 
 				case INACTIVE:
-					mm.addMarker(false, this, Severity.Warning, "Science program is inactive.", v, a);
+					mm.addMarker(false, this, Severity.Error, "Science program is inactive.", v, a);
```
jdnavarro: Change confirmed by @andrewwstephens

PR #1323, opened 2017-08-28
Updated p1-monitor configuration

cquiroz:
Forgot to update the p1-monitor configuration for the PIT 2018A release


jdnavarro:
Is this specific of the candidate release? Will those test proposals be stored in the same host used for production?

cquiroz:
We usually have only one configuration for the p1-monitors so when it comes the time to test we change the path. This could probably be improved

The hosts is the same for test and real proposals. but different dirs

PR #1324, opened 2017-08-29
REL-2838 Make PWFS2 p2checker agree with the PWFS2 AGS

jdnavarro:
It turns out the PWFS2 p2checker was ruling out the guide stars that where between the inner limit and the inner boundary which should have been picked, as the AGS is already doing.

We also introduced an enumeration class for the results of the PWFS2 p2checker so that it's easier to understand.

### DIFF
--- a/bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/pwfs/PwfsRule.java
+++ b/bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/pwfs/PwfsRule.java
@@ -45,14 +45,13 @@ public class PwfsRule implements IRule {
                 try {
                     pwfs.checkBoundaries(primary, ctx).foreach(bs -> {
                         switch (bs) {
-                            case inside:
+                            case vignetting:
                                 problems.addWarning(PREFIX + "WARN", "The " + pwfs.getKey() + WARN, targetNode);
                                 break;
-                            case innerBoundary:
+                            case inRange:
                                 // OK, this is where it should be: between inner and outer limits
                                 break;
-                            case outerBoundary:
-                            case outside:
+                            case outOfRange:
                                 problems.addError(PREFIX + "ERROR", "The " + pwfs.getKey() + ERROR, targetNode);
                                 break;
                         }
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/obsComp/PwfsGuideProbe.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/obsComp/PwfsGuideProbe.java
@@ -306,7 +306,7 @@ public enum PwfsGuideProbe implements ValidatableGuideProbe, OffsetValidatingGui
             // check positions against corrected outer patrol field bounds
             return getCorrectedPatrolField(ctx).map(pf -> {
                 final BoundaryPosition bp = pf.checkBoundaries(coords, baseCoordinates, positionAngle, sciencePositions);
-                if (bp != BoundaryPosition.inside) {
+                if ((bp != BoundaryPosition.inside) && (bp != BoundaryPosition.innerBoundary)) {
                     return BoundaryPosition.outside;
                 }
 
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/obsComp/PwfsGuideProbe.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/obsComp/PwfsGuideProbe.java
@@ -19,7 +19,6 @@ import java.awt.geom.Ellipse2D;
 import java.awt.geom.GeneralPath;
 import java.util.Set;
 
-
 /**
  * The Peripheral guide probes.
  */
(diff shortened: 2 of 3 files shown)


bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/pwfs/PwfsRule.java
```diff
@@ -45,14 +45,14 @@ public IP2Problems check(ObservationElements elements)  {
                 try {
                     pwfs.checkBoundaries(primary, ctx).foreach(bs -> {
                         switch (bs) {
-                            case inside:
+                            case nearBoundary:
                                 problems.addWarning(PREFIX + "WARN", "The " + pwfs.getKey() + WARN, targetNode);
```
swalker2m: So, `checkBoundaries` was returning `inside` to mean inside the inner vignetting limit.  It looks like you've renamed this `nearBoundary` but the warning message is still about vignetting.  I think you want the `vignetting` enum value here.

bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/pwfs/PwfsRule.java
```diff
@@ -45,14 +45,14 @@ public IP2Problems check(ObservationElements elements)  {
...
                                 break;
-                            case innerBoundary:
+                            case inRange:
                                 // OK, this is where it should be: between inner and outer limits
                                 break;
-                            case outerBoundary:
-                            case outside:
+                            case vignetting:
```
swalker2m: Vignetting should trigger a warning, not an error.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/guide/BoundaryCheck.java
```diff
@@ -0,0 +1,6 @@
+package edu.gemini.spModel.guide;
+/**
+   * Enumeration of possible results when checking boundaries.
+   */
+  public enum BoundaryCheck{inRange, nearBoundary, outOfRange, vignetting}
```
swalker2m: This should be specific to PWFS since it is the only one with a vignetting boundary.  I would put it in the same package as the `PwfsGuideProbe` and call it something like `PwfsProbeRangeArea` to tie it to PWFS.  Maybe we only include `nearBoundary` if we're going to distinguish that region for a distinct warning.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/obsComp/PwfsGuideProbe.java
```diff
@@ -284,41 +284,47 @@ public GuideStarValidation validate(Coordinates coords, ObsContext ctx) {
...
             final Angle positionAngle = ctx.getPositionAngleJava();
             final Set<Offset> sciencePositions = ctx.getSciencePositions();
 
             // check positions against corrected outer patrol field bounds
             return getCorrectedPatrolField(ctx).map(pf -> {
                 final BoundaryPosition bp = pf.checkBoundaries(coords, baseCoordinates, positionAngle, sciencePositions);
-                if (bp != BoundaryPosition.inside) {
-                    return BoundaryPosition.outside;
+                if ((bp != BoundaryPosition.inside) && (bp != BoundaryPosition.innerBoundary)) {
+                    return BoundaryCheck.nearBoundary;
```
swalker2m: This should continue to be `outside`, I believe.  If it isn't one of the inside positions, it's outside.  If we need to distinguish `innerBoundary` though we could switch on `bp` here I guess.  I can't remember the syntax but something like:

```java
switch (bp) {
    case BoundaryPosition.outside:
    case BoundaryPosition.outerBoundary:
        return outOfRange;
    case BoundaryPosition.innerBoundary:
        return nearBoundary;
    case BoundaryPosition.inner:
       // here we have to check whether we're vignetting to decide between inRange and vignetting
      ....
}

```

bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/pwfs/PwfsRule.java
jdnavarro: I thought it was the other way around 🤦‍♂️. I'll fix it.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/guide/BoundaryCheck.java
jdnavarro: I'll remove `nearBoundary`, it's coming from my misunderstanding of how to handle vignetting.

swalker2m:
Maybe instead of another `enum` that is nearly the same as `BoundaryPosition`, maybe just have the phase 2 check do the vignetting check if the result is `inner`?

jdnavarro:
> Maybe instead of another enum that is nearly the same as BoundaryPosition, maybe just have the phase 2 check do the vignetting check if the result is inner?

A distinct enum for the PWFS2 p2checker result made the it easier to understand for me, but I could change it if you really think I should. Now that is already written is not a big deal to remove the new enum.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/obsComp/PwfsGuideProbe.java
```diff
@@ -284,41 +283,48 @@ public GuideStarValidation validate(Coordinates coords, ObsContext ctx) {
...
+                        final double minLimit = getVignettingClearance(ctx).toArcsecs().getMagnitude();
+                        final Ellipse2D e = new Ellipse2D.Double(-minLimit, -minLimit, minLimit * 2.0, minLimit * 2.0);
+                        final PatrolField p = getCorrectedPatrolField(new PatrolField(e, e, e), ctx);
+                        if (p.anyInside(coords, baseCoordinates, positionAngle, sciencePositions)) {
+                            return PwfsProbeRangeArea.vignetting;
+                        }
                 }
-                return BoundaryPosition.innerBoundary;
-            }).getOrElse(BoundaryPosition.outside);
+                return PwfsProbeRangeArea.inRange;
```
jdnavarro: I tried to include this return statement in the `inside`/`innerBoundary` switch case but it didn't work.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/obsComp/PwfsGuideProbe.java
```diff
@@ -284,41 +283,48 @@ public GuideStarValidation validate(Coordinates coords, ObsContext ctx) {
...
-                final Ellipse2D e = new Ellipse2D.Double(-minLimit, -minLimit, minLimit * 2.0, minLimit * 2.0);
-                final PatrolField p = getCorrectedPatrolField(new PatrolField(e, e, e), ctx);
-                if (p.anyInside(coords, baseCoordinates, positionAngle, sciencePositions)) {
-                    return BoundaryPosition.inside;
+                switch (bp) {
+                    case outside:
+                    case outerBoundary:
+                        return PwfsProbeRangeArea.outOfRange;
+                    case inside:
+                    case innerBoundary:
```
swalker2m: It can't be both `innerBoundary` and in the vignetting area so you could short-circuit this case a bit and leave the vignetting check for `inside`.  On the other hand, it is rare that a guide star falls in that tiny `innerBoundary` area anyway.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/obsComp/PwfsGuideProbe.java
swalker2m: Perhaps the way to do it in Java is to declare a `final PwfsProbeRangeArea result;` before the switch statement, assign it in all cases inside the switch statement, and just have the one return outside.

swalker2m:
Seems like it should work. 👍

PR #1325, opened 2017-08-31
REL-3220: NPE on p1-monitor

cquiroz:
Last minute fix. There can be a race condition that makes the dir be created after update is called


cquiroz:
it is odd indeed

PR #1326, opened 2017-08-31
REL-1121: Add GMOS timing overheads

jdnavarro:
Adding the GMOS overheads was the primary goal but most of the changes in this PR are about the tooltip message over the timeline which now shows the configurations that didn't contribute to the maximum timings.

tooltip

The code for the tooltip is quite dirty but not I'm not sure how to do it more cleanly in Java.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gmos/InstGmosCommon.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gmos/InstGmosCommon.java
@@ -1281,15 +1281,6 @@ public abstract class InstGmosCommon<
     private static final CategorizedTime DHS_WRITE = CategorizedTime.fromSeconds(Category.DHS_WRITE, 10.0);
 
     public CategorizedTimeGroup calc(Config cur, Option<Config> prev) {
-        final ItemKey FPU_KEY = new ItemKey("instrument:fpu");
-        final double FPU_OVERHEAD = 60.0;
-
-        final ItemKey FILTER_KEY = new ItemKey("instrument:filter");
-        final double FILTER_OVERHEAD = 20.0;
-
-        final ItemKey DISPERSER_KEY = new ItemKey("instrument:disperser");
-        final double DISPERSER_OVERHEAD = 90.0;
-
         final Collection<CategorizedTime> times = new ArrayList<>();
 
         double exposureTime = ExposureCalculator.instance.exposureTimeSec(cur);
@@ -1305,14 +1296,20 @@ public abstract class InstGmosCommon<
         }
         times.add(CategorizedTime.fromSeconds(Category.EXPOSURE, exposureTime));
 
-        if (PlannedTime.isUpdated(cur, prev, FPU_KEY)) {
-            times.add(CategorizedTime.fromSeconds(Category.CONFIG_CHANGE, FPU_OVERHEAD, "FPU"));
+        if (PlannedTime.isUpdated(cur, prev, GmosCommonType.FPUnit.KEY)) {
+            times.add(CategorizedTime.fromSeconds(
+                    Category.CONFIG_CHANGE, GmosCommonType.FPUnit.CHANGE_OVERHEAD, "FPU")
+            );
         }
-        if (PlannedTime.isUpdated(cur, prev, FILTER_KEY)) {
-            times.add(CategorizedTime.fromSeconds(Category.CONFIG_CHANGE, FILTER_OVERHEAD, "Filter"));
+        if (PlannedTime.isUpdated(cur, prev, GmosCommonType.Filter.KEY)) {
+            times.add(CategorizedTime.fromSeconds(
+                    Category.CONFIG_CHANGE, GmosCommonType.Filter.CHANGE_OVERHEAD, "Filter")
+            );
         }
-        if (PlannedTime.isUpdated(cur, prev, DISPERSER_KEY)) {
-            times.add(CategorizedTime.fromSeconds(Category.CONFIG_CHANGE, DISPERSER_OVERHEAD, "Disperser"));
+        if (PlannedTime.isUpdated(cur, prev, GmosCommonType.Disperser.KEY)) {
+            times.add(CategorizedTime.fromSeconds(
+                    Category.CONFIG_CHANGE, GmosCommonType.Disperser.CHANGE_OVERHEAD, "Disperser")
+            );
         }
 
         double readoutTime = GmosReadoutTime.getReadoutOverhead(cur, getCustomROIs());
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gmos/InstGmosCommon.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gmos/InstGmosCommon.java
@@ -26,6 +26,7 @@ import edu.gemini.spModel.inst.ScienceAreaGeometry;
 import edu.gemini.spModel.inst.VignettableScienceAreaInstrument;
 import edu.gemini.spModel.obs.plannedtime.CommonStepCalculator;
 import edu.gemini.spModel.obs.plannedtime.ExposureCalculator;
+import edu.gemini.spModel.obs.plannedtime.PlannedTime;
 import edu.gemini.spModel.obs.plannedtime.PlannedTime.CategorizedTime;
 import edu.gemini.spModel.obs.plannedtime.PlannedTime.CategorizedTimeGroup;
 import edu.gemini.spModel.obs.plannedtime.PlannedTime.Category;
@@ -1280,6 +1281,17 @@ public abstract class InstGmosCommon<
     private static final CategorizedTime DHS_WRITE = CategorizedTime.fromSeconds(Category.DHS_WRITE, 10.0);
 
     public CategorizedTimeGroup calc(Config cur, Option<Config> prev) {
+        final ItemKey FPU_KEY = new ItemKey("instrument:fpu");
+        final double FPU_OVERHEAD = 60.0;
+
+        final ItemKey FILTER_KEY = new ItemKey("instrument:filter");
+        final double FILTER_OVERHEAD = 20.0;
+
+        final ItemKey DISPERSER_KEY = new ItemKey("instrument:disperser");
+        final double DISPERSER_OVERHEAD = 90.0;
+
+        final Collection<CategorizedTime> times = new ArrayList<>();
+
         double exposureTime = ExposureCalculator.instance.exposureTimeSec(cur);
 
         if (_useNS && isNodAndShuffleableObsType(cur)) {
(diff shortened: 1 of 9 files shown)


bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gmos/InstGmosCommon.java
```diff
@@ -1280,6 +1281,17 @@ public static boolean isNodAndShuffleableObsType(Config c) {
...
 
     public CategorizedTimeGroup calc(Config cur, Option<Config> prev) {
+        final ItemKey FPU_KEY = new ItemKey("instrument:fpu");
+        final double FPU_OVERHEAD = 60.0;
+
+        final ItemKey FILTER_KEY = new ItemKey("instrument:filter");
+        final double FILTER_OVERHEAD = 20.0;
+
+        final ItemKey DISPERSER_KEY = new ItemKey("instrument:disperser");
+        final double DISPERSER_OVERHEAD = 90.0;
```
jdnavarro: I guess these constants should go somewhere else but not sure where.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/obs/plannedtime/PlannedTime.java
```diff
@@ -223,17 +222,30 @@ public static CategorizedTimeGroup apply(CategorizedTime... col) {
...
+                }
+
+                if (cur == null) {
                     max.put(ct.category, ct);
+                } else if (cur.compareTo(ct) < 0) {
+                    max.put(ct.category, ct);
+                    losers.put(ct.category, currLosers.append(cur));
+                }
+                else {
+                    losers.put(ct.category, currLosers.append(ct));
```
jdnavarro: This is so dirty because I was missing some method in `TreeMap` to *mappend* values when the key is present, I guess there is a better way but didn't find it. Also I don't know how to handle those null cases better.

cquiroz: i don't think there is a way around on `TreeMap`. Possibly `ConcurrentMap` has a `putIfPresent/Missing` or something like that

bundle/jsky.app.ot/src/main/java/jsky/app/ot/editor/seq/ObserveTimeLineNode.java
```diff
@@ -90,20 +91,45 @@ public static long getDuration(PlannedTime pt, int step) {
         } else {
             buf.append("<p align=\"center\"><b>").append(SequenceTabUtil.shortDatasetLabel(plannedTime.sequence.getStep(step))).append("</b></p>");
 
-            buf.append("<table><tr><th>Event</th><th>Sec</th></tr>");
+            buf.append("<table><tr><th colspan=\"2\">Event</th><th colspan=\"2\">Sec</th></tr>");
```
cquiroz: Sorry you had to resort to this 😄 
Using html in swing controls is quite fragile though often is the only way around it. This changes probably need to be tested across platforms though

jdnavarro: Ugly, I know 😳. I can test the OT on Linux, are you worried this won't be displayed properly on Windows?

arturog8m: That's a good point. Recently we discovered some swing issues under windows 10, so this will need to be tested. I don't see anything that raises a flag, but if we can make a version of this for Linux/OSX/Windows and let Bryan try it on the three platforms, that would remove the concern.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gmos/InstGmosCommon.java
swalker2m: Let's define them with their respective definitions like we do in `Flamingos2`.  Because of the whacky way that GMOS enums are split over north and south versions, I think this means in the interface definitions like `GmosCommonType.FPUnit` etc.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/obs/plannedtime/PlannedTime.java
```diff
@@ -223,17 +222,30 @@ public static CategorizedTimeGroup apply(CategorizedTime... col) {
             return times.hashCode();
         }
 
-        public Map<Category, CategorizedTime> maxTimes() {
+        public Pair<Map<Category, CategorizedTime>, Map<Category, ImList<CategorizedTime>>> maxTimes() {
```
swalker2m: `maxTimes` is not quite accurate anymore since we're returning everything, just partitioned. Maybe a way to simplify would be to "groupBy" `Category`, returning just one `Map<Category, ImList<CategorizedTime>>` with everything in it (no finding the max here).  You could call it `groupTimes` or something.

Then in the `ObserveTimeLineNode` where you need to pick out the max, you can use `ImList`'s `sort` method to find it.  i.e.

```java
  final ImList<CategorizedTime> s = catTimes.sort(Comparator.naturalOrder()); // I think ...
  final CategorizedTime max = s.head();
  final ImList<CategorizedTime> losers = s.tail();
```

swalker2m: Well, except that I suppose you'd be picking the min that way so it'd need to be reversed or you'd have to get the `last` and `init` instead of `head` and `tail`!

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gmos/InstGmosCommon.java
jdnavarro: I was reluctant to touch those GMOS enums because I needed to reshuffle several types around and was worried about breaking something fundamental in the core. But I'll give it a try, I'll be counting on your review for reassurance 😄 .

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/obs/plannedtime/PlannedTime.java
jdnavarro: So you can do all that with `ImList`s 😮? That's great.

I'm realizing `ImList` provides a very decent functional API. I guess I might be underestimating Java as a capable FP language.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/editor/seq/ObserveTimeLineNode.java
jdnavarro: Alright I have a spare Windows 10 license, I'll use it in a VM to it. I guess we'll need more Windows 10 testing in the future.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/obs/plannedtime/PlannedTime.java
swalker2m: Thinking about it more, you don't really need to sort them, there's also `max`:

```
final CategorizedTime max = catTimes.max(Comparator.naturalOrder());
```

You want all the items including the max anyway for the sub-table.

There is no `groupBy` in `ImList` but I guess we could add it for this.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/editor/seq/ObserveTimeLineNode.java
```diff
@@ -93,13 +94,13 @@ public static long getDuration(PlannedTime pt, int step) {
...
             buf.append("<table><tr><th colspan=\"2\">Event</th><th colspan=\"2\">Sec</th></tr>");
             Step s = plannedTime.steps.get(step);
-            Map<Category, CategorizedTime> m = s.times.maxTimes()._1();
-            Map<Category, ImList<CategorizedTime>> losers = s.times.maxTimes()._2();
+            Map<Category, ImList<CategorizedTime>> m = s.times.groupTimes();
             for (Category c : Category.values()) {
-                CategorizedTime ct = m.get(c);
-                if (ct == null) continue;
+                ImList<CategorizedTime> cts = m.get(c);
+                if (cts == null) continue;
```
jdnavarro: Is there a better way to handle these nulls?

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/obs/plannedtime/PlannedTime.java
```diff
@@ -223,15 +222,13 @@ public static CategorizedTimeGroup apply(CategorizedTime... col) {
...
-                }
-            }
-            return max;
+        public HashMap<Category, ImList<CategorizedTime>> groupTimes() {
+            Function1<CategorizedTime, Category> f = new Function1<CategorizedTime, Category>() {
+                @Override
+                public Category apply(CategorizedTime ct) { return ct.category; }
+            };
+
+            return DefaultImList.create(times).groupBy(f);
```
swalker2m: I think you can use a lambda here:

```
return DefaultImList.create(times).groupBy(t -> t.category);
```

bundle/edu.gemini.shared.util/src/main/java/edu/gemini/shared/util/immutable/DefaultImList.java
```diff
@@ -348,6 +348,21 @@ public T max(final Comparator<? super T> c) {
...
+    public <K> HashMap<K, ImList<T>> groupBy(Function1<? super T, K> f) {
+        HashMap<K, ImList<T>> m = new HashMap<>();
+        for (final T elem : backingList) {
+            final K key = f.apply(elem);
+            final ImList<T> val = m.get(key);
+            if (val == null) {
+                m.put(key, create(elem));
+            } else {
+                m.put(key, val.append(elem));
+            }
```
swalker2m: I think this can be:

```
m.put(key, ImOption.apply(m.get(key)).map(lst-> lst.append(elem)).getOrElse(() -> create(elem));
```

Though maybe not any better.

bundle/edu.gemini.shared.util/src/main/java/edu/gemini/shared/util/immutable/ImCollections.java
```diff
@@ -215,6 +217,11 @@ public Object max(final Comparator<? super Object> c) {
         }
 
         @Override
+        public <K> HashMap<K, ImList<Object>> groupBy(final Function1<? super Object, K> f) {
+            return new HashMap<>();
```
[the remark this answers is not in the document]
jdnavarro: That's what I initially tried but I'm getting:

```no instance(s) of type variable(s) K,V exist so that java.util.Map<K,V> conforms to java.util.HashMap<K,edu.gemini.shared.util.immutable.ImList<java.lang.Object>>```

bundle/edu.gemini.shared.util/src/main/java/edu/gemini/shared/util/immutable/DefaultImList.java
jdnavarro: Actually, this made realize `ImOption.apply(null)` returns `None`, so I will use this technique to handle all those nulls when the key is not present.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/editor/seq/ObserveTimeLineNode.java
jdnavarro: I considered wrapping the `m.get` with `ImOption` but in the end I don't think it's worth exchanging the penalty in readability for the extra (little) safety that we get.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/editor/seq/ObserveTimeLineNode.java
jdnavarro: I tested it Windows and Linux. In Arch Linux with GNOME, it looks fine. In Windows 10 all the fonts look very pixelated (maybe it's because of my Remote Desktop compression settings or something), but aside of that, the alignment of the columns for the tooltip looks alright.

PR #1327, opened 2017-09-05
SEQNG-312: Display file id of already executed steps

cquiroz:
On the new seqexec we'd like to display the file id of already executed steps. This information is on the program and can be transferred to the seqexec when reading the configuration sequencep

### DIFF
--- a/bundle/edu.gemini.seqexec.odb/src/main/scala/edu/gemini/seqexec/odb/SeqExecService.scala
+++ b/bundle/edu.gemini.seqexec.odb/src/main/scala/edu/gemini/seqexec/odb/SeqExecService.scala
@@ -89,10 +89,7 @@ object SeqExecService {
     }
 
   private def extractName(obs: ISPObservation): TrySeq[String] = trySeq {
-    obs.getDataObject match {
-      case o: SPObservation => o.getTitle.right
-      case _                => SeqException(new RuntimeException("Cannot read the obs title")).left
-    }
+    obs.getDataObject.getTitle.right
   }
 
   private def extractSequence(obs: ISPObservation): TrySeq[ConfigSequence] =
--- a/bundle/edu.gemini.seqexec.odb/src/main/scala/edu/gemini/seqexec/odb/SeqExecService.scala
+++ b/bundle/edu.gemini.seqexec.odb/src/main/scala/edu/gemini/seqexec/odb/SeqExecService.scala
@@ -11,16 +11,19 @@ import edu.gemini.spModel.io.SpImportService
 import java.io.{BufferedReader, InputStreamReader, StringReader}
 import java.net.{HttpURLConnection, URL}
 import java.net.HttpURLConnection.{HTTP_NOT_FOUND, HTTP_OK}
-import java.time.Duration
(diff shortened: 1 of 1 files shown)


bundle/edu.gemini.seqexec.odb/src/main/scala/edu/gemini/seqexec/odb/SeqExecService.scala
```diff
@@ -83,25 +88,43 @@ object SeqExecService {
       }
     }
 
+  private def extractName(obs: ISPObservation): TrySeq[String] = trySeq {
+    obs.getDataObject match {
+      case o: SPObservation => o.getTitle.right
```
swalker2m: `getTitle` is defined in the `ISPDataObject` from which `SPObservation` inherits, so you wouldn't need this match unless you just want to verify that the data object is in fact an `SPObservation`.  If it isn't though, nothing else will work anyway.

cquiroz: Right, do you think I should just call it on the `DataObject`?

bundle/edu.gemini.seqexec.odb/src/main/scala/edu/gemini/seqexec/odb/SeqExecService.scala
```diff
@@ -83,25 +88,43 @@ object SeqExecService {
...
   private def extractSequence(obs: ISPObservation): TrySeq[ConfigSequence] =
     catchingAll {
       ConfigBridge.extractSequence(obs, null, IDENTITY_MAP, true)
     }
 
+  private def extractExecutedDatsets(obs: ISPObservation): TrySeq[Map[Int, ExecutedDataset]] =
+    catchingAll {
+      Option(ObsLog.getIfExists(obs)) match {
+        case Some(obsLog) => obsLog.getAllDatasetRecords.asScala.map(_.exec.dataset).map { d =>
+          d.getIndex -> ExecutedDataset(Instant.ofEpochMilli(d.getTimestamp), d.getDhsFilename)
```
cquiroz: It seems that `getIndex` is the step where the file was executed right. There is another `index` on `DatasetLabel`. @swalker2m  Do you know if they are the same?

bundle/edu.gemini.seqexec.odb/src/main/scala/edu/gemini/seqexec/odb/SeqExecService.scala
swalker2m: Yeah, that's probably what I'd do.

bundle/edu.gemini.seqexec.odb/src/main/scala/edu/gemini/seqexec/odb/SeqExecService.scala
swalker2m: I usually key datasets by their full dataset label, which includes the index.  In reality in the current OCS these must be the same.  That's the source of innumerable problems that motivated the need for a new sequence model.

PR #1328, opened 2017-09-05
REL-3251: Fix parsing of F2 and Michelle engineering enums

swalker2m:
This PR corrects an issue with parsing of F2 and Michelle engineering enums that resulted in data loss.  In particular, when reading an optional enum tag from XML, we make the call:

```java
        v = Pio.getValue(paramSet, WINDOW_COVER_PROP.getName());
        if (v != null) setWindowCover(WindowCover.valueOf(v, getWindowCover()));
```

where the value of `getWindowCover()`, intended to be used as the default value in case `v` is not readable, is usually `None` when initializing a program.  Then in the `WindowCover` parser we would inexplicably map the default value to the actual lookup of `v` resulting in `None`.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -4,7 +4,7 @@ name := "ocs"
 
 organization in Global := "edu.gemini.ocs"
 
-ocsVersion in ThisBuild := OcsVersion("2017B", false, 1, 1, 3)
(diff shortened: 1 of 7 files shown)


tpolecat:
https://www.youtube.com/watch?v=A5xA21Qoxaw&t=183

PR #1331, opened 2017-09-05
REL-2834: Allow deactivation of AGS

jdnavarro:
There are still some potentially confusing UI behavior that still needs to be addressed, but with this PR is now possible to deactivate the guide star.

### DIFF
--- a/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsAnalysis.scala
+++ b/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsAnalysis.scala
@@ -149,7 +149,7 @@ object AgsAnalysis {
    * Analysis of the given guide star in the given context, regardless of which
    * guide star is actually selected in the target environment.
    */
-  protected [ags] def analysis(ctx: ObsContext, mt: MagnitudeTable, guideProbe: ValidatableGuideProbe, guideStar: SiderealTarget, bands: BandsList): Option[AgsAnalysis] = {
+  def analysis(ctx: ObsContext, mt: MagnitudeTable, guideProbe: ValidatableGuideProbe, guideStar: SiderealTarget, bands: BandsList): Option[AgsAnalysis] = {
     val spTarget = new SPTarget(SiderealTarget.empty.copy(coordinates = Coordinates(guideStar.coordinates.ra, guideStar.coordinates.dec)))
 
     if (guideProbe.validate(spTarget, ctx) != GuideStarValidation.VALID) Some(NotReachable(guideProbe, guideStar, bands))
--- a/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsAnalysis.scala
+++ b/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsAnalysis.scala
@@ -57,45 +57,45 @@ object AgsGuideQuality {
 
 sealed trait AgsAnalysis {
   def quality: AgsGuideQuality = Unusable
-  def probeBands: BandsList
   def message(withProbe: Boolean): String
 }
 
 object AgsAnalysis {
-  case class NoGuideStarForProbe(guideProbe: GuideProbe, probeBands: BandsList) extends AgsAnalysis {
+  case class NoGuideStarForProbe(guideProbe: GuideProbe) extends AgsAnalysis {
     override def message(withProbe: Boolean): String = {
       val p = if (withProbe) s"${guideProbe.getKey} " else ""
       s"No ${p}guide star selected."
     }
   }
 
-  case class NoGuideStarForGroup(guideGroup: GuideProbeGroup, probeBands: BandsList) extends AgsAnalysis {
+  case class NoGuideStarForGroup(guideGroup: GuideProbeGroup) extends AgsAnalysis {
     override def message(withProbe: Boolean): String =
       s"No ${guideGroup.getKey} guide star selected."
   }
 
-  case class MagnitudeTooFaint(guideProbe: GuideProbe, target: SiderealTarget, probeBands: BandsList) extends AgsAnalysis {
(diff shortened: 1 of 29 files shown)


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/TargetFeedback.scala
```diff
@@ -207,7 +208,9 @@ object BagsFeedback {
     case RunningState(_,_,_)   => RunningStateRow
     case RunningEditedState(_) => RunningStateRow
     case FailureState(_,why)   => FailureStateRow(why)
-    case IdleState(_,_) if ctx.exists(_.getTargets.getGuideEnvironment.guideEnv.auto === AutomaticGroup.Initial) => NoStarsRow
+    case IdleState(_,_) if ctx.exists(c =>
+      AgsRegistrar.currentStrategy(c).exists(_.key.id =/= OffStrategy.key.id) &&
```
jdnavarro: I was about to create a proper Eq instance but thought it might not be worth the trouble if using the AgsStrategy.Key is good enough.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/TargetFeedback.scala
```diff
@@ -194,6 +195,7 @@ object BagsFeedback {
   }
 
   case object NoStarsRow      extends BagsStateRow(BANANA.some, "Automatic guide star lookup found nothing.", warningIcon)
+  case object NoGuideStateRow extends BagsStateRow(BANANA.some, "Observation unguided.", warningIcon)
```
swalker2m: Just a small nit to pick about the name of this `case object`.  The other *`StateRow`s correspond to actual BAGS manager states but this one doesn't.  Maybe a more appropriate name might be `AgsOffRow`.

What happens if you have manual guide stars?  Do you still see "Observation unguided."?  That would be misleading and we would need to at least change the message to something like "Auto guide star search disabled." if so.

jdnavarro: > Just a small nit to pick about the name of this case object. The other *StateRows correspond to actual BAGS manager states but this one doesn't. Maybe a more appropriate name might be `AgsOffRow`.

I didn't know. I'll change it to `AgsOffRow`.

> What happens if you have manual guide stars? Do you still see "Observation unguided."? That would be misleading and we would need to at least change the message to something like "Auto guide star search disabled."

The message is shown only when selecting "base" or the previously created guide star. It's not shown when clicking on the rows "Auto" or "Manual". It's a bit weird, I may have to record another screen capture.

swalker2m: > I'll change the message to "Auto guide star search disabled", but there is still something weird going on. The message is shown only when selecting "base" or the previously created guide star. It's not shown when clicking on the rows "Auto" or "Manual". I may have to record another screen capture.

I think that may be correct.  What still concerns me though is that the message appears when auto guide search is off and yet there are manual guide stars?  It seems like it should only be shown if there isn't a selected manual guide star group.

sraaphorst:
I was just looking at this and agree with @swalker2m as in I am impressed that so few changes needed to be made to accommodate this. Great stuff.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsAnalysis.scala
```diff
@@ -149,7 +149,7 @@ object AgsAnalysis {
    * Analysis of the given guide star in the given context, regardless of which
    * guide star is actually selected in the target environment.
    */
-  protected [ags] def analysis(ctx: ObsContext, mt: MagnitudeTable, guideProbe: ValidatableGuideProbe, guideStar: SiderealTarget, bands: BandsList): Option[AgsAnalysis] = {
+  def analysis(ctx: ObsContext, mt: MagnitudeTable, guideProbe: ValidatableGuideProbe, guideStar: SiderealTarget, bands: BandsList): Option[AgsAnalysis] = {
```
swalker2m: I think this all looks quite plausible.  One detail that might be improved is that since now guide probes know about which bands are relevant we shouldn't need to pass the `BandsList` to this method.  Instead, it seems like it can just ask the `guideProbe` directly.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
```diff
@@ -821,7 +823,11 @@ private void setSelectionFromNode() {
             if (selectionIsBasePosition()) {
                 DialogUtil.error("You can't remove the base position.");
             } else if (selectionIsAutoTarget()) {
-                DialogUtil.error("You can't remove automatic guide stars.");
+                AgsStrategyUtil.setSelection(
+                        getContextObservation(),
+                        ImOption.fromScalaOpt(AgsRegistrar.lookup(AgsStrategyKey$.MODULE$.fromStringOrNull("Off")))
```
swalker2m: Could you just pass in the `AgsStrategKey.OffKey` directly or is it a problem accessing that in Java.

The `Off` strategy is just stubbed out so it seems like it could easily go in the `AgsStrategy` companion object or even in the `edu.gemini.ags.api` package as a top-level class and then you wouldn't need to look it up at all.

jdnavarro:
@swalker2m Please have a final look at this before merging. I removed the `fromStringOrNull` hack and replaced with some nice ugly syntax 😄

I also tried your suggestion of moving `OffStrategy` to `api.AgsStrategy` but ran into some implicit imports problems I couldn't fix right away. I git stashed this attempt in case you think it's worthy trying to make it work.

PR #1338, opened 2017-10-24
REL-3255: Updating QPT icon for 2018A.

sraaphorst:
Self-explanatory.


sraaphorst:
Not for the QPT, no.
For the OT, yes, but that is a separate PR.

PR #1341, opened 2017-10-24
REL-3268: Update NIRI f/6 J-Band exposure time

swalker2m:
I reformatted the `expMap` initialization code but there is just one small substantive change.

I also updated the `NIRI_exptimes.xsl` spreadsheet with exposure times, though we don't read it in the code and it's in a binary format so we can't even use it to compare one version to another.  I would support just deleting it.


cquiroz:
I support deleting it

PR #1345, opened 2017-10-25
REL-3272: Added two new filters each to GMOS-N and GMOS-S.

sraaphorst:
As per the JIRA task, we have two new filters for the upcoming semester for GMOS-N and GMOS-S.
This adds them, with Julia's specifications and on where she would like to see them added in the list for GMOS-N.


cquiroz:
This seems bine but I don't recall those being added to the PIT. Will be they used in engineering only?

sraaphorst:
Seems like they'll only be available for 2018 Q2 for FT programs, and then phased in for all programs. I'm guessing, then, that this can wait until the next PITeration?

PR #1346, opened 2017-10-30
REL-2834 Add AGS Off p2checker rule

jdnavarro:
When AGS is Off there is now p2checker warning.

This was necessary to close REL-2834. The tests are still missing but I have some work in progress stashed away in case we want to include tests in this PR.

### DIFF
--- a/bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/general/GeneralRule.java
+++ b/bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/general/GeneralRule.java
@@ -1,5 +1,6 @@
 package edu.gemini.p2checker.rules.general;
 
+import edu.gemini.ags.api.AgsRegistrar;
 import edu.gemini.p2checker.api.IP2Problems;
 import edu.gemini.p2checker.api.IRule;
 import edu.gemini.p2checker.api.ObservationElements;
@@ -7,11 +8,11 @@ import edu.gemini.p2checker.api.P2Problems;
 import edu.gemini.pot.sp.*;
 import edu.gemini.shared.util.immutable.ImList;
 import edu.gemini.shared.util.immutable.Option;
+import edu.gemini.spModel.ags.AgsStrategyKey;
 import edu.gemini.spModel.config2.Config;
 import edu.gemini.spModel.config2.ConfigSequence;
 import edu.gemini.spModel.config2.ItemKey;
 import edu.gemini.spModel.core.ProperMotion;
-import edu.gemini.spModel.core.SiderealTarget;
 import edu.gemini.spModel.core.Site;
 import edu.gemini.spModel.core.Target;
 import edu.gemini.spModel.gemini.altair.AltairAowfsGuider;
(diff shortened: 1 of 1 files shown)


bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/general/GeneralRule.java
```diff
@@ -588,6 +587,32 @@ private boolean isAcceptable(final SPSiteQuality expected, final SPSiteQuality a
...
+        public IP2Problems check(final ObservationElements elements)  {
+            if (elements.getTargetObsComp().isEmpty()) return null; // can't perform this check without a target environment
+
+            final P2Problems problems = new P2Problems();
+
+            if (!hasScienceObserves(elements.getSequence())) return null; //if there are not observes, ignore this check (SCT-260)
+
+            final Boolean isOff = elements.getObsContext().flatMap(
+                    ctxt -> AgsRegistrar.currentStrategyForJava(ctxt)
+            ).map(st -> st.key().id().equals("Off")).getOrElse(false);
```
sraaphorst: Couldn't this be done with `exists` instead of `map` and `gerOrElse(false)`?

Final line:
`.exists(_.key().id().equals("Off"));`

sraaphorst: Oooops... Java code, so you would need the `st -> st.....`

cquiroz: Why use a string "Off" as the key? Could we at least replace it with a constant?

sraaphorst: Sure... something like `st -> st.key().id().equals(AgsStrategyKey.OffKey.id)`?
Since `OffKey` is a `case object`, it will be more fiddly than that, I am sure.

bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/general/GeneralRule.java
```diff
@@ -588,6 +588,32 @@ private boolean isAcceptable(final SPSiteQuality expected, final SPSiteQuality a
...
+        public IP2Problems check(final ObservationElements elements)  {
+            if (elements.getTargetObsComp().isEmpty()) return null; // can't perform this check without a target environment
+
+            final P2Problems problems = new P2Problems();
+
+            if (!hasScienceObserves(elements.getSequence())) return null; //if there are not observes, ignore this check (SCT-260)
+
+            final Boolean isOff = elements.getObsContext().flatMap(
+                    ctxt -> AgsRegistrar.currentStrategyForJava(ctxt)
+            ).exists(st -> st.key().equals(AgsStrategyKey.OffKey$.MODULE$));
```
jdnavarro: There you go, still not pretty but more concise now.

PR #1349, opened 2017-11-02
REL-1485: Cleanup and sync QPT and OT colours

sraaphorst:
This is a major cleaning up / modernization of the classes used in the QPT Visualizer (that I did as I went through the code to understand it better), along with changing the colours of the QPT so that:
1. The daytime is now white, to be in line with the OT;
2. The sun is darkened slightly, to be visible against the white; and
3. The moon is darkened ever so slightly, to also be visible against the white.


cquiroz:
Is this part of release tasks? I'm a bit worried by the "major" word so near the release

tpolecat:
Agree, I don't think a 700-line cosmetic change is a great idea right before the release. Can we sit on this until after we get a release out, or are you working on release tasks out ahead of this?

sraaphorst:
It's just very basic cleanup: moving to more functional constructs instead of loops, refactoring a small amount, and adding `final` everywhere. (The `final` probably constitutes most of the bulk of the changes.)

But yes, if it makes people uncomfortable, I can make the bare changes (which would only be a few lines) and we can hold off on this until after release.

(I'm done my release tasks and cleaned as I researched: hence the changes.)

sraaphorst:
Oh yes, and deparameterizing instantiations of typed classes, removal of old comments, and some unused variables.

I was aiming to reduce the number of warnings.

I'll make a new PR with the minimal changes in a moment.

PR #1356, opened 2017-11-13
REL-1438: add tag to TCC config

swalker2m:
This is a small update to populate the `tag` parameter of TCC target elements.  I'm not sure why we were passing an empty String here, but we have the available information.  @jluhrs can you confirm that this will be okay?

Ideally there should be an associated test case but the TCC XML and dom4j APIs are difficult to use.  After a few hours trying to get a reasonable test, I decided that a visual inspection of the output was probably sufficient.

### DIFF
--- a/bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/TargetConfig.java
+++ b/bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/TargetConfig.java
@@ -47,7 +47,7 @@ public final class TargetConfig extends ParamSet {
         return String.format("%s (%d)", tag, position);
     }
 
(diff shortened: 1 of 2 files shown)


cquiroz:
What's `tag` supposed to contain?

swalker2m:
Historically the tag specified what kind of thing the target represented.  I can't remember why we stopped writing it though.  Now with various kinds of user-defined targets, I'm hoping to revive it to distinguish between blind-offset, tuning star, etc.  I need @jluhrs to comment on the ramifications of data appearing in that field again.

jluhrs:
If I remember correctly, `tag` was used to identify a target as a base, p1, p2, oi or ao target. That use was replaced by the introduction of target groups. All references to that `param` were removed from TCC.

swalker2m:
Merging to repurpose the existing tag field for identifying the type of user-targets.

PR #1358, opened 2017-11-13
REL-1485: Added time labels.

sraaphorst:
Added time labels to the OT elevation plot in accordance with the discussion on REL-1485.

Also removed warnings from the code and deleted commented out code.

Added `@Overrides` as well.

You can see the times indicated in the attached PNG for sunrise, sunset, and 12-deg twilight.
labels

### DIFF
--- a/bundle/jsky.app.ot/src/main/java/jsky/app/ot/viewer/SPElevationPlotPlugin.java
+++ b/bundle/jsky.app.ot/src/main/java/jsky/app/ot/viewer/SPElevationPlotPlugin.java
@@ -25,7 +25,9 @@ import java.awt.event.ItemListener;
 import java.io.Serializable;
 
 import java.util.Hashtable;
+import java.util.Set;
 import java.util.TreeMap;
+import java.util.TreeSet;
 import java.util.function.Consumer;
 import java.util.function.Function;
 
@@ -214,18 +216,20 @@ public class SPElevationPlotPlugin implements ChangeListener, Storeable {
     // 1. Executing an optional preprocessing function (to precalculate some colors and create some default legend items); and
     // 2. Executing an extractor for each selected observation, which produces an optional ID. If the ID is defined, create
     //    a new legend item for it.
+    // We also ensure that IDs for LegendItems are unique so that we don't, for example, end up with multiple entries for
+    // "Band 4" if observations are labeled by band, or repeated priorities if observations are labeled by priority.
     private LegendItemCollection createLegendItemsPreOpt(final Option<Function<TreeMap<String, Paint>, LegendItemCollection>> pref,
                                                          final Function<ISPObservation, Option<String>> extractor) {
         final Paint[] colors = new Paint[_selectedObservations.length];
         final TreeMap<String,Paint> paintMap = new TreeMap<>();
-        final TreeMap<String,LegendItem> licTree = new TreeMap<>();
+        final Set<String> licIds = new TreeSet<>();
         final LegendItemCollection lic = new LegendItemCollection();
 
         // Perform the preprocessing function, if it exists.
         pref.foreach(f -> lic.addAll(f.apply(paintMap)));
         for (int i=0 ; i < lic.getItemCount(); ++i) {
             final LegendItem li = lic.get(i);
-            licTree.put(li.getLabel(), li);
+            licIds.add(li.getLabel());
         }
 
         // Iterate over the selected observations.
@@ -235,10 +239,10 @@ public class SPElevationPlotPlugin implements ChangeListener, Storeable {
             final Paint p = idOpt.map(id -> paintMap.computeIfAbsent(id, s -> ColorManager.instance.nextColor())).getOrElse(Color.BLACK);
             colors[obsIdx] = p;
             idOpt.foreach(id -> {
-                if (!licTree.keySet().contains(id)) {
+                if (!licIds.contains(id)) {
                     final LegendItem li = new LegendItem(id, p);
                     lic.add(li);
-                    licTree.put(id, li);
+                    licIds.add(id);
                 }
             });
         }
@@ -308,7 +312,7 @@ public class SPElevationPlotPlugin implements ChangeListener, Storeable {
                 break;
 
             case _COLOR_CODE_NONE:
-                // Special case.
+                // Special case: no legend, and all lines are drawn in black.
                 lic = new LegendItemCollection();
                 final Paint[] colors = new Paint[_selectedObservations.length];
                 java.util.Arrays.fill(colors, Color.BLACK);
--- a/bundle/jsky.app.ot/src/main/java/jsky/app/ot/viewer/SPElevationPlotPlugin.java
+++ b/bundle/jsky.app.ot/src/main/java/jsky/app/ot/viewer/SPElevationPlotPlugin.java
@@ -259,7 +259,7 @@ public class SPElevationPlotPlugin implements ChangeListener, Storeable {
                 paintMap.put(queueBand, color);
                 lic.add(new LegendItem(queueBand, color));
                 for (int i = 1; i < 5; i++) {
-                    color = _COLORS[colorIndex++];
+                    color = _COLORS[colorIndex++ % _COLORS.length];
                     queueBand = "Band " + i;
                     paintMap.put(queueBand, color);
                     lic.add(new LegendItem(queueBand, color));
@@ -276,7 +276,7 @@ public class SPElevationPlotPlugin implements ChangeListener, Storeable {
                     else
                         queueBand = "Band " + queueBand;
                     if ((colors[i] = paintMap.get(queueBand)) == null) {
-                        color = _COLORS[colorIndex++];
+                        color = _COLORS[colorIndex++ % _COLORS.length];
                         paintMap.put(queueBand, color);
                         colors[i] = color;
                         lic.add(new LegendItem(queueBand, color));
@@ -287,7 +287,7 @@ public class SPElevationPlotPlugin implements ChangeListener, Storeable {
             case _COLOR_CODE_PRIORITY:
                 // get the order of items right
                 for (SPObservation.Priority pr: _PRIORITIES) {
-                    final Paint c = _COLORS[colorIndex++];
+                    final Paint c = _COLORS[colorIndex++ % _COLORS.length];
                     paintMap.put(pr.displayValue(), c);
                     String label = pr.displayValue() + " Priority";
                     lic.add(new LegendItem(label, c));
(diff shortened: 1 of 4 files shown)


bundle/jsky.elevation.plot/src/main/java/jsky/plot/ElevationPanel.java
```diff
@@ -588,28 +572,48 @@ public void setItemColors(Paint[] colors) {
...
             IntervalMarker m1 = new IntervalMarker(start, last, color, DARKNESS_STROKE,
                     color, DARKNESS_STROKE, alpha);
-            xyPlot.addDomainMarker(m1);
+            xyPlot.addDomainMarker(m1, Layer.BACKGROUND);
+            xyPlot.addAnnotation(createAnnotation(String.format("%s: %s", startDateName, df.format(startDate)), start));
             IntervalMarker m2 = new IntervalMarker(first, end, color, DARKNESS_STROKE,
                     color, DARKNESS_STROKE, alpha);
-            xyPlot.addDomainMarker(m2);
+            xyPlot.addDomainMarker(m2, Layer.BACKGROUND);
+            xyPlot.addAnnotation(createAnnotation(String.format("%s: %s", endDateName, df.format(endDate)), first));
```
sraaphorst: The labels were added using `addAnnotation`, which was widely recommended on `JFreeChart` discussions online.

cquiroz:
Ouch, that's a lot of labels

tpolecat:
I'm assuming this is a degenerate case where you have selected way too many observations.

swalker2m:
Maybe we should limit it to a half-dozen or so.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/viewer/SPElevationPlotPlugin.java
```diff
@@ -234,98 +219,103 @@ private void _setColorCode(String label) {
...
+                    ISPProgram prog = db.lookupProgram(progKey);
+                    // LORD OF DESTRUCTION: DataObjectManager get without set
+                    SPProgram spProg = (SPProgram) prog.getDataObject();
+                    queueBand = spProg.getQueueBand();
+                    if (queueBand == null || queueBand.length() == 0)
+                        queueBand = "Default Band";
+                    else
+                        queueBand = "Band " + queueBand;
+                    if ((colors[i] = paintMap.get(queueBand)) == null) {
+                        color = _COLORS[colorIndex++];
```
tpolecat: I think this needs to be `++colorIndex % _COLORS.length` here and elsewhere. It seems to be wrong in every usage. The array must be big enough that we have never hit the end of it.

bundle/jsky.elevation.plot/src/main/java/jsky/plot/ElevationPanel.java
```diff
@@ -127,33 +126,30 @@
...
-                                       FieldPosition fieldPosition) {
-                if (number > 90.) {
-                    return result; // used to leave blank space for labels above 90 deg mark
-                }
-                return super.format(number, result, fieldPosition);
+    // Used to display the degrees values in the primary axis in the range 0..90
+    private static final NumberFormat _yAxisNumberFormat = new DecimalFormat() {
+        @Override
+        public StringBuffer format(double number, StringBuffer result, FieldPosition fieldPosition) {
+            if (number > 90.) {
```
tpolecat: `90.0` please, here and elsewhere.

sraaphorst:
No no... the only labels added are the four for sunset, sunrise, and the two twilights.
The other labels were all already there.

This is just a random program I picked and opened with the elevation plot for all observations.

sraaphorst:
Here is a before-and-after shot of a program with fewer observations shown for comparison purposes.

Before: top.
After: bottom.
screen shot 2017-11-13 at 4 03 01 pm

bundle/jsky.elevation.plot/src/main/java/jsky/plot/ElevationPanel.java
sraaphorst: Last time I touched additional code, given proximity to release, I was told to make the minimal changes. All of this is existing code. I'm more than happy to check it and change it, provided my efforts will not be wasted.

tpolecat: That's a reasonable point. Maybe just change it in your code. Allan used to do it all the time and we need to get out of the habit.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/viewer/SPElevationPlotPlugin.java
sraaphorst: This is worth looking into. I'll check it.

tpolecat: Maybe print out the length of `_COLORS` and see what we're dealing with. My guess is 256 ;-)

sraaphorst: I'm surprised this hasn't caused an `ArrayIndexOutOfBoundsException` that has been noted before.

sraaphorst:
Changes as per @tpolecat's request (good catch on the `_COLORS`) made.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/viewer/SPElevationPlotPlugin.java
```diff
@@ -234,98 +219,103 @@ private void _setColorCode(String label) {
...
+                    ISPProgram prog = db.lookupProgram(progKey);
+                    // LORD OF DESTRUCTION: DataObjectManager get without set
+                    SPProgram spProg = (SPProgram) prog.getDataObject();
+                    queueBand = spProg.getQueueBand();
+                    if (queueBand == null || queueBand.length() == 0)
+                        queueBand = "Default Band";
+                    else
+                        queueBand = "Band " + queueBand;
+                    if ((colors[i] = paintMap.get(queueBand)) == null) {
+                        color = _COLORS[colorIndex++ % _COLORS.length];
```
tpolecat: Post-increment is a red flag that says we're doing something confusing, but it's not necessary here so I suggest switching to pre-increment. I may be over-stressed about this but indexing really gives me the heebiejeebies.

tpolecat: Or at least open an issue to revisit this. It's not urgent.

sraaphorst: Alright. I did some refactoring of the code in `SPElevationPlotPlugin` as it was horrible, with assignments in `if` conditions (e.g. `if ((variable = newValue) != null)`), lots of repeated code, and as @tpolecat pointed out, array indexing that generated a red flag.

I think it is much nicer now, and much safer.

I realize that it is a substantial change, but I spent several hours testing it for various programs and selections of observations through all the different menu option combinations in the elevation plot against what we had in `2018A-test.1.4.1`, and the results were consistently identical as expected, so I'm hoping that despite the scope of the changes, we can get this merged.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/viewer/SPElevationPlotPlugin.java
```diff
@@ -232,106 +212,122 @@ private void _setColorCode(String label) {
...
+
+            default:
+                // Shouldn't ever get here.
+                lic = null;
+                _elevationPanel.setItemColors(null);
+        }
+
+        return lic;
+    }
+
```
sraaphorst: Essentially, the pattern repeated here was to:
1. Possibly pre-create some legend items (e.g. based on priority or band); and then
2. Iterate over selected observations, creating legend items for them if necessary (depending on the menu options), and then assigning them colors based on some ID string extracted from the observation. (If this corresponded to an existing legend item, such as band, then the appropriate band color was chosen).

The pattern was repeated in each block of the switch / if, and the code was quite terrible and, as Rob pointed out, array indexing was done in a way that was incorrect with regards to fetching the next color, so I extracted out the pattern into the `createLegendItem...` methods and now delegate to them with lambda functions to do (1) the preprocessing, and (2) the extraction of some kind of ID string from an observation.

Colors are assigned using a `ColorManager` so that this code is isolated from the rest and used consistently, whereas before we had things like `_COLORS[nextIdx++ % _COLORS.length]`, with the mod operation sometimes being forgotten.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/viewer/SPElevationPlotPlugin.java
```diff
@@ -232,106 +212,122 @@ private void _setColorCode(String label) {
...
-    }
+    // Create a LegendItemCollection by:
+    // 1. Executing an optional preprocessing function (to precalculate some colors and create some default legend items); and
+    // 2. Executing an extractor for each selected observation, which produces an optional ID. If the ID is defined, create
+    //    a new legend item for it.
+    // We also ensure that IDs for LegendItems are unique so that we don't, for example, end up with multiple entries for
+    // "Band 4" if observations are labeled by band, or repeated priorities if observations are labeled by priority.
+    private LegendItemCollection createLegendItemsPreOpt(final Option<Function<TreeMap<String, Paint>, LegendItemCollection>> pref,
+                                                         final Function<ISPObservation, Option<String>> extractor) {
+        final Paint[] colors = new Paint[_selectedObservations.length];
```
sraaphorst: The ultimate goal is to get a color for each selected observation, and a collection of legend items describing those colors.

sraaphorst:
👍 to @swalker2m's comment, and thanks to both you and @tpolecat for looking at this.
I'll make it very explicit in the test release announcement tomorrow that the elevation plot features require detailed, careful testing.

andrewwstephens:
Now that we are printing the times in the elevation plot, I vote to remove them from the title.

sraaphorst:
I'll look into this quickly.

sraaphorst:
Okay, having looked into this, I think before I make any changes, we have to have a quick discussion: in the elevation plot, the same title generator is used for all three tabs, so if we remove it from the elevation plot, it will also be removed from the observation table.

I'm not sure that's what we want, so I'm leaving it alone for now.

Also - ha - I know @cquiroz will want me to switch from `SimpleDateFormat` to `DateTimeFormatter`, and I have to remember how to do that properly, since we're using `long`s here for date.

(Obviously, that's easily changed and we can use one title for the elevation plot and a different one for the observation table, but I don't want to make the change more than once, so message me when you get in, @andrewwstephens, and we can discuss further. For now I'll proceed with the test release.)

PR #1359, opened 2017-11-14
REL-1485: Adjust horizon for sunrise/sunset

andrewwstephens:
This PR applies a geometric correction to the horizon for sunrise and sunset time calculations.

This PR is NOT ready to merge.  Someone who knows what they are doing needs to change SiteElevation to be "altitude" from Site.java

### DIFF
--- a/bundle/edu.gemini.util.skycalc/src/main/java/edu/gemini/skycalc/TwilightBoundType.java
+++ b/bundle/edu.gemini.util.skycalc/src/main/java/edu/gemini/skycalc/TwilightBoundType.java
@@ -15,8 +15,12 @@ public final class TwilightBoundType implements Comparable<TwilightBoundType>, S
 
     private static final long serialVersionUID = 1;
 
+    // Horizon geometric correction from p. 24 of the Skycalc manual: sqrt(2 * elevation / Re) (radians)
+    static double Re = 6378140; // Radius of the Earth (meters)
+    static double SiteElevation = 4213; // Mauna Kea elevation (meters)
+
     public static final TwilightBoundType OFFICIAL =
-            new TwilightBoundType("Official", 50.0/60.0);
+            new TwilightBoundType("Official", 50.0/60.0 + Math.sqrt(2.0 * SiteElevation / Re) * 180./Math.PI);
 
     public static final TwilightBoundType CIVIL =
             new TwilightBoundType("Civil", 6.0);


bundle/edu.gemini.util.skycalc/src/main/java/edu/gemini/skycalc/TwilightBoundType.java
```diff
@@ -15,8 +15,12 @@
 
     private static final long serialVersionUID = 1;
 
+    // Horizon geometric correction from p. 24 of the Skycalc manual: sqrt(2 * elevation / Re) (radians)
+    static double Re = 6378140; // Radius of the Earth (meters)
+    static double SiteElevation = 4213; // Mauna Kea elevation (meters)
```
cquiroz: Is there already an altittude on `Site`?
Possibly we should use `squants` distances in meters rather than `double`

andrewwstephens:
For the failing tests, skycalc predicts:

For the night of: Tue, 2017 Oct 31 ---> Wed, 2017 Nov 1
MK: Sunset ( 4215 m horizon):   17 57 HST  Sunrise:   6 14 HST
CP: Sunset ( 2215 m horizon):   20 12 CDT; Sunrise:   6 41 CDT

For the night of: Sat, 2014 Nov 1 ---> Sun, 2014 Nov 2
MK: Sunset ( 4215 m horizon):   17 57 HST; Sunrise:   6 15 HST
CP: Sunset ( 2215 m horizon):   20 13 CDT; Sunrise:   6 40 CDT

jdnavarro:
The current Skycalc implementation already takes into account the altitude in some methods but I'm not sure if is this geometric correction is already taking place or not. Can you please have a quick look at the methods in [ImprovedSkyCalcMethods.java](https://github.com/andrewwstephens/ocs/blob/116b2bce3e44ff13bff0080622645d082082564e/bundle/edu.gemini.util.skycalc/src/main/java/edu/gemini/skycalc/ImprovedSkyCalcMethods.java) where the altitude is being used? The methods are called in [SkyCalcMethods.java](https://github.com/andrewwstephens/ocs/blob/116b2bce3e44ff13bff0080622645d082082564e/bundle/edu.gemini.util.skycalc/src/main/java/edu/gemini/skycalc/ImprovedSkyCalc.java). What about [ha_alt](https://github.com/andrewwstephens/ocs/blob/116b2bce3e44ff13bff0080622645d082082564e/bundle/edu.gemini.util.skycalc/src/main/java/edu/gemini/skycalc/ImprovedSkyCalcMethods.java#L533-L559)?

If the geometric correction is not applied anywhere in `ImprovedSkyCalcMethods`, where would you plug it in? If we make it part of those methods we can fix this issue for all types of twilight, sunset/rise.

Thanks for helping with this issue. Until last week I thought Skycalc was the name of a subcontracted company or something 😃.

bundle/edu.gemini.util.skycalc/src/main/java/edu/gemini/skycalc/TwilightBoundType.java
```diff
@@ -15,8 +15,12 @@
...
+
     public static final TwilightBoundType OFFICIAL =
-            new TwilightBoundType("Official", 50.0/60.0);
+            new TwilightBoundType("Official", 50.0/60.0 + Math.sqrt(2.0 * SiteElevation / Re) * 180./Math.PI);
```
swalker2m: I don't think this belongs here.  I'm not sure where the "Official" twilight definition comes from, but in general these are supposed to be site/altitude definitions of twilight that should be adjusted for altitude elsewhere.  For example, perhaps the `TwilightBoundedNight` calculation needs to take this into account if it isn't being done in skycalc.

sraaphorst:
I'm also guessing (in line with @jdnavarro) that this may belong in `ImprovedSkyCalcMethods`, which suggests from method names and descriptions that it takes altitude into account / calculates altitudes.

These methods are also used in `TwilightBoundedNight`, which @swalker2m mentioned. I'm just not quite sure where.

sraaphorst:
Perhaps `SunRiseSet` should figure into this?
It receives a `Site`, so we have the necessary information regarding altitude at this point to make the geometric correction.

sraaphorst:
There are a lot of concerns about this PR, for multiple reasons:
1. Why is the geometric adjustment applied only to one of the four twilight formulae? Why not `CIVIL`, `NAUTICAL`, and `ASTRONOMICAL`?
2. Also, changing the twilight like this will affect time accounting from all previous semesters.

Is this a firm requirement for 2018A? Why hasn't it been an issue before this point?

It feels like this is something that requires much more discussion and consideration as to (and identification of) the snowball effects it will have.

jdnavarro:
> It feels like this is something that requires much more discussion and consideration as to (and identification of) the snowball effects it will have.

The sunrise/set were slightly different between OT and QPT before 2018A, but now they are the same. However there is still a difference between the sunrise/set in the nightlog and in the QPT/OT.

We discussed briefly with @arturog8m and came to the conclusion of opening a new ticket and leave it for next semester because apparently there are several more skycalc implementations laying around in Gemini. See REL-3295.

I mean, if this was easily fixed now I'd welcome the PR, but it still seems tricky to get it exactly right.

andrewwstephens:
Hi all,

Some aggregated answers to your questions:

Before I started,  I changed the altitude of Maunakea to 0 meters above sea level, and the resulting sunrise and sunset times did *not* change, indicating that the observer's altitude was *not* being taken into account when calculating sunrise and sunset.

The twilight times are simple to calculate because they are just when the center of the sun is at a predefined angle below the nominal horizon.  However, the sunrise and sunset are when the limb of the sun is just below the actual horizon, which is dependent on the altitude of the observer, making this calculation a little more tricky.

Skycalc adjusts the horizon for sunrise and sunset, so I implemented the same calculation here.

This should not affect time accounting since the twilight times are correct and unchanged.

I am happy to wait to implement this, but I don't think it hurts anything, and I don't like showing the current incorrect values for the sunrise and sunset in the elevation plot.

bundle/edu.gemini.util.skycalc/src/main/java/edu/gemini/skycalc/TwilightBoundType.java
tpolecat: Earth radius is already defined on `ImprovedSkyCalcMethods` and yes, elevations are defined on `Site`.

bundle/edu.gemini.util.skycalc/src/main/java/edu/gemini/skycalc/TwilightBoundType.java
tpolecat: Something may need to be parameterized by site but we certainly don't want to hardcode it to MK do we?

swalker2m:
I think we can close this in favor of #1361.

PR #1361, opened 2017-11-14
REL-1485 Make sunrise/set take altitude into account

jdnavarro:
Also remove LocalSunriseSet.java fix from QPT since the TwilightBoundedNight
already takes care of the altitude.

I'm still fixing the tests.


bundle/edu.gemini.util.skycalc/src/main/java/edu/gemini/skycalc/TwilightBoundedNight.java
```diff
@@ -150,7 +150,24 @@ public TwilightBoundedNight(TwilightBoundType type, long time, Site site) {
...
+        // -- Andrew Stephens 2017-11-14
+
+        if (type == TwilightBoundType.OFFICIAL) {
+            // Horizon geometric correction from p. 24 of the Skycalc manual: sqrt(2 * elevation / Re) (radians)
+           final double angle = type.getHorizonAngle() +
+                   Math.sqrt(2.0 * site.altitude / ImprovedSkyCalcMethods.EQUAT_RAD) *
+                           ImprovedSkyCalcMethods.DEG_IN_RADIAN;
+            _calcTimes(angle, jdmid, site);
+        } else {
+            _calcTimes(type.getHorizonAngle(), jdmid, site);
```
swalker2m: Perhaps we should make `TwilightBoundType` and `enum` now that we don't need any other instances?  Then I guess you could switch on `type` to compute the `angle`.

tpolecat:
What's the deal with the tests? Do they just need new baseline values?

jdnavarro:
> What's the deal with the tests? Do they just need new baseline values?

The expected results will be differ, but we are disabling them temporarily so that Andy and Bryan can verify manually the new expected results are really correct.

sraaphorst:
Wouldn't it just be better to use the `@Ignore` directive before the tests than to entirely comment them out?

jdnavarro:
I forgot the switch case suggestion. I'll try to include it when restoring the tests.

PR #1363, opened 2017-11-15
REL-3277: update GRACES template XML

swalker2m:
No logic changes, XML updated to latest version.


sraaphorst:
Looks good. Unfortunately, I don't think this will make it into the current test release, which is already well underway.

PR #1364, opened 2017-11-15
REL-1485: Removed times from elevation titles and fixed legend appearance.

sraaphorst:
This PR does two things:

1. Removes the times from the titles of the elevation plot and the observation chart as per my conversation with @andrewwstephens. (This information is now displayed directly in the plot, and thus is no longer useful in the title.)

2. Fixes the previously used hideous legends by adding spacing between legend items in the legend.

@cquiroz, I tried to switch from `java.util` to `java.time` classes, but there was just way too much conversion that needed to be done back and forth, and `JFreeChart` uses the `java.util` classes, so it just became way too cumbersome.

Here is a before and after demonstrating the title and legend difference:
before

after

### DIFF
--- a/bundle/jsky.app.ot/src/main/java/jsky/app/ot/viewer/SPElevationPlotPlugin.java
+++ b/bundle/jsky.app.ot/src/main/java/jsky/app/ot/viewer/SPElevationPlotPlugin.java
@@ -24,10 +24,8 @@ import java.awt.*;
 import java.awt.event.ItemListener;
 import java.io.Serializable;
 
-import java.util.Hashtable;
-import java.util.Set;
-import java.util.TreeMap;
-import java.util.TreeSet;
+import java.util.*;
+import java.util.List;
 import java.util.function.Consumer;
 import java.util.function.Function;
 
@@ -219,6 +217,7 @@ public class SPElevationPlotPlugin implements ChangeListener, Storeable {
     // We also ensure that IDs for LegendItems are unique so that we don't, for example, end up with multiple entries for
     // "Band 4" if observations are labeled by band, or repeated priorities if observations are labeled by priority.
     private LegendItemCollection createLegendItemsPreOpt(final Option<Function<TreeMap<String, Paint>, LegendItemCollection>> pref,
+                                                         final Comparator<ISPObservation> comparator,
                                                          final Function<ISPObservation, Option<String>> extractor) {
         final Paint[] colors = new Paint[_selectedObservations.length];
         final TreeMap<String,Paint> paintMap = new TreeMap<>();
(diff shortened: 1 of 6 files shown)


sraaphorst:
Ignore the colour differences. That's a result of the implementation of `SPElevationPanelPlugin` and the fact that in one OT, I had opened multiple elevation panels.

sraaphorst:
Or I guess, it appears that `JFreeChart`'s `LegendItemCollection` doesn't necessarily respect order?

sraaphorst:
Actually, it was our list of selected observations that was completely unordered.
I have added a `Comparator` to the legend item creation function to put some sanity in place in these legends, using the already implemented `Comparable`s when possible.
screen shot 2017-11-15 at 5 53 25 pm

bundle/jsky.app.ot/src/main/java/jsky/app/ot/viewer/SPElevationPlotPlugin.java
```diff
@@ -274,14 +281,26 @@ private LegendItemCollection getLegendItems() {
         final LegendItemCollection lic;
         switch (_colorCodeTrajectories) {
             case _COLOR_CODE_OBS:
-                lic = createLegendItems(obs -> obs.getObservationIDAsString("unknown"));
+                lic = createLegendItems(
+                        Comparator.comparing(ISPObservation::getObservationID),
```
swalker2m: Cool, I didn't know about `Comparator.comparing`.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/viewer/SPElevationPlotPlugin.java
```diff
@@ -274,14 +281,26 @@ private LegendItemCollection getLegendItems() {
...
                 break;
 
             case _COLOR_CODE_PROG:
-                lic = createLegendItemsOpt(obs -> ImOption.apply(obs.getProgramID()).map(SPProgramID::stringValue));
+                lic = createLegendItemsOpt(
+                        (o1,o2) -> {
+                            final SPProgramID id1 = o1.getProgramID();
+                            final SPProgramID id2 = o2.getProgramID();
+                            if (id1 == null) return (id2 == null) ? 0 : 1;
+                            return (id2 == null) ? -1 : id1.compareTo(id2);
```
swalker2m: There seems to be a `Comparator.nullsFirst` that handles `null`.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/viewer/SPElevationPlotPlugin.java
```diff
@@ -308,7 +328,13 @@ private LegendItemCollection getLegendItems() {
                         plic.add(new LegendItem(prioStr, p));
                     }
                     return plic;
-                }, obs -> String.format("%s Priority", ((SPObservation) obs.getDataObject()).getPriority().displayValue()));
+                }, (o1, o2) -> {
+                            final SPObservation.Priority p1 = ((SPObservation) o1.getDataObject()).getPriority();
+                            final SPObservation.Priority p2 = ((SPObservation) o2.getDataObject()).getPriority();
+                            return p1.compareTo(p2);
```
swalker2m: `Enum` implements `Comparable` so could you use your `Comparator.comparing` trick here as well?  Something like `Comparator.comparing(o -> ((SPObservation) o.getDataObject()).getPriority)`?

bundle/jsky.app.ot/src/main/java/jsky/app/ot/viewer/SPElevationPlotPlugin.java
sraaphorst: Nice! Did not know about that. I'll make the change.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/viewer/SPElevationPlotPlugin.java
sraaphorst: I'll give it a try. Thanks for looking at this PR.

swalker2m:
Looks good 👍

PR #1368, opened 2017-11-17
REL-1485 Use switch case for TwilightBoundType choice

jdnavarro:
I forgot to include this in #1361


sraaphorst:
Maybe I'm missing something, but I'm not clear what the difference is here except for style.

jdnavarro:
> Maybe I'm missing something, but I'm not clear what the difference is here except for style.

It's just style but here is the full boring thought process:

It was a review comment that I forgot to include in a previous PR and hold on it for the PR where we'd restore the sunrise/set. I thought it was not worth cluttering the history with another merge commit, but since I learned yesterday, with the squash-merge discussion, how to *merge* without an actual merge commit, I just went ahead with yet another PR.

swalker2m:
> Maybe I'm missing something, but I'm not clear what the difference is here except for style.

It is pretty trivial but the change factors out the common call to `_calcTimes` and matches a bit more closely what you might do in scala:

```
val correction = boundsType match {
  case Official => math.toDegrees(math.sqrt(2.0 * site.altitude / EQUAT_RAD))
  case _        => 0
}
_calcTimes(boundsType.getHorizonAngle + correction, jmid, site)
```

PR #1369, opened 2017-11-20
REL-1485: Change to unicode arrow in date range titles.

sraaphorst:
As per Andy's comment on REL-1485, the elevation panel and observation chart would formerly use titles of the form:
"Night of 2017 Nov 16 ---> 2017 Nov 17"

This makeshift arrow looked ugly, especially in certain cases, so I have changed it to a proper unicode long arrow:
"Night of 2017 Nov 16 ⟶ 2017 Nov 17"

screen shot 2017-11-20 at 8 04 14 am


swalker2m:
The original screen shot showed Nov 21 -> Nov 21?

sraaphorst:
Yes, this is because we were looking at GS programs from Mauna Kea in GS time, so the "night" was happening entirely during a "day" here at GS as you can see from the horizontal axis times; because I don't think anyone would pick this combination of options since it doesn't make sense (and was just from me testing), I removed the screenshot and replaced with a more sensible selection.

The "Night of..." in this case is confusing. I don't know if this is something we want to concern ourselves with.

PR #1371, opened 2017-11-20
REL-3310: Addition of spdb init.d script, and removal of obsolete JVM parameters.

sraaphorst:
This PR does two things:
1. Removes the MaxPermSize parameter from the `build.sbt` file for the `spdb`, since this parameter is obsolete in Java 8.0, which is the requirement for OCS.
2. Introduces the `init.d` script, `spdb`, which can `start`, `stop`, `restart`, and `status` query the ODB, as per Chris Stark's request for server shutdowns.

I would appreciate any suggestions for the `spdb` script if anyone has any ideas for improvements, or sees any possible issues with it.

### DIFF
--- a/app/spdb/scripts/spdb
+++ b/app/spdb/scripts/spdb
@@ -112,8 +112,8 @@ determine_spdb_directory ()
 # Returns 0 if running, and nonzero (1) otherwise.
 find_running_spdb ()
 {
-    SPDB_PID=`psjava | awk '{ printf $2; }'`
-    SPDB_VER=`psjava | sed "s/^.*\(spdb_[^_]*\).*/\1/"`
+    SPDB_PID=`psjava | grep spdb | awk '{ printf $2; }'`
+    SPDB_VER=`psjava | grep spdb | sed "s/^.*\(spdb_[^_]*\).*/\1/"`
     if [[ -n "$SPDB_PID" ]]; then
 	return 0
     else
--- a/app/spdb/scripts/spdb
+++ b/app/spdb/scripts/spdb
@@ -1,0 +1,256 @@
+#!/bin/bash
+# 
+# This script manages the startup / shutdown of the ODB.
+
+# RETVALS:
+# 0. Expected.
+# 1. Usage displayed.
+# 2. Could not start.
+# 3. Could not stop.
+# 4. Status: not running.
+
+# Source function library.
+. /etc/rc.d/init.d/functions
+
+# psjava access.
+export PATH=$PATH:/home/software/bin
+
+# System variables.
+SERVER=`hostname`
+PORT=8224
+
+
+# Display usage information.
+usage ()
+{
+    echo $"Usage: $0 {start|stop|status|restart}" 1>&2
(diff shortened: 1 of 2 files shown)


app/spdb/scripts/spdb
```diff
@@ -0,0 +1,256 @@
...
+    SPDB_SEMESTER=0000A
+    SPDB_VERSION=(0 0 0)
+    
+    for CURR_DIR in "${SPDB_DIRS[@]}"
+    do
+	# Extract the version so that we can find the most recent.
+	# We don't use sort on the ls output for this because it will compare by string
+	# instead of by number, and we want to avoid things like 1.10.1 being less than 1.2.1.
+	CURR_SEMESTER=`echo $CURR_DIR | sed "s/^.*spdb_\([^-.]*\).*$/\1/"`
+	if [[ "$CURR_SEMESTER" < "$SPDB_SEMESTER" ]]; then
```
sraaphorst: Lexicographic string ordering should be correct for semesters.

swalker2m: This comment doesn't change anything about the `init.d` script because we'd still need to figure out the current version anyway, but we should do a better job cleaning up old installations.  We really only need the last installation before an update, and only for a few days.  Once a a bit of time has passed we can't easily go back to a previous version anyway.

app/spdb/scripts/spdb
```diff
@@ -0,0 +1,256 @@
...
+
+
+# If there is already a running SPDB:
+# 1. Determine its PID and set SPDB_PID to it.
+# 2. Determine the version (e.g. 2018A-test.1.6.1) and set SPDB_VER to it.
+# Returns 0 if running, and nonzero (1) otherwise.
+find_running_spdb ()
+{
+    SPDB_PID=`psjava | awk '{ printf $2; }'`
+    SPDB_VER=`psjava | sed "s/^.*\(spdb_[^_]*\).*/\1/"`
```
swalker2m: We're assuming that the only java process running will be the `spdb`.  I think that's true for now but would silently break if we ever run another java process?

swalker2m:
Must have been annoying to write but it looks very clean.

swalker2m:
I think we want to be careful not to run the Data Manager in the test machine.  At the very least it will add a lot of queries to the archive and fits storage services.

app/spdb/scripts/spdb
sraaphorst: This is a good point. We aren't great at cleaning up old installations. I wrote this so that it could operate on all ODB machines, including `gsodbtest` and `gnodbtest`, where I must admit that I let old versions linger too long. I'll be more diligent about this and do some cleanup in preparation for the release.

You are right about the `psjava` issue. I may as well make it a `psspdb` call that filters only for spdb.

sraaphorst:
Sounds good. The check has been added.
I'd still like to verify connection to ensure successful startup, so I just sent a trivial "echo" command to a running test spdb (instead of "frameworklevel 100") to make sure we can connect properly.

PR #1372, opened 2017-11-21
REL-2784: Fixing exceptions arising from not setting ZoneIds.

sraaphorst:
The obslog was generating exceptions from the update from `SimpleDateFormat` to `DateTimeFormatter`, because `DateTimeFormatter` needs a specified `ZoneId` to be able to extract certain fields from an `Instant` or other `TemporalAccessor`.

I also modernized `ObservationState` to use new `java.time` constructs instead of the old `java.util.Calendar`. See below in comments for more notes on this.

### DIFF
--- a/bundle/edu.gemini.obslog/src/main/resources/openLog.jsp
+++ b/bundle/edu.gemini.obslog/src/main/resources/openLog.jsp
@@ -38,7 +38,7 @@
                 </a>
             </display:column>
 
-            <display:column title="Last Modified" property="lastmodified" sortable="true"/>
+            <display:column title="Last Modified (UTC)" property="lastmodified" sortable="true"/>
 
         </display:table>
 
--- a/bundle/edu.gemini.spdb.reports.collection/src/main/java/edu/gemini/dbTools/odbState/ObservationState.java
+++ b/bundle/edu.gemini.spdb.reports.collection/src/main/java/edu/gemini/dbTools/odbState/ObservationState.java
@@ -173,7 +173,8 @@ public final class ObservationState implements Serializable {
         final StringBuilder buf = new StringBuilder();
         buf.append(startDay).append("/").append(endDay).append(" ");
 
-        final DateTimeFormatter f = DateTimeFormatter.ofPattern("MMM yyyy").withZone(ZoneId.of("UTC"));
+        // Do not need to explicitly set a ZoneId since we are only accessing MMM and yyyy.
+        final DateTimeFormatter f = DateTimeFormatter.ofPattern("MMM yyyy");
         buf.append(f.format(end));
 
         return buf.toString();
--- a/bundle/edu.gemini.spdb.reports.collection/src/main/java/edu/gemini/dbTools/odbState/ObservationState.java
+++ b/bundle/edu.gemini.spdb.reports.collection/src/main/java/edu/gemini/dbTools/odbState/ObservationState.java
@@ -1,6 +1,3 @@
-//
-// $Id: ObservationState.java 46768 2012-07-16 18:58:53Z rnorris $
-//
 package edu.gemini.dbTools.odbState;
 
 import edu.gemini.pot.sp.ISPObservation;
@@ -15,10 +12,12 @@ import org.dom4j.DocumentFactory;
 import org.dom4j.Element;
 
 import java.io.Serializable;
-import java.util.Calendar;
-import java.util.GregorianCalendar;
-import java.text.DateFormat;
-import java.text.SimpleDateFormat;
+import java.time.Instant;
+import java.time.LocalDateTime;
+import java.time.ZoneId;
+import java.time.format.DateTimeFormatter;
+import java.time.temporal.ChronoField;
+import java.time.temporal.ChronoUnit;
 
 public final class ObservationState implements Serializable {
     private static final long serialVersionUID = 2;
@@ -144,42 +143,38 @@ public final class ObservationState implements Serializable {
     }
 
     /**
-     * Gets a formated string that indicates the night during which the
+     * Gets a formatted string that indicates the night during which the
      * observation took place.
      * @param time utc time of the last observation event message
      * @return formatted string indicating the night during which the
      * observation took place.
      */
     private static String _getNight(final long time) {
-        final Calendar cal = GregorianCalendar.getInstance();
-        cal.setTimeInMillis(time);
+        // Determine when, in the local time zone, the time started:
+        // If the time is before noon, then the night started on the previous day.
+        // If the time is after noon, then the night ends on the next day.
+        // Note: AM is 0, PM is 1 by documentation for ChronoField.
+        final LocalDateTime dt = LocalDateTime.ofInstant(Instant.ofEpochMilli(time), ZoneId.systemDefault());
+        final int ampm = dt.get(ChronoField.AMPM_OF_DAY);
 
-        // The calendar is in the local time zone.  If the time is before
-        // noon, then the night started on the previous day.  If the time
-        // is after noon, then the night ends on the next day.
-        final int ampm = cal.get(Calendar.AM_PM);
-
-        final Calendar endCal;
-
-        final int startDay;
-        final int endDay;
-        if (ampm == Calendar.AM) {
-            endCal = (Calendar) cal.clone();
-            endDay = cal.get(Calendar.DAY_OF_MONTH);
-            cal.add(Calendar.DAY_OF_MONTH, -1);
-            startDay = cal.get(Calendar.DAY_OF_MONTH);
+        final LocalDateTime start;
+        final LocalDateTime end;
+        if (ampm == 0) {
+            start = dt.minus(1, ChronoUnit.DAYS);
+            end   = dt;
         } else {
-            startDay = cal.get(Calendar.DAY_OF_MONTH);
-            cal.add(Calendar.DAY_OF_MONTH, 1);
-            endCal = cal;
-            endDay = cal.get(Calendar.DAY_OF_MONTH);
+            start = dt;
+            end   = dt.plus(1, ChronoUnit.DAYS);
         }
+
+        final int startDay = start.get(ChronoField.DAY_OF_MONTH);
+        final int endDay   = end.get(ChronoField.DAY_OF_MONTH);
 
         final StringBuilder buf = new StringBuilder();
         buf.append(startDay).append("/").append(endDay).append(" ");
 
-        final DateFormat f = new SimpleDateFormat("MMM yyyy");
-        buf.append(f.format(endCal.getTime()));
+        final DateTimeFormatter f = DateTimeFormatter.ofPattern("MMM yyyy").withZone(ZoneId.of("UTC"));
+        buf.append(f.format(end));
 
         return buf.toString();
     }
--- a/bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/sitequality/SiteQualityPanel.java
+++ b/bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/sitequality/SiteQualityPanel.java
@@ -8,6 +8,7 @@ import jsky.util.gui.Resources;
 
 import javax.swing.*;
 import javax.swing.table.DefaultTableModel;
+import javax.swing.table.TableRowSorter;
 import javax.swing.text.DefaultFormatter;
 import java.awt.*;
 import java.beans.PropertyChangeEvent;
@@ -15,8 +16,10 @@ import java.beans.PropertyChangeListener;
 import java.text.NumberFormat;
 import java.time.Instant;
 import java.time.ZoneId;
+import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.List;
 
 final class SiteQualityPanel extends JPanel {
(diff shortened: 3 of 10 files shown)


bundle/edu.gemini.spdb.reports.collection/src/main/java/edu/gemini/dbTools/odbState/ObservationState.java
```diff
@@ -144,42 +143,38 @@ public TooType getTooType() {
...
         final StringBuilder buf = new StringBuilder();
         buf.append(startDay).append("/").append(endDay).append(" ");
 
-        final DateFormat f = new SimpleDateFormat("MMM yyyy");
-        buf.append(f.format(endCal.getTime()));
+        final DateTimeFormatter f = DateTimeFormatter.ofPattern("MMM yyyy").withZone(ZoneId.of("UTC"));
+        buf.append(f.format(end));
 
         return buf.toString();
     }
```
sraaphorst: I was slightly worried that I might have screwed something up here, so I executed the following test to compare the old and new output strings from now to 10,000 hours from now, which is over 416 days, and the comparisons always succeeded.

```scala
import java.time._
import java.time.temporal._
import java.time.format._
import java.text._
import java.util._

def tOld(time: Long): String = {
  val dt = LocalDateTime.ofInstant(Instant.ofEpochMilli(time), ZoneId.systemDefault())
  val ampm = dt.get(ChronoField.AMPM_OF_DAY)

  val start = if (ampm == 0) dt.minus(1, ChronoUnit.DAYS) else dt
  val end   = if (ampm == 0) dt else dt.plus(1, ChronoUnit.DAYS)

  val startDay = start.get(ChronoField.DAY_OF_MONTH)
  val endDay   = end.get(ChronoField.DAY_OF_MONTH)

  val df = DateTimeFormatter.ofPattern("MMM yyyy").withZone(ZoneId.systemDefault())
  s"$startDay/$endDay ${df.format(end)}"
}

def tNew(time: Long): String = {
  val cal = Calendar.getInstance()
  cal.setTimeInMillis(time.toLong)

  // The calendar is in the local time zone.  If the time is before
  // noon, then the night started on the previous day.  If the time
  // is after noon, then the night ends on the next day.
  val ampm: Int = cal.get(Calendar.AM_PM)

  var endCal: Calendar = null

  var startDay: Int = 0
  var endDay: Int = 0
  if (ampm == Calendar.AM) {
    endCal = cal.clone.asInstanceOf[Calendar]
    endDay = cal.get(Calendar.DAY_OF_MONTH)
    cal.add(Calendar.DAY_OF_MONTH, -1)
    startDay = cal.get(Calendar.DAY_OF_MONTH)
  }
  else {
    startDay = cal.get(Calendar.DAY_OF_MONTH)
    cal.add(Calendar.DAY_OF_MONTH, 1)
    endCal = cal
    endDay = cal.get(Calendar.DAY_OF_MONTH)
  }

  val f: DateFormat = new SimpleDateFormat("MMM yyyy")
  s"$startDay/$endDay ${f.format(endCal.getTime)}"
}

for (i <- Range(0,10000)) {
  val time = Instant.now().plus(i, ChronoUnit.HOURS).toEpochMilli
  assert(tOld(time).equals(tNew(time)))
}
```

sraaphorst: The `var endCal` is because I copy / pasted from Java and allowed it to just convert. ☺️

bundle/edu.gemini.spdb.reports.collection/src/main/java/edu/gemini/dbTools/odbState/ObservationState.java
```diff
@@ -144,42 +143,38 @@ public TooType getTooType() {
...
+        final int startDay = start.get(ChronoField.DAY_OF_MONTH);
+        final int endDay   = end.get(ChronoField.DAY_OF_MONTH);
+
         final StringBuilder buf = new StringBuilder();
         buf.append(startDay).append("/").append(endDay).append(" ");
 
-        final DateFormat f = new SimpleDateFormat("MMM yyyy");
-        buf.append(f.format(endCal.getTime()));
+        final DateTimeFormatter f = DateTimeFormatter.ofPattern("MMM yyyy").withZone(ZoneId.of("UTC"));
+        buf.append(f.format(end));
```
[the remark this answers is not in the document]
sraaphorst: The old code seemed odd to me. We got a time in UTC, and then shoved it into a `GregorianCalendar` that has the local time zone set, I guess it was to determine if the time in the local timezone was AM / PM.

We then print using the formatter.
Actually, this seems to work properly if we cut off the `DateTimeFormatter...withZone` just fine, so perhaps we should leave it off to ensure consistency.

The hours test I included passes either way. Seems strange, but I won't pretend to fully understand the dynamics of the new Java time API.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/util/TimeZonePreference.scala
```diff
@@ -8,9 +9,13 @@ import java.util.prefs.Preferences
...
-  val pref = Preferences.userNodeForPackage(TimeZonePreference.getClass)
+  val pref: Preferences = Preferences.userNodeForPackage(TimeZonePreference.getClass)
 
   def get: TimeZone = TimeZone.getTimeZone(pref.get("TimeZone", "UTC"))
 
   def set(tz: TimeZone): Unit = pref.put("TimeZone", tz.getID)
+
+  def getZoneId: ZoneId = get.toZoneId
+
+  def setZoneId(zi: ZoneId): Unit = pref.put("TimeZone", TimeZone.getTimeZone(zi).getID)
```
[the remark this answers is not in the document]
sraaphorst: Sure. One calling the other makes more sense. I'll make that change.

sraaphorst:
Hmmm... I don't think we should use `TimeZonePreference` as I did, since the only place `set` is ever called is in the `ParallacticAngleDialog`, and that doesn't seem right. I think the `ImagePrintDialog`, to be consistent, should be using the default `ZoneId`.

sraaphorst:
Yes, using the default `ZoneId` instead for `ImagePrintDialog` yields consistent results with the `OT 2017B.1.1.1`, so let's go with that.

bundle/edu.gemini.spdb.reports.collection/src/main/java/edu/gemini/dbTools/odbState/ObservationState.java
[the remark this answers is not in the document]
swalker2m: I suppose it works without the time zone because the time zone information is incorporated in the `LocalDateTime`.  What I'm not sure of is what time zone is used for the formatting in this case.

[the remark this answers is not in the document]
sraaphorst: Investigating further, it seems like in the original case (as well as the new case whether I specify with `ZoneId.of("UTC")` or nothing) the time is being printed in UTC.

bundle/edu.gemini.obslog/src/main/resources/openLog.jsp
```diff
@@ -38,7 +38,7 @@
                 </a>
             </display:column>
 
-            <display:column title="Last Modified" property="lastmodified" sortable="true"/>
+            <display:column title="Last Modified (UTC)" property="lastmodified" sortable="true"/>
```
swalker2m: I suppose the sorting here will also be screwed up, but also think it isn't very high priority.

bundle/edu.gemini.spdb.reports.collection/src/main/java/edu/gemini/dbTools/odbState/ObservationState.java
```diff
@@ -144,42 +143,39 @@ public TooType getTooType() {
...
 
+        final int startDay = start.get(ChronoField.DAY_OF_MONTH);
+        final int endDay   = end.get(ChronoField.DAY_OF_MONTH);
+
         final StringBuilder buf = new StringBuilder();
         buf.append(startDay).append("/").append(endDay).append(" ");
 
-        final DateFormat f = new SimpleDateFormat("MMM yyyy");
-        buf.append(f.format(endCal.getTime()));
+        // Do not need to explicitly set a ZoneId since we are only accessing MMM and yyyy.
```
swalker2m: I think the reason a `ZoneId` isn't needed is because we're formatting a `LocalDateTime`.  Even the month and year will differ for an instant in distinct time zones at some times.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/sitequality/SiteQualityPanel.java
```diff
@@ -296,7 +301,9 @@ public void updateEnabledState(boolean enabled) {
     private static final long MS_PER_MINUTE = MS_PER_SECOND * 60;
     private static final long MS_PER_HOUR = MS_PER_MINUTE * 60;
 
-    private static final DateTimeFormatter dateFormat = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss z").withZone(ZoneId.of("UTC"));
+    private static final DateTimeFormatter dateFormat = DateTimeFormatter.ofPattern("yyyy-MMM-dd HH:mm:ss z")
+            .withZone(ZoneId.of("UTC"));
+    public static final Comparator<String> START_DATE_COMPARATOR = Comparator.comparingLong(s -> ZonedDateTime.parse(s, dateFormat).toEpochSecond());
```
swalker2m: I think the `TimingWIndowTableModel` should provide the `Instant` in the call for `getValueAt` instead of a formatted `String` and then we wouldn't need a special comparator that has to parse the formatted `String`.  Unfortunately that's a bigger change though so it's not a good idea to do it now.  Perhaps we could though use the *same* `dateFormat` used by the `TimingWindowTableModel` so that there isn't a hidden dependency between the sorter and the formatter.  We'd need to extract the `DateTimeFormatter` from `TimingWindowTableModel` so it could be used by both.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/too/TooAlertPanel.java
```diff
@@ -12,17 +8,16 @@
...
-import java.text.DateFormat;
-import java.text.SimpleDateFormat;
 
 /**
  * A panel that displays a TOO alert according to the spec in SCT-233.
  */
 public final class TooAlertPanel extends JPanel {
     private static final String TIME_ONLY_PATTERN = "HH:mm:ss";
-    private static final String TIME_DATE_PATTERN = "yyyy/MM/dd HH:mm:ss";
+    private static final String TIME_DATE_PATTERN = "YYYY-MMM-dd HH:mm:ss";
```
swalker2m: We should probably follow up on this after the release to extract this pattern in the OT so we don't have to repeat it in a lot of places.

bundle/edu.gemini.spdb.reports.collection/src/main/java/edu/gemini/dbTools/odbState/ObservationState.java
```diff
@@ -144,42 +143,39 @@ public TooType getTooType() {
...
+        final DateTimeFormatter f = DateTimeFormatter.ofPattern("MMM yyyy");
+        buf.append(f.format(end));
```
swalker2m: > Investigating further, it seems like in the original case (as well as the new case whether I specify with ZoneId.of("UTC") or nothing) the time is being printed in UTC.

I think this version of the code with the new time API is better than the previous one but I don't understand why it would format the date in UTC.  It appears to me that both cases should be in local time.  For Gemini South that is going to correspond to the same "night" as the time in UTC given that we finish observing before noon UTC.  At Gemini North, I'm not so sure.

At any rate, it doesn't appear that this formatted `String` or the corresponding night is ever used by anything (the `_night` field is private and there is no accessor) so it probably doesn't matter.

bundle/edu.gemini.obslog/src/main/resources/openLog.jsp
sraaphorst: Good catch.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/too/TooAlertPanel.java
sraaphorst: Absolutely. This is repeated over and over again in so many places. `DateUtil` should be reworked, I'm thinking, to have `DateTimeFormatter`s so that we don't have to declare these over and over again.

I'll make a task for this so we don't forget it.

bundle/edu.gemini.spdb.reports.collection/src/main/java/edu/gemini/dbTools/odbState/ObservationState.java
sraaphorst: I was a bit confused by how this was supposed to work and why we need a night in the first place, but it seemed "wonky" to me in its original format, so I just tried to modernize without changing the functionality, since it didn't seem logical to me.

It does seem like perhaps this can all just be removed, but as we're six days from release, I don't want to muck something unexpected up by doing so.

sraaphorst:
Sure. I added the JIRA task as REL-3311 and will make the changes for the `DateTimeFormatter` and we can hopefully produce a test release this afternoon.

PR #1374, opened 2017-11-30
REL-3317: don't clobber ToO 'override rapid'

swalker2m:
This PR addresses a bug in which all standard ToO observations are reset to rapid ToO in a rapid ToO program if the program admin dialog is opened and the "OK" button is clicked.

Determining the ToO status of an observation involves consulting the ToO designation of the program as a whole and the value of a "override rapid too" flag in the observation itself.  The combination of these values determines the ToO status of the observation.  It is probably most concisely described in code:

```java
    public static TooType get(ISPObservation obs) {
        final TooType progType = get(obs.getProgram());
        if (progType == TooType.rapid) {
            final boolean override = ((SPObservation) obs.getDataObject()).isOverrideRapidToo();
            return override ? TooType.standard : TooType.rapid;
        } else {
            return progType;
        }
    }
```

This PR avoids attempting to update the ToO status of the program unless it has actually changed, and removes code that would reset the override flag in every observation to `false` every time regardless.  This means that now when a program toggles from `TooType.rapid` to either `standard` or `none` and then back to `rapid`, observations that were previously set to `standard` remain `standard`.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/too/Too.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/too/Too.java
@@ -50,30 +50,12 @@ public final class Too {
     /**
      * Sets the ToO type of the program and all of its observations to match.
      */
-    public static boolean set(ISPProgram prog, TooType newType) {
+    public static void set(ISPProgram prog, TooType newType) {
         // REL-538 Move TooType to SPProgram
-        SPProgram spProgram = (SPProgram) prog.getDataObject();
-        spProgram.setTooType(newType);
-        prog.setDataObject(spProgram);
-
-        updateObservations(prog);
-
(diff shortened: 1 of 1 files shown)


sraaphorst:
This looks right to me. I'm guessing it's something we're going to want to release in a 1.1.2 patch at some point? Am I right in assuming it only affects the apps and not the ODB?

swalker2m:
> I'm guessing it's something we're going to want to release in a 1.1.2 patch at some point?

I'm not sure.  It has apparently been broken for years so maybe it can wait until June.  For now I'll just put it in the develop branch.

> Am I right in assuming it only affects the apps and not the ODB?

For the most part.  It would have an impact any place that we set the ToO status.  In practice that is the OT Admin dialog and the skeleton creator, which is part of the ODB itself.  In the case of skeleton creation though the program is new and so either version would be fine.

PR #1375, opened 2017-11-30
REL-3311: Cleaning up date time formatting, util functions, and timing window table.

sraaphorst:
This is a first crack at the following:

1. Removing cruft from both the `DateUtil` and `TimeUtils` class and making a new scala class, `DateTimeUtils` to replace these both.

2. Removing reliance on mathematical calculations in `TimeUtils` such as `TimeUtils.hours` and `TimeUtils.asHours` to convert to / from hours and ms by instead using `scala.concurrent.duration._` calls instead for clarity.

3. Removed much reliance on old `java.util` time classes, such as `Calendar` and `SimpleDateFormat` in lieu of `ZonedDateTime`, `Instant`, and `DateTimeFormatter`.

4. Repaired the `TimingWindowTableModel` to no longer inherently format the `Instant` that defines the timing window. We instead use `TableCellRenderer` to render this instant, which allows us to sort on the instant without requiring any special handling.


bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/ui/html/ScheduleDocument.java
```diff
@@ -175,10 +175,10 @@ private String getLaserLimits(Observation observation, ClearanceWindow clearance
         String fmt = "HH:mm:ss";
         LaserTarget scienceTarget = observation.getScienceTarget().getLaserTarget();
         for (Visibility.Interval visibility : scienceTarget.getVisibility().getAboveLaserLimit()) {
-            long vStart = visibility.getStart().getTime() / 1000; // get rid of milliseconds
-            long vEnd   = visibility.getEnd().getTime() / 1000;
-            long cStart = clearance.getStart().getTime() / 1000;
-            long cEnd   = clearance.getEnd().getTime() / 1000;
+            long vStart = visibility.getStart().getTime() / DateTimeUtils.MillisecondsPerSecond(); // get rid of milliseconds
```
cquiroz: Why is `MillisecodnsPerSecond` a function. Sounds like it should be a constant

cquiroz: Ahh, because it is called from Java

bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/charts/HoursChart.scala
```diff
@@ -79,14 +83,9 @@ case class HoursChart(ctx: QvContext, nights: Seq[Night], observations: Set[Obs]
...
-  private def moonHours(n: Night): Option[Double] = {
-    val dur = n.moonAboveHorizon.restrictTo(n.scienceTime).duration // moon hours during science time
-    Some(TimeUtils.asHours(dur))
-  }
-
-  private def hours(o: Obs, s: Solution)(n: Night): Option[Double] = {
-    val hrs = TimeUtils.asHours(s.restrictTo(n.interval).duration)
-    Some(hrs)
-  }
+  private def moonHours(n: Night): Option[Double] =
```
cquiroz: Should we use a Time based type here instead of `Double`?

bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/selector/ScheduleEditor.scala
```diff
@@ -245,8 +252,8 @@ class ScheduleEditor(ctx: QvContext, scheduleCache: ScheduleCache) extends Dialo
       start <- cd.startDate
       end <- cd.endDate
     } yield {
-      val s = start + TimeUtils.hours(14)
-      val e = end + TimeUtils.hours(14) + TimeUtils.days(2)
+      val s = start + 14.hours.toMillis
```
cquiroz: What's the 14 hours? Does it mean 2PM as the time our day shifts?
Could we use a time based type rather than millis?

bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/selector/TimeRangeSelector.scala
```diff
@@ -1,25 +1,28 @@
...
 object TimeRangeSelector {
   abstract class RangeType(val duration: Long, val increment: Long)
-  object Day extends RangeType(TimeUtils.days(1), TimeUtils.days(1))
-  object Week extends RangeType(TimeUtils.days(7), TimeUtils.days(1))
-  object Month extends RangeType(TimeUtils.days(30), TimeUtils.days(1))
-  object Quarter extends RangeType(TimeUtils.days(92), TimeUtils.days(30))
-  object HalfYear extends RangeType(TimeUtils.days(183), TimeUtils.days(30))
-  object Year extends RangeType(TimeUtils.days(365), TimeUtils.days(30))
-  object Custom extends RangeType(0, 0)
+  object Day      extends RangeType(1.day.toMillis,     1.day.toMillis)
```
cquiroz: I think RangeType could also use `Duration` rather than millis? i'm not sure as you'd be carrying nanos but maybe it is a good idea

bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/util/SemesterData.scala
```diff
@@ -111,7 +111,7 @@ object SemesterData {
 
   def current(site: Site, currentTime: Long = System.currentTimeMillis()): SemesterData = semesters(site, currentTime, currentTime).head
 
-  def next(site: Site): SemesterData = semesters(site, System.currentTimeMillis(), System.currentTimeMillis() + TimeUtils.days(185))(1)
+  def next(site: Site): SemesterData = semesters(site, System.currentTimeMillis(), System.currentTimeMillis() + 185.days.toMillis)(1)
```
cquiroz: 185 days? Maybe we should have a constant for that

bundle/edu.gemini.shared.util/src/main/scala/edu/gemini/shared/util/DateTimeUtils.scala
```diff
@@ -0,0 +1,86 @@
...
+import java.util.{Comparator, Date, TimeZone}
+import java.util.function.ToLongFunction
+
+import scala.concurrent.duration._
+import scalaz._
+import Scalaz._
+
+object DateTimeUtils {
+  val UTC: ZoneId         = ZoneId.of("UTC")
+  val DefaultZone: ZoneId = ZoneId.systemDefault()
```
cquiroz: This will use the timezone of the system this is being run. e.g. in the case of the OT it will be the user's TZ. This maybe what you intend but I just want to be sure as the result could be surprising

bundle/edu.gemini.shared.util/src/main/scala/edu/gemini/shared/util/DateTimeUtils.scala
```diff
@@ -0,0 +1,86 @@
...
+    val neg = span < 0
+    val d   = Math.abs(span).milliseconds
+    val hh  = d.toHours
+    val mm  = d.toMinutes % MinutesPerHour
+    val ss  = d.toSeconds % SecondsPerMinute
+    f"${neg ? "-" | ""}%s$hh%02d:$mm%02d:$ss%02d"
+  }
+
+  // Some useful constants.
+  val StartOfDayHour:        Int  = 14
```
cquiroz: Let's use this constant on the various places `14` is used

cquiroz:
Thanks a lot for this heroic task 😄 
As a general comment I'd promote to avoid having miliseconds or hours as doubles and try to use higher level types where possible. We have also magic numbers in many places which could be replaced by constants instead

swalker2m:
> As a general comment I'd promote to avoid having miliseconds or hours as doubles and try to use higher level types where possible.

I completely, 100% agree for new code.  That could be a pretty big change that ripples through the old codebase though.  At any rate, this is already a huge PR so maybe we should save any additional refactoring for future PRs?

bundle/edu.gemini.shared.util/src/main/scala/edu/gemini/shared/util/DateTimeUtils.scala
sraaphorst: Yes, this is actually deliberate, and has been extracted from a couple of places in the code.

sraaphorst:
Agree with Shane: this is a task that could go on for as long as we want it to, really. I'll add these notes to the JIRA task as I think they're all good.

bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/selector/ScheduleEditor.scala
sraaphorst: I think the 14 hours refers to the start of day hour as per later on.
You're right that it should be changed to use the constant.
There are so many repeated constants.

sraaphorst:
More than anything, I wanted to get a start on this task and make a PR to work off of where I knew the test cases were passing and the basic framework for moving forward was in place; thus, really, I'm looking for any comments about `DateTimeUtils` and using `scala.concurrent.duration._` instead of how we were doing things before.

cquiroz:
Yes it could spiral to ripple across the whole codebase. So let's do it in an best effort basis.
This is already a great improvement

bundle/edu.gemini.shared.util/src/main/scala/edu/gemini/shared/util/DateTimeUtils.scala
sraaphorst: Agreed that this is a good idea, and making the change now.

bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/selector/ScheduleEditor.scala
tpolecat: Seems like this will be wrong twice a year at the DST switchover because the observing day changes at 2pm *local* time … you can't express it in epoch time. If this is the case maybe open another issue, it seems minor.

bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/selector/TimeRangeSelector.scala
tpolecat: Agree, and it seems like this needs to be parameterized by year to deal with leap years. Maybe investigate as a follow-up because this PR is pretty big already.

bundle/edu.gemini.shared.util/src/main/scala/edu/gemini/shared/util/DateTimeUtils.scala
tpolecat: Can we rename this `SystemDefault` so it's a little bit more clear what it means? It should rarely be used.

bundle/edu.gemini.shared.util/src/main/scala/edu/gemini/shared/util/DateTimeUtils.scala
```diff
@@ -0,0 +1,86 @@
...
+object DateTimeUtils {
+  val UTC: ZoneId         = ZoneId.of("UTC")
+  val DefaultZone: ZoneId = ZoneId.systemDefault()
+
+  // Standard UTC formatters.
+  val YYYY_MMM_DD_Formatter: DateTimeFormatter = DateTimeFormatter.ofPattern("yyyy-MMM-dd").withZone(UTC)
+  val YYYY_MMM_DD_HHMM_Formatter: DateTimeFormatter = DateTimeFormatter.ofPattern("yyyy-MMM-dd HH:mm").withZone(UTC)
+  val YYYY_MMM_DD_HHMMSS_Formatter: DateTimeFormatter = DateTimeFormatter.ofPattern("yyyy-MMM-dd HH:mm:ss").withZone(UTC)
+  val YYYY_MMM_DD_HHMMSS_Z_Formatter: DateTimeFormatter = DateTimeFormatter.ofPattern("yyyy-MMM-dd HH:mm:ss z").withZone(UTC)
+  val HHMMSS_Formatter: DateTimeFormatter = DateTimeFormatter.ofPattern("HH:mm:ss").withZone(UTC)
```
tpolecat: Would it make sense to have a `Formatters` case class parameterized by zone, and then instances for UTC and system default? You could then say `DateTimeUtils.Formatters.UTC.YYYY_MMM_DD_HHMMSS_Z` and it would be pretty clear what's happening at the call site. As-is you just kind of have to know that it's UTC.

bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/util/SemesterData.scala
tpolecat: Is this correct? What about leap years?

swalker2m:
> We should make a list of things touched here because I doubt test coverage is sufficient to reveal mistakes. Or add more tests I guess, but it may be hard.

I want to second this.  This is great work but I don't feel confident that there are no new problems as a result of these changes.  It'll likely lie dormant until May and even then people won't know to test all the things that this will have touched.

bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/selector/TimeRangeSelector.scala
sraaphorst: I created a separate JIRA task for this:
http://jira.gemini.edu:8080/browse/REL-3319

bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/util/SemesterData.scala
sraaphorst: I'm not even sure if the `semesters` method is doing what it claims to be doing:
```scala
  /**
   * Gets all available semesters covering the given time frame.
   * @return
   */
  def semesters(site: Site, start: Long, end: Long): Seq[SemesterData] =
    semesters(site).
      filter(sd => sd.end > start && sd.start < end)
```

It seems like this actually gets all the semesters intersecting the given time frame, rather than **covering** the given time frame?

I would think a semester would cover `[start, end]` iff `sd.start <= start && sd.end >= end`?

I'm kind of confused by this original logic.

bundle/edu.gemini.shared.util/src/main/scala/edu/gemini/shared/util/DateTimeUtils.scala
sraaphorst: @tpolecat That makes more sense. I'll do that.

bundle/edu.gemini.shared.util/src/main/scala/edu/gemini/shared/util/DateTimeUtils.scala
sraaphorst: Unfortunately, this isn't available to `edu.gemini.spModel.core.Semester`, where I'm guessing `SWITCH_OVER_HOUR = 14` means the same thing.

bundle/edu.gemini.shared.util/src/main/scala/edu/gemini/shared/util/DateTimeUtils.scala
sraaphorst: Hmmm. I think I like this idea better. I'll take a look at that.

bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/charts/HoursChart.scala
```diff
@@ -79,14 +83,9 @@ case class HoursChart(ctx: QvContext, nights: Seq[Night], observations: Set[Obs]
...
+    (n.moonAboveHorizon.restrictTo(n.scienceTime).duration.toDouble / DateTimeUtils.MillisecondsPerHour).some // moon hours during science time
 
+  private def hours(o: Obs, s: Solution)(n: Night): Option[Double] =
+    (s.restrictTo(n.interval).duration.toDouble / DateTimeUtils.MillisecondsPerHour).some
```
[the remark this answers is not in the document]
sraaphorst: I was wondering the same thing myself.

bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/filter/core/Filter.scala
```diff
@@ -243,7 +247,7 @@ object Filter {
   // =======================  *** Set time filter ***
   case class SetTime(ctx: QvContext, min: Double = 0, max: Double = 15) extends SimpleRangeFilter {
     def label = "Set Time Hrs"
-    def getter = (o: Obs, ctx: QvContext) => TimeUtils.asHours(SolutionProvider(ctx).remainingHours(ctx, o).getOrElse(min.toLong))
+    def getter = (o: Obs, ctx: QvContext) => SolutionProvider(ctx).remainingHours(ctx, o).getOrElse(min.toLong).toDouble / DateTimeUtils.MillisecondsPerHour
```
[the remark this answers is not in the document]
sraaphorst: Sure thing. Happy to clarify anything I can.

bundle/edu.gemini.services.server/src/main/scala/edu/gemini/services/server/telescope/LttsService.scala
```diff
@@ -15,21 +19,22 @@ import scala.xml.XML
...
-    val e = TimeUtils.print(range.end, site.timezone, "yyyyMMdd")
+    val df = DateTimeUtils.YYYYMMDD_Formatter.withZone(site.timezone.toZoneId)
+    val s  = df.format(Instant.ofEpochMilli(range.start))
+    val e  = df.format(Instant.ofEpochMilli(range.end))
 
 
     val client = new DefaultHttpClient()
     val request = new HttpGet(s"http://$lttsHost:8080/ltts/services/nights?from=$s&to=$e")
     val response = client.execute(request)
     val sdf = new SimpleDateFormat("yyyy-MM-dd'T'HHmmss.SSSZ")
```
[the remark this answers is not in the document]
sraaphorst: Ooops.... no, I don't think we do. Good point. I was keeping it around for comparing results.
May as well get rid of it.

bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/util/SemesterData.scala
sraaphorst: I think 185 would always work, no? Even for leap years, 366 / 2 = 183, so 185 should be sufficient to advance a semester. My concern would be the potential to advance two semesters accidentally if we were on the last day of a semester.

This code definitely doesn't seem appropriate, but I think trying to look at its actual functionality and correct it if it is wrong should be a separate JIRA task.

bundle/edu.gemini.shared.util/src/main/scala/edu/gemini/shared/util/DateTimeUtils.scala
sraaphorst: Okay, I parameterized the formatters, and I like it better this way: this forces a time zone to be used and avoids repetitive declarations.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/obs/TimingWindowSolver.java
```diff
@@ -16,7 +16,7 @@
 	private final boolean value;
 
 	public TimingWindowSolver(List<TimingWindow> windows, boolean value) {
-		super(TimeUtils.MS_PER_HOUR / 4, TimeUtils.MS_PER_MINUTE);
+		super(DateTimeUtils.MillisecondsPerHour() / 4, DateTimeUtils.MillisecondsPerMinute());
```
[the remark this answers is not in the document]
sraaphorst: Yes, this is much better, and I switched it over in the recent commit.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/ui/html/ScheduleDocument.java
```diff
@@ -268,11 +268,7 @@ public boolean isEmpty(Object o) {
 
     public SortedSet<Marker> getMarkers(Alloc a) {
         SortedSet<Marker> ret = a.getMarkers();
-        for (Iterator<Marker> it = ret.iterator(); it.hasNext(); ) {
-            Marker marker = it.next();
-            if (marker.isQcOnly() && !showQcMarkers)
-                it.remove();
-        }
+        ret.removeIf(m -> m.isQcOnly() && !showQcMarkers);
```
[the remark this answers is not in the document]
sraaphorst: I'm not 100% sure what you mean about pulling it out. Could you elucidate a bit further?

bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/charts/HistogramChart.scala
```diff
@@ -66,11 +69,11 @@ class HistogramChart(ctx: QvContext, xAxis: Axis, yAxis: Axis, calculation: Calc
 
     // try to use the same color for each group even if some groups are added/removed
     // (use the indices of the visible groups in the seq of all groups on the y-axis)
-    visibleYGroups.zipWithIndex.foreach({case (f, ix) => {
+    visibleYGroups.zipWithIndex.foreach({case (f, ix) =>
```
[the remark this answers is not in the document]
sraaphorst: 👍

bundle/edu.gemini.services.server/src/main/scala/edu/gemini/services/server/telescope/LttsService.scala
```diff
@@ -15,21 +19,22 @@ import scala.xml.XML
...
+    val dfp = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HHmmss.SSSZ")
 
     // Get the response
-    val ltts = XML.load(new InputStreamReader(response.getEntity().getContent()))
+    val ltts = XML.load(new InputStreamReader(response.getEntity.getContent))
     val nights = (ltts \\ "night").map(night => {
-      val start = sdf.parse((night \ "start" text) replaceAllLiterally (":", "")).getTime
-      val end   = sdf.parse((night \ "end" text) replaceAllLiterally (":", "")).getTime
+      val start = ZonedDateTime.parse((night \ "start" text) replaceAllLiterally (":", ""), dfp).toInstant.toEpochMilli
+      val end   = ZonedDateTime.parse((night \ "end" text) replaceAllLiterally (":", ""),   dfp).toInstant.toEpochMilli
```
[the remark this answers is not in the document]
sraaphorst: I'm going to look at adding some test cases for a few of the more substantial changes, I think.

sraaphorst:
I cleaned up quite a bit with regards to the very useful comments (thanks all for looking at this) and will make a list of things that it touches that will be required for testing and attach it to the JIRA task and to this PR.

I'm also going to adjust REL-3319 to include a more detailed examination of date handling in the QV, as we seem to all agree that this is probably not being handled correctly.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/ui/html/ScheduleDocument.java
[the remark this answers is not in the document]
swalker2m: Sorry, I meant that `showQcMarkers` doesn't depend on anything in the collection so wouldn't need to be checked with every element.  It's no big deal, and maybe the code is worse but as long as we're mutating anyway:

```java
if (!showQcMarkers) {
}

sraaphorst:
Ooops. I forgot to add the new `DateFormatting` trait to github, hence the Travis errors.
Hopefully this will compile and tests will pass now.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/ui/html/ScheduleDocument.java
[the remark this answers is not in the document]
sraaphorst: 👍

PR #1381, opened 2017-12-07
REL-3323: Fix QPT NPE after deleting a variant

sraaphorst:
In recent changes, when a variant in the QPT is deleted, the `CandidateObsComparator` causes a `NullPointerException` due to the fact that the selected `Variant` is then null and it tries to retrieve a comparator from it.

This change just instead uses a dummy comparison, since there are no `Obs` to be compared when there is no selected `Variant`.


sraaphorst:
Something weird going on with this PR (8 commits - should be 3).
I will fix that and re-create it.

PR #1382, opened 2017-12-07
REL-3323: Fix QPT NPE after deleting a variant (for 2018A)

sraaphorst:
In recent changes, when a variant in the QPT is deleted, the `CandidateObsComparator` causes a `NullPointerException` due to the fact that the selected `Variant` is then `null` and it tries to retrieve a comparator from it.

This change just instead uses a dummy comparison, since there are no `Obs` to be compared when there is no selected `Variant`.

### DIFF
--- a/bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/ui/view/candidate/CandidateObsComparator.java
+++ b/bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/ui/view/candidate/CandidateObsComparator.java
@@ -34,7 +34,7 @@ public class CandidateObsComparator extends MouseAdapter implements GComparator<
 			return direction * ((ret != 0) ? ret : -o1.compareTo(o2));
 		}
 
-		// If we don't have a variant, then we don't have a middle point.
+        // If we don't have a variant, then we don't have a middle point.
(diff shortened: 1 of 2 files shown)


bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/ui/view/candidate/CandidateObsComparator.java
```diff
@@ -33,7 +33,13 @@ public int compare(Obs o1, Obs o2) {
 			int ret = (int) Math.signum(variant.getScore(o1) - variant.getScore(o2));
 			return direction * ((ret != 0) ? ret : -o1.compareTo(o2));
 		}
-		
+
+		// If we don't have a variant, then we don't have a middle point.
+        // This will happen when we've just deleted a variant, for example.
```
cquiroz: Indentation seems off

sraaphorst:
Odd. It looked fine in my source code, and when I deleted / repasted the line, the git diff was the same. I finally just cut / repasted the whole comment block and now the indentation seems fine.

👍  for noticing this.

tpolecat:
Grim. Programming is hard 👍

PR #1384, opened 2017-12-07
REL-3310: Fixing script to run spdb as software instead of root.

sraaphorst:
The original `spdb` `init.d` script accidentally started the `spdb` as `root` when it should have been started as `software`. This should fix that according to my research.

(Note: I cannot test this directly because `/sbin/runuser` is only executable as `root`.)

### DIFF
--- a/app/spdb/scripts/spdb
+++ b/app/spdb/scripts/spdb
@@ -159,7 +159,7 @@ start ()
 
 	    # Attempt to start.
 	    printf "Starting ODB ${SPDB_VERSION_STR}"
-	    "${SPDB_EXEC}" 2>&1 > "${SPDB_DIR}/sysout.txt" &
+	    /sbin/runuser software -s /bin/bash -c "nohup ${SPDB_EXEC} 2>&1 > ${SPDB_DIR}/sysout.txt &"
 	    sleep_dots
 	    printf "\n"
 


app/spdb/scripts/spdb
```diff
@@ -159,7 +159,7 @@ start ()
 
 	    # Attempt to start.
 	    printf "Starting ODB ${SPDB_VERSION_STR}"
-	    "${SPDB_EXEC}" 2>&1 > "${SPDB_DIR}/sysout.txt" &
+	    /sbin/runuser software -s /bin/bash -c "nohup ${SPDB_EXEC} 2>&1 > ${SPDB_DIR}/sysout.txt &"
```
cquiroz: Is it possible to use `sudo -u software` and test it that way?

sraaphorst: I was reading up on it and apparently, in some cases, `sudo` asks for a password (even when run as root), which obviously can't be provided in an init.d script, although that was one solution that I did see proposed. The advantage of `runuser` is that it never asks for a password as it can only be run as `root`.

cquiroz: ok, sounds good. i guess this is difficult to test in so perhaps you can put the new script in gsnodb test and restart to verify

sraaphorst:
I think that's the best idea: I'll ask ITS to put the script in gnodbtest or gsodbtest and restart the VM.

PR #1386, opened 2017-12-13
REL-2928: respect ITAC allocated time

swalker2m:
This is a bit of a rush bug fix for ITAC proposal imports.  REL-2928 began keeping track of program vs. partner time.  It also introduced a bug in that the total awarded time is no longer being taken from the ITAC award itself, but rather from the sum of observation times.


tpolecat:
Hours as `Double` makes my hands sweat, but this looks fine.

sraaphorst:
Agreed with @tpolecat. There are a good number of places where we have hours expressed as `Double` and cannot convert to durations because we need the representation in hours and need to multiply before rounding. Not the best way to do things.

swalker2m:
Thanks guys.  It looks like we're not going to try to fix this in production after all.  I'll make an identical PR for the develop branch though.

swalker2m:
It was decided to just fix the awarded time manually for now.  The same update is, however, merged into the `develop` branch so that they won't have to manually adjust the times next semester as well.

PR #1388, opened 2017-12-14
REL-3332: Change ITC image URLs from HTTP to HTTPS

andrewwstephens:
Update URLs for images displayed in the ITC web form to avoid browser security warnings.


cquiroz:
I checked some of the links and they seem to be working just fine

PR #1389, opened 2017-12-15
REL-3333: Remove visualization of CWFS probe arms and regions covered by probes.

sraaphorst:
This addresses REQ-OT-002 of the NGS2 software document, i.e. it removes the CWFS probe arms and the region covered by each probe from the TPE.

As per Bryan's comment, these can be removed for existing observations.

This allowed considerable simplification of the CWFS code and a bit of cleanup of the TPE, as there were some items (such as a key legend indicating the CWFS1 vs. CWFS2 probe ranges) that were only used by CWFS.

Please see further comments below in the code.


bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/flamingos2/Flamingos2OiwfsGuideProbe.java
```diff
@@ -178,7 +175,4 @@ public VignettingCalculator calculator(ObsContext ctx) {
                 F2OiwfsProbeArm$.MODULE$,
                 F2ScienceAreaGeometry$.MODULE$);
     }
-
-    @Override
-    public BandsList getBands() { return RBandsList.instance(); }
```
sraaphorst: Since most of the probes have this exact method definition for `getBands`, I moved this as the `default` implementation in the `GuideProbe` class to clean up repeated code.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gems/Canopus.java
```diff
@@ -28,111 +25,36 @@
...
     /**
      * Canopus guide probe options.
-     * <br>
-     * Note for validate:<br>
-     * CWFS1 can vignette both CWFS2 and CWFS3<br>
-     * CWFS2 can vignette only CWFS3<br>
-     * CWFS3 does not vignette other probe arms<br>
+     * Vignetting is no longer an issue due to move to NGS2.
      */
     public enum Wfs implements GuideProbe, ValidatableGuideProbe, OffsetValidatingGuideProbe {
```
sraaphorst: Given the changes, there is now very little that is unique to each of the CWFS probes, so I was able to simplify these dramatically.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gems/Canopus.java
```diff
@@ -227,154 +145,99 @@ public GuideStarValidation validate(SPTarget guideStar, ObsContext ctx) {
             ).getOrElse(GuideStarValidation.UNDEFINED);
         }
 
-        private static GuideStarValidation validateVignetting(Wfs wfs, SPTarget guideStar, ObsContext ctx) {
```
sraaphorst: Since we are not using probe arms anymore, we can get rid of the vignetting code in validation.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gems/Canopus.java
```diff
@@ -227,154 +145,99 @@ public GuideStarValidation validate(SPTarget guideStar, ObsContext ctx) {
...
-                        double p = -dis.p().toArcsecs().getMagnitude();
-                        double q = -dis.q().toArcsecs().getMagnitude();
-                        Set<Offset> offsets = new TreeSet<>();
-                        offsets.add(offset);
-
-                        Area a = Canopus.instance.offsetIntersection(ctx, offsets);
-                        return a != null && a.contains(p, q);
-                    }))
-                .getOrElse(false);
+        public boolean inRange(final ObsContext ctx, final Offset offset) {
```
sraaphorst: This just cleans up the code to use `Option`s more cleanly instead of having to do calls like `opt.isEmpty` and `opt.getValue`.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gems/Canopus.java
```diff
@@ -227,154 +145,99 @@ public GuideStarValidation validate(SPTarget guideStar, ObsContext ctx) {
...
 
     // Gets the primary CWFS 3 guide star, if any.
-    private Option<SPTarget> getPrimaryCwfs3(ObsContext ctx) {
-        TargetEnvironment env = ctx.getTargets();
-        if (env == null) return None.instance();
-
-        Option<GuideProbeTargets> gtOpt = env.getPrimaryGuideProbeTargets(Canopus.Wfs.cwfs3);
-        if (gtOpt.isEmpty()) return None.instance();
-        return gtOpt.getValue().getPrimary();
+    private Option<SPTarget> getPrimaryCwfs3(final ObsContext ctx) {
```
sraaphorst: Again, cleaner use of `Option`s and elimination of `null`.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gems/Canopus.java
```diff
@@ -227,154 +145,99 @@ public GuideStarValidation validate(SPTarget guideStar, ObsContext ctx) {
...
-    static {
-        // REL-286: port orientation
-        rotation[IssPort.SIDE_LOOKING.ordinal()] = new Angle(90.0, Angle.Unit.DEGREES);
-        rotation[IssPort.UP_LOOKING.ordinal()] = new Angle(0.0, Angle.Unit.DEGREES);
+        );
     }
 
-    public synchronized static Angle getRotationConfig(IssPort port) {
-        return rotation[port.ordinal()];
+    private static final Map<IssPort,Angle> rotation = new HashMap<IssPort,Angle>(IssPort.values().length) {{
```
sraaphorst: I thought it made far more sense to use a `Map` here rather than an array.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gems/Canopus.java
```diff
@@ -384,27 +247,21 @@ private Area probe3DependentRange(ObsContext ctx, Point2D.Double mid, Tuple2<Dou
...
         // star, then we translate the rectangle to there.  Otherwise, we assume
         // the center of the probe 3 range computed above.
-        Point2D xlat = new Point2D.Double(center.getX(), center.getY());
-
-        // Find the cwfs3 star, if any.
-        Option<Point2D> cwfs3Opt = getPrimaryCwfs3Offset(ctx);
-        if (!cwfs3Opt.isEmpty()) {
-            Point2D cwfs3 = cwfs3Opt.getValue();
-
+        final Point2D xlat = getPrimaryCwfs3Offset(ctx).map(cwfs3 -> {
```
sraaphorst: Again, this is just a cleaner way of handling the calculations. There was no reason to create `xlat` twice when only one of the two is used.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gems/Canopus.java
```diff
@@ -414,48 +271,46 @@ private Area probe3DependentRange(ObsContext ctx, Point2D.Double mid, Tuple2<Dou
...
-    *
-    * @param offsets positions to include in the intersection
-    */
-    public Area offsetIntersection(ObsContext ctx, Set<Offset> offsets) {
+     * Gets the intersection of the FOV and the offsets.
+     *
+     * @param offsets positions to include in the intersection
+     */
+    public Area offsetIntersection(final ObsContext ctx, final Set<Offset> offsets) {
+        // Note: If offsets are ever empty, this will return null, and this is not checked for anywhere.
```
sraaphorst: This makes me uncomfortable and could be done instead by using `map` and `reduce` over `offsets`, but that gives us an `Optional<Area>`, which makes things more complicated.

There is an assumption that `offsets` will always be non-empty: if it ever is empty, this method returns `null` and will cause an NPE in a couple places.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/gems/CanopusFeature.java
```diff
@@ -4,70 +4,31 @@
...
-import java.awt.image.BufferedImage;
 import java.beans.PropertyChangeListener;
 import java.util.Collection;
 import java.util.Collections;
 
 /**
  * Draws the Canopus AO field of view and probe ranges.
  */
-public final class CanopusFeature extends TpeImageFeature implements PropertyWatcher, TpeModeSensitive, TpeDragSensitive {
+public final class CanopusFeature extends TpeImageFeature implements PropertyWatcher {
```
sraaphorst: We no longer need to be `TpeModeSensitive` or `TpeDragSensitive` since we are no longer drawing the CWFS1 and CWFS2 ranges inside the CWFS patrol field.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/gems/CanopusFeature.java
```diff
@@ -4,70 +4,31 @@
...
-    private static final Composite BLOCKED = AlphaComposite.getInstance(AlphaComposite.SRC_OVER, 0.5F);
-
-    // Property used to control drawing of the probe ranges.
-    private static final BasicPropertyList props = new BasicPropertyList(CanopusFeature.class.getName());
-    private static final String PROP_SHOW_RANGES = "Show Probe Ranges";
-    static {
-        props.registerBooleanProperty(PROP_SHOW_RANGES, true);
-    }
-
-    private enum RangeDisplayMode {
```
sraaphorst: We no longer need a `RangeDisplayMode` due to elimination of CWFS1 and 2 ranges.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/gems/CanopusFeature.java
```diff
@@ -250,166 +135,23 @@ public void draw(Graphics g, TpeImageInfo tii) {
         g2d.setColor(AO_FOV_COLOR);
```
sraaphorst: All we need to draw now is the red circle for Canopus.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/gems/CanopusFeature.java
```diff
@@ -250,166 +135,23 @@ public void draw(Graphics g, TpeImageInfo tii) {
...
-                Paint p = createProbeRangeKeyPaint(g2d, slant);
-                g2d.setPaint(p);
-                g2d.fill(new Rectangle2D.Double(1, 1, 16, 16));
-            }
-            g2d.setPaint(origPaint);
-        }
-    }
-
     @Override
-    public Option<Component> getKey() {
```
sraaphorst: This was just a legend for the CWFS1 and 2 probe ranges; thus, it is no longer needed.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/tpe/TpeEditorTools.java
```diff
@@ -66,7 +63,6 @@ static ButtonState get(JToggleButton btn) {
             setToolTipText("Switch to browse mode");
             addActionListener(e -> {
                     _current = this;
-                    modeChange(TpeMode.BROWSE, None.instance());
```
sraaphorst: The only thing formerly receiving mode change notifications was `CanopusFeature`, so we can do entirely away with this.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/tpe/TpeImageFeature.java
```diff
@@ -220,20 +220,6 @@ public boolean isEnabledByDefault() {
...
-    /**
-     * Gets the key, or legend, associated with this image feature.  For
-     * example to explain the meaning of one or more elements that it draws.
-     * This method returns {@link None} by default.  A subclass must override
-     * to return a specific legend.
-     *
-     * @return {@link None}; override to provide a component that contains the
-     * legend
-     */
-    public Option<Component> getKey() {
```
sraaphorst: Again, only used by Canopus, so we can get rid of this entirely.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/tpe/TpeModeSensitive.java
```diff
@@ -1,25 +0,0 @@
...
-
-package jsky.app.ot.tpe;
-
-import edu.gemini.shared.util.immutable.Option;
-
-/**
- * An interface that identifies a TPE element that is sensitive to
- * {@link TpeMode mode} changes.
- */
-public interface TpeModeSensitive {
```
sraaphorst: Only used by Canopus, so we can throw this away.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/guide/GuideProbe.java
```diff
@@ -122,7 +119,9 @@ public Boolean apply(GuideProbe guideProbe) {
 
     /**
      * Indicates the bands that will be used for a given probe.
-     *
+     * Use R-band by default.
      */
-    BandsList getBands();
+    default BandsList getBands() {
+        return RBandsList.instance();
+    }
```
swalker2m: It's certainly true that this removes duplicate code, but it also allows you to not think about it when you implement a new guide probe.  If there is no default you're forced to provide an implementation and figure out what it should be.

One of the main drivers for the `default` feature was to allow interfaces that have existing implementations to be extended without requiring every implementation be updated.  We aren't really providing libraries and have control over all the implementations so that doesn't apply to us.

I haven't thought about this much but maybe a good rule would be to only use this feature for convenience methods that you don't expect or want to be overridden in the implementations?

sraaphorst: Your first paragraph had occurred to me. I have no objections changing it back, which will just take a few minutes anyway. I'll go ahead and do that.

sraaphorst:
Looks like I need to fix a few test cases as well.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gems/Canopus.java
```diff
@@ -227,154 +152,99 @@ public GuideStarValidation validate(SPTarget guideStar, ObsContext ctx) {
...
-
-    // Probe arm end, as distance in arcsec from the CWFS guide star
-    private static final double PROBE_ARM_END = 8.9;
-
-
     /**
      * Returns a Shape that defines the AO port in arcsecs, centered at
      * <code>(0,0)</code>.
      */
     private static final Shape AO_PORT_SHAPE = new Ellipse2D.Double(AO_PORT.getX(), AO_PORT.getY(), AO_PORT.getWidth(), AO_PORT.getHeight());
```
swalker2m: I wonder what the difference between the `AO_PORT_SHAPE` and `AO_PORT` is?  This is also now the "probe range" as I understand it.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gems/Canopus.java
```diff
@@ -227,154 +152,99 @@ public GuideStarValidation validate(SPTarget guideStar, ObsContext ctx) {
...
 
-
-    // Probe ranges expressed in arcsec
-//    private static final Tuple2<Integer, Integer> PROBE1_RANGE = new Pair<Integer, Integer>(-66, 24);
-//    private static final Tuple2<Integer, Integer> PROBE2_RANGE = new Pair<Integer, Integer>(-24, 66);
-
-    // REL-297:
     // P1 can access a square field of 3.7 x 3.0 armcin, but the center is at (+0.11,+0.3)
     // P2 can access a square field of 4.2 x 3.0 arcmin, but the center is at (0.0,-0.33)
     private static final Point2D.Double PROBE1_CENTER = new Point2D.Double(0.11 * 60.0, 0.3 * 60.0);
```
swalker2m: We don't need these definitions any more either?

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gems/Canopus.java
```diff
@@ -227,154 +152,99 @@ public GuideStarValidation validate(SPTarget guideStar, ObsContext ctx) {
...
-            asterism.getSkycalcCoordinates(when).flatMap(bc ->
-            target.getSkycalcCoordinates(when).map(tc -> {
-                CoordinateDiff diff = new CoordinateDiff(bc, tc);
-                Offset o = diff.getOffset();
-                double p = -o.p().toArcsecs().getMagnitude();
-                double q = -o.q().toArcsecs().getMagnitude();
-                return new Point2D.Double(p, q);
-            }));
-    }
+    private Option<Point2D> getPrimaryCwfs3Offset(final ObsContext ctx) {
```
swalker2m: The reason for doing this was to define the probe ranges for CWFS1 and CWFS2 -- they moved along with CWFS3.  Maybe we don't need this at all anymore?

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gems/Canopus.java
```diff
@@ -227,154 +152,99 @@ public GuideStarValidation validate(SPTarget guideStar, ObsContext ctx) {
...
 
-    public synchronized static void setRotationConfig(IssPort port, Angle rotation) {
-        Canopus.rotation[port.ordinal()] = rotation;
+    public synchronized static void setRotationConfig(final IssPort port, final Angle angle) {
+        Canopus.rotation.put(port, angle);
     }
 
     // Returns the probe range given the context, mid point and dimensions in arcsec
-    private Area probe3DependentRange(ObsContext ctx, Point2D.Double mid, Tuple2<Double, Double> dim) {
+    private Area probe3DependentRange(final ObsContext ctx, final Point2D.Double mid, final Tuple2<Double, Double> dim) {
```
swalker2m: Now there are no probe 3-dependent ranges.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gems/Canopus.java
swalker2m: This seems wrong to me too.  If there are no offset positions then we should have the unclipped whole AO port. at the base.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/tpe/TpeEditorTools.java
swalker2m: This is true and yet I'm hesitant about getting rid of it.  I guess because it was such a hassle getting it all to work.  But you're right, it's probably best to delete it if we aren't using it.

sraaphorst:
Thanks, @swalker2m, for the heaps of valuable feedback! I'll take a look at all the things you suggested. I think this may also address a chunk of REQ-OT-001 based on the test case for `GemsStrategySpec` for M6, but I'm still not certain. After this is done, I may have a couple questions for you.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gems/Canopus.java
sraaphorst: They were commented out in the original code, so I just removed them since they were no longer being used.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gems/Canopus.java
```diff
@@ -28,110 +27,40 @@
...
-            @Override public Option<PatrolField> getCorrectedPatrolField(ObsContext ctx) {
-                return correctedPatrolField(ctx);
-            }
-
-            @Override public BandsList getBands() { return RBandsList.instance(); }
         };
 
-        private static Option<PatrolField> correctedPatrolField(ObsContext ctx) {
+        @Override
+        public Option<PatrolField> getCorrectedPatrolField(final ObsContext ctx) {
```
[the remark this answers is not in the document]
sraaphorst: Excellent. I wasn't sure about this and was wondering the same thing myself. I'll go ahead and change this to the AO window.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gems/Canopus.java
sraaphorst: Excellent. Thanks. I really didn't know what to do about this. This will clean things up considerably and eliminate the possibility of an NPE here.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gems/Canopus.java
sraaphorst: Ooops. I understand now what you meant: we don't need these other definitions anymore.

sraaphorst:
I don't think we even need the `rotation` map anymore: this was just used to calculate the CWFS1 and CWFS2 probe ranges, and since the patrol field is an entirely symmetric circle, I think we can get rid of this entirely.

sraaphorst:
Apparently, I missed a couple of `Wfs` conversions. (Not sure why it compiled locally.)
I'll fix this on Monday morning and hopefully that will wrap up this part of the NGS2 change.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gems/CanopusWfs.java
```diff
@@ -0,0 +1,203 @@
...
+
+                                    // Get offset and switch it to be defined in the same coordinate
+                                    // system as the shape.
+                                    final Offset dis = diff.getOffset();
+                                    final double p = -dis.p().toArcsecs().getMagnitude();
+                                    final double q = -dis.q().toArcsecs().getMagnitude();
+                                    final Set<Offset> offsets = new TreeSet<>();
+                                    offsets.add(offset);
+
+                                    return offsetIntersection(ctx, offsets).contains(p, q);
```
sraaphorst: I think we can simplify this even further, too, with `areProbesInRange`.

sraaphorst:
I question the validity of the `UCAC3RegressionSpec` test cases... every time we make any changes to GeMS, we have to change this code just to get the test cases to pass so that we can merge a PR.

I'm sure these will change again during the NGS2 implementation.

cquiroz:
@sraaphorst I added those regressions tests when we did a large refactoring. Before that there were no tests at all so it was impossible to know if the refactoring would break it. The tests were done picking some targets and conditions, then running them through the OT and capturing the output

i'm fairly sure they will be invalid on NGS2

sraaphorst:
@cquiroz: I thought it was something to do with the catalog changes, given the name and the functionality.

Do you think we should keep them? Now that the catalog code is done, maybe we should discard them? Or at least somehow mark them the equivalent of `@Ignore` unless otherwise needed?

cquiroz:
We should probably remove them for ngs2. Perhaps you want to add something similar later once NGS2 is ready but I don't know for sure

sraaphorst:
Okay, great. I'll remove them as a separate PR so that we can revert easily if we need to.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gems/CanopusWfs.java
```diff
@@ -0,0 +1,191 @@
...
+    @Override
+    public PatrolField getPatrolField() {
+        return patrolField;
+    }
+
+    // The patrol field for Canopus is simply a fully symmetric circle, so we don't need to apply any rotations
+    // to it to correct it.
+    @Override
+    public Option<PatrolField> getCorrectedPatrolField(final ObsContext ctx) {
+        return ctx.getAOComponent().filter(ado -> ado instanceof Gems).map(a -> patrolField);
```
cquiroz: I miss `collect`...

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gems/CanopusWfs.java
```diff
@@ -0,0 +1,191 @@
...
+    private static Area offsetIntersection(final ObsContext ctx, final Set<Offset> offsets) {
+        final double t = ctx.getPositionAngle().toRadians();
+
+
+        return offsets.stream().map(pos -> {
+            final double p = pos.p().toArcsecs().getMagnitude();
+            final double q = pos.q().toArcsecs().getMagnitude();
+
+            final AffineTransform xform = new AffineTransform();
+            if (t != 0.0) xform.rotate(-t);
```
cquiroz: This could be slightly clearer if we use directly the position angle like
`if (pa != PositionAngle.zero) xform.rotate(-pa.toRadians)`

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gsaoi/Gsaoi.java
```diff
@@ -424,8 +424,6 @@ public static OdgwSize valueOf(String name, OdgwSize nvalue) {
 
     }
 
-    private static final String VERSION = "2011A-1";
```
cquiroz: Is it right to remove the version?

bundle/jsky.app.ot/src/main/java/jsky/app/ot/tpe/TpeEditorTools.java
cquiroz: Hopefully we won't need them again but I can imagine the effort to get this right.

cquiroz:
This is a great endeavor, canopus is a scary part of the code and I'm glad you could manage. I don't see any big problems though of course the OT could surprise us again

LGTM

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gsaoi/Gsaoi.java
swalker2m: GSAOI seems to be missing the call to `setVersion` that the others do.  The version is written out to XML and used during migration to detect whether to apply updates to the XML before importing.  Instead of deleting the VERSION we should probably `setVersion` in the constructor like the other instrument data objects.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gems/CanopusWfs.java
sraaphorst: I agree. I'll make that change.

sraaphorst: Kind of annoying, because now we are using both the `edu.gemini.spModel.core.Angle` and the `edu.gemini.skycalc.Angle`, and this is Java, so I will have to preface the former with the fully qualified package name.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gsaoi/Gsaoi.java
sraaphorst: Okay, I'll restore that. Is 2011A-1 indeed the correct version?

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gems/CanopusWfs.java
```diff
@@ -183,9 +182,9 @@ private static Area offsetIntersection(final ObsContext ctx, final Set<Offset> o
             xform.translate(-p, -q);
 
             return patrolField.getArea().createTransformedArea(xform);
-        }).reduce(patrolField.getArea(), (a1, a2) -> {
+        }).reduce((a1, a2) -> {
             a1.intersect(a2);
             return a1;
```
swalker2m: I'm trying to puzzle out how much to worry about the fact that the combining operation actually mutates one of the parameters.  I think I'd feel better if it were, say,

```java
final Area result = a1.clone();
result.intersect(a2);
return result;
```

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gsaoi/Gsaoi.java
swalker2m: It was picking up the default, 2009A-1, so we might as well use that.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gems/CanopusWfs.java
cquiroz: Interesting. I was somehow thinking that `a1.intersect` would return another instance

sraaphorst: 👍 
Yes, this seems like a better idea.

sraaphorst: How I wish that Java `Area` and `AffineTransform` objects returned new instances when combined...

PR #1391, opened 2017-12-18
REL-3336: Fixing broken test case that was causing issues sometimes when Travis was run.

sraaphorst:
One of the test cases for the ODB browser improvements (i.e. moving away from `StringUtil` to proper regular expressions) was improperly formed, sometimes failing when Travis was run.

### DIFF
--- a/bundle/edu.gemini.lchquery.servlet/src/test/scala/edu/gemini/lchquery/servlet/LchQueryRegexSpec.scala
+++ b/bundle/edu.gemini.lchquery.servlet/src/test/scala/edu/gemini/lchquery/servlet/LchQueryRegexSpec.scala
@@ -13,23 +13,22 @@ import scalaz._
 import Scalaz._
 
 class LchQueryRegexSpec extends Specification with ScalaCheck {
-  import ValueMatcher.ToRegex
   import LchQueryRegexSpec._
 
   "toRegEx and StringUtil.match" should {
     "always match exact strings the same" in {
-      forAll (queryGen) { s => s.matchesRe(s) should beTrue }
+      forAll (queryGen) { s => s.matchesAgree(s) should beTrue }
     }
 
     "always match case insensitively the same" in {
-      forAll (queryGen) { s => s.matchesRe(s.randomizeCase) should beTrue }
+      forAll (queryGen) { s => s.matchesAgree(s.randomizeCase) should beTrue }
     }
 
     "always match everything in an OR list the same" in {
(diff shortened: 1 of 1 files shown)


bundle/edu.gemini.lchquery.servlet/src/test/scala/edu/gemini/lchquery/servlet/LchQueryRegexSpec.scala
```diff
@@ -58,7 +57,7 @@ class LchQueryRegexSpec extends Specification with ScalaCheck {
   "toRegEx" should {
     "not match something not in an OR list" in {
       forAll (Gen.nonEmptyListOf(queryGen), queryGen) {
-        (lst, other) => !lst.contains(other) ==> (lst.mkString("|").toRegex.matches(other) should beFalse)
+        (lst, other) => !lst.exists(_.matchesLCH(other)) ==> (lst.mkString("|").matchesLCH(other) should beFalse)
```
swalker2m: I suppose this is the substantive change.  Maybe a bit of explanation would help me understand what it is.

sraaphorst: Ah, sure. The problem before was the use of `contains`, which is case-sensitive on strings, so if we had:

`lst = List("N", "M")` and
`other = "n"`,

then `lst.contains(other)` would be `false`, and hence its negation `true`.

As the ODB matching is case insensitive, the regular expression generated by `lst`, namely `N|M`, would match `n`. Thus, the test case would fail.

Thus, we need to switch to use the regular expression of the ODB browser (i.e. `matchesLCH`) when doing the left side of the implication check on `lst`.

Hopefully I explained that in a way that makes sense?

sraaphorst: Also, this is a test strictly on the ODB browser regex format, so `contains` was naturally the wrong choice here: both sides of the implication should use the ODB browser regex format to even have the condition make sense.

sraaphorst: Essentially, the idea here is to test that if something does not match anything in a list of regular expressions, then it does not match the OR combination of all of those regular expressions.

PR #1392, opened 2017-12-19
REL-3337: Removed UCAC3 regression tests.

sraaphorst:
The UCAC3 regression tests were made as a sanity check when a refactor was done to ensure that the refactor did not break the lookup.

Now, with the changes being made for NGS2, the UCAC3 regression tests will continue to fail and need updating, and are no longer useful. This should also speed up Travis slightly as they can also be somewhat time-consuming to run.


cquiroz:
I just remembered that i spent maybe a week building those 😄 
But good riddance

sraaphorst:
Huh. Well, that's irritating: I did a safe delete in IDEA and there were no reports of `MascotGuideStarSpec` going through.  I'm guessing that we're going to want to keep this class, so maybe we should just restore `UCAC3Regression` and leave out the `Spec`s?

sraaphorst:
And ha - sorry to throw away a week's worth of work!

PR #1393, opened 2017-12-19
REL-3339: Update CWFS limiting magnitudes for NGS2

sraaphorst:
The CWFS mag limits for NGS2 are changing due to the new EMCCD focal plane array.

The brightness limit is 2.5 mag larger than previously, and the faintness limit is 1.5 mag larger than previously.

Additionally, this obviously required correcting a number of GeMS test specs. This is annoying, as they will likely change again when I implement REL-3338, but at least it forces me to check correctness at each stage of the implementation.


bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsMagnitudeTable.scala
```diff
@@ -102,12 +102,16 @@ object GemsMagnitudeTable extends MagnitudeTable {
...
 
+  // Values provided by science are at:
+  //      SB Any, CC 70, IQ 70.
+  // Add 0.8 to faintness / brightness limits to get values for the calculator, which assumes:
+  //      SB  20, CC 50, IQ 70.
   lazy val CanopusWfsMagnitudeLimitsCalculator = new CanopusWfsCalculator {
-    override def gemsMagnitudeConstraint(starType: GemsGuideStarType, nirBand: Option[MagnitudeBand]) =
-      magLimits(RBandsList, 16.3, 8.8)
+    override def gemsMagnitudeConstraint(starType: GemsGuideStarType, nirBand: Option[MagnitudeBand]): MagnitudeConstraints =
+      magLimits(RBandsList, 17.8, 11.3)
```
cquiroz: Maybe we should use named constants here? specially given they are used twice

sraaphorst: I am a big fan of names constants... having to search for methods to figure out what their unnamed parameters mean is often too time consuming.

I like this idea, so I think I will do it for both Canopus and ODGW.

sraaphorst:
I think incrementally updating the test cases is not the worst idea: it forces me to sanity check that, when the guide stars change, it is for the right reasons: in this case because of the mag change.

I suspect there will only be one other major change, and that is when MASCOT is updated to only allow a difference of mag 3 between all the guide stars in an asterism.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsMagnitudeTable.scala
```diff
@@ -102,12 +102,19 @@ object GemsMagnitudeTable extends MagnitudeTable {
     def getNominalMagnitudeConstraints(cwfs: CanopusWfs): MagnitudeConstraints
   }
 
+  // Values provided by science are at:
+  //      SB Any, CC 70, IQ 70.
+  // Add 0.8 to faintness / brightness limits to get values for the calculator, which assumes:
+  //      SB  20, CC 50, IQ 70.
```
swalker2m: I can't remember any of this stuff but this really doesn't sound familiar.  Could you point out where this difference is coming from to refresh my memory?

sraaphorst: I found this incredibly confusing to determine. I looked at the old mag limits here and compared them to what science gave us, and the ones here were fainter by 0.8 mag. It took some mucking around in the `edu.gemini.catalog` api to figure out how to explain the difference.

Now that I'm looking again, this may not even be the correct explanation, as both CC70 and SB20 could explain the 0.3 difference, and SB Any or IQ85 could explain the 0.5 difference.

In any case, I think I should eliminate this comment as I'm not convinced of it now, and just note that these are 0.8 mag different than the values typically supplied by science.

cquiroz: i vaguely remember a google excel sheet where those numbers were stored. Do you have access to that?

sraaphorst: There is a CSV file where some of the values are stored, but it doesn't include the values for any GeMS guide probes.

sraaphorst: See here:
https://github.com/gemini-hlsw/ocs/blob/develop/bundle/edu.gemini.ags/src/main/resources/edu/gemini/ags/conf/Guide%20Limits%20-%20OT%20Config.csv

sraaphorst: Also, from the top of this file, the comment:

"A magnitude table defined in the same way as we have done since before 2015A (that is, with predefined magnitude limits)."

I can't exactly remember how this worked either as it's been a couple years, but the values here required different handling than other probe limits.

swalker2m: So, for the other guide probes we are explicitly provided values for all the SB, and IQ combinations, and make an adjustment for CC.

sraaphorst: @swalker2m Yes, exactly. For GeMS, we only get one set of mags, which we adjust using the code in:
https://github.com/gemini-hlsw/ocs/blob/develop/bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/api/package.scala

A bit confusing, really, especially given that science did not provide the conditions for the mag limits in the NGS2 OT doc.

swalker2m: Let's get feedback on what the assumed conditions are for the new limits, but we can merge this in the meantime and come back to it if we need to make a change. 👍

sraaphorst: So it turns out that the values provided by science are SB ANY, CC50, IQ70.
Thus, I think the mag limits for Canopus in previous versions of the OT are wrong and should have been -0.3 mag what they actually are.

We must add +0.5 mag to the limits given by science in order for them to account for SB ANY, which results in a -0.5 mag during condition adjustment.

cquiroz: Are those new values for NGS2?

sraaphorst: Whoops. I didn't see this before I merged. Sorry, @cquiroz.

Yes, those are the new NGS2 values as confirmed by science for the conditions in the comment.

cquiroz: Excellent, I was worried they were the current valid ones and we have had them wrong in the OT forever

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsMagnitudeTable.scala
```diff
@@ -103,11 +103,16 @@ object GemsMagnitudeTable extends MagnitudeTable {
   }
 
   lazy val CanopusWfsMagnitudeLimitsCalculator = new CanopusWfsCalculator {
-    override def gemsMagnitudeConstraint(starType: GemsGuideStarType, nirBand: Option[MagnitudeBand]) =
-      magLimits(RBandsList, 16.3, 8.8)
+    // The values provided by science are assumed to be at SB ANY, CC50, and IQ70.
+    // Since SB ANY results in an adjustment of -0.5 mag, we add 0.5 mag to account for this.
+    private val FaintLimit  = 17.5
+    private val BrightLimit = 11.0
```
swalker2m: So the values they actually specified are (10.5, 17.0), which for our reference conditions of SB50, CC50, IQ70 would translate to (11.0,17.5)?  Is that correct?  It's surprising that they give values in a different set of reference conditions than in the past.

A comment about where the adjustment is defined might be good (package object `edu.gemini.catalog.api`) -- I mean to piggy pack on a future PR.

sraaphorst: Yes, that is correct. I agree that it is odd and I'm not sure why this is the case, but it was confirmed by Rodrigo in the NGS2 OT document.

Agreed on the comment. I'll make that in the PR I'm working on now regarding the MASCOT changes.

PR #1394, opened 2017-12-22
REL-3341: Added CWFS window marker geometry and drawing.

sraaphorst:
This PR aims to satisfy the remaining parts of REQ-OT-002 in the NGS2 OT document, i.e. for CWFS guide stars, mark them with a fixed 2" or 4" square or circle.

I am very iffy about this PR, given that the requirements are still not entirely clear, but I've designed it so that we can change size, shape, and color of the CWFS markers easily enough.

Here is an image of some "dummy stars" for M6 to show that the translation for the markers works properly and to show what a star that is out of range for Canopus looks like. I am not very happy with the way that this looks, and think it appears quite awkward in the TPE.

It is still unknown if these markers have any physical significance with regards to NGS2 or if they are simply there to mark the stars now that the probe arms have been removed: to finalize this requirement, once clear, we may need to:
1. Filter out asterisms where the markers overlap (which will be done in MASCOT); and
2. Filter out guide stars that do not fall completely within the CWFS patrol field (which will be done in `CanopusWfs` when validating a guide star).

Any feedback would be much appreciated.
screen shot 2017-12-22 at 11 17 19 am

### DIFF
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gems/CanopusWfs.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gems/CanopusWfs.java
@@ -14,6 +14,7 @@ import edu.gemini.spModel.obs.context.ObsContext;
 import edu.gemini.spModel.target.SPTarget;
 import edu.gemini.spModel.target.env.GuideProbeTargets;
 
+import java.awt.*;
 import java.awt.geom.*;
 import java.util.*;
 
@@ -30,6 +31,15 @@ public enum CanopusWfs implements GuideProbe, ValidatableGuideProbe, OffsetValid
     static {
         final Ellipse2D AO_PORT = new Ellipse2D.Double(-RADIUS_ARCSEC, -RADIUS_ARCSEC, RADIUS_ARCSEC * 2, RADIUS_ARCSEC * 2);
         patrolField = new PatrolField(AO_PORT);
+    }
+
+    // According to NGS2, guide stars should be centered in 2" or 4" windows. We select 4".
+    private static final double GS_WINDOW_SIZE_ARCSEC = 4.0;
+    private static final Area   GS_WINDOW;
+
+    static {
+        final Shape GS_WINDOW_SHAPE = new Rectangle2D.Double(-GS_WINDOW_SIZE_ARCSEC / 2.0, -GS_WINDOW_SIZE_ARCSEC / 2.0, GS_WINDOW_SIZE_ARCSEC, GS_WINDOW_SIZE_ARCSEC);
+        GS_WINDOW = new Area(GS_WINDOW_SHAPE);
     }
 
     @Override
@@ -115,6 +125,10 @@ public enum CanopusWfs implements GuideProbe, ValidatableGuideProbe, OffsetValid
     @Override
     public Option<PatrolField> getCorrectedPatrolField(final ObsContext ctx) {
         return ctx.getAOComponent().filter(ado -> ado instanceof Gems).map(a -> patrolField);
+    }
+
+    public static Area getGuideStarWindow() {
+        return GS_WINDOW;
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


cquiroz:
Do you mean that you either mark them with a square or a circle, or that in certain cases you use a square and in others a circle? how do you choose to use 2 or 4?

sraaphorst:
I'm rather unhappy because at 1x zoom, it looks very odd:
screen shot 2017-12-22 at 11 36 38 am

At 2x zoom, the markers are the same size as the green squares that we currently use to indicate guide stars, which means if the PA % 90 == 0, then they overwrite the green squares, and if the PA % 90 != 0, then we get some weird rotated marker on top of the green square.

sraaphorst:
@cquiroz These are all good questions that I myself have asked.

The answers I've gotten seems like a 4" square is what science would "prefer," which leads me to think that there is no actual physical significance to these, which leads to me wondering why we need them at all when we already indicate guide stars with a green square.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gems/CanopusWfs.java
```diff
@@ -117,6 +127,10 @@ public PatrolField getPatrolField() {
         return ctx.getAOComponent().filter(ado -> ado instanceof Gems).map(a -> patrolField);
     }
 
+    public static Area getGuideStarWindow() {
+        return GS_WINDOW;
```
sraaphorst: I guess we should probably `clone()` this since it's mutable.

cquiroz: 👍

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/gems/CanopusFeature.java
```diff
@@ -128,6 +135,35 @@ public void draw(final Graphics g, final TpeImageInfo tii) {
...
+                                                final AffineTransform tx = AffineTransform.getTranslateInstance(base.x, base.y);
+                                                tx.scale(ppa, ppa);
+                                                tx.translate(p, q);
+                                                tx.rotate(-tii.getCorrectedPosAngleRadians());
+
+                                                final Shape window = tx.createTransformedShape(flipArea(CanopusWfs.getGuideStarWindow()));
+                                                if (((ValidatableGuideProbe) gpt.getGuider()).validate(t, ctx) == GuideStarValidation.VALID)
+                                                    g2d.setColor(GS_WINDOW_COLOR);
+                                                else
+                                                    g2d.setColor(OtColor.makeSlightlyTransparent(GS_WINDOW_COLOR));
```
cquiroz: `makeSlightlyTransparent` 😄

sraaphorst: One has to wonder what:
`OtColor.makeSlightlyTransparent(OtColor.makeSlightlyTransparent(c))`
would do. 😸

sraaphorst: I guess it cuts the RGB values in half, so we could get fully transparent if we continued to infinity.

sraaphorst: This is also where I really miss Scala `for` constructions for `Option`. All the nested `forEach` calls are kind of horrible.

cquiroz: we should add `makeSlightlyLessTransparent/makeSlightlyMoreTransparent` 😄

swalker2m:
> Filter out asterisms where the markers overlap (which will be done in MASCOT)

Presumably we'll need at least a P2 check for this as well since you can manually select guide stars.

swalker2m:
I agree, it isn't clear what the requirements are or why 2" vs 4", circle vs square are interchangeable.  The code looks fine but maybe we should wait until we know more before merging this?

sraaphorst:
Agree with you, @swalker2m... I'm not in a rush to merge this. It just seemed like a straightforward enough task to knock REQ-OT-002 off my list and get something for science to look at so they can decide if this is the right direction to take things.

Good thinking on the P2 check.

sraaphorst:
This will be added to the NGS2 branch instead, with modifications.

PR #1395, opened 2017-12-29
REL-3338: NGS2: Change MASCOT so that guide stars in an asterism have difference in brightness of at most 3 mag in R band

sraaphorst:
This is to satisfy REQ-OT-005 in the NGS OT requirements doc.

Now that we are moving to a EMCCD focal plane array instead of using probe arms, CWFS guide stars in an asterism can only differ by at most 3 mag in the R band.

This change implements that by adding an "asterism pre-filter" to MASCOT so that any inadmissible candidates can be filtered out before the computationally intensive Strehl calculations are run.

Due to the different ways that MASCOT is called (which vary in production and test code, where both take different pathways and maintain different information at different times), this necessitated adding a `asterismPreFilter` to `GemsGuideProbeGroup` and also a reference in `GuideStarType` to the `GemsGuideProbeGroup` to propagate the pre-filter to MASCOT along the two different pathways.

I also added a P2 check for this in `GemsGuideStarRule`.

Here is an example that shows the effect of the filtering. In the first image, you can see the old asterism that would have been chosen before the filtering was done. Notice that the R band magnitude between CWFS1 and CWFS3 is more than 3, and thus by the new NGS rules, it should be invalid. In the second image, we see what would now be picked instead as the old asterism would be filtered out.

Original, without filtering:
old

New, with filtering:
new

While working on this, I also included some debugging output to show what asterisms were being filtered out and their R-band extracted values. Here is the debugging output that corresponds to the aforementioned case:

filtered

I also added this as a test case to `GemsStrategySpec`.


bundle/edu.gemini.ags/src/test/scala/edu/gemini/ags/impl/GemsStrategySpec.scala
```diff
@@ -43,9 +43,6 @@ class GemsStrategySpec extends Specification {
       val target = new SPTarget(ra.toDegrees, dec.toDegrees)
       val env = TargetEnvironment.create(target)
       val inst = new Gsaoi <| {_.setPosAngle(0.0)} <| {_.setIssPort(IssPort.UP_LOOKING)}
-      val gsaoi = GsaoiOdgw.values().toList
-      val canopus = CanopusWfs.values().toList
-      val pwfs1 = List(PwfsGuideProbe.pwfs1)
```
sraaphorst: These values were never used, so I removed them.

sraaphorst:
Note this PR looks deceptively larger than it actually is because I included the XML dump of a catalog lookup for the test case, namely `ngs2_mag.xml`.

sraaphorst:
I figured this might break some test cases, but I can't seem to run `test` on my machine anymore.

I'll look at this after lunch.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsResultsAnalyzer.scala
```diff
@@ -61,10 +61,12 @@ object GemsResultsAnalyzer {
     obsContext.getBaseCoordinates.asScalaOpt.map(_.toNewModel).map { base =>
       // gemsGuideStars needs to be mutable to support updates from inside mascot
       val gemsGuideStars = TiptiltFlexurePair.pairs(catalogSearch.asScala.toList).foldLeft(List.empty[GemsGuideStars]) { (gemsGuideStars, pair) =>
+        // tiptiltGroup and flexureGroup are each one of CanopusWfs.Group and GsaoiOdgw.group.
         val tiptiltGroup = pair.tiptiltResults.criterion.key.group
         val flexureGroup = pair.flexureResults.criterion.key.group
         val tiptiltTargetsList = filter(obsContext, pair.tiptiltResults.results, tiptiltGroup, posAngles.asScala.toSet)
         val flexureTargetsList = filter(obsContext, pair.flexureResults.results, flexureGroup, posAngles.asScala.toSet)
+        val tiptiltGroupPreAsterismFilter: (List[SiderealTarget] => Boolean) = lst => tiptiltGroup.asterismPreFilter(lst.asImList)
```
cquiroz: What's the role of the `preFilter`

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/mascot/Mascot.scala
```diff
@@ -77,23 +81,29 @@ object Mascot {
...
+   * @param filter a filter function that returns false if the Star should be excluded
+   * @param asterismPreFilter a filter function that returns false if a candidate asterism should be excluded
    * @return a tuple: (list of stars actually used, list of asterisms found)
    */
   def findBestAsterism(starList: List[Star],
                        factor: Double = defaultFactor,
                        progress: ProgressFunction = defaultProgress,
-                       filter: Star => Boolean = defaultFilter)
+                       filter: Star => Boolean = defaultFilter,
+                       asterismPreFilter: List[SiderealTarget] => Boolean = defaultAsterismPreFilter)
```
cquiroz: Should the argument of this filter be `Asterism` (i don't know if that exists)?

bundle/edu.gemini.ags/src/test/scala/edu/gemini/ags/gems/GemsResultsAnalyzerSpec.scala
```diff
@@ -71,7 +71,7 @@ class GemsResultsAnalyzerSpec extends MascotProgress with SpecificationLike with
       }
 
       LOGGER.info("gems results: size = " + gemsGuideStars.size)
-      gemsGuideStars should have size 252
+      gemsGuideStars should have size 228
```
cquiroz: is this value to make the test pass or was it derived from somewhere?

bundle/edu.gemini.ags/src/test/scala/edu/gemini/ags/impl/GemsStrategySpec.scala
```diff
@@ -633,6 +630,73 @@ class GemsStrategySpec extends Specification {
...
       Await.result(estimate, 1.minute) should beEqualTo(Estimate.GuaranteedSuccess)
     }
+
+    // Previously, before the mag 3 difference limit was imposed in guide stars in an asterism due to the NGS2 changes,
+    // MASCOT would have chosen the asterism:
+    // CWFS1: 289-128992, R=11.238
+    // CWFS2: 289-129007, R=14.006
+    // CWFS3: 289-129036, R=14.849
+    // However, this asterism is no longer valid as the maximum mag difference is 14.849 - 11.238 = 3.611 > 3.
+    "select CWFS guide stars being at maximum 3 mag difference for NGS2" in {
```
cquiroz: Great that we test this

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/GemsResultsAnalyzer.scala
sraaphorst: `preFilter` is just the way to pass on the filtering to be done to asterism candidates to cut them out from consideration entirely before Strehl calculations are done. In the case of CWFS, this is to limit the diff by R-mag 3.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/mascot/Mascot.scala
sraaphorst: Looking at the `Asterism` class, I'm not sure we have a full implementation yet that allows construction of asterisms with more than a single `SPTarget`, unless I'm misunderstanding... but yes, that might be a better way to go in the long run through all of MASCOT if we can do it.

bundle/edu.gemini.ags/src/test/scala/edu/gemini/ags/gems/GemsResultsAnalyzerSpec.scala
sraaphorst: I checked this with my debug filtering as per the terminal screenshot example above to see how many asterisms were being filtered out.

bundle/edu.gemini.ags/src/test/scala/edu/gemini/ags/impl/GemsStrategySpec.scala
sraaphorst: I really wanted to add a test for the non-filtered asterisms as well, but that involved replicating a huge amount of the code that prepared the necessary parameters to MASCOT.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/mascot/Mascot.scala
```diff
@@ -77,23 +81,29 @@ object Mascot {
...
-
     val ns = filteredStarList.length
-    val trips = AllPairsAndTriples.allTrips(filteredStarList)
-    val pairs = AllPairsAndTriples.allPairs(filteredStarList)
-    val total = trips.length + pairs.length + ns
+
+    // Create all candidate asterisms, filtering out those that are inadmissible.
+    val sings = filteredStarList.filter(s => asterismPreFilter(List(s.target)))
+    val trips = AllPairsAndTriples.allTrips(filteredStarList).
+      filter { case (s1,os2, os3) => asterismPreFilter(List(s1.target.some, os2.map(_.target), os3.map(_.target)).flatten) }
```
swalker2m: If it is `allTrips` (i.e., all triples) why are stars 2, and 3 wrapped in an `Option`?  I would think a triple where one or more of the three is `None` would stop being a triple.  Same for `allPairs` below.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/mascot/MascotCat.scala
```diff
@@ -49,10 +51,11 @@ object MascotCat {
                        centerRA: Double, centerDec: Double,
                        factor: Double = Mascot.defaultFactor,
                        progress: ProgressFunction = defaultProgress,
-                       filter: Star => Boolean = Mascot.defaultFilter)
+                       filter: Star => Boolean = Mascot.defaultFilter,
+                       asterismPreFilter: List[SiderealTarget] => Boolean = Mascot.defaultAsterismPreFilter)
```
swalker2m: Is there ever any other "pre filter" besides the magnitude check?  I guess it's kind of nice to have more flexibility I'm just wondering if we need it.  We end up having to pass this around a lot.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/mascot/Mascot.scala
sraaphorst: I was indeed wondering the exact same thing when I came across this code.
I'm really not sure why a triple is a `(Star, Option[Star], Option[Star])` and a pair is a `(Star, Option[Star])'`.

I was hoping that maybe you knew? 😸

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/gems/mascot/MascotCat.scala
sraaphorst: No, this is the only check.

From my understanding, we wouldn't want to do a check when looking for ODGW guide stars, but we aren't using those anymore, correct?

swalker2m: Right, I see why you did that then.  Let's just leave it this way for now because I believe Bryan wants to leave the door open for GSAOI asterisms in the future.

PR #1396, opened 2018-01-04
REL-3342: Restored magnitude property settings to GpiCB.

sraaphorst:
This restores the accidentally removed setting of magnitudes for H and I band in `GpiCB.java`, and has been rewritten to make it clear that these properties are being set instead of relying on unusual uses of `ApplyOp`.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gpi/GpiCB.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gpi/GpiCB.java
@@ -74,7 +74,7 @@ public class GpiCB extends AbstractObsComponentCB {
                     final ImList<Magnitude> magnitudes = base.getMagnitudesJava();
                     MAG_PROPERTIES.forEach((band, propName) ->
                         magnitudes.find(m -> m.band().equals(band)).
-                                foreach(mag -> config.putParameter(systemName, DefaultParameter.getInstance(propName, mag))));
+                                foreach(mag -> config.putParameter(systemName, DefaultParameter.getInstance(propName, mag.value()))));
                 });
             }
         }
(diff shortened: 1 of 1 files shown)


bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gpi/GpiCB.java
```diff
@@ -44,50 +44,46 @@ protected void thisReset(Map<String, Object> options) {
...
+        final ISPObsComponent targetcomp = SPTreeUtil.findTargetEnvNode(getObsComponent().getContextObservation());
+        if (targetcomp != null) {
+            final TargetObsComp toc = (TargetObsComp) targetcomp.getDataObject();
+            if (toc != null) {
+                // The Asterism here should be Single(base), so we only need to concern ourselves with the head target.
+                toc.getAsterism().allSpTargetsJava().headOption().foreach(base -> {
+                    final ImList<Magnitude> magnitudes = base.getMagnitudesJava();
+                    MAG_PROPERTIES.forEach((band, propName) ->
+                        magnitudes.find(m -> m.band().equals(band)).
+                                foreach(mag -> config.putParameter(systemName, DefaultParameter.getInstance(propName, mag))));
```
swalker2m: Haven't we changed the type of the object that we're storing in the sequence?  Before it was

```java
    config.putParameter(systemName, DefaultParameter.getInstance(magProp, magnitude.value()));
```

which is a `Double` and now it is the `Magnitude` instance itself?

swalker2m: Having the `Magnitude` is objectively better than a `Double`, and I don't think the property is used anywhere else except for the seqexec.  But how does the sequence XML compare from the 2017B OT and the OT with this update?  We need to make sure that what we send to the seqexec doesn't change.

sraaphorst: I'll double-check this.

sraaphorst: Yes, you are right. We were writing a `Double` and now a `Magnitude`.

I think for now, we should just write the `Double` so we can push out the test release that Bryan wants and if we want to change this later, we can do so. What do you think?

PR #1401, opened 2018-01-05
REL-3188: Overheads. Test version (GMOS-N, GMOS-S, GNIRS, NIFS, NIRI)

olesjasmirnova:
Added table with observation overhead info to the web-ITC result page. The overhead estimations should match those in the OT, with exception of re-acquisitions and recentering times, which are not calculated by the OT.

Other new features:
* Added "Telescope offsets" to the ITC web form.
* Added "Coadds" to the ITC web form (GNIRS, NIRI, NIFS, GSAOI).
* Signal plot should reflect now the signal for single coadd, whereas the SNR plot - for single frame and all frames.
* Added "ROI" and "Sub Array" to GMOS and NIRI web forms.
* Number of acquisitions (number of visits) is calculated as one per every two hours of *science* time.
* Number of reacquisitions (recenterings) is calculated as one per every hour of spectroscopy *science* time if PWFS is used.
* Updated readout time tables and/or DHS write time for GNIRS, NIRI and NIFS.


bundle/edu.gemini.itc.shared/src/main/java/edu/gemini/itc/shared/ConfigCreator.java
```diff
@@ -0,0 +1,201 @@
...
+    public static final ItemKey CcdYBinning = new ItemKey("instrument:ccdYBinning");
+    public static final ItemKey AmpGain = new ItemKey("instrument:gainChoice");
+    public static final ItemKey AmpCount = new ItemKey("instrument:ampCount");
+    public static final ItemKey DetectorManufacturerKey = new ItemKey("instrument:detectorManufacturer");
+    public static final ItemKey GuideWithPWFS2Key = new ItemKey("telescope:guideWithPWFS2");
+    public static final ItemKey TelescopeQKey = new ItemKey("telescope:q");
+    public static final ItemKey TelescopePKey = new ItemKey("telescope:p");
+    public static final ItemKey AOGuideStarTypeKey = new ItemKey("adaptive optics:guideStarType");
+    public static final ItemKey AOSystemKey = new ItemKey("adaptive optics:aoSystem");
+    public static final ItemKey AOFieldLensKey = new ItemKey("adaptive optics:fieldLens");
```
sraaphorst: I grepped through the codebase and these are defined multiple times in multiple files, which makes me a bit uncomfortable. For example, lots of these are already in `jsky.app.it.editor.seq.Keys`.

I wonder if we should be reusing these instead of redefining them repeatedly?

bundle/edu.gemini.itc.shared/src/main/java/edu/gemini/itc/shared/ConfigCreator.java
```diff
@@ -0,0 +1,201 @@
...
+        int numberCoadds = calcMethod.coaddsOrElse(1);
+        Config[] dc = new DefaultConfig[numberExposures];
+        String offset = Double.toString(calcMethod.offset());
+        // for spectroscopic observations we consider ABBA offset pattern
+        List<String> spectroscopyOffsets = new ArrayList<>(Arrays.asList("0", offset, offset, "0"));
+        // for imaging observations we consider ABAB offset pattern
+        List<String> imagingOffsets = new ArrayList<>(Arrays.asList("0", offset, "0", offset));
+        List<String> offsetList = new ArrayList<>();
+
+        for (int i = 0; i < (1 + numberExposures / 4); i++) {
```
sraaphorst: Are we certain that `numberExposures` will always be a multiple of 4? Do we really want multiple copies of either `spectroscopyOffsets` or `imagingOffsets` added to `offsetList`?

bundle/edu.gemini.itc.shared/src/main/java/edu/gemini/itc/shared/ConfigCreator.java
```diff
@@ -0,0 +1,201 @@
...
+            step.putItem(DisperserKey, (gmosParams.grating().displayValue()));
+            GmosCommonType.Binning xbin = GmosCommonType.Binning.getBinningByValue(gmosParams.spectralBinning());
+            GmosCommonType.Binning ybin = GmosCommonType.Binning.getBinningByValue(gmosParams.spatialBinning());
+            step.putItem(CcdXBinning, xbin);
+            step.putItem(CcdYBinning, ybin);
+            step.putItem(AmpGain, (gmosParams.ampGain()));
+            if (gmosParams.ccdType().displayValue().equals("E2V")) {
+                step.putItem(AmpCount, GmosCommonType.AmpCount.SIX);
+            } else {
+                step.putItem(AmpCount, GmosCommonType.AmpCount.TWELVE);
```
sraaphorst: I guess `AmpCount.THREE` is only kept around for legacy reasons.
So the assumption is if not `E2V` then `Hamamatsu`?

bundle/edu.gemini.itc.shared/src/main/java/edu/gemini/itc/shared/ConfigCreator.java
olesjasmirnova: I really wanted to reuse the ItemKeys, since as you noticed they are indeed defined multiple times throughout the ocs. However there are only few packages which are accessible from the ITC packages, and the OT packages are not among them. I remember introducing some ”circular dependencies”, probably while trying to add more dependencies to the ITC. So I went for redefining, as I couldn’t come up with a better solution. If you could suggest a way to do it otherwise, I can change it.

bundle/edu.gemini.itc.shared/src/main/java/edu/gemini/itc/shared/ConfigCreator.java
olesjasmirnova: You are right, this is a horrible way of doing this. I will try to come up with something better.

bundle/edu.gemini.itc.shared/src/main/java/edu/gemini/itc/shared/ConfigCreator.java
olesjasmirnova: Yes, in the ITC we only allow for the maximum number of amplifiers, so it’s either E2V with 6 amps, or Hamamatsu with 12 amps. I will change this to make it more specific, thanks.

sraaphorst:
BTW, I will look over this more, but it looks good for the most part!
I hope my comments are helpful: they are not intended to be critical, and part of it is me wanting to learn a bit more about the ITC code and keep up with the changes.

olesjasmirnova:
Thanks for reviewing the code Sebastian, your comments are appreciated! For a new programmer like me any constructive criticism is very helpful.

bundle/edu.gemini.itc.shared/src/main/java/edu/gemini/itc/shared/ConfigCreator.java
```diff
@@ -0,0 +1,201 @@
...
+            step.putItem(ExposureTimeKey, obsDetailParams.exposureTime());
+            step.putItem(ObserveTypeKey, "OBJECT");
+            step.putItem(ObserveClassKey, "SCIENCE");
+            step.putItem(CoaddsKey, numberCoadds);
+            step.putItem(TelescopePKey, "0");
+            step.putItem(TelescopeQKey, offsetList.get(i));
+
+            if (itcParams.telescope().getWFS().equals(GuideProbe.Type.PWFS)) {
+                step.putItem(GuideWithPWFS2Key, "guide");
+            }
```
sraaphorst: I'm not 100% sure about this. You continue to create the `DefaultConfig` with the same parameter values over and over again, with only different offsets.

What I would recommend is, before the `for` loop, creating the `DefaultConfig` with all the parameters that don't change.

Something like this, instead of line 77 - 91:
```java
final Config stepCommon = new DefaultConfig();
stepCommon.putItem(ExposureTimeKey, obsDetailsParams.exposureTime());
stepCommon.putItem(ObserveTypeKey, "OBJECT");
stepCommon.putItem(ObserveClassKey, "SCIENCE");
stepCommon.putItem(CoaddsKey, numberCoadds);
stepCommon.putItem(TelescopePKey, "0");

for (int i=0; i < dc.length; ++i) {
  final Config step = new DefaultConfig(stepCommon);
  step.putItem(TelescopeQKey, offset.get(i));

  if (itcParams.telescope().getWFS().equals(GuideProbe.Type.PWFS)) {
    step.putItem(GuideWithPWFS2Key, "guide");
  }

  dc[i] = step;
}
```

bundle/edu.gemini.itc.shared/src/main/java/edu/gemini/itc/shared/ConfigCreator.java
```diff
@@ -0,0 +1,201 @@
...
+            step.putItem(SlitWidthKey, gnirsParams.slitWidth().logValue());
+
+            if (gnirsParams.altair().isDefined()) {
+                AltairParameters altairParameters = gnirsParams.altair().get();
+                step.putItem(AOGuideStarTypeKey, altairParameters.wfsMode().displayValue());
+                step.putItem(AOSystemKey, "Altair");
+                step.putItem(AOFieldLensKey, altairParameters.fieldLens().displayValue());
+            }
+        }
+        return conf;
```
sraaphorst: This could again be simplified using a `Config` with the repeated values in the `for` loop and then for each step, using `putAll`, e.g.:

```java
final Config commonStep = new DefaultConfig();
commonStep.putItem(ReadModeKey, (gnirsParams.readMode()));
commonStep.putItem(InstInstrumentKey, "GNIRS");
commonStep.putItem(SlitWidthKey, gnirsParams.slitWidth().logValue());
if (gnirsParams.altair().isDefined()) {
  AltairParameters altairParameters = gnirsParams.altair().get();   step.putItem(AOGuideStarTypeKey, altairParameters.wfsMode().displayValue());
  commonStep.putItem(AOSystemKey, "Altair");
  commonStep.putItem(AOFieldLensKey, altairParameters.fieldLens().displayValue());
}

final Config[] conf = createCommonConfig(numExp);
for (final Config step : conf) {
  step.putAll(commonStep);
}
```

Again, though, I'm not clear why you're putting the exact same step repeatedly into an array. That may be the proper behaviour but it seems a but unusual to me.

sraaphorst: Also, same for the below `createXXXXConfig` methods.

bundle/edu.gemini.itc.shared/src/main/java/edu/gemini/itc/shared/ConfigCreator.java
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    public Config[] createGmosConfig(GmosParameters gmosParams, int numExp) {
+        Config[] conf = createCommonConfig(numExp);
+
+        for (Config step : conf) {
+            System.out.println("SIte: " + gmosParams.site().displayName);
```
sraaphorst: Typo: **Site**, and not SIte.

bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/Parameters.scala
```diff
@@ -38,7 +38,14 @@ sealed trait Spectroscopy
 
 sealed trait CalculationMethod {
   def exposureTime: Double
+  def coadds: Option[Int]
+  /*
+     It seems that using .getOrElse is not straightforward to use from Java,
+     so we can use this "coaddsOrElse" to simplify it.
+   */
+  def coaddsOrElse(d: Int): Int = coadds.getOrElse(d)
```
sraaphorst: What happens, out of curiosity, when you try to use `getOrElse` from Scala?
I would think maybe you need to turn Scala `Int` to Java `java.lang.Integer`?
This should be fine, though.

bundle/edu.gemini.itc.web/src/main/java/edu/gemini/itc/web/html/NifsPrinter.java
```diff
@@ -108,4 +125,23 @@ else if (instrument.getIFUMethod() instanceof IfuSummed)
...
+            numberExposures = ((SpectroscopyS2N) calcMethod).exposures();
+        }
+
+        ConfigCreator cc    = new ConfigCreator(p);
+        config              = cc.createNifsConfig(instr, numberExposures);
+        pta                 = PlannedTimeCalculator.instance.calc(config, new InstNIFS());
+
+        if (numberExposures < 2) step = 0; else step = 1;
+    }
+
```
sraaphorst: This `getOverheadTableParams` method seems repetitive from one `PrinterBase` to the next.

I would instead define it in `PrinterBase` as follows:
```java
public void getOverheadTableParams(Result result, ObservationDetails obs) {
         final int numberExposures;
         final CalculationMethod calcMethod = obs.calculationMethod();
 
         if (calcMethod instanceof ImagingInt) {
             numberExposures = (int)(((ImagingResult) result).is2nCalc().numberSourceExposures() * obs.sourceFraction());
         } else if (calcMethod instanceof ImagingS2N) {
             numberExposures = ((ImagingS2N) calcMethod).exposures();
         } else if (calcMethod instanceof SpectroscopyS2N) {
             numberExposures = ((SpectroscopyS2N) calcMethod).exposures();
         } else {
             numberExposures = 1;
         }
         config            = createInstConfig(numberExposures);
         pta                 = PlannedTimeCalculator.instance.calc(config, getInst());
         if (numberExposures < 2) step = 0; else step = 1;
     }

abstract Config[] createInstConfig(int numberExposures);

abstract SPInstObsComp getInst();
```

And then define just the different code of `createInstConfig` and `getInst()` in subclasses, e.g. here:

```java
Config[] createInstConfig(int numberExposures) {
  ConfigCreator cc = new ConfigCreator(p);
  return cc.createNifsConfig(instr, numberExposures);
}

SPInstObsComp getInst() {
  return new InstNIFS();
}
```

That way you don't have to write the same code over and over again, which is very unmaintainable.

bundle/edu.gemini.itc.web/src/main/java/edu/gemini/itc/web/html/NiriPrinter.java
```diff
@@ -153,5 +165,24 @@ private String niriToString(final Niri instrument) {
...
+        }
+
+        ConfigCreator cc    = new ConfigCreator(p);
+        config              = cc.createNiriConfig(instr, numberExposures);
+        pta                 = PlannedTimeCalculator.instance.calc(config, new InstNIRI());
+        NiriReadoutTime nrt = NiriReadoutTime.lookup(instr.builtinROI(), instr.readMode()).getValue();
+        readoutTimePerCoadd = nrt.getReadout(1);
 
+        if (numberExposures < 2) step = 0; else step = 1;
+    }
```
sraaphorst: Again :-).
If you're copying code this much, you really need to refactor to put most of the code in the superclass. If you need to change this code at any point, it will be a nightmare to ensure it is changed everywhere consistently and correctly. Code reuse, not code duplication.

bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/Parameters.scala
olesjasmirnova: I think the problem was that getOrElse required a scala.Function0<…> argument, and I was not sure how to implement that.

bundle/edu.gemini.itc.shared/src/main/java/edu/gemini/itc/shared/ConfigCreator.java
olesjasmirnova: Ok, got it. I'll change it (here and further) according to your suggestions, thank you for recommending an actual way of fixing the duplications. Something for me to keep in mind in future.

bundle/edu.gemini.itc.shared/src/main/java/edu/gemini/itc/shared/ConfigCreator.java
olesjasmirnova: Well, as you noted above, the steps are not "exactly the same", the offsets are different. For the GSAOI (still in progress) there will be additional differences. I did it this way in order to reuse the OT methods for overhead calculations with as little hacking as possible. After a lot of consideration I decided that this way (OT sequence-like) it would be easier to maintain and change, in case of extra requirements, or if the ITC gets more sophisticated.

sraaphorst:
Hi there Olesja,

I was just wondering if you needed any help with this PR.
If you do, please let me know!

olesjasmirnova:
Hi Sebastian,

I had a mild crisis indeed, but I think I'm over it, so I hope to be done with this soon. If I don't manage, I'll ask for help, thank you!

sraaphorst:
Hi Olesja,

So sorry to hear about your crisis. I hope it's resolved and that it resolved well. 😞 
Take care and be well!

olesjasmirnova:
Hi Sebastian,

I did some code refactoring to get rid of duplications, and hopefully to make it look better in general.  A lot of changes at once, sorry (I know I need to learn to make smaller commits...). I have added all the instruments, so this is more or less it. Travis has failed the checks, but I don't think it's because of my changes, since it's in bundle_edu_gemini_horizons_api, which I didn't touch. 
Thanks, Olesja

sraaphorst:
Ha! I am still bad about making my pull requests way too big, so don't worry about it.

I'll look at this tomorrow. Did we want to get this into the 2018B release? @andrewwstephens ?

sraaphorst:
There's a failed test case here, but it isn't from this PR. I seem to be getting it in my local build as well.

tpolecat:
As discussed on Stride, this failure is due to horizons renaming a target, and there are no longer any known ways to get it to produce the output we're looking for. So it's ok to comment that test out.

sraaphorst:
Hi @olesjasmirnova... it looks like some of our test cases broke, and it's preventing the continuous integration system Travis from approving your code as successful. Could you update your code and rebase? If you don't know how to do it, it's along the lines of:

```
git checkout develop
git pull central develop
git push origin develop
git checkout feature/REL-3188
git rebase develop
git push origin feature/REL-3188
```

I think that should do the trick.

olesjasmirnova:
Thanks @sraaphorst , it worked this time.

bundle/edu.gemini.itc.shared/src/main/java/edu/gemini/itc/shared/ConfigCreator.java
```diff
@@ -0,0 +1,370 @@
...
+         */
+
+        if ((sourceFraction != 1.0) && (sourceFraction != 0.5)) {
+            result.addWarning("Warning: Observation overheads can only be calculated for the fraction of exposures on source 1.0 or 0.5.");
+            error = true;
+        }
+
+        if (!error) {
+            if (numLargeOffsets == 0) {
+                if (sourceFraction == 1) {
```
cquiroz: This probably should be `== 1.0` Still all exact comparison with doubles are a bit worrisome as there could be rounding errors

bundle/edu.gemini.itc.web/src/main/java/edu/gemini/itc/web/html/GmosPrinter.java
```diff
@@ -235,4 +248,23 @@ protected void _printPeakPixelInfo(final scala.Option<ItcCcd> ccd, final GmosSat
...
         }
     }
+
+    public ConfigCreator.ConfigCreatorResult createInstConfig(int numberExposures) {
+        ConfigCreator cc = new ConfigCreator(p);
+        return cc.createGmosConfig(instr, numberExposures);
+    }
+
+    public PlannedTime.ItcOverheadProvider getInst() {
+        if (instr.site().displayName.equals("Gemini North")) {
```
cquiroz: you should compare directly with `Site.GN` rather than with the string

bundle/edu.gemini.itc.web/src/main/scala/edu/gemini/itc/web/ITCRequest.scala
```diff
@@ -60,6 +60,15 @@ sealed abstract class ITCRequest {
       }
   }
 
+  /** Gets the named value as an integer. */
+  def intParameterCoadd(name: String): Int = parameter(name).trim() match {
+    case ""   => 0
+    case i    =>
+      try i.toInt catch {
+        case _: IllegalArgumentException => 1
```
cquiroz: so if there is an error we default to 1. Is that the desired behavior?

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/flamingos2/Flamingos2.java
```diff
@@ -879,6 +900,10 @@ public static double getCustomMaskSetupSec() {
         return 30 * 60;
     }
 
+    public double getReacquisitionTime() { return 6 * 60; }
```
cquiroz: I guess you can use `IMAGING_SETUP_TIME_PWFS2` here

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/niri/InstNIRI.java
```diff
@@ -263,13 +270,43 @@ public double getSetupTime(ISPObservation obs) {
         return (altair == null) ? getSetupTime(mode) : getSetupTime(mode, altair.getMode());
     }
 
+    public double getSetupTime(Config[] conf) {
+        String aoSystem = (String) conf[0].getItemValue(AOConstants.AO_SYSTEM_KEY);
```
cquiroz: Why is it hardcoded to be `conf[0]`?

bundle/edu.gemini.itc.shared/src/main/java/edu/gemini/itc/shared/ConfigCreator.java
```diff
@@ -0,0 +1,370 @@
...
+
+        public String getOffsetMessage() {
+            return offsetMessage;
+        }
+
+        public void setOffsetMessage(String offsetMessage) {
+            this.offsetMessage = offsetMessage;
+        }
+
+        public Config[] getConfig() {
```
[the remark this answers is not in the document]
olesjasmirnova: Made config private.

bundle/edu.gemini.itc.shared/src/main/java/edu/gemini/itc/shared/ConfigCreator.java
```diff
@@ -0,0 +1,370 @@
...
+    public static final ItemKey AOSystemKey = new ItemKey("adaptive optics:aoSystem");
+
+    private final ItcParameters itcParams;
+    private final ObservationDetails obsDetailParams;
+
+    public static final double GSAOI_SMALL_SKY_OFFSET = 120.0; // arcsec (assumed in case of sky offset <5')
+    public static final double GSAOI_LARGE_SKY_OFFSET = 310.0; // arcsec (assumed in case of sky offset >5')
+
+    public class ConfigCreatorResult {
+        private List<String> warnings;
```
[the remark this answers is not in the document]
olesjasmirnova: Fixed.

bundle/edu.gemini.itc.shared/src/main/java/edu/gemini/itc/shared/ConfigCreator.java
```diff
@@ -0,0 +1,370 @@
...
+        }
+        return result;
+    }
+
+    public ConfigCreatorResult createGmosConfig(GmosParameters gmosParams, int numExp) {
+        ConfigCreatorResult result = createCommonConfig(numExp);
+
+        for (Config step : result.getConfig()) {
+            if (gmosParams.site().displayName.equals(Site.GN.displayName)) {
+                step.putItem(InstInstrumentKey, SPComponentType.INSTRUMENT_GMOS.readableStr);
```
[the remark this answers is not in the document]
olesjasmirnova: Dunno, that's how it is defined in SPComponentType.java: INSTRUMENT_GMOSSOUTH for GMOS-S, and INSTRUMENT_GMOS for GMOS-N...

bundle/edu.gemini.itc.web/src/main/java/edu/gemini/itc/web/html/GmosPrinter.java
olesjasmirnova: Fixed.

bundle/edu.gemini.itc.web/src/main/scala/edu/gemini/itc/web/ITCRequest.scala
olesjasmirnova: Oops, this method is actually not used anywhere, I'll remove it.

bundle/edu.gemini.itc.shared/src/main/java/edu/gemini/itc/shared/ConfigCreator.java
olesjasmirnova: Fixed this. sourceFraction is a parameter introduced by user (not a calculated one), can it be still be affected by rounding errors? If so, do you prefer me to compare the difference between the two numbers instead?

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/niri/InstNIRI.java
olesjasmirnova: conf[0] is used to determine which setup time to use for the sequence (setup time for some instruments can vary depending on the observing mode/components used). I'm assuming that the sequence always has at least one step; this first step has all the information needed to determine the type of setup. Now that you asked, I'm thinking I should add a check that there is actually at least one step in that sequence...

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/flamingos2/Flamingos2.java
olesjasmirnova: Not really, IMAGING_SETUP_TIME_PWFS2 and reacquisition time are different things (it's a coincidence that they are equal here). Reacquisition time is used only for spectroscopy.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/flamingos2/Flamingos2.java
```diff
@@ -852,8 +856,25 @@ public static double getFpuChangeOverheadSec() {
...
+    private static final ItemKey GUIDE_WITH_OIWFS_KEY = new ItemKey(TELESCOPE_KEY, TargetObsCompConstants.GUIDE_WITH_OIWFS_PROP);
+
     public static double getImagingSetupSec(ISPObservation obs) {
-        return 60 * (usesF2Oiwfs(obs) ? 15 : 6);
+        return usesF2Oiwfs(obs) ? IMAGING_SETUP_TIME_OIWFS : IMAGING_SETUP_TIME_PWFS2;
+    }
+
+
+    public static double getImagingSetupSec(Config[] conf) {
+        if (conf[0].containsItem(GUIDE_WITH_OIWFS_KEY)) {
```
swalker2m: Since these methods work with just one `Config` instance, we should probably change the signature to reflect that fact.  This is, push up the extraction of the first step as much as possible.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/flamingos2/Flamingos2.java
```diff
@@ -852,8 +856,25 @@ public static double getFpuChangeOverheadSec() {
         return CommonStepCalculator.instance.calc(cur, prev).addAll(times);
     }
 
+    private static final double IMAGING_SETUP_TIME_OIWFS = 60 * 15;
+
+    private static final double IMAGING_SETUP_TIME_PWFS2 = 60 * 6;
```
swalker2m: I realize these are (a) trivial calculations that are immediately obvious, and (b) perfectly consistent with how these things are defined all over the OCS codebase.  Still, the new time API in Java offers a slight improvement here that we have lately been taking advantage of:

```java
    import java.time.Duration
    ...
    private static final Duration IMAGING_SETUP_TIME_OIWFS = Duration.ofMinutes(15);
    private static final Duration IMAGING_SETUP_TIME_PWFS2 = Duration.ofMinutes(6);
    ...
    public static double getImagingSetupSec(ISPObservation obs) {
        return (usesF2Oiwfs(obs) ? IMAGING_SETUP_TIME_OIWFS : IMAGING_SETUP_TIME_PWFS2).getSeconds();
    }
```

Really the `getImagingSetupSec` should return a `Duration` but that would be a bigger refactoring that we can skip since we should do this better in the new model.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/flamingos2/Flamingos2.java
```diff
@@ -889,13 +914,26 @@ public double getSetupTime(ISPObservation obs) {
...
      */
     private boolean isImaging() {
         return (_fpu == FPUnit.FPU_NONE) && (_disperser == Disperser.NONE);
     }
 
+    private boolean isImaging(Config[] conf) {
+        _fpu = (FPUnit) conf[0].getItemValue(FPUnit.KEY);
+        _disperser = (Disperser) conf[0].getItemValue(Disperser.KEY);
+
+        return (_fpu == FPUnit.FPU_NONE) && (_disperser == Disperser.NONE);
```
swalker2m: Why does this method mutate the value of `_fpu` and  `_disperser`?  It seems like it should be a `static` method that takes a `Config` and returns a value.

```java
private static boolean isImagingConfig(FPUnit fpu, Disperser disperser) {
    return (fpu == FPUnit.FPU_NONE) && (disperser == Dispserser.NONE);
}

private boolean isImagingConfig(Config conf) {
    return isImagingConfig(
        (FPUnit) conf.getItemValue(FPUnit.KEY),
        (Disperser) conf.getItemValue(Disperser.KEY));
}

private boolean isImaging() {
   return isImagingConfig(_fpu, _disperser);
}
```

olesjasmirnova: Ok, I have changed this as suggested.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/flamingos2/Flamingos2.java
olesjasmirnova: Thanks, I have fixed this for all methods using conf[0].

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gmos/InstGmosCommon.java
```diff
@@ -423,9 +431,22 @@ private void readObject(ObjectInputStream in) throws IOException, ClassNotFoundE
...
-        if (_disperser.isMirror()) return 6 * 60;
-        if (_fpu.isSpectroscopic() || _fpu.isNSslit()) return 16 * 60;
-        return 18 * 60.;
+        if (_disperser.isMirror()) return SETUP_TIME_IMAGING;
+        if (_fpu.isSpectroscopic() || _fpu.isNSslit()) return SETUP_TIME_LS_SPECTROSCOPY;
+        return SETUP_TIME_IFU_MOS;
+    }
+
+    public double getSetupTime(Config conf) {
+        String fpu_name = (String) conf.getItemValue(FPU_KEY);
```
swalker2m: I'm surprised this is a `String`.  Or "shocked but not surprised" or whatever they say these days.  Are we sure it isn't either a `GmosSouthType.FPUnitSouth` or `GmosNorthType.GPUnitNorth`?  Hopefully it is because then we can just directly ask whether it is IFU, etc. I would expect this method to be implemented something like:

```java
import java.util.Optional;
...
public double getSetupTime(Config conf) {
    return Optional.ofNullable(conf.getItemValue(FPU_KEY))
                   .map(c -> (GmosCommonType.FPUnit) c)
                   .filter(f -> !f.isImaging())
                   .map(f -> f.isIFU() ? SETUP_TIME_IFU_MOS : SETUP_TIME_LS_SPECTROSCOPY)
                   .orElse(SETUP_TIME_IMAGING);
 }
```

olesjasmirnova: Indeed... Thanks for pointing this out, I will fix some other stray Strings as well.

swalker2m: Whether the value in the `Config` is a `String` or an `FPUnit` will depend on how the sequence was created.  It appears that [this happens in `ItcParametersProvider`](https://github.com/gemini-hlsw/ocs/blob/develop/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcParametersProvider.scala#L82-L84).  If I'm correct then the use of the `IDENTITY_MAP` there means that it will be an `FPUnit`.

olesjasmirnova: Actually, since people are in a hurry to test this feature, I will stop looking for things to improve (there are too many of them), and add just the changes you requested.

andrewwstephens:
Hi, is this PR ready to be merged?  It would be great if we could make a test release and get feedback from the instrument scientists.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gnirs/InstGNIRS.java
```diff
@@ -222,10 +242,27 @@ public double getSetupTime(ISPObservation obs) {
...
         }
 
-        return (_slitWidth == SlitWidth.IFU) ? (20 * 60) : (15 * 60);
+        return (_slitWidth == SlitWidth.IFU) ? (getSetupTimeIfu()) : (getSetupTime());
+    }
+
+    public double getSetupTime(Config conf) {
+        String aoSystem = (String) conf.getItemValue(AOConstants.AO_SYSTEM_KEY);
+        String guideStarType = (String) conf.getItemValue(AOConstants.AO_GUIDE_STAR_TYPE_KEY);
+        String slitWidth = (String) conf.getItemValue(GNIRSConstants.SLIT_WIDTH_KEY);
```
swalker2m: This should be a `GNIRSParams.SlitWidth` I believe, which can be checked to see whether it equals `GnirsParams.SlitWidth.IFU`.

bundle/edu.gemini.itc.shared/src/main/java/edu/gemini/itc/shared/ConfigCreator.java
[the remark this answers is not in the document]
swalker2m: @cquiroz yeah it isn't consistent.  This is really old code that must have been created before GMOS-S was in use.  It should have been updated but never was.

bundle/edu.gemini.itc.shared/src/main/java/edu/gemini/itc/shared/ConfigCreator.java
```diff
@@ -0,0 +1,374 @@
...
+    /**
+     * Instrument-specific configs
+     */
+    public ConfigCreatorResult createGnirsConfig(GnirsParameters gnirsParams, int numExp) {
+        ConfigCreatorResult result = createCommonConfig(numExp);
+
+        for (Config step : result.getConfig()) {
+            step.putItem(ReadModeKey, (gnirsParams.readMode()));
+            step.putItem(InstInstrumentKey, SPComponentType.INSTRUMENT_GNIRS.readableStr);
+            step.putItem(SlitWidthKey, gnirsParams.slitWidth().logValue());
```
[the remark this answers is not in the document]
swalker2m: `Config`, and the sequence model as a whole, was not a great API unfortunately.  There are no guarantees about which keys will be defined. It's impossible to know what the types of the various keys are without checking.  I'd be more comfortable moving all the new API methods that accept a `Config` out of the `spModel` packages and into the ITC itself.  At that point I'd still recommend using the actual typed values from the model (like the `GNIRSParams.SlithWidth` instead of adding `String`s).  That way you can use the APIs they provide instead of doing `String` comparisons.  Nevertheless if it is all internal to the ITC I'm less concerned about it.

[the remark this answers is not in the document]
olesjasmirnova: Thanks for reviewing. I was using Strings originally because I thought I *had* to use Strings for Config (I looked at the Seqexec config indeed). I replaced some of them with typed values after I realized I was wrong, but some were left over. I can try to fix it. I don't think I'll have time for major refactoring like taking all the methods out of spModel. If the whole idea of simulating Config and sequence doesn't seem great anyway, maybe it's now worth the effort too. I have two 1.5 project day left, I'll do what I can.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gnirs/InstGNIRS.java
```diff
@@ -235,7 +272,11 @@ public double getSetupTime(ISPObservation obs) {
      * @return time in seconds
      */
     public double getReacquisitionTime(ISPObservation obs) {
-        return 6 * 60; // 6 minutes as defined in REL-1346
+        return getReacquisitionTime();
+    }
+
+    public double getReacquisitionTime () {
+        return REACQUISITION_TIME;
```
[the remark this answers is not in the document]
olesjasmirnova: Indeed, getReacquisitionTime implements PlannedTime.ItcOverheadProvider

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gnirs/GnirsReadoutTime.java
```diff
@@ -11,23 +11,36 @@
...
+ *
+ *  2017-12-15: Modified according to the latest read time measurements.
  */
 public final class GnirsReadoutTime {
-    //Maps from the read mode to an array containing the overhead per frame and the overhead per coadd per frame
-    private static final Map<GNIRSParams.ReadMode, double[]> map;
+    private static final double DHS_WRITE_TIME = 8.5;
+
+    //Maps from the read mode to an array containing the overhead per coadd per frame
+    private static final Map<GNIRSParams.ReadMode, Double> map;
```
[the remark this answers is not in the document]
olesjasmirnova: Fixed

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gnirs/InstGNIRS.java
olesjasmirnova: Replaced all Strings with the typed values from the model.

bundle/edu.gemini.itc.shared/src/main/java/edu/gemini/itc/shared/ConfigCreator.java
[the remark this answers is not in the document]
olesjasmirnova: Hi Shane, I have fixed all the Strings in Config and spModel, using typed values instead, in the latest PR. I didn't move anything out of the spModel though. I hope this is ok for testing.

sraaphorst:
Given that @olesjasmirnova's last day at Gemini was yesterday, I suppose we should decide how to move forward with this PR.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/flamingos2/Flamingos2.java
sraaphorst: We can't make this `static` since it overrides a method in `ItcOverheadProvider`.
I marked it `@Override` instead.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/flamingos2/Flamingos2.java
```diff
@@ -889,11 +911,27 @@ public double getSetupTime(ISPObservation obs) {
...
+
+    private static boolean isImagingConfig(Config conf) {
+        return isImagingConfig(
+                (FPUnit) conf.getItemValue(FPUnit.KEY),
+                (Disperser) conf.getItemValue(Disperser.KEY));
+    }
+
     private boolean isImaging() {
-        return (_fpu == FPUnit.FPU_NONE) && (_disperser == Disperser.NONE);
+        return isImagingConfig(_fpu, _disperser);
```
sraaphorst: I really don't see an easy way to extract these changes from the model, since the original model methods have been changed to use them as well, so I'll leave them for now.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gnirs/InstGNIRS.java
```diff
@@ -222,10 +243,28 @@ public double getSetupTime(ISPObservation obs) {
...
+            }
+        }
+
+        SlitWidth slitWidth = (SlitWidth) conf.getItemValue(GNIRSConstants.SLIT_WIDTH_KEY);
+
+        if (slitWidth.equals(GNIRSParams.SlitWidth.IFU)) {
+            return getSetupTimeIfu();
+        } else  {
+            return getSetupTime();
+        }
```
sraaphorst: This method could be factored out, if it wasn't part of `ItcOverheadProvider`.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/flamingos2/Flamingos2.java
sraaphorst: This makes me raise an eyebrow. We have:

`SPInstObsComp.reacquisitionTime(ISPObservation)` which returns 0 and is not overridden here.

Then we have `ItcOverheadProvider.reacquisitionTime()` which reutrns 6 minutes.

Is this correct? Should the two not be the same? This seems dubious.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gsaoi/Gsaoi.java
```diff
@@ -726,24 +745,59 @@ private static boolean isExpensiveOffset(Config cur, Option<Config> prev) {
...
+        }
+        return ctg;
     }
 
 
+    public double readout(int coadds, int lowNoiseReads) {
+        return 21 + 2.8 * lowNoiseReads * coadds + 6.5 * (coadds - 1);
+    }  // REL-445
+
     @Override public CategorizedTimeGroup calc(Config cur, Option<Config> prev) {
```
sraaphorst: This method is making me really uncomfortable, because it's declared in two interfaces (`StepCalculator` and `ItcOverheadsProvider`) that are both inherited by this class, and some others.

Having the same method coming from two interfaces seems like a design flaw.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/obs/plannedtime/PlannedTime.java
```diff
@@ -404,4 +420,67 @@ public static PlannedTime apply(Setup setup, List<Step> steps, ConfigSequence se
...
+    }
+
+    // total time with acquisitions and re-acquisitions.
+    // Uses the parameter because of GSAOI LGS re-acquisitions.
+    public long totalTimeWithReacq(int numReacq) {
+        long totalTimeWithReacq = setup.time * numAcq() + setup.reacquisitionTime * numReacq;
+        for (Step step : steps) totalTimeWithReacq += step.totalTime();
+        return totalTimeWithReacq;
+    }
+
```
sraaphorst: I guess that these could probably be moved to a different class. I'll do that, at least.

PR #1402, opened 2018-01-05
REL-3342: Restored magnitude property settings to GpiCB.

sraaphorst:
This restores the accidentally removed setting of magnitudes for H and I band in `GpiCB.java`, and has been rewritten to make it clear that these properties are being set instead of relying on unusual uses of `ApplyOp`.


cquiroz:
Looks fine to me

PR #1406, opened 2018-01-05
REL-3342: Reverting to fix the break in serialization from deleting inner classes.

sraaphorst:
RFC, and then we'll do a test release to make sure this works as expected when we are ready.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gpi/GpiCB.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gpi/GpiCB.java
@@ -1,6 +1,8 @@
 package edu.gemini.spModel.gemini.gpi;
 
 import edu.gemini.pot.sp.ISPObsComponent;
+import edu.gemini.shared.util.immutable.ApplyOp;
+import edu.gemini.shared.util.immutable.Function1;
 import edu.gemini.shared.util.immutable.ImList;
 import edu.gemini.spModel.config.AbstractObsComponentCB;
 import edu.gemini.spModel.core.Magnitude;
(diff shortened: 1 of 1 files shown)


bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gpi/GpiCB.java
```diff
@@ -44,46 +44,61 @@ protected void thisReset(Map<String, Object> options) {
...
             if (toc != null) {
-                // The Asterism here should be Single(base), so we only need to concern ourselves with the head target.
-                toc.getAsterism().allSpTargetsJava().headOption().foreach(base -> {
-                    final ImList<Magnitude> magnitudes = base.getMagnitudesJava();
-                    MAG_PROPERTIES.forEach((band, propName) ->
-                        magnitudes.find(m -> m.band().equals(band)).
-                                foreach(mag -> config.putParameter(systemName, DefaultParameter.getInstance(propName, mag.value()))));
-                });
+                if (toc.getAsterism().allSpTargetsJava().head() != null) {
+                    ImList<Magnitude> magnitudes = toc.getAsterism().allSpTargetsJava().head().getMagnitudesJava();
```
sraaphorst: This line and the preceding one needed to be changed for the new way targets are handled, but everything else is as it was back beforehand.

swalker2m:
Did you verify that serialization doesn't break?

sraaphorst:
Yes. I copied over the `.sp` files from `gsodb` that were reported as being unserializable directly into my `spdb` directory and started my local `spdb`, and they are all listed in `lsprogs` with no errors reported.

PR #1407, opened 2018-01-05
REL-3341: Implement 4 arcsec squares around active guide probe targets for CWFS

sraaphorst:
This code is an extension of my earlier PR that I released for NGS2 regarding OT_REQ_002, which requests squares of 4 arcsecs drawn around targets. Discussions from Thursday revealed that science staff is stating that these squares are not just for display in the TPE and have physical significance, i.e. they must fall completely in the Canopus patrol field and cannot overlap.

This PR extends on my earlier work to implement the code for target validation, asterism filtering, and Phase-2 checks to take these physical constraints into account.

Unfortunately, to determine if two squares overlap, we need the position angle, as two squares next to each other may intersect when rotated at certain angles, and not intersect otherwise, so I had to change how the position angle was handled a bit in the way it was passed around in MASCOT, but this was fairly minor.

Here is a screenshot of some manual guide stars, showing:
1. CWFS1's window falls outside of the Canopus patrol field; and
2. The guiding window for CWFS2 and CWFS3 overlap slightly, generating a P2 error.

screen shot 2018-01-05 at 5 38 25 pm

### DIFF
--- a/bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/gems/GemsGuideStarRule.java
+++ b/bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/gems/GemsGuideStarRule.java
@@ -7,10 +7,7 @@ import edu.gemini.p2checker.api.P2Problems;
 import edu.gemini.p2checker.util.PositionOffsetChecker;
 import edu.gemini.pot.sp.ISPProgramNode;
 import edu.gemini.pot.sp.SPComponentType;
-import edu.gemini.shared.util.immutable.DefaultImList;
-import edu.gemini.shared.util.immutable.ImList;
-import edu.gemini.shared.util.immutable.ImOption;
-import edu.gemini.shared.util.immutable.Option;
+import edu.gemini.shared.util.immutable.*;
 import edu.gemini.spModel.core.SiderealTarget;
 import edu.gemini.spModel.gemini.flamingos2.Flamingos2;
 import edu.gemini.spModel.gemini.flamingos2.Flamingos2OiwfsGuideProbe;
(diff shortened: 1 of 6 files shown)


bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/gems/GemsGuideStarRule.java
```diff
@@ -109,20 +107,29 @@ public IP2Problems check(ObservationElements elements)  {
...
+                final CanopusWfs c2 = cwfsTargets.get(t2);
+                problems.addError(PREFIX + "GuideWindowOverlapError", String.format(WindowOverlap, c1.getKey(), c2.getKey()), targetNode);
+            });
 
             // Check for magnitude violations.
-            final ImList<SiderealTarget> cwfsTargetsImList = DefaultImList.create(cwfsTargets);
-            final ImList<Double> cwfsMagnitudes = CanopusWfs.Group.instance.extractRMagnitudes(cwfsTargetsImList);
-            if (!cwfsTargets.isEmpty() && !cwfsMagnitudes.isEmpty() && !CanopusWfs.Group.instance.checkAsterismMagnitude(DefaultImList.create(cwfsTargets)))
+            final ImList<Double> cwfsMagnitudes = CanopusWfs.Group.extractRMagnitudes(cwfsTargetsImList);
+            if (!cwfsTargets.isEmpty() && !cwfsMagnitudes.isEmpty() && !CanopusWfs.Group.checkAsterismMagnitude(cwfsTargetsImList))
```
swalker2m: Is it possible to have non-empty `cwfsMagnitudes` without `cwfsTargets`?

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gems/CanopusWfs.java
```diff
@@ -70,14 +75,64 @@ public boolean asterismPreFilter(final ImList<SiderealTarget> targets) {
...
+            // Must use indices here to avoid double pairs of the form (t1,t2) and (t2,t1).
+            for (int i=0; i < targets.size(); ++i) {
+                final SiderealTarget t1 = targets.get(i);
+                final Area a1 = windowFromCoordinates(posAngle, t1.coordinates());
+                for (int j=i+1; j < targets.size(); ++j) {
+                    final SiderealTarget t2 = targets.get(j);
+                    final Area a2 = windowFromCoordinates(posAngle, t2.coordinates());
+
+                    // To check overlap, take a1 xor a2 and compare to a1 union a2.
+                    // If they are not equal, they overlap.
```
swalker2m: Would it not just be: `!a1.intersect(a2).isEmpty` implies overlap?

bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/gems/GemsGuideStarRule.java
sraaphorst: This only came up when I just made some dummy manual targets and didn't specify any magnitudes for them. In that case, `CanopusWfs.Group.checkAsterismMagnitude` returns `false`. Without the check for non-empty `cwfsMagnitudes`, the error would be generated, and it didn't seem right to me to issue an error about magnitudes being too far from one another when there were no magnitudes specified at all.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gems/CanopusWfs.java
sraaphorst: Yes, that's much better (although `intersect` returns `void` and mutates the original object). 
I'll clone `a1`, intersect with `a2`, and then check `!intArea.isEmpty()`.

I'll make the change and test it. Thanks for helping me to simplify this.

bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/gems/GemsGuideStarRule.java
swalker2m: My question is whether we need both `!cwfsTargets.isEmpty() && !cwfsMagnitudes.isEmpty()`.  The magnitudes can only be non-empty if there are targets, right?

sraaphorst: Oh... ha! Yes. That's a good point. Okay, I'll push another commit to this PR and reopen it.

PR #1412, opened 2018-01-19
REL-3557: Change p1monitor to use the old style for DD proposals

cquiroz:
This PR changes the template used for DD and PW to include the full info unlike the current status where the pdf was formatted with the unbiased version

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -6,7 +6,7 @@ organization in Global := "edu.gemini.ocs"
 
 ocsVersion in ThisBuild := OcsVersion("2016A", true, 1, 1, 1)
 
-pitVersion in ThisBuild := OcsVersion("2018A", false, 1, 1, 0)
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -6,7 +6,7 @@ organization in Global := "edu.gemini.ocs"
 
 ocsVersion in ThisBuild := OcsVersion("2016A", true, 1, 1, 1)
 
-pitVersion in ThisBuild := OcsVersion("2018A", false, 1, 1, 0)
+pitVersion in ThisBuild := OcsVersion("2018A", false, 1, 1, 1)
```
sraaphorst: Why are we adding another point version to the PIT? (Not that I think it's a bad idea... just curious.)

cquiroz: Because i'm releasing a minor fix for p1-monitor who uses the same version as the pit

sraaphorst: Oh, of course. Sorry for the stupid question. 😳

PR #1413, opened 2018-01-25
REL-3333: NGS2 Canopus manual revert.

sraaphorst:
This is the final revert to develop for NGS2.

Github wouldn't let me autorevert, and reverting the commits individually seemed like it would be a nightmare. Since almost all of this, except the change from `Canopus` to `CanopusWfs` and the changes to `CanopusFeature`, was basically trivial altering of files to use `CanopusWfs` instead of `Canopus.Wfs`, it was just easier to manually revert.

I did leave out reverting some minor changes that won't affect us, like the unused UCAC regression tests, but otherwise just restored the raw files from the previous version.


swalker2m:
> I did leave out reverting some minor changes that won't affect us, like the unused UCAC regression tests, but otherwise just restored the raw files from the previous version.

I'm wondering about the plan going forward.  You made a separate branch that includes all these changes already?  Into that you plan to merge from the develop branch along the way and then finally after June merge the separate branch back to develop?  If so, I'm worried about the conflict mess you might be facing.

sraaphorst:
Yes, I have all of the changes in a separate branch, NGS2.
The plan is to _try_ to merge this when it finally comes time to release NGS2, which I suspect won't even be until June 2019, IIRC.

I am, frankly, somewhat terrified of the conflict mess I will be facing, but I'm not quite sure the best way to handle it apart from just muddle through it when the time comes.

sraaphorst:
I think maybe the best thing to do is periodically rebase NGS2 on develop (e.g. whenever something major or that touches the same code) and fix any merge conflicts that come up in the NGS2 branch?

swalker2m:
I think a source of conflicts would be avoided if the separate branch were created after all of the unwinding of this feature finishes.  But I suppose it will be equally painful to add the changes back to the separate branch if you went that way.

sraaphorst:
Agreed, but it's a bit late for that now, since I already created the new branch when we were first told this shouldn't go into develop because it wasn't scheduled to be released for at least another year. Hindsight.

I think once this is done, I'll rebase develop on NGS2. Wouldn't that have a similar effect, but with the caveat that there might be merge conflicts to fix now?

swalker2m:
> Agreed, but it's a bit late for that now, since I already created the new branch when we were first told this shouldn't go into develop because it wasn't scheduled to be released for at least another year. Hindsight.

You can always create a new branch of course.  I'm not sure it will be any easier to add the changes back to the new branch, but it would avoid some conflicts I think.

> I think once this is done, I'll rebase develop on NGS2. Wouldn't that have a similar effect, but with the caveat that there might be merge conflicts to fix now?

You mean you'll rebase NGS2 on develop?  I think that's what you'll have to do and I think it will produce some difficult conflicts.  If you find that it is too much of a mess you can always create a new branch starting from the end of the revert process and re-do your changes afterwards.

sraaphorst:
Agreed on the largeness of the PR. I think part of it was that the change from `Canopus` to `CanopusWfs` touched so many files that it was going to be large no matter what, but I probably should have made the minimal changes to `Canopus` and `CanopusFeature` first.

I'm definitely guilty of large PRs, and I'll try to keep them smaller in the future for everyone's sake. 😊

PR #1416, opened 2018-01-30
REL-3273: Add OVI and OVIC filters for GMOS-N and GMOS-S to PIT.

sraaphorst:
This mirrors task REL-3272, i.e. adding the OVI and OVIC filters for GMOS-N and GMOS-S to the OT, which has already been done.

The PIT must support these new filters for 2018B. This adds the filters and their internal names to the XJB and XSD files, as well as to the immutable classes.

I think we will also need to modify the template creation process, e.g. `GmosNImaging.defaultExposures`:
https://github.com/gemini-hlsw/ocs/blob/f6e26b07e05f7ba3c69b4850534f3303cfe56153/bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/template/factory/impl/gmos/GmosNImaging.scala#L40

### DIFF
--- a/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/package.scala
+++ b/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/package.scala
@@ -130,9 +130,9 @@ package object immutable {
       Ha_G0310,
       HaC_G0311,
       SII_G0317,
-      DS920_G0312,
       OVI_G0345,
       OVIC_G0346,
+      DS920_G0312,
       USER_SUPPLIED
     )
     val ALTAIR_IMAGING = List(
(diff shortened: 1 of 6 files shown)


cquiroz:
Did you check they appear on the intstrument selection wizard?

sraaphorst:
@cquiroz: Yes, they appear for both. I'm home now, but can post a screenshot tomorrow.

sraaphorst:
Here are a couple screenshots, with the filter in the GMOS-N filter selection list, and as it appears in the Observations pane as a resource. (It also is available for GMOS-S.)
screen shot 2018-01-31 at 9 24 19 am
screen shot 2018-01-31 at 9 24 35 am

swalker2m:
I think you're right about the default exposures.  Perhaps those values should go directly into the filter enumeration values themselves though instead of being maintained as a separate map.  With an `imagingExposure` or something.

sraaphorst:
I think that's a worthwhile change, given that there's no reason to maintain this as a separate map as far as I can see. I'm going to make a JIRA task to add this for OVI / OVIC and I'll make a note of that there.

PR #1418, opened 2018-02-01
REL-3349: PIT visitor instrument changes.

sraaphorst:
The following have been listed as requirements for the 2018B PIT:

1. Remove TEXES from the list of instruments.
2. Add Phoenix to the list of instruments (GS only).
3. Make DSSI GS only, as Alopeke will be used for GN instead.

In this PR, I adjust the instrument list accordingly, i.e. removing TEXES and adding Phoenix back, and change DSSI to be GS only. Here is the instrument list now:
screen shot 2018-02-01 at 2 55 59 pm

Here is what you now see when you select DSSI. (This is how Phoenix works, despite the strangeness of a one option selection box, so I followed suit to indicate that it is only at GS, especially important now since this has not always been the case.)
screen shot 2018-02-01 at 2 56 20 pm

The UpConverter now takes DSSI observations at GN and issues a warning, and changes them to GS with the information that Alopeke should be used for GN observations instead.
screen shot 2018-02-01 at 2 57 33 pm

Additionally, the UpConverter also assigns Phoenix proposals to GS. This was done in an earlier upconversion, so I simply grabbed the code out of there and moved it to the 2018B conversion.

Finally, the list of problems now generates issues if the configuration includes TEXES, or Phoenix at GN, or DSSI at GN. This should never happen based on the upconversion, but I thought it prudent to include it anyway.

### DIFF
--- a/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/transform/UpConverter.scala
+++ b/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/transform/UpConverter.scala
@@ -265,7 +265,7 @@ case object SemesterConverter2016ATo2016B extends SemesterConverter {
     }
 
   val dssiSite: TransformFunction = {
-    case <dssi>{ns @ _*}</dssi> if (ns \ "site").isEmpty =>
+    case p @ <dssi>{ns @ _*}</dssi> if (p \\ "Dssi" \\ "site").isEmpty =>
       object DssiSiteTransformer extends BasicTransformer {
         override def transform(n: xml.Node): xml.NodeSeq = n match {
           case p @ <Dssi>{q @ _*}</Dssi> => <Dssi id={p.attribute("id")}>{q.map(transform) +: <site>{Site.GS.name}</site>}</Dssi>
--- a/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/transform/UpConverter.scala
+++ b/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/transform/UpConverter.scala
@@ -138,17 +138,17 @@ case class LastStepConverter(semester: Semester) extends SemesterConverter {
  * This converter will upgrade to 2018B.
  */
 case object SemesterConverter2018ATo2018B extends SemesterConverter {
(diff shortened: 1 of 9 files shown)


bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/transform/UpConverter.scala
```diff
@@ -138,7 +138,38 @@ case class LastStepConverter(semester: Semester) extends SemesterConverter {
...
+    case <dssi>{ns @ _*}</dssi> =>
+      object DssiSiteTransformer extends BasicTransformer {
+        override def transform(n: xml.Node): xml.NodeSeq = n match {
+          case p @ <Dssi>{q @ _*}</Dssi> => <Dssi id={p.attribute("id")}>{q.map(transform) +: <site>{Site.GS.name}</site>}</Dssi>
+          case <name>{name}</name>       => <name>DSSI {Site.GS.name}</name>
+          case elem: xml.Elem            => elem.copy(child = elem.child.flatMap(transform))
+          case _                         => n
+        }
+      }
+      StepResult("DSSI proposal has been assigned to Gemini South. Use ʻAlopeke instead for Gemini North.", <dssi>{DssiSiteTransformer.transform(ns)}</dssi>).successNel
```
cquiroz: Why would you say `use Alopeke` when converting DSSI?

sraaphorst: From conversations, I was under the impression that Alopeke was being offered as a "better" DSSI, but I might be wrong. I'll double check that with Andy.

sraaphorst: Okay, talking to @andrewwstephens, it looks like Alopeke in speckle mode is an enhanced DSSI, so DSSI and GN proposals should migrate to that instead, with a message being issued. I'll make that change.

sraaphorst:
@cquiroz: Now I changed the code so that DSSI GN proposals automatically get changed to Alopeke in speckle mode. I added a test case to the spec file to test this.

cquiroz:
Thanks for the fixes

PR #1420, opened 2018-02-02
REL-3351: Made Alopeke a visitor instrument.

sraaphorst:
I noticed during testing that both DSSI and TEXES proposals had a tag in the instrument node with `<visitor>true</visitor>` and that Alopeke had `<visitor>false</visitor>`, which is wrong and something I missed in the initial implementation.

This fixes that.


cquiroz:
That's an important bit on the model. Perhaps we should have a test to verify it

sraaphorst:
Sure, I'll add something.

sraaphorst:
I also noticed while making `AlopekeBlueprintSpec` that I didn't include Alopeke in `BlueprintBase`, so this now fixes that as well.

cquiroz:
Great, serialization tests should be always done as it's easy to miss something on the output

PR #1423, opened 2018-02-06
REL-3349: Fixed visitor instruments in PIT

sraaphorst:
I had forgotten to:
1. Remove the code from an earlier upconversion that removed Phoenix. Since it is offered this semester, that was problematic.
2. Texes was not being removed.

Phoenix now remains and Texes is now removed.
Furthermore, the code to assign Phoenix to GS is now only run if the site is missing or set to GN.


cquiroz:
It maybe good to have a test case for these. Otherwise it looks 👍

PR #1425, opened 2018-02-06
REL-3363: GMOS-N does not support Altair in 2018B

sraaphorst:
Andy pointed out that GMOS-N does not support Altair for 2018B. Thus, I:
1. Removed the Altair node from the GMOS-N instrument tree, instead forcing `AltairNone` to be used.
2. Removed any Altair mode other than `AltairNone` during upconversion, issuing a warning.
3. Removed references to Altair modes in GMOS-N names (as they will always be `None`, which would be confusing).
4. Added a case to the `ProblemRobot` (which should never happen) that reports an error if one tries to use GMOS-N with an Altair mode.
5. Added a test case to `UpConverterSpec` to make sure the Altair mode was set to `AltairNone`.

### DIFF
--- a/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/transform/UpConverter.scala
+++ b/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/transform/UpConverter.scala
@@ -142,6 +142,21 @@ case class LastStepConverter(semester: Semester) extends SemesterConverter {
  * This converter will upgrade to 2018B.
  */
 case object SemesterConverter2018ATo2018B extends SemesterConverter {
+  // REL-3363: No GMOS-N + Altair offered for 2018B. Remove Altair from component.
+  // The name of the instrument is auto-generated from its properties, so we don't worry about the <name>...</name> tag.
+  lazy val gmosnAltairRemoverMessage = "GMOS-N does not offer Altair in 2018B. Altair component removed."
+  val gmosnAltairRemover: TransformFunction = {
+    case p @ <gmosN>{ns @ _*}</gmosN> if (p \\ "altair" \ "lgs").nonEmpty || (p \\ "altair" \\ "ngs").nonEmpty =>
+      object GmosNAltairRemover extends BasicTransformer {
+        override def transform(n: xml.Node): xml.NodeSeq = n match {
+          case <altair>{_ @ _*}</altair> => <altair><none/></altair>
+          case elem: xml.Elem            => elem.copy(child = elem.child.flatMap(transform))
+          case _                         => n
+        }
+      }
+      StepResult(gmosnAltairRemoverMessage, <gmosN>{GmosNAltairRemover.transform(ns)}</gmosN>).successNel
+  }
+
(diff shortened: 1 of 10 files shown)


bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/transform/UpConverter.scala
```diff
@@ -142,6 +142,21 @@ case class LastStepConverter(semester: Semester) extends SemesterConverter {
  * This converter will upgrade to 2018B.
  */
 case object SemesterConverter2018ATo2018B extends SemesterConverter {
+  // REL-3363: No GMOS-N + Altair offered for 2018B. Remove Altair from component.
+  // The name of the instrument is auto-generated from its properties, so we don't worry about the <name>...</name> tag.
```
cquiroz: Is that so? In some cases we do transform the name, are you sure it isn't necessary in this case?

bundle/edu.gemini.model.p1/src/test/scala/edu/gemini/model/p1/immutable/transform/UpConverterSpec.scala
```diff
@@ -559,6 +559,17 @@ class UpConverterSpec extends Specification with SemesterProperties with XmlMatc
...
       }
     }
+    "proposal with GmosN with Altair mode must have it set to AltairNone for 2018B, REL-3363" in {
+      val xml = XML.load(new InputStreamReader(getClass.getResourceAsStream("proposal_with_gmosn_altair_lgs.xml")))
+
+      val converted = UpConverter.convert(xml)
+      converted must beSuccessful.like {
+        case StepResult(changes, result) =>
+          changes must contain(SemesterConverter2018ATo2018B.gmosnAltairRemoverMessage)
+          result \\ "altair" must \\("none")
```
cquiroz: Shall we test the `name` node too?

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/transform/UpConverter.scala
sraaphorst: In the cases I tested, it wasn't necessary, but I may have missed a case.

Are there any instances in where we will upconvert just the XML and not actually create the blueprints from it?

sraaphorst: I'm going to go ahead and add the name change as well, but unfortunately, it's going to be a bit of a hassle, I think.

sraaphorst: Okay, so not nearly as bad as expected.

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/transform/UpConverter.scala
```diff
@@ -142,6 +142,24 @@ case class LastStepConverter(semester: Semester) extends SemesterConverter {
...
```
sraaphorst: Still need to remove this comment, as it is no longer applicable.

PR #1427, opened 2018-02-14
REL-2659: Complete ITC support for summing GMOS IFU elements

jocelynferrara:
This is the continuation of the upgrade to the GMOS ITC to support a new IFU analysis method to sum all apertures with a specified radius of the center, e.g.:
Sum all IFU elements within ____ arcsec of the center

This resolves the issues of:
Center and pack fibers using correct geometry
Display the number of summed elements in the generated plot title
Display the analysis method in the GMOS ITC page

### DIFF
--- a/bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gmos/IFUComponent.java
+++ b/bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gmos/IFUComponent.java
@@ -69,9 +69,9 @@ public final class IFUComponent extends TransmissionElement {
 
         int Nelements = 0;
         for (int i = 0; i < numX; i++) {
-            double x = (i - (numX/2.)) * (0.75*IFU_DIAMETER + IFU_SPACING);
+            double x = (i - (numX/2.)) * ((Math.sqrt(3)/2.)*IFU_DIAMETER + IFU_SPACING);
             for (int j = 0; j < numY; j++) {
-                double y = (2*j - (numY - Math.abs(i)%2 + 1)) * ((Math.sqrt(3)/4.)*IFU_DIAMETER + IFU_SPACING);
+                double y = (2*j - (numY - Math.abs(i)%2 + 1)) * (IFU_DIAMETER/2. + IFU_SPACING);
                 double r = Math.sqrt(Math.pow(x, 2) + Math.pow(y, 2));
                 if (r < radius) {
                     IFUApertures.addAperture(new HexagonalAperture(x, y, IFU_DIAMETER));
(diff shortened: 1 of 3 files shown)


cquiroz:
I can't comment much about the contents of the change but it seems fine code-wise

andrewwstephens:
This looks really good!

I may be missing something, or have my orientation flipped, but I expected something like
`x = (i - numX/2) * sqrt(3)/2 * (IFU_DIAMETER + IFU_SPACING)`
`y = (2j - numY - abs(i)%2 + 1) * (IFU_DIAMETER + IFU_SPACING)/2`

Would it be possible to help me understand?

jocelynferrara:
Andy, yes, first there is the matter of orientation. My research on the IFU showed that this would be the correct way to pack the IFU lenslets, if we are in fact setting the ranges with the maximum numX = 20 or 40 for one or two slit mode, respectively.

As for your actual calculation concern -- this is set up to get to the center point of each hexagonal lenslet packed together with a *constant* IFU_SPACING in between (so no multiplication factor is applied to that value). 
The multiplication factors themselves may seem off to you because we are working with a hexagonal diameter (vertex to opposite vertex), so it's not a "neat" factor of 1/2 in the x direction, rather 3/4. Similarly, an increase by one row in the y direction for one row is sqrt(3)/4 (rather than over 2) simply because we have a diameter and not a radius.

andrewwstephens:
Thanks for the explanation!  :+1:

jocelynferrara:
Turns out vertex-to-opposite-vertex was the wrong hexagonal diameter definition, latest commit reflects side-to-opposite-side definition confirmed by instrument scientist.

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gmos/IFUComponent.java
```diff
@@ -69,9 +69,9 @@ public IFUComponent(final String prefix, final double radius, final Boolean isIf
 
         int Nelements = 0;
         for (int i = 0; i < numX; i++) {
-            double x = (i - (numX/2.)) * (0.75*IFU_DIAMETER + IFU_SPACING);
+            double x = (i - (numX/2.)) * ((Math.sqrt(3)/2.)*IFU_DIAMETER + IFU_SPACING);
             for (int j = 0; j < numY; j++) {
-                double y = (2*j - (numY - Math.abs(i)%2 + 1)) * ((Math.sqrt(3)/4.)*IFU_DIAMETER + IFU_SPACING);
+                double y = (2*j - (numY - Math.abs(i)%2 + 1)) * (IFU_DIAMETER/2. + IFU_SPACING);
```
sraaphorst: A nitpicky point, but we've been trying to purge the code of things like `2.` for `double`s when we find them, and instead change them to `2.0` for clarity.

Otherwise, this looks fine to me. The only thing I can think of is that calculating `Math.sqrt(3)/2.0` in a loop is a bit redundant, as the value doesn't change, so it would be better to calculate it prior to the loop. Same with `IFU_DIAMETER/2.0`.

These are not serious issues, though.

andrewwstephens:
:+1:

PR #1428, opened 2018-02-22
REL-3351: Changed okinas to apostrophes to avoid unicode issues.

sraaphorst:
Reports with the Hawaiian okina "apostrophe-like" character were making this display as `#Alopeke`. I changed to the apostrophe to fix this.


sraaphorst:
Whoops. Conflicts. I thought I rebased properly, but apparently not.
Will fix and recreate.

PR #1429, opened 2018-02-22
REL-3351: Changed okinas to apostrophes to avoid unicode issues, take 2.

sraaphorst:
Reports with the Hawaiian okina "apostrophe-like" character were making this display as `#Alopeke`. I changed to the apostrophe to fix this.


cquiroz:
Is the `#` displayed on the pdfs? I bet it is a font issue, but I think using `apostrophe` is simpler

sraaphorst:
Yes, that is exactly it. I don't think it was supported in the font, and I figured using a weird, seldom seen unicode character would cause problems later on down the road.

Andy said an apostrophe was fine, hence the switch.

PR #1430, opened 2018-02-23
REL-3290: Add an empty "dummy" observation when there are no observations

sraaphorst:
This task is to help guide users of the PIT how to create observations: essentially, the goal is to create a completely empty observation if there are no observations to guide the user through the various fields required.

The mechanism behind this is to shove a dummy observation into the observation list if it is ever empty. There was concern that this would cause problems with ITAC, but as an empty observation is missing fields, users are not allowed to submit until those fields are complete, so this should be completely transparent from ITAC's point of view.

Additionally, as per Bryan's request, I added "fixers" for cases where multiple observations are missing fields. Beforehand, if you double clicked on a message like:

"2 observations have no instrument configuration."

it would simply highlight an empty field in the observation table, whereas now it will open an editor for it.

I will add some notes to the code to clarify design choices.

Here is a screenshot as to how things now look when opening the PIT, or when deleting all observations (and having no other data):
screen shot 2018-02-23 at 4 45 57 pm


bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Proposal.scala
```diff
@@ -74,12 +76,16 @@ object Proposal {
     List.empty[Keyword],
     Investigators.empty,
     List.empty[Target],
-    List.empty[Observation],
+    List(Observation.empty),
```
sraaphorst: Now instead of an empty `List[Observation]`, we initialize a new `Proposal` with a `List[Observation]` with an empty observation.

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Proposal.scala
```diff
@@ -115,6 +121,13 @@ case class Proposal(meta:Meta,
 
   def resetObservationMeta:Proposal = copy(observations = observations.map(_.copy(meta = None)))
 
+  // REL-3290: Now that we always have an observation (an empty one if there are no others), when we access the
+  // list of observations for processing, we want to filter this dummy observation out.
+  def nonEmptyObservations: List[Observation] = observations.filterNot(_ == Observation.empty)
+
+  // REL-3290: Check to see if there are observations that are non-empty.
+  def hasNonEmptyObservations: Boolean = nonEmptyObservations.nonEmpty
+
```
sraaphorst: These are just convenience methods for reuse to made sure the empty observation isn't propagated to places it shouldn't be.

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
```diff
@@ -433,29 +432,26 @@ class ProblemRobot(s: ShellAdvisor) extends Robot {
...
-
-      def check[A](s: String, g: ObsListGrouping[A]) =
-        p.observations.filter(g.lens.get(_).isEmpty) match {
+      def check[A](msg: String, g: ObsListGrouping[A], adder: ObsListView => Unit) =
+        //p.observations.filter(o => o.nonEmpty && g.lens.get(o).isEmpty) match {
+        p.observations.filter(o => g.lens.get(o).isEmpty) match {
           case Nil => None
-          case h :: Nil => Some(new Problem(Severity.Error, s"One observation has no $s.", "Observations", fix(h.band, g)))
-          case h :: tail => Some(new Problem(Severity.Error, s"${1 + tail.length} observations have no $s.", "Observations", fix(h.band, g)))
+          case o :: Nil => Some(new Problem(Severity.Error, s"One observation has no $msg.", "Observations", s.inObsListView(o.band, adder)))
```
sraaphorst: We now have an `adder` param to fix missing details using an editor that adds them instead of simply highlighting them, as per Bryan's request.

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
```diff
@@ -433,29 +432,26 @@ class ProblemRobot(s: ShellAdvisor) extends Robot {
...
+          case o :: tail => Some(new Problem(Severity.Error, s"${1 + tail.length} observations have no $msg.", "Observations", s.inObsListView(o.band, adder)))
         }
 
       List(
-        check("instrument configuration", ObsListGrouping.Blueprint),
-        check("target", ObsListGrouping.Target),
-        check("observing conditions", ObsListGrouping.Condition)).flatten
+        check("instrument configuration", ObsListGrouping.Blueprint, _.Fixes.addBlueprint()),
+        check("target", ObsListGrouping.Target, _.Fixes.addTarget()),
+        check("observing conditions", ObsListGrouping.Condition, _.Fixes.addConditions())).flatten
```
sraaphorst: These now pass in new fix-functions for the aforementioned `adder` parameter.

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/view/obs/ObsListModel.scala
```diff
@@ -181,15 +181,18 @@ case class ObsListModel(all:List[Observation], band:Band, gs:(ObsListGrouping[_]
...
 
-  def cut(elem:ObsListElem):ObsListModel = elem match {
-    case target:ObsGroup[_] =>
-      val (_, toClear) = Seq(gs._1, gs._2, gs._3).splitAt(depth(target))
-      def clear(o:Observation) = (o.copy(progTime = None) /: toClear)((o, g) => g.clear(o))
       val os = selected(target)
-      (this /: os.zip(os.map(clear)))(_.replacePair(_))
-    case o:ObsElem          => this - o
+      os.zip(os.map(clear)).foldLeft(this)(_.replacePair(_))
+    case o:ObsElem          =>
```
sraaphorst: I loathe `:/` and `/:`, and I believe current Scala style guides recommend against using them.

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Proposal.scala
```diff
@@ -74,12 +76,16 @@ object Proposal {
...
-    List.empty[Observation],
+    List(Observation.empty),
     ProposalClass.empty,
     currentSchemaVersion)
 
   def apply(m:M.Proposal) = new Proposal(m)
 
+  // REL-3290: Create a non-empty list of observations, i.e. make sure that if there are no defined observations,
+  // then an empty one is provided to serve as a dummy observation for users to edit.
+  def nonEmptyObsList(olist: List[Observation]): List[Observation] =
```
cquiroz: Maybe we should use `NonEmptyList` here

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/view/obs/ObsListView.scala
```diff
@@ -339,42 +339,45 @@ class ObsListView(shellAdvisor:ShellAdvisor, band:Band, queueLookup: Target => U
...
 
-    // Our double-click handler for editing
-    onDoubleClick {
-
-      // Edit handlers for groups
-      case cg:ConditionGroup => edit(cg, p => ConditionEditor.open(cg.c, canEdit, panel), Observation.condition)
-      case bg:BlueprintGroup => edit(bg, BlueprintEditor.open(_, bg.b, canEdit, panel), Observation.blueprint)
-      case tg:TargetGroup    =>
+    // Method for handling the editing of the target group, since this is quite unique.
+    def editTargetGroup(tg: TargetGroup): Unit = {
```
sraaphorst: While this looks deceptively complicated, it is just a slight refactoring. I thought shoving the code for `case tg:TargetGroup` into this was messy, so I factored it out into its own method, `editTargetGroup`.

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/view/obs/ObsListView.scala
```diff
@@ -666,54 +669,78 @@ class ObsListView(shellAdvisor:ShellAdvisor, band:Band, queueLookup: Target => U
 
     val lens = olmLens
 
+    // General fix-it method.
+    def fixer(findAndProcess: ObsListElem => Boolean): Unit = {
+      model.foreach(_.elems.find(findAndProcess))
+    }
```
sraaphorst: I made `fixer` to avoid the repeated code of:
```
model.foreach { m =>
  m.elems.find {
    // some boolean function
  }
}
```

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/view/obs/ObsListView.scala
```diff
@@ -666,54 +669,78 @@ class ObsListView(shellAdvisor:ShellAdvisor, band:Band, queueLookup: Target => U
...
-      }
+    def fixEmpty[A](grouping:ObsListGrouping[A]) = fixer {
+      case g: ObsGroup[_] if (g.grouping == grouping) && g.grouping.get(g).isEmpty =>
+        viewer.selection = Some(g)
+        true
+      case _ => false
+    }
+
+    // Add missing conditions.
+    def addConditions() = fixer {
```
sraaphorst: Now we have `addConditions`, `addBlueprint`, and `addTarget` to add missing elements to an observation.

This doesn't work with obs time, as that dialog is treated differently and can only be filled in after other parameters are, so that is still just highlighted if the message that it is missing is double clicked.

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Proposal.scala
sraaphorst: I did try that at first, actually, but `NonEmptyList` seems to be missing a bunch of convenient methods (from what I could tell) that broke a lot of code.

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
```diff
@@ -433,29 +432,26 @@ class ProblemRobot(s: ShellAdvisor) extends Robot {
...
 
     private lazy val missingObsElementCheck = {
-      def fix[A](b: Band, g: ObsListGrouping[A]) {
-        s.inObsListView(b, _.Fixes.fixEmpty(g))
-      }
-
-      def check[A](s: String, g: ObsListGrouping[A]) =
-        p.observations.filter(g.lens.get(_).isEmpty) match {
+      def check[A](msg: String, g: ObsListGrouping[A], adder: ObsListView => Unit) =
+        //p.observations.filter(o => o.nonEmpty && g.lens.get(o).isEmpty) match {
```
cquiroz: Let's get rid of old code

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
cquiroz: i'm not sure what's the role of `adder`

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
cquiroz: ok, now i get it. It will open the UI to edit the obs when double clicking on problems

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
sraaphorst: Ooops. I missed that. Yes, absolutely.

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/view/obs/ObsListModel.scala
```diff
@@ -181,15 +181,18 @@ case class ObsListModel(all:List[Observation], band:Band, gs:(ObsListGrouping[_]
 
   }
 
+  def cut(elem: ObsListElem): ObsListModel = elem match {
+    case target: ObsGroup[_] =>
+      val (_, toClear) = Seq(gs._1, gs._2, gs._3).splitAt(depth(target))
 
+      def clear(o: Observation) =
+        toClear.foldLeft(o.copy(progTime = None))((o, g) => g.clear(o))
```
cquiroz: Can this be done as `_.clear(_)`? just to confuse people in the future 😄

sraaphorst: I would love to, but as we reverse the parameter order, I think I can't be evil.

sraaphorst:
The issue with Travis was that `ObservationSpec` was failing. It was creating a "dummy" observation (which was equal to `Observation.empty`) to test a setting on observations.

Now that we filter out empty observations, this is not the right way to do it, so I added a visitor instrument blueprint to the observation in `ObservationSpec` so it was not the empty observation any longer.

This makes me slightly nervous, because I'm not entirely sure how the `enabled` flag works and if we can set it on empty observations. (I would think not, but I'm going to look into it.)

PR #1433, opened 2018-02-27
REL-3362: Configure PIT for 2018B release.

sraaphorst:
These are the changes necessary to move the PIT from test build to official release.

### DIFF
--- a/app/pit/build.sbt
+++ b/app/pit/build.sbt
@@ -38,7 +38,7 @@ def common(pv: Version) = AppConfig(
     "org.osgi.framework.storage.clean"        -> "onFirstInit",
     "org.osgi.framework.startlevel.beginning" -> "100",
     "org.osgi.framework.bootdelegation"       -> "*",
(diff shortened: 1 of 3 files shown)


app/pit/build.sbt
```diff
@@ -38,7 +38,7 @@ def common(pv: Version) = AppConfig(
     "org.osgi.framework.storage.clean"        -> "onFirstInit",
     "org.osgi.framework.startlevel.beginning" -> "100",
     "org.osgi.framework.bootdelegation"       -> "*",
-    "edu.gemini.pit.test"                     -> "true",
+    "edu.gemini.pit.test"                     -> "false",
```
cquiroz: I don’t recall all the detail, but it seems this should be removed for non test, we check the presence not the value

sraaphorst: Actually, now that you mention it, I think you may be right.

We should fix that as it's kind of ridiculous.

PR #1434, opened 2018-02-28
REL-3384: Added warning to PIT build.sbt file.

sraaphorst:
The PIT `build.sbt` file has a property to indicate whether it is a test build or not.

After some investigation, I found that even if the property is changed from `true` to `false` or vice versa, there are many instances in which the compiled code retains the former value, and this has been quite problematic in the past with regards to making release builds.

The most reliable way to make a release build is to simply comment the line out as it defaults to `false`: hence the addition of the warning comment here.

It is used in `AppMode.scala`:
https://github.com/gemini-hlsw/ocs/blob/develop/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/model/AppMode.scala

### DIFF
--- a/app/pit/build.sbt
+++ b/app/pit/build.sbt
@@ -44,7 +44,7 @@ def common(pv: Version) = AppConfig(
     // line out for PIT production releases.
     // (I think it seems to work if you quit sbt, change it, restart sbt,
     // clean, compile, and then build the PIT.)
-    "edu.gemini.pit.test"                     -> "true",
+    //"edu.gemini.pit.test"                     -> "true",
     
     "edu.gemini.ags.host"                     -> "gnauxodb.gemini.edu",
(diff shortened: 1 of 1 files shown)


cquiroz:
This is odd. I thought there was only one place where the property is set. I think i'd push to get this fixed properly. As you mentioned it has been problemmatic in the past

sraaphorst:
This IS the only place where the property is set. It is then only read in in `AppMode.scala`, which uses it for the entire PIT application through `AppMode`.

The problem seems to be that when the flag's value is changed, `sbt` does not reliably (and not even predictably, from what I can tell) reread the new value. I have managed to:

1. Have the flag set to `true`, but `AppMode` pick it up as `false`;and
2. Have the flag set to `false`', but `AppMoe` picks it up as `true`.

I am really not sure what a "proper fix" would constitute at this point.

PR #1445, opened 2018-03-19
REL-3390: GHOST Asterism model updates.

sraaphorst:
This PR does three things:

1. Makes (hopefully all of) the model updates for GHOST targets / asterisms for now as per discussions with Steve.
2. Turns (I don't know if this is a bad idea due to possibly differing epochs) `ProperMotion` into a `Monoid`. Either way, I suspect we're going to have to assume same epochs at some point, as we were before in `DualTarget`, or throw an exception. This greatly simplifies adding `Option[ProperMotion]`s.
3. Refactors the code for making magnitude adjustments based on conditions out of both `edu.gemini.spModel.gemini.ghost` and the `edu.gemini.catalog` API and puts it into `SPSiteQuality`, where it probably belongs.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostAsterism.scala
+++ b/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostAsterism.scala
@@ -6,8 +6,6 @@ import edu.gemini.spModel.target.SPTarget
 import edu.gemini.spModel.target.env.Asterism
 import java.time.Instant
 
-import edu.gemini.spModel.gemini.ghost.GhostAsterism.GuideFiberState.Disabled
-
 import scalaz._
 import Scalaz._
 
@@ -132,7 +130,7 @@ object GhostAsterism {
       StandardResolution.guideFiberState(targets.ifu1, cc)
 
     def ifu2GuideFiberState(cc: CloudCover): GuideFiberState = {
-      targets.ifu2.map(t => StandardResolution.guideFiberState(t, cc)).getOrElse(Disabled)
+      targets.ifu2.map(t => StandardResolution.guideFiberState(t, cc)).getOrElse(GuideFiberState.Disabled)
     }
 
     override def copyWithClonedTargets: Asterism =
@@ -141,7 +139,7 @@ object GhostAsterism {
 
   object StandardResolution {
     def guideFiberState(e: Either[Coordinates, GhostTarget], cc: CloudCover): GuideFiberState =
-      e.rightMap(t => GhostTarget.standardResGuideFiberState(t, cc)).right.getOrElse(Disabled)
+      e.rightMap(t => GhostTarget.standardResGuideFiberState(t, cc)).right.getOrElse(GuideFiberState.Disabled)
   }
 
   /** GHOST standard resolution asterism types.
--- a/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostAsterism.scala
+++ b/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostAsterism.scala
@@ -2,15 +2,16 @@ package edu.gemini.spModel.gemini.ghost
 
 import edu.gemini.spModel.core._
 import edu.gemini.spModel.gemini.obscomp.SPSiteQuality.CloudCover
-import CloudCover.{PERCENT_70 => CC70, PERCENT_80 => CC80, PERCENT_90 => CC90, ANY => CCAny}
 import edu.gemini.spModel.target.SPTarget
 import edu.gemini.spModel.target.env.Asterism
+import java.time.Instant
 
-
-import java.time.{Duration, Instant}
+import edu.gemini.spModel.gemini.ghost.GhostAsterism.GuideFiberState.Disabled
 
 import scalaz._
 import Scalaz._
+
+// TODO: Unsure if we should return sky positions in allSpTargets as dummy targets.
 
 /** Base trait for the three GHOST asterism types: two target, beam switching,
   * and high resolution.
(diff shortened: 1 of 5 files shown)


bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostAsterism.scala
```diff
@@ -87,242 +83,132 @@ object GhostAsterism {
...
+                                   sky: Coordinates,
+                                   base: Option[Coordinates]) extends GhostAsterism {
 
     override def allSpTargets: NonEmptyList[SPTarget] =
       NonEmptyList(ghostTarget.spTarget)
 
     /** Defines the base position to be the same as the target position. */
     override def basePosition(when: Option[Instant]): Option[Coordinates] =
-      ghostTarget.coordinates(when)
+      base orElse ghostTarget.coordinates(when)
```
swalker2m: I wonder if `HighResolution` really will always have a sky position.  Certainly it was created that way in the previous version.  It isn't clear to me whether the sky position is in the second IFU or not despite, or maybe because of, the comment I wrote above the `HighResolution` class.  Let's check with Steve.  If it really is assigned to the second IFU, we should probably interpolate the coordinates as you did with the standard res target options.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostAsterism.scala
```diff
@@ -87,242 +83,132 @@ object GhostAsterism {
...
     * The target is always observed using the high resolution IFU1.  The sky
     * coordinates are observed using the sky fibers of IFU2, not SRIFU2. The
     * guide fibers will be used by default because the target must be bright,
     * but can be explicitly turned off.
     */
-  final case class HighRes(
-                     mode:        HighResMode,
-                     ghostTarget: GhostTarget,
-                     sky:         Coordinates) extends GhostAsterism {
+  sealed case class HighResolution(ghostTarget: GhostTarget,
```
swalker2m: Did you mean to change this to `sealed`?  I think it should be left as `final` because there should be no subclasses.

swalker2m:
Oh, I notice these changes are going into `develop`.  That's probably fine and probably the right thing to do actually because we don't use `GhostAsterism` yet in the rest of the codebase.  It might as well go into develop and we'll have less to merge later.  Maybe some of the other low-level updates can go into `develop` as well.  Once we start touching the UI though, I suppose we should use a separate 👻 branch.

sraaphorst:
Ooops. I created a `feature/GHOST` branch and thought I had made the PR on that. Mistake on my part. Yes, I think we can merge this into `develop` since it doesn't yet make any substantial changes (only the `Conditions` adjustment refactor) that will affect anything in `develop`.

Strange that I got an error from Travis now when I pushed a simple method rename. Doesn't look like a code issue, so I'll make the changes you request, delete the `feature/GHOST` branch, create a new one when we merge this, and start there on the rest of the changes.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostAsterism.scala
sraaphorst: Ah, I see what I did. I just deleted the code up to `HighResMode` and renamed that. Yes, it should be `final`. I'll make those changes.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostAsterism.scala
```diff
@@ -87,242 +83,132 @@ object GhostAsterism {
...
-      * can be explicitly set to a particular duration.
-      *
-      * (TODO: Do we need to be able to turn off the lamp altogether?  Perhaps
-      * that is the same as setting the duration to 0.)
-      */
-    final case class Prv(agitator: FiberAgitatorState, calDuration: Option[Duration]) extends HighResMode
+    case class SingleTarget(target: GhostTarget) extends GhostStandardResTargets
+    case class DualTarget(target1: GhostTarget, target2: GhostTarget) extends GhostStandardResTargets
+    case class TargetPlusSky(target: GhostTarget, sky: Coordinates) extends GhostStandardResTargets
+    case class SkyPlusTarget(sky: Coordinates, target: GhostTarget) extends GhostStandardResTargets
```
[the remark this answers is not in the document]
sraaphorst: 👍

sraaphorst:
I heard from Steve this morning, and the sky position for high resolution mode is optional. I'll change that as well.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostAsterism.scala
sraaphorst: Sure thing. I asked him about the sky position earlier, but I just asked about the interpolation.

sraaphorst: From Steve:

"For HR or SR single target mode, the sky fibers should not used to pick the base position.  The default base position will be the same as the object IFU and the sky fibers will be at TBD default position."

I'll make those changes.

swalker2m: I'm still confused about this. There are sky fibers at fixed positions, if I understand it correctly, and there are observations of sky positions at given coordinates? If you have one IFU at a given set of coordinates and the other at a different set of coordinates, why wouldn't the base position be in the middle.  If the sky is at a fixed position why do you supply coordinates?

swalker2m:
I went ahead and approved because I don't see any reason to hold up the changes made so far.  I just need a bit of help understanding sky positions and sky fibers.  Let's walk through all the options with Steve in the next meetup.

PR #1446, opened 2018-03-20
REL-3365: engineering program ids in QPT

swalker2m:
This PR implements a request to accept engineering program ids of the form, for example, `GS-2018A-ENG-1` in the QPT. To accomplish this I removed the old `StructuredProgramID` in favor of `edu.gemini.spModel.core.ProgramId` which does a better job parsing the parts of the id.  Unfortunately since the QPT is Java, this made it awkward to use.

There aren't as many substantive changes as there appear to be.  I foolishly decided to reformat some of the QPT classes to get rid of the tab characters.  We should probably do a PR to just de-tabbify QPT.  At any rate, you can look at the commits one by one and just skip the second one altogether.


bundle/edu.gemini.spModel.core/src/main/java/edu/gemini/spModel/core/ProgramTypeEnum.java
```diff
@@ -0,0 +1,18 @@
...
+    CAL,
+    DD,
+    DS,
+    ENG,
+    FT,
+    LP,
+    Q,
+    SV
+    ;
+}
```
swalker2m: This may seem like a maintenance issue, since it has to parallel the Scala `ProgramType`.  `ProgramType` has a `typeEnum` method to return the corresponding enum element.  If we add a new `ProgramType` we also have to implement `typeEnum` for it in order for it to compile.

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/ProgramType.scala
```diff
@@ -7,50 +7,96 @@ sealed trait ProgramType extends Ordered[ProgramType] {
...
+
   case object Calibration extends ProgramType {
     val name = "Calibration"
     val abbreviation = "CAL"
     override val isScience = false
+    val typeEnum = ProgramTypeEnum.CAL
   }
+
+  /** Access from Java code. */
+  def CAL: ProgramType = Calibration
```
swalker2m: I could not figure out another way to reference `Calibration` from Java.  If somebody knows what the trick is, we can remove these.

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/ProgramId.scala
```diff
@@ -21,10 +21,10 @@ sealed trait ProgramId {
   def shortName: String =
     this match {
       case ProgramId.Science(site, semester, ptype, index) =>
-        s"${site.abbreviation.drop(1)}-${semester.toShortString}-${ptype.abbreviation}-$index"
+        s"${site.abbreviation.drop(1)}${semester.toShortString}-${ptype.abbreviation}-$index"
 
       case ProgramId.Daily(site, ptype, year, month, day) =>
-        f"${site.abbreviation.drop(1)}-${ptype.abbreviation}${year.toString.takeRight(2)}$month%02d$day%02d"
+        f"${site.abbreviation.drop(1)}${year.toString.takeRight(2)}$month%02d$day%02d-${ptype.abbreviation}"
```
swalker2m: It wasn't clear what to do about the short name for `Daily` program ids.  Maybe it isn't an issue. We could just shorten say `GS-ENG20180320` to `SENG180320`, but I went with rearranging it a bit to correspond to the science ids: `S180320-ENG`.  Maybe this never comes up in the QPT.

bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/table/renderer/ObservationsRenderer.scala
```diff
@@ -86,9 +87,13 @@ object EncodedObservationsRenderer extends CellRenderer {
     /** Encodes the observation id (i.e. program number and obs id) */
     private def encObsId(o: Obs): String = {
       val pid = o.getProg.getStructuredProgramId
-      Option(pid.getSemester).map(_.takeRight(2)).getOrElse("") +
-        pid.getType +
-        pid.getNumber + "-" +
+      pid.semester.map(_.toString.takeRight(2)).getOrElse("") +
```
swalker2m: It seemed odd to me to just take the last two characters of `Semester` but I suppose that's how it is done in the QV?  So "2018A" becomes "8A"?

sraaphorst: Ugh... that's awful. The last three characters makes sense, but the last two sounds like a terrible idea.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/ui/ShellAdvisor.java
```diff
@@ -321,79 +320,79 @@ private void hookMacQuitMenu() {
...
-				mgr.addSeparator(rel, path);
-			}
-		}
-	}
+                mgr.addSeparator(rel, path);
+            }
+        }
+    }
 
 }
```
sraaphorst: Well, that was a horrible diff.

bundle/edu.gemini.spModel.core/src/main/java/edu/gemini/spModel/core/ProgramTypeEnum.java
sraaphorst: At least that will force us to remember the maintenance issue if we ever need to add more `ProgramType`s.

PR #1452, opened 2018-03-27
REL-3406: Fix AlmostEqual for Angles.

sraaphorst:
In the previous implementation:
`!(Angle.fromDegrees(359.999999) ~= Angle.zero)`.

This fixes that.


tpolecat:
Can you add a test for this?

sraaphorst:
Will do. I'll tests in the morning.

sraaphorst:
Tests added. The issue was due to rounding in floating point arithmetic.

PR #1453, opened 2018-03-28
REL-3393 (A). Refactor node initialization, observation creation

swalker2m:
This PR contains a number of mechanical changes to science program node initialization.  The overarching goal was to provide instrument-specific initialization of observations in `ISPFactory`.  To get there, I

* removed the old `spdb-initializer.conf` configuration file and the reflection code that read it and registered initializers with the factory.  Instead, references to all node initializers are collected in `NodeInitializers` and used directly by the factory.

* simplified initializer creation, boiling it down to one line for most normal initializers.  That's the reason there is so much deleted code.

* added an `Instrument` enumeration and an `Option<Instrument>` parameter to the `ISPFactory`.  It uses this to get the instrument-specific initializer for the observation.

This will ultimately allow Ghost observations to be created with an initial Ghost asterism, once the Ghost-specific initializer is written.


swalker2m:
This PR is obviously too long and has the potential to break things.  I couldn't think of a good alternative to instrument-specific observation initialization but I'm happy to consider other ideas if anybody has any.

tpolecat:
> removed the old spdb-initializer.conf configuration file and the reflection code that read it and registered initializers with the factory.

praise jeebus

cquiroz:
Can I just LGTM it?

bundle/edu.gemini.p2checker/src/test/scala/edu/gemini/p2checker/rules/NonSiderealTargetRulesSpec.scala
```diff
@@ -132,7 +132,6 @@ class NonSiderealTargetRulesSpec extends RuleSpec {
       }
       e.setDataObject(toc)
     }
-    f.createObsComponent(p, inst, null) <| { i => o.addObsComponent(i) }
```
tpolecat: The equivalent line in `RuleSpec.scala:30` (the next file down) remains … is this because we have an instrument here but not in `RuleSpec`?

bundle/edu.gemini.p2checker/src/test/scala/edu/gemini/p2checker/rules/NonSiderealTargetRulesSpec.scala
```diff
@@ -117,7 +117,7 @@ class NonSiderealTargetRulesSpec extends RuleSpec {
   def obs(target: Target, sb: Option[SchedulingBlock], inst: SPComponentType = InstGmosSouth.SP_TYPE): ISPObservation = {
     val f = POTUtil.createFactory(UUID.randomUUID())
     val p = f.createProgram(null, SPProgramID.toProgramID("no-site"))
-    val o = f.createObservation(p, null) <| { on =>
+    val o = f.createObservation(p, Instrument.fromComponentType(inst), null) <| { on =>
       p.addObservation(on)
       on.getDataObject.asInstanceOf[SPObservation] <| { o =>
```
sraaphorst: Just as a random aside of curiosity, I always see that `IdOpts` `<|` will be becoming obsolete soon. Should this be something we should concern ourselves with?

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/init/NodeInitializers.java
```diff
@@ -0,0 +1,209 @@
...
+    public final ISPNodeInitializer<ISPTemplateFolder, TemplateFolder> templateFolder =
+            TemplateFolder.NI;
+
+    public final ISPNodeInitializer<ISPTemplateGroup, TemplateGroup> templateGroup =
+            TemplateGroup.NI;
+
+    public final ISPNodeInitializer<ISPTemplateParameters, TemplateParameters> templateParameters =
+            TemplateParameters.NI;
+
+    public final Map<SPComponentType, ISPNodeInitializer<ISPObsComponent, ? extends ISPDataObject>> obsComp =
```
tpolecat: Do we have a unit test asserting that these maps are sufficiently full? It would be an easy place to make an error in the future.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/obslog/ObsQaLog.java
```diff
@@ -18,6 +21,9 @@
 
     public static final SPComponentType SP_TYPE = SPComponentType.OBS_QA_LOG;
 
+    public static final ISPNodeInitializer<ISPObsQaLog, ObsQaLog> NI =
```
tpolecat: Kind of makes you miss typeclasses doesn't it.

bundle/edu.gemini.p2checker/src/test/scala/edu/gemini/p2checker/rules/NonSiderealTargetRulesSpec.scala
tpolecat: With some luck we won't ever have to upgrade scalaz. But if we do we can add this syntax back ourselves.

tpolecat:
This is a really good simplification. I wonder if it would make sense to do a test release from your branch and have science go through the test matrix … I worry that it will be hard to revert if we end up finding problems at runtime.

bundle/edu.gemini.pot/src/main/java/edu/gemini/pot/sp/Instrument.java
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * An enumeration of all instruments, which is a subset of all component types.
+ */
+public enum Instrument {
+    AcquisitionCamera(INSTRUMENT_ACQCAM),
+    Bhros(INSTRUMENT_BHROS),
+    Flamingos2(INSTRUMENT_FLAMINGOS2),
+    GmosNorth(INSTRUMENT_GMOS),
+    GmosSouth(INSTRUMENT_GMOSSOUTH),
+    Gnirs(INSTRUMENT_GNIRS),
```
sraaphorst: I guess wewill have to add `Ghost` here. I have been trying to mark such areas of the code as ``TODO:GHOST` so that we can grep through and not forget.

swalker2m:
> This is a really good simplification. I wonder if it would make sense to do a test release from your branch and have science go through the test matrix … I worry that it will be hard to revert if we end up finding problems at runtime.

That's a good idea if they can spare a bit of testing effort now.  The release would also include the new GS-2018B-ENG-1 format they requested so that would be examined as well.  I'll check with Bryan and @andrewwstephens.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/bhros/SeqConfigBHROS.java
```diff
@@ -18,6 +21,9 @@
 
     public static final SPComponentType SP_TYPE = SPComponentType.ITERATOR_BHROS;
 
+    public static final ISPNodeInitializer<ISPSeqComponent, SeqConfigBHROS> NI =
```
sraaphorst: All these reductions down to `NI` and elimination of subclasses is very nice indeed.

sraaphorst:
Wow... this is indeed impressive. Thanks for making these changes.
As Rob says, this probably merits a test release. I can certainly arrange that, regardless of whether we merge this into `develop` (which seems what you are looking to do) or `feature/GHOST`.

bundle/edu.gemini.p2checker/src/test/scala/edu/gemini/p2checker/rules/RuleSpec.scala
```diff
@@ -26,7 +26,7 @@ abstract class RuleSpec extends Specification {
   def setup[I <: ISPDataObject](instrument: SPComponentType, programId: String = "")(mod: I => Unit): ISPObservation = {
     val f = POTUtil.createFactory(UUID.randomUUID())
     val p = f.createProgram(null, SPProgramID.toProgramID(programId))
-    val o = f.createObservation(p, null) <| p.addObservation
+    val o = f.createObservation(p, Instrument.none, null) <| p.addObservation
     val i = f.createObsComponent(p, instrument, null) <| o.addObsComponent
```
swalker2m: > The equivalent line in RuleSpec.scala:30 (the next file down) remains … is this because we have an instrument here but not in RuleSpec?

Here I left the explicit observation component creation for the instrument because the test case needs a references to it in lines 36-38.  Otherwise I could have done

```scala
val o = f.createObservation(p, Instrument.fromComponentType(instrument), null) <| p.addObservation
```

as I did in `NonSiderealTargetRulesSpec` above.  It could be done this way anyway, but then I'd need to fish out the instrument observation component from the observation.

swalker2m: Of course, the drawback of separating the instrument node creation from the observation is that any observation-specific initialization (like setting up the Ghost asterism) won't be done.

bundle/edu.gemini.pot/src/main/java/edu/gemini/pot/sp/Instrument.java
swalker2m: I'll add that here and add a completeness test case as well.

bundle/edu.gemini.pot/src/test/scala/edu/gemini/pot/sp/TestInstrumentCompleteness.scala
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * Tests that the Instrument enumeration agrees with SPComponentType
+ */
+final class TestInstrumentCompleteness {
+
+  @Test
+  def testInstrumentEnumCompleteness(): Unit = {
+    val all = SPComponentType.values.toSet.filter(t => t.broadType == INSTRUMENT) -
+                SPComponentType.QPT_CANOPUS -
+                SPComponentType.QPT_PWFS
```
swalker2m: These are kind of annoying and seem to have been introduced by mistake.  It would be good to get rid of them.  They aren't actually instruments or program nodes of any kind.

bundle/edu.gemini.p2checker/src/test/scala/edu/gemini/p2checker/rules/NonSiderealTargetRulesSpec.scala
cquiroz: I doubt we'll ever upgrade scalaz in ocs

cquiroz:
I'm declaring review bankruptcy at around 1/3rd of the PR, but so far it seemed ok. I'm very confident it does everything correctly 👍

sraaphorst:
@cquiroz We're in the midst of making a test release off this branch.

swalker2m:
I want ahead and merged this since it will be holding us back soon.  I don't think the test release will get much attention from science until May, but having it merged now will at least obligate us to use it during development.

PR #1454, opened 2018-03-28
REL-3360: Added Alopeke code to template generator.

sraaphorst:
This adds the basics of the `Alopeke` code to the template generator so that P1 proposals from ITAC can be imported into the ODB.

### DIFF
--- a/bundle/edu.gemini.model.p1/src/test/scala/edu/gemini/model/p1/dtree/RootSpec.scala
+++ b/bundle/edu.gemini.model.p1/src/test/scala/edu/gemini/model/p1/dtree/RootSpec.scala
@@ -43,6 +43,10 @@ class RootSpec extends Specification {
       val root = new Root(Semester(2016, A))
       root.choices must not contain Instrument.Trecs
     }
+    "include Alopeke" in {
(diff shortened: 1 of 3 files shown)


swalker2m:
Were you able to import an Alopeke proposal into the ODB?

sraaphorst:
@swalker2m Yes. I should have posted the process. Here is an Alopeke P1 proposal:
screen shot 2018-03-29 at 10 20 02 am

I followed this up by:

`skeleton alopeke.xml alopeke.pdf GN-2108B-Q-1`

or something to that extent, which gave:

screen shot 2018-03-29 at 10 19 05 am

And finally:

screen shot 2018-03-29 at 10 19 49 am

swalker2m:
Looks good, but should the name field in the OT be a lot wider?  It looks like it is cutting off a lot of information in order to align with the numeric fields below it.

sraaphorst:
Yes, let's make it wider. I'll fiddle with that now.

andrewwstephens:
It's not that important, so let's go with that for now.

sraaphorst:
I went and fixed the field lengths since this was easy:
screen shot 2018-04-02 at 10 47 08 am

PR #1455, opened 2018-03-29
REL-3411: synchronize access to formatter

swalker2m:
This PR synchronizes access to a `DateFormat` object to prevent a potential test case failure.  I had originally intended to switch to `java.time.format.DateTimeFormatter` so that synchronization wouldn't be necessary but the existing `SPTargetPio.parseDate` method seemed complicated and would have to be modified.

### DIFF
--- a/bundle/edu.gemini.spModel.io/src/test/scala/edu/gemini/spModel/io/impl/migration/to2016A/PlutoDemotionTest.scala
+++ b/bundle/edu.gemini.spModel.io/src/test/scala/edu/gemini/spModel/io/impl/migration/to2016A/PlutoDemotionTest.scala
@@ -7,6 +7,8 @@ import edu.gemini.spModel.target.{SPTargetPio, SPTarget}
 import edu.gemini.spModel.target.obsComp.TargetObsComp
 import org.junit.{Assert, Test}
 
+import java.util.Date
+
 import scala.collection.JavaConverters._
 
 class PlutoDemotionTest extends MigrationTest {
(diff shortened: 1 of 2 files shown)


bundle/edu.gemini.spModel.io/src/test/scala/edu/gemini/spModel/io/impl/migration/to2016A/PlutoDemotionTest.scala
```diff
@@ -28,7 +30,7 @@ class PlutoDemotionTest extends MigrationTest {
             e.toList match {
               case List((t, c)) =>
 
-                  Assert.assertEquals("ValidAt", "10/28/15 7:39:58 PM UTC", SPTargetPio.formatter.format(t))
+                  Assert.assertEquals("ValidAt", "10/28/15 7:39:58 PM UTC", SPTargetPio.formatDate(new Date(t)))
```
cquiroz: If this is the only place we are using it perhaps we should just move it to test, or is it used internally in `SPTargetPio`? If so, then all calls should go through `formatDate`, shouldn't they?

swalker2m: Good question.  It's used internally in `SPTargetPio`.  It was `public` before and that's how it was visible to the test case.  Inside of `SPTargetPio` it is only used for parsing dates and that usage already was synchronized via the `parseDate` method.  The only usage for formatting was in the test case and now that has been synchronized as well so that it cannot be used simultaneously with parsing.

cquiroz: Thanks for the clarification

PR #1447, opened 2018-03-20
REL-3391: GHOST instrument

sraaphorst:
This code adds the new instrument `Ghost` to the model. This adds the following:

1. a new class, `Ghost`, that contains the configuration for the instrument (right now this is just a position angle);
2. the necessary code to other classes to handle the new class `Ghost`;
3. the AGS code for `Ghost`; and
4. an `AsterismType` for all instruments to determine what asterism type a particular instrument can support, along with its default type.

We may have to move the `AsterismType` to an external class of `Map[SPComponentType, ImList[AsterismType]]`, probably best put in `AsterismType` itself, as I think we will need to be able to access this code in a static way independent of an existing instance of `SPInstObsComp`.

PR #1457, opened 2018-04-02
REL-3391: GHOST instrument

sraaphorst:
Moving this code to the `develop` branch as it doesn't change anything for June and we want to minimize the number of commits required to the `feature/GHOST` branch so that we don't end up with a million merge conflicts.

See discussion here:
https://github.com/gemini-hlsw/ocs/pull/1447


bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/AsterismType.java
```diff
@@ -0,0 +1,57 @@
...
+        this.tag = tag;
+    }
+
+    public abstract Asterism createEmptyAsterism();
+
+    // Return the asterism types supported by the different instruments.
+    // We need to do this here because we want these statically accessible.
+    // Assume the head of the list is the default for the instrument.
+    public static ImList<AsterismType> supportedTypesForInstrument(final SPComponentType instType) {
+        if (!instType.broadType.equals(SPComponentBroadType.INSTRUMENT))
```
swalker2m: Once my big initializer PR is merged, let's change this method to take an `Instrument` so there's no need to check or throw a `RuntimeException`.

sraaphorst: Sounds good. I'll make a note of this.

PR #1458, opened 2018-04-02
REL-3390: GHOST asterism arbitraries, lenses, PIO, and test cases

sraaphorst:
This should be the final PR that we can push to `develop` without affecting releases.


bundle/edu.gemini.pot/src/test/scala/edu/gemini/spModel/target/env/Arbitraries.scala
```diff
@@ -159,7 +160,153 @@ trait Arbitraries extends edu.gemini.spModel.core.Arbitraries {
...
+
+    // This generates a list of n coordinates, wherein the last n-1 are within distance [lower, upper) from the first.
+    def genNCoordsWithinDistance(n: Int, lower: Double, upper: Double): Gen[List[Coordinates]] =
+      for {
+        c  <- arbitrary[Coordinates]
+        cs <- listOfN(n-1, genCoordsWithinDistance(c, lower, upper))
+      } yield c :: cs
+
+    // TODO:GHOST Change these to GHOST constants as we learn more, and move to Ghost, probably.
+    val inRangeDistance: Double =   5.0 // in arcsecs
```
swalker2m: They should also be expressed in terms of `Angle` instead of `Double`, I think.

bundle/edu.gemini.spModel.core/src/test/scala/edu/gemini/spModel/core/CoordinatesSpec.scala
```diff
@@ -55,6 +55,16 @@ object CoordinatesSpec extends Specification with ScalaCheck with Arbitraries wi
 
   }
 
+  "Coordinates Angular Offset" should {
+
+    "result in an angular separation equal to distance, regardless of bearing" !
+      forAll { (c: Coordinates, bearing: Angle, distance: Angle) =>
+        val newCoordinates = c.angularOffset(bearing, distance)
+        (c.angularDistance(newCoordinates) ~= distance) || (newCoordinates.angularDistance(c) + distance ~= Angle.zero)
```
swalker2m: I would think `angularDistance` would always be positive.  Is that not the case?

bundle/edu.gemini.spModel.core/src/test/scala/edu/gemini/spModel/core/Arbitraries.scala
```diff
@@ -50,6 +50,16 @@ trait Arbitraries {
...
     }
 
+  // A way of picking coordinates within an offset distance in arcseconds from a given coordinate.
+  def genCoordsWithinDistance(c: Coordinates, lower: Double, upper: Double): Gen[Coordinates] =
+    for {
+      dRaSgn <- oneOf(-1.0, 1.0)
+      dDecSgn <- oneOf(-1.0, 1.0)
+      dRa <- choose(lower, upper).map(a => Angle.fromArcsecs(dRaSgn * a))
+      dDec <- choose(lower, upper).map(a => Angle.fromArcsecs(dDecSgn * a))
+    } yield c.offset(dRa, dDec)
```
swalker2m: Shouldn't this use the new `angularOffset(bearing, distance)` method?

bundle/edu.gemini.spModel.core/src/test/scala/edu/gemini/spModel/core/CoordinatesSpec.scala
sraaphorst: Hmmm... I thought I did that, but maybe it got lost in the git confusion. I'll look at these issues.

PR #1460, opened 2018-04-03
REL-3279: Use full exposure time for GMOS in ITC

swalker2m:
This PR updates the OT ITC to use full nod and shuffle time for GMOS observations.  Normally, the ITC reports single exposure time from the sequence, but the total time on source for nod and shuffle should be multiplied by the cycle count and the number of positions (usually 2).

Unfortunately after a couple of days of working on this, I couldn't come up with a more general solution. The ITC works with config steps so I initially added a `fullExposureTime` property to the sequence for GMOS and had ITC prefer `fullExposureTime` to `exposureTime` when present.  Unfortunately that adds another "exposure time" to the sequence display tables which gets confusing, particularly in the ITC.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/gmos/FullExposureTime.scala
+++ b/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/gmos/FullExposureTime.scala
@@ -1,0 +1,76 @@
+package edu.gemini.spModel.gemini.gmos
+
+import edu.gemini.spModel.config2.{ Config, ItemKey }
+import edu.gemini.spModel.gemini.calunit.calibration.CalDictionary.OBS_TYPE_ITEM
+import edu.gemini.spModel.gemini.gmos.InstGmosCommon.{ NS_STEP_COUNT_PROP_NAME, NUM_NS_CYCLES_PROP, USE_NS_PROP }
+import edu.gemini.spModel.obscomp.InstConstants.{ EXPOSURE_TIME_PROP, OBSERVE_TYPE_PROP }
+
+import edu.gemini.spModel.seqcomp.SeqConfigNames.{ INSTRUMENT_CONFIG_NAME, OBSERVE_CONFIG_NAME }
+
+import java.time.Duration
+
+import scalaz._
+import Scalaz._
+
+/** Adds support for calculating and inserting "full" exposure time for GMOS,
+  * even for nod and shuffle steps.  Nod and shuffle takes a number of exposures
+  * per step (cycle count * offset position step count) so the full exposure
+  * time is the nominal instrument exposure time times the nod and shuffle count.
+  */
+object FullExposureTime {
+
+  private object keys {
+    val StepCount:        ItemKey = new ItemKey(s"$INSTRUMENT_CONFIG_NAME:$NS_STEP_COUNT_PROP_NAME")
+    val CycleCount:       ItemKey = new ItemKey(s"$INSTRUMENT_CONFIG_NAME:${NUM_NS_CYCLES_PROP.getName}")
+    val UseNs:            ItemKey = new ItemKey(s"$INSTRUMENT_CONFIG_NAME:${USE_NS_PROP.getName}")
+    val ExposureTime:     ItemKey = new ItemKey(s"$INSTRUMENT_CONFIG_NAME:$EXPOSURE_TIME_PROP")
+    val ObsType:          ItemKey = OBS_TYPE_ITEM.key
+  }
+
+  private implicit class ConfigOps(c: Config) {
+
+    import keys._
+
+    def value(k: ItemKey): Option[Object] =
+      Option(c.getItemValue(k))
+
+    def useNs: Boolean =
+     value(UseNs).map {
+       case b: java.lang.Boolean => b.booleanValue
+       case _                    => false
+     }.getOrElse(false)
+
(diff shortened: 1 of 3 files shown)


bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/gmos/FullExposureTime.scala
```diff
@@ -0,0 +1,76 @@
...
+  * per step (cycle count * offset position step count) so the full exposure
+  * time is the nominal instrument exposure time times the nod and shuffle count.
+  */
+object FullExposureTime {
+
+  private object keys {
+    val StepCount:        ItemKey = new ItemKey(s"$INSTRUMENT_CONFIG_NAME:$NS_STEP_COUNT_PROP_NAME")
+    val CycleCount:       ItemKey = new ItemKey(s"$INSTRUMENT_CONFIG_NAME:${NUM_NS_CYCLES_PROP.getName}")
+    val UseNs:            ItemKey = new ItemKey(s"$INSTRUMENT_CONFIG_NAME:${USE_NS_PROP.getName}")
+    val ExposureTime:     ItemKey = new ItemKey(s"$INSTRUMENT_CONFIG_NAME:$EXPOSURE_TIME_PROP")
```
tpolecat: How do you know whether to use the `$FOO_PROP` vs. `${FOO_PROP.getName}` here?

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/gmos/FullExposureTime.scala
```diff
@@ -0,0 +1,76 @@
...
+
+    import keys._
+
+    def value(k: ItemKey): Option[Object] =
+      Option(c.getItemValue(k))
+
+    def useNs: Boolean =
+     value(UseNs).map {
+       case b: java.lang.Boolean => b.booleanValue
+       case _                    => false
```
tpolecat: Is a bogus value here not an error condition (and similarly with other props below)?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcUniqueConfig.scala
```diff
@@ -37,9 +37,13 @@ case class ItcUniqueConfig(count: Int, labels: String, configs: NonEmptyList[Con
...
-  /** Instruments other than TReCS and Michelle have a defined exposure time per image. */
-  private val singleTime = Option(config.getItemValue(INST_EXP_TIME_KEY)).map(_.asInstanceOf[Double])
-
+  /** Instruments other than TReCS and Michelle have a defined exposure time per
+    * image.  In the case of GMOS, this may include nod and shuffle steps so
+    * full exposure time is preferred.
+    */
+  private val singleTime: Option[Double] =
+    if (ItcUniqueConfig.isGmos(config)) Some(GmosExposureTime.asDoubleSecs(config))
+    else Option(config.getItemValue(INST_EXP_TIME_KEY)).map(_.asInstanceOf[Double])
```
tpolecat: Above you're handling the case where it's something other than a `Double` but you're assuming it here.

tpolecat:
This could probably use a test.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/gmos/FullExposureTime.scala
swalker2m: Depends on the type of the thing being referred to.  Some are beans `PropertyDescriptor` and some are just `String`.  I guess I could go through the old constant names and try to make them consistent.  It seems like a task for another PR if it's worth doing at this point though.

tpolecat: Gotcha.

tpolecat:
This part of OCS gives me the heebiejeebies. It's like programming in Javascript.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/gmos/FullExposureTime.scala
swalker2m: That's a good point, I'll just throw an exception if there is an unexpected type in the configuration.  Likely it won't get this far anyway but it makes more sense than just ignoring the problem.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/seq/ItcUniqueConfig.scala
swalker2m: The `else` clause is just "whatever it was doing before".  `totalTime` and `singleTime` were just cast before and I was trying to disturb the ITC stuff as little as possible.

tpolecat: Alright.

swalker2m:
I think I need about another day to write test cases.  As you might guess they're difficult to put together.  I also want to see about generalizing and pulling some of the `Config` ops into a reusable enrichment ops class.  I'll come back to this at the end of the cycle if there is time.

PR #1461, opened 2018-04-04
REL-3394: GHOST configuration UI

sraaphorst:
This is the basic configuration UI for GHOST. Right now it simply handles position angle and displays the components for standard resolution vs. high resolution: the logic for setting the asterism will come in a subsequent PR.

I have commented out the entry for GHOST from `UIInfo.xml` so that it is not available to be added: this way, we can merge this code into `develop` without any issues.

Here is a picture of the UI with standard resolution mode.
screen shot 2018-04-04 at 1 31 17 pm


bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostMixin.java
```diff
@@ -0,0 +1,10 @@
+package edu.gemini.spModel.gemini.ghost;
+
+import edu.gemini.pot.sp.SPComponentType;
+
+// We need to do this because of reflection. This is a static property of instruments, but since Ghost is
```
cquiroz: Does it make any sense attempting to remove reflection?

cquiroz:
I don't see code touching the TPE, Was that done earlier?

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostMixin.java
```diff
@@ -0,0 +1,10 @@
...
+// coded in Scala, values in the companion object are called as methods, which is incompatible with reflection
+// on fields.
+public interface GhostMixin {
+    SPComponentType SP_TYPE = SPComponentType.INSTRUMENT_GHOST;
```
sraaphorst: This can be read as a static field in `Ghost` from Java classes, but not from Scala, which looks at the companion object when you access `Ghost`. Even changing it to:

`object Ghost extends GhostMixin`

doesn't allow us to access `SP_TYPE` through `Ghost` in Scala, so we need to access `GhostMixin.SP_TYPE` directly, as seen below in `GhostNI.scala`.

cquiroz: And the other way around? Maybe put it on `object Ghost`? and call it from Java? or is that the reflection bit?

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostMixin.java
sraaphorst: I would love to see reflection go... I'm not sure how difficult that would be to eliminate in this case since `UIInfo.xml` contains a good amount of data. @swalker2m? What do you think?

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostMixin.java
cquiroz: ok, nevermind, it isn't worth the effort. could you make sure this would compile on scala 2.12? We'll have to use these classes from the seqexec

sraaphorst:
As for the TPE, there isn't a lot we'll have to do for that: most of what we need to do is already handled. We'll probably make very minor changes since GHOST can have "sky targets" instead of actual target objects, so we might want to draw these differently, and the base position is not necessarily the science target position, so we will probably do something for that, too.

I'm not sure any of this is needed for the July testing, though, so we don't yet have a task for it.

swalker2m:
I wonder if it would be simpler to just add "High Resolution" to the menu and get rid of the radio buttons.  Something like a menu of

* Single Target (Standard Resolution)
* Dual Target (Standard Resolution)
* Target + Sky (Standard Resolution)
* Sky + Target (Standard Resolution)
* High Resolution

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostMixin.java
swalker2m: Agree that reflection is bad but I don't think we want to take that on at this point.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostMixin.java
sraaphorst: @cquiroz or @swalker2m, is there any way you know of to try compiling these files with Scala 2.12? I know OCS2 doesn't compile with it, but I don't know any easy way to try compiling only certain files, especially given how many dependencies `Ghost.scala` has.

PR #1463, opened 2018-04-06
REL-3394: GHOST asterism conversions

sraaphorst:
This PR simply adds the ability to convert between different types of GHOST asterisms.

This will happen when the GHOST instrument configuration is changed to use a different configuration, in which case, we want to maintain as much information as possible from the previous configuration.

The test cases are incomplete, but this is getting way out of control and beyond the scope of what we want for July, so I am putting a pin in it here and will come back to fill out the test cases in more detail if / when time permits.

This required moving some code around and adding a few things to:
1. Be able to create `TargetEnvironments` with different types of `Asterism`s (so that Scalacheck could get enough valid test cases to pass).
2. Be able to do an `AlmostEqual` operation on `TargetEnvironment`s.
3. Add the code to convert between asterisms. The goal is to not drop any targets / coordinates, and add them as user targets to preserve them.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/AsterismConverters.scala
+++ b/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/AsterismConverters.scala
@@ -20,19 +20,22 @@ object AsterismConverters {
 
   sealed trait AsterismConverter {
     def name: String
-    def convert(env: TargetEnvironment): String \/ TargetEnvironment
+    def convert(env: TargetEnvironment): Option[TargetEnvironment]
   }
 
   sealed trait GhostAsterismConverter extends AsterismConverter {
-    override def convert(env: TargetEnvironment): String \/ TargetEnvironment = env.getAsterism match {
-        case StandardResolution(SingleTarget(t), b)     => creator(env, t,  None,    None,   b).right
-        case StandardResolution(DualTarget(t1, t2), b)  => creator(env, t1, t2.some, None,   b).right
-        case StandardResolution(TargetPlusSky(t, s), b) => creator(env, t,  None,    s.some, b).right
-        case StandardResolution(SkyPlusTarget(s, t), b) => creator(env, t,  None,    s.some, b).right
-        case HighResolution(t, s, b)                    => creator(env, t,  None,    s,      b).right
-        case _                                          => s"Could not convert to $name".left
+    override def convert(env: TargetEnvironment): Option[TargetEnvironment] = env.getAsterism match {
+        case StandardResolution(SingleTarget(t), b)     => creator(env, t,  None,    None,   b).some
+        case StandardResolution(DualTarget(t1, t2), b)  => creator(env, t1, t2.some, None,   b).some
+        case StandardResolution(TargetPlusSky(t, s), b) => creator(env, t,  None,    s.some, b).some
+        case StandardResolution(SkyPlusTarget(s, t), b) => creator(env, t,  None,    s.some, b).some
+        case HighResolution(t, s, b)                    => creator(env, t,  None,    s,      b).some
+        case _                                          => None
       }
 
+    /** This is just a method that takes the targets / coordinates common to all GHOST asterism types so that we can
+      * more easily extract parameters from one asterism type and turn it into another without having to have n*n
+      * cases, were n is the number of asterism types. */
     protected def creator(env: TargetEnvironment, t1: GhostTarget, t2: Option[GhostTarget], s: SkyPosition, b: BasePosition): TargetEnvironment
   }
 
@@ -93,7 +96,7 @@ object AsterismConverters {
   private def appendTarget(userTargets: ImList[UserTarget], t: Option[UserTarget]): ImList[UserTarget] =
     t.map(userTargets.append).getOrElse(userTargets)
 
-  /** Convert an SPTarget to an empty UserTarget. */
+  /** Convert an SPTarget to a UserTarget. */
   private def t2UT(t: SPTarget): UserTarget =
     new UserTarget(UserTarget.Type.other, t)
 
@@ -103,6 +106,7 @@ object AsterismConverters {
     t2UT(new SPTarget(t))
   }
 
+  /** Convert a GhostTarget to a UserTarget. */
   private def gT2UT(tOpt: Option[GhostTarget]): Option[UserTarget] =
     tOpt.map(t => t2UT(t.spTarget))
 }
--- a/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/AsterismConverters.scala
+++ b/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/AsterismConverters.scala
@@ -24,13 +24,17 @@ object AsterismConverters {
   }
 
   sealed trait GhostAsterismConverter extends AsterismConverter {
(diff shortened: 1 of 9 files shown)


bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/AsterismConverters.scala
```diff
@@ -0,0 +1,108 @@
...
+    override def convert(env: TargetEnvironment): String \/ TargetEnvironment = env.getAsterism match {
+        case StandardResolution(SingleTarget(t), b)     => creator(env, t,  None,    None,   b).right
+        case StandardResolution(DualTarget(t1, t2), b)  => creator(env, t1, t2.some, None,   b).right
+        case StandardResolution(TargetPlusSky(t, s), b) => creator(env, t,  None,    s.some, b).right
+        case StandardResolution(SkyPlusTarget(s, t), b) => creator(env, t,  None,    s.some, b).right
+        case HighResolution(t, s, b)                    => creator(env, t,  None,    s,      b).right
+        case _                                          => s"Could not convert to $name".left
+      }
+
+    protected def creator(env: TargetEnvironment, t1: GhostTarget, t2: Option[GhostTarget], s: SkyPosition, b: BasePosition): TargetEnvironment
```
sraaphorst: This is just a method that takes the targets / coordinates common to all asterism types so that we can more easily extract parameters from one asterism type and turn it into another without having to have n^2 cases, were n is the number of asterism types.

bundle/edu.gemini.pot/src/test/scala/edu/gemini/spModel/target/env/AsterismParamSetCodecSpec.scala
```diff
@@ -20,76 +20,7 @@ import scalaz._
 import Scalaz._
 
 
-class AsterismParamSetCodecSpec extends Specification with ScalaCheck with Arbitraries {
```
sraaphorst: We need all of these outside of just testing `ParamSetCodecSpec`, so I moved them to `Almosts` since there is no reason not to make them more available.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/AsterismConverters.scala
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  sealed trait GhostAsterismConverter extends AsterismConverter {
+    override def convert(env: TargetEnvironment): String \/ TargetEnvironment = env.getAsterism match {
+        case StandardResolution(SingleTarget(t), b)     => creator(env, t,  None,    None,   b).right
+        case StandardResolution(DualTarget(t1, t2), b)  => creator(env, t1, t2.some, None,   b).right
+        case StandardResolution(TargetPlusSky(t, s), b) => creator(env, t,  None,    s.some, b).right
+        case StandardResolution(SkyPlusTarget(s, t), b) => creator(env, t,  None,    s.some, b).right
+        case HighResolution(t, s, b)                    => creator(env, t,  None,    s,      b).right
+        case _                                          => s"Could not convert to $name".left
```
tpolecat: It would be nice if we could compute this in terms of a total match on `GhostAsterism` because we should be able to get exhaustive checking. I realize it's awkward because of the way the `TargetEnvironment` is structured. What if we did something like

```scala
env.getAsterism match {
  case ga: GhostAsterism =>
    ga match {
      case StandardResolution(...) => ...
      // ... etc., with no catch-all so we get exhaustive checking for this clause
    }
  case _ => // fail
```

We might also consider just using `Option` here since there's only one failure case and it always means the same thing. In general I don't like generating end-user messages this deep in the model. Debatable.

bundle/edu.gemini.pot/src/test/scala/edu/gemini/spModel/target/env/Arbitraries.scala
```diff
@@ -309,4 +312,11 @@ trait Arbitraries extends edu.gemini.spModel.core.Arbitraries {
       arbitrary[Asterism.Single],
       arbitrary[GhostAsterism]
     ))
+
+  // Specific target environment types to avoid Scalacheck giving up.
+  // These can't go in GhostGens or we get a stack overflow error.
+  val genStandardResAsterismTargetEnvironment: Gen[TargetEnvironment] =
```
tpolecat: If you make these return the subtype `GhostAsterism.StandardResolution`, etc. then you won't need the casts in the tests above. You can widen these via `.map[TargetEnvironment](identity)` in cases where you need the wider type.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/AsterismConverters.scala
sraaphorst: 😊 
Good idea.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/AsterismConverters.scala
sraaphorst: Agreed. The error message, if we need it, could easily be constructed when the `Option` is accessed, and this will simplify things.

bundle/edu.gemini.pot/src/test/scala/edu/gemini/spModel/target/env/Arbitraries.scala
sraaphorst: This one I'm a little confused about and not sure if I'm misunderstanding. You're referring to the tests in `AsterismConverterSpec` and the casts I'm doing there, right?

I want `TargetEnvironment`, preferably populated with an `Asterism` subclass of the correct type, a `GuideEnvironment` and possibly some `UserTarget`s, and `GhostAsterism.StandardResolution` is a subclass of `Asterism`... so I'm not sure how I can go from a `Gen[GhostAsterism.StandardResolution]` or an `Arbitrary[GhostAsterism.StandardResolution]` to a full-fledged `TargetEnvironment` without doing something like I'm doing here.

Wouldn't the widening require `GhostAsterism.StandardResolution` to be a subclass of `TargetEnvironment`?

The casting is making me feel dirty, though, so if I can eliminate it, that would be excellent.

(Sorry if I'm being a bit dense here, and I definitely appreciate the help.)

tpolecat: Oh right, I'm sorry, I misread your code. Yeah I think that's probably the best you can do with the way things are structured now. Nevermind.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/AsterismConverters.scala
sraaphorst: Somehow I glazed over the earlier part of this comment in lieu of the `Option` part. Friday.
I'm fine with doing the exhaustive checking as you said: I'll make this change now and check it in momentarily.

Sorry about that!

sraaphorst: Incidentally, thanks for checking this PR. Much appreciated. It's a lot of crud that's probably overkill at this point, but will hopefully make things easier later on.

PR #1464, opened 2018-04-09
REL-3394: GHOST instrument editor completion

sraaphorst:
This PR is an attempt to complete the UI work and wiring for the GHOST instrument configuration.

In this PR:

1. I removed the radio boxes (as per an earlier suggestion by @swalker2m) and added `High Resolution` mode to the combo box instead.

2. The asterism type combo box is now wired to perform conversions between the different asterism types.

3. The instrument editor is properly initialized from the `Ghost` instrument configuration passed to it.

Small concerns:

1. I don't like the way of initializing the asterism combo box, but I haven't deduced a better way to do it.

2. Error handling from asterism issues: Right now, I display dialogs, and possibly disable the asterism combo box. I'm not sure if this is how we want to handle this.

Here is a screenshot of the editor:

screen shot 2018-04-09 at 4 40 00 pm


sraaphorst:
I squashed down the commits because all but the first were trivial changes.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/ghost/GhostEditor.scala
```diff
@@ -56,87 +59,99 @@ final class GhostEditor extends ComponentEditor[ISPObsComponent, Ghost] {
...
-        highResolution.doClick()
-      else
-        standardResolution.doClick()
+    /** Convert the asterism to the new type, and set the new target environment. */
+    def convertAsterism(converter: GhostAsterismConverter): Unit = Swing.onEDT {
+      val o = getContextObservation
+      val oc = getContextTargetObsComp
+
+      val newToc = for {
+        toc <- o.findTargetObsComp
```
swalker2m: If you've got `oc` above (which could be `null` so you should check for that) then you can just ask it for its data object:  

```scala
val action = for {
  oc  <- Option(getContextTargetObsComp)
  toc <- Option(oc.getDataObject).collect { case t: TargetObsComp => t }`
  env <- converter.convert(toc.getTargetEnvironment)
} yield {
  // Here I want something like cats.Eval.later() but since we don't have that
  () => { toc.setTargetEnvironment(env); oc.setDataObject(toc) }
}
```

This assumes you want to show the dialog box if something goes wrong.  You'd apply the action if it is defined or else show the dialog.  I'm wondering about this though.  Why can the conversion fail?  Is it just for the case where you're dealing with a `SingleTargetAsterism` that shouldn't be there?  If so that case should probably just disable the combobox in the first place (and show a Phase 2 check error), in which case you wouldn't `yield` anything above and instead just apply the action directly.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/ghost/GhostEditor.scala
```diff
@@ -56,87 +59,99 @@ final class GhostEditor extends ComponentEditor[ISPObsComponent, Ghost] {
...
+            }
+          case HighResolution(_, _, _) => Some(4)
+          case _                       =>
+            Dialog.showMessage(ui, "Illegal asterism type for GHOST observation.",
+              "Asterism Error", Dialog.Message.Error)
+            ui.asterismComboBox.enabled = false
+            None
+        }
+      }
+      asterismComboBox.selection.index = selectionIdx.getOrElse(0)
```
swalker2m: What is in the combo box model?  I wonder if you could handle it the same way you did for the position angle constraint combo in `PositionAnglePanel` where the model is the list of enum values.  Then when it comes time to change the selection it you can set the `selection.item` to an actual value instead of having to use the index:

`ui.positionAngleConstraintComboBox.selection.item = instrument.getPosAngleConstraint`

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/Ghost.scala
```diff
@@ -52,6 +58,10 @@ final class Ghost extends SPInstObsComp(GhostMixin.SP_TYPE) with PropertyProvide
     sc.putParameter(DefaultParameter.getInstance(Ghost.POS_ANGLE_PROP, getPosAngle))
     sc
   }
+
+  /** Take the current asterism, convert it to the new type, and set the target environment. */
+  def convertAsterism(converter: GhostAsterismConverter): Unit = {
+  }
```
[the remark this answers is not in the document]
sraaphorst: Ooops... 😊 
No... I originally wanted to gather all the target environment code in the instrument, like having the `OBSERVATION_NI` here, but realized it couldn't be done and forgot to take this out.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/ghost/GhostEditor.scala
```diff
@@ -56,87 +59,99 @@ final class GhostEditor extends ComponentEditor[ISPObsComponent, Ghost] {
...
-      "Dual target"   -> (() => GhostStandardResTargets.emptyDualTarget),
-      "SRIFU1 target, SRIFU2 sky position" -> (() => GhostStandardResTargets.emptyTargetPlusSky),
-      "SRIFU1 sky position, SRIFU2 target" -> (() => GhostStandardResTargets.emptySkyPlusTarget)
+    /** A map between GHOST asterism names and converters to those types. */
+    val asterismMap: Map[String, GhostAsterismConverter] = ListMap(
+      "Single target" -> GhostSingleTargetConverter,
+      "Dual target" -> GhostDualTargetConverter,
+      "SRIFU1 target, SRIFU2 sky position" -> GhostTargetPlusSkyConverter,
+      "SRIFU1 sky position, SRIFU2 target" -> GhostSkyPlusTargetConverter,
+      "High resolution" -> GhostHighResolutionConverter
```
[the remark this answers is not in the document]
sraaphorst: 👍

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/ghost/GhostEditor.scala
sraaphorst: There we go. Now `AsterismType`, which you can get from `Asterism`, is much more useful: it bundles together the `tag` for XML generation, the `name` of the asterism type, an empty `Asterism` creator for each type, and the `AsterismConverter` that is used to transform another `Asterism` into that type.

I like this much better, as `AsterismType` wasn't very useful before and only would have been used for P2 checks, really, and some very basic XML generation stuff.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostAsterism.scala
```diff
@@ -49,7 +48,7 @@ object GhostAsterism {
       Equal.equalA[GuideFiberState]
 
     def fromString(s: String): Option[GuideFiberState]
-      = All.findLeft(_.name == s)
+    = All.findLeft(_.name == s)
```
sraaphorst: Stupid IDEA reformatting.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/AsterismType.java
```diff
@@ -2,38 +2,58 @@
...
-    GhostHighResolution("ghostHighRes") {
-        @Override
-        public Asterism createEmptyAsterism() {
-            return GhostAsterism$.MODULE$.createEmptyHighResolutionAsterism();
-        }
-    },
+    /**
+     * Note all GHOST standard resolution asterisms have the same tag because we use this to output them to XML
+     * in a category together.
+     */
```
swalker2m: Hmm, I wonder if it would have been better to make different actual `Asterism` instances for all the different standard res options instead of having them all grouped under standard resolution.  Then you wouldn't have two levels of stuff and `AsterismType` would be 1:1 with `Asterism` instances.  That seems cleaner to me now.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/AsterismType.java
```diff
@@ -44,7 +64,12 @@ public Asterism createEmptyAsterism() {
 
         switch (instType) {
             case INSTRUMENT_GHOST:
-                return DefaultImList.create(GhostStandardResolution, GhostHighResolution);
+                return DefaultImList.create(
```
swalker2m: There's now an `Instrument` enum.  We could make `instType` be an `Instrument` instead of an `SPComponentType` and then you wouldn't have to check whether it is valid.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/AsterismType.java
```diff
@@ -2,38 +2,58 @@
...
+            new Some<>(AsterismConverters.GhostSkyPlusTargetConverter$.MODULE$),
+            GhostAsterism$.MODULE$::createEmptyStandardResolutionSkyPlusTargetAsterism),
+
+    GhostHighResolution("ghostHighRes", "High Resolution",
+            new Some<>(AsterismConverters.GhostHighResolutionConverter$.MODULE$),
+            GhostAsterism$.MODULE$::createEmptyHighResolutionAsterism),
     ;
 
     public final String tag;
+    public final Option<AsterismConverters.AsterismConverter> converter;
```
swalker2m: Something about the `converter` is troubling me.  In general there are _O(n!)_ combinations of pairs of asterisms that you could want to convert.  I suppose most of the conversions are just not specified though and return `None`?

sraaphorst: How do you see O(n!) ?

I would think it would be O(n^2), as we'd want to convert between any pair of asterism types, so assuming n asterism types, nC2 = n(n-1)/2 ?

sraaphorst: Actually, nix that. It should be O(n^2), but n(n-1) and not divided by two, obviously.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/AsterismType.java
sraaphorst: I think that this would make it difficult when we do the P2 check, since we have the `SpComponentType` of the instrument, but I don't think we'd have the `Instrument` enum value readily available?

The only time this code will be likely be used is for P2 checks, I think. (At least, that's the only time we're using it now.)

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/AsterismType.java
sraaphorst: Yeah, I agree that the extra layer adds a certain amount of possibly unnecessary complexity.

I guess we could consider making `StandardResolution` a `sealed trait` or `abstract sealed class`, and then make four `case class`es off of that, one for each asterism type?

PR #1466, opened 2018-04-11
REL-3395: GHOST P2 check to ensure supported asterism type

sraaphorst:
This adds a Phase 2 check that determines if the current `Asterism` in the `TargetEnvironment` is compatible with the instrument. If not, an error is generated.

I also added an old-school test case to the existing `GeneralRule` test suite to make sure that when a `Ghost` observation is configured with an unsupported asterism type (`Single`), then it generates the appropriate error.

### DIFF
--- a/bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/general/GeneralRule.java
+++ b/bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/general/GeneralRule.java
@@ -31,9 +31,7 @@ import edu.gemini.spModel.obsclass.ObsClass;
 import edu.gemini.spModel.obscomp.InstConstants;
 import edu.gemini.spModel.obscomp.SPInstObsComp;
 import edu.gemini.spModel.target.SPTarget;
-import edu.gemini.spModel.target.env.GuideGroup;
-import edu.gemini.spModel.target.env.GuideProbeTargets;
-import edu.gemini.spModel.target.env.TargetEnvironment;
+import edu.gemini.spModel.target.env.*;
 import edu.gemini.spModel.target.obsComp.GuideSequence;
 import edu.gemini.spModel.target.obsComp.PwfsGuideProbe;
 import edu.gemini.spModel.target.obsComp.TargetObsComp;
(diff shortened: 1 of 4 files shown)


cquiroz:
Looks good though i don't know enough about GHOST to comments on the internals

bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/general/GeneralRule.java
```diff
@@ -326,6 +324,32 @@ public IP2Problems check(final ObservationElements elements)  {
...
+
+        final SPInstObsComp inst = elems.getInstrument();
+        if (inst == null)
+            return null;
+
+        final P2Problems problems = new P2Problems();
+        tocOpt.foreach(toc -> {
+            final SPComponentType instType = inst.getType();
+            final AsterismType    at       = toc.getAsterism().asterismType();
+            final ImList<AsterismType> supportedAsterismTypes = AsterismType.supportedTypesForInstrument(instType);
```
swalker2m: I think we should add this to `SPInstObsComp`:

```java
    public Instrument getInstrument() {
        return Instrument.fromComponentType(getType()).getValue();
    }
```

That would throw an exception if the `SPComponentType` is not in fact an instrument component type, but nothing will be working anyway because we'd be using an `SPInstObsComp` that isn't in fact an instrument.  At any rate, then you can pass the `Instrument` directly.

Also, it seems like supported asterism types should be a `Set`?  Of course, we don't have an immutable `Set` so it'd have to be wrapped with the `Collections.unmodifiable` thing.  Or maybe the order matters in some other context?

bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/general/GeneralRule.java
```diff
@@ -326,6 +324,32 @@ public IP2Problems check(final ObservationElements elements)  {
...
+            if (!supportedAsterismTypes.contains(at))
+                problems.addError(PREFIX + "ASTERISM_TYPE_RULE",
+                        instType.readableStr + " observation is configured to use an unsupported asterism of type " + at.name + ".",
+                        elems.getTargetObsComponentNode().getValue());
```
swalker2m: It seems like we should indicate how to fix this problem if we can?  It might not be easy to fix of course.  There's no way in the target component to change the asterism type.  Even if you delete the target component and add another you likely won't have fixed it, unless we hack around in the OT to check and possibly change the asterism type when you add a target component.

In the merge conflict notes that we add, there are links to fix problems automatically.  I wonder if we could do something like that.  We can come back to this problem later I suppose though, since it really isn't very important for September.

bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/general/GeneralRule.java
swalker2m: If we do this, we could check that the component type resolves to an `Instrument` in the constructor as well so that at least we know at construction time.

bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/general/GeneralRule.java
swalker2m: Could you add a task for this?

sraaphorst: Sure thing! I'll add a GHOST subtask.

bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/general/GeneralRule.java
sraaphorst: I'll make these changes prior to committing.

PR #1467, opened 2018-04-11
REL-3390B: GHOST major refactor

sraaphorst:
This is a major refactoring of `GhostAsterism` to divvy up `StandardAsterism` into four subclasses, one for each type of standard asterism as per discussion with @swalker2m earlier today.

(It also includes REL-3395, as I branched from that PR to create this one.)

Everything else in here is a consequence of that refactor.

I think this is an improvement, especially to `AsterismType`, and simplifies a number of things.

This PR is largely to ensure that all test cases still pass and to get feedback from @swalker2m.


sraaphorst:
Excellent. Thanks!

PR #1469, opened 2018-04-14
REL-3412: GMOS-S ITC fails with power-law input SED

andrewwstephens:
The power-law generator was occasionally going 1 step beyond the predefined dimension of the array due to imprecision in representing the looped variable.  In the case I investigated (B600 @ 500nm) the loop was set to stop at 1181.97, but on what should have been the last step the wavelength was only at 1181.969999999 so it tried to loop through one more time and crashed.  To avoid this I changed the for loop to increment the index rather than the wavelength.

### DIFF
--- a/bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/base/PowerLawSpectrum.java
+++ b/bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/base/PowerLawSpectrum.java
@@ -26,14 +26,9 @@ public class PowerLawSpectrum extends DefaultSampledSpectrum {
 
         final int n = (int) ((_end - _start) / _sampling + 1);
         double[] fluxArray = new double[n];
-
-
-        int i = 0;
-
-        for (double lam = _start; lam <= _end; lam += _sampling) {
+        for (int i = 0; i < n ; i++) {
+            double lam = _start + i * _sampling;
             fluxArray[i] = _pLawFlux(lam, powerLawIndex);
-            i++;
-
         }
 
         reset(fluxArray, _start, _sampling);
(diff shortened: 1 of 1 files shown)


andrewwstephens:
I looked at this a little more closely trying to understand why this failed in some circumstances and not in others, and it came down to the calculation of the array dimension, namely the truncation of a double to an int.  In the case I tested (GMOS-S, B600 @ 500nm) the dimension n = 22266.999999999996 was truncated to 22266.  In my second commit I add a Math.round(n) so that the wavelength range returned more closely matches the wavelength range requested.

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/base/PowerLawSpectrum.java
```diff
@@ -24,16 +24,11 @@ public PowerLawSpectrum(final double powerLawIndex, final double start, final do
         _start /= (1 + z);
         _end /= (1 + z);
 
-        final int n = (int) ((_end - _start) / _sampling + 1);
+        final int n = (int) Math.round((_end - _start) / _sampling + 1);
```
cquiroz: 👍

sraaphorst: I don't think there should be any need to cast to `int` after `Math.round`, which, in this case, would return a `long`. We could just make `i` a `long` and then proceed as is being done.

andrewwstephens: I think `n` must be an integer since it is used to define the size of `fluxArray` in line 28?

sraaphorst: You're right. I didn't know that you were obligated to use an `int` for size when creating a new array, but apparently, you are. 😊

PR #1470, opened 2018-04-14
REL-3403: Trim NIRI H2O and Ch4ice filters

andrewwstephens:


andrewwstephens:
A user reported that the NIRI ITC gave errors when using a custom SED with the H2O and CH4ice filters.  These filters extended down to 300nm, blueward of the detector sensitivity, and therefore required that the input SED extend to the same lower limit.  Here I simply trimmed the lower limits to 850nm, which is still a bit below the detector sensitivity limit, so the output results should remain unchanged.

PR #1473, opened 2018-04-18
Initial ICTD / QPT integration

swalker2m:
Integrates the ICTD / QPT as outlined in REL-902.  The implementation is split over several bundles:

* `pot / spModel` - contains the mapping from ICTD table rows to `enum` values.  This could have been done with a typeclass I imagine without touching the `enum` definitions themselves.  I thought it best though to keep the mapping with the individual `enum` elements so that it is clear that it needs to be defined if you modify anything there.

* `ictd` - contains the DAO for reading ICTD information and mapping it to OCS enums, etc.

* `qpt` - reading ICTD data and updating "facilities" and filtering on custom mask availability.

__Note:__ this PR is missing database testing.  My plan is to switch to other June tasks and come back to this during the May testing period.  I want to see if using H2 in mysql mode is sufficient for this because we don't want to require all developers to install mysql and setup the ICTD locally.


bundle/edu.gemini.ictd/src/main/scala/edu/gemini/ictd/FeatureAvailability.scala
```diff
@@ -0,0 +1,155 @@
...
+      )
+  }
+
+  // Given a map of String key to Availability drawn from the ICTD, and an Enum
+  // type A, find the availability of each instance of A.
+  private def resolve[A <: java.lang.Enum[A] with IctdType : ClassTag](
+    tab: AvailabilityTable
+  ): Map[A, Availability] = {
+
+    val c  = classTag[A].runtimeClass.asInstanceOf[Class[A]]
```
cquiroz: Reflection is often problematic. Do you see any risks here?

bundle/edu.gemini.ictd/src/main/scala/edu/gemini/ictd/FeatureTables.scala
```diff
@@ -0,0 +1,36 @@
...
+  longslit:   LongslitTable,
+  nsLongslit: NSLongslitTable
+)
+
+
+object FeatureTables {
+
+  val empty: FeatureTables = {
+    def emptyTable[T]: AvailabilityTable @@ T =
+      tag[T][AvailabilityTable](Map.empty)
```
cquiroz: Could you explain what does `tag` do?

bundle/edu.gemini.ictd/src/main/scala/edu/gemini/ictd/FeatureAvailability.scala
tpolecat: If you had a typeclass instance for each `A` that provided the values you would't need to use a tag, but this is probably good enough. We're not looking things up by name, which is where you can get into trouble.

bundle/edu.gemini.ictd/src/main/scala/edu/gemini/ictd/dao/FeatureTablesDao.scala
```diff
@@ -0,0 +1,111 @@
...
+
+  final case class Feature(
+    name:       String,
+    instrument: Instrument,
+    location:   Location
+  )
+
+  // Tag the various kinds of Feature to distinguish them.  They are otherwise
+  // identical triplets of (name, Instrument, Location).
+  type Circular   = Feature @@ CircularTag
```
cquiroz: Are you using this like `newtype` or opaque types I guess

bundle/edu.gemini.pot/src/main/java/edu/gemini/pot/sp/Instrument.java
```diff
@@ -5,36 +5,47 @@
...
     Trecs(INSTRUMENT_TRECS),
-    Visitor(INSTRUMENT_VISITOR)
+    Visitor(INSTRUMENT_VISITOR, GN, GS)
     ;
 
     public final SPComponentType componentType;
+    public final Set<Site> sites;
 
-    Instrument(SPComponentType type) {
+    Instrument(SPComponentType type, Site... sites) {
```
cquiroz: oh, model change. I'll have to update seqexec after we merge this in

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/flamingos2/Flamingos2.java
```diff
@@ -127,26 +128,26 @@ public String toString() {
     /**
      * Filters
      */
-    public enum Filter implements DisplayableSpType, SequenceableSpType, LoggableSpType, ObsoletableSpType {
+    public enum Filter implements DisplayableSpType, SequenceableSpType, LoggableSpType, ObsoletableSpType, IctdType {
 
-        OPEN("Open", "Open", new Some<>(1.6)) {
+        OPEN("Open", "Open", new Some<>(1.6),                     Ictd.unavailable()) {
```
cquiroz: I wonder if this will bring into the seqexec, all the dependencies. I think scalaz is not a problem but would doobie and shapeless come along to the seqexec?

cquiroz:
First of all, this seems a great work, it's difficult to bridge to an application like ICTD but you still managed it. I can't comment on the QPT part but the rest seems quite good.

I have a concern on whether the model changes will carry `shapeless` and `doobie` to the seqexec. That would be problematic as we have `shapeless` as an indirect dependency and we may add `doobie` in the near future. Or are they restricted to the `ictd` package? Maybe we should test this before merging

tpolecat:
Would it make sense to move the contents of `package object ictd` and `FeatureTables` down into the `dao` package since they're really irrelevant for end users (who only deal with the fully-deswizzled `FeatureAvailability`)?

swalker2m:
I changed `project/OcsCredentials.scala` for this, which breaks Travis apparently.  What do we need to do to get that working?   I failed to update `project/OcsCredentials.scala.template`.  Maybe that's it?

cquiroz:
Most likely that is the reason. travis uses `OcsCredentials.scala.template` to get around the need to access the internal svn repo

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/flamingos2/Flamingos2.java
swalker2m: No, the `doobie` and `shapeless` stuff are not in `pot` so they shouldn't get pulled in.

bundle/edu.gemini.pot/src/main/java/edu/gemini/pot/sp/Instrument.java
swalker2m: The `Instrument` enum was only added recently for GHOST so I doubt seqexec uses it for anything.  Can't believe we made it this long without an enumeration of the instruments.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/flamingos2/Flamingos2.java
cquiroz: `pot` is certainly a seqexec dependency

bundle/edu.gemini.ictd/src/main/scala/edu/gemini/ictd/dao/FeatureTablesDao.scala
swalker2m: Right, without the tags everything is just a `Feature` and you can easily mix them up.

bundle/edu.gemini.ictd/src/main/scala/edu/gemini/ictd/dao/FeatureTablesDao.scala
```diff
@@ -0,0 +1,111 @@
...
+    ns: List[NSLongslit]
+  ): Map[Instrument, FeatureTables] = {
+
+    def toAvailabilityTable[T](features: List[Feature @@ T]): AvailabilityTable @@ T =
+      tag[T][AvailabilityTable](features.map(f => (f.name, f.location.availability)).toMap)
+
+    def group[T](features: List[Feature @@ T]): Map[Instrument, AvailabilityTable @@ T] =
+      features.groupBy(_.instrument)
+              .mapValues(toAvailabilityTable)
+              .withDefaultValue(tag[T][AvailabilityTable](Map.empty))
```
tpolecat: When you use `.withDefaultValue` it breaks a lot of map invariants, so it's safer in these cases to widen to a function type. Would things work if the return type here were `Instrument => AvailabilityTable @@ T` ?

Same comment with the use of `withDefault` elsewhere.

swalker2m:
> Would it make sense to move the contents of package object ictd and FeatureTables down into the dao package since they're really irrelevant for end users (who only deal with the fully-deswizzled FeatureAvailability)?

I think that makes sense yes.

bundle/edu.gemini.ictd/src/main/scala/edu/gemini/ictd/dao/FeatureTablesDao.scala
tpolecat: With scalaz tags it's opaque: `A @@ T` and `A` are unrelated. With shapeless `@@` (as here) it's a refinement: `A @@ T` is a subtype of `A`. Both are reasonable, depends on how you plan to use them.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/flamingos2/Flamingos2.java
swalker2m: Doobie and shapeless are already used in the ocs codebase, in the `edu.gemini.util.security` bundle.  I'm using the same versions in `edu.gemini.ictd`, which is not referenced from `edu.gemini.pot`. All that is in `pot` is a description of how to lookup corresponding values in the database -- basically a listing of the associated `String` keys per `enum` value.

It does use a scalaz type, `\&/`.  Do you think that will be problematic for the seqexec?

cquiroz: `scalaz` is not an issue as we are moving away from it. It will remain a runtime dependency but that's fine. `shapeless` and `doobie` are not directly used at the moment in `ocs3` so it is not an issue but if we were to add it could cause trouble. I've checked and at the moment we get `shapeless` from `monocle`. We don't import `edu.gemini.util.security`

cquiroz: It seems then it shouldn't be a problem if we don't import `ictd` right. Do you get `shapeless` or `doobie` packages on `pot`'s osgi manifest?

swalker2m: No, neither appears in the manifest.

cquiroz: I think that solves my worry

PR #1474, opened 2018-04-19
REL-1714 Extra notice messages for constraints

jdnavarro:
This PR is still not complete but it's intended as a request for feedback. These are the main points of contention.

- Multiple constraint intervals. Right now it only uses the first interval of the night, but how do we really want to represent these cases? I'd try to show the interval that affects more allocations but I'm not sure how to do it and that might require much more time than are willing to spend on.

- Open ended constraints. I'd use the start/end of the night, but would that be correct?

- Conditions to show messages. Right now they are always shown when there are constraints present, but I guess maybe we don't want to show them when the condition is not met, is that the case?

- Background constraint. It's still pending but since it seems the trickiest I'm leaving it until the end, after confirming the way I'm doing the other constraints is correct.

I had several issues with null exceptions when dragging allocations around. I think I ironed out most cases but I can't be completely sure I got them all, so please let me know if you catch any.

Sorry for the mess with the tabs/spaces, I realized late there was already a mix of them in this file. I decided not to convert them all to spaces in this PR because it would make harder to review it. I can do it in a different PR.

### DIFF
--- a/bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
+++ b/bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
@@ -2,20 +2,26 @@ package edu.gemini.qpt.core.listeners;
 
 import java.beans.PropertyChangeEvent;
 import java.util.*;
-import java.util.function.Predicate;
+import java.util.function.*;
+import java.time.Instant;
+import java.time.format.DateTimeFormatter;
+import java.util.TimeZone;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
 
 import edu.gemini.qpt.core.Alloc;
+import edu.gemini.qpt.core.Marker;
 import edu.gemini.qpt.core.Variant;
 import edu.gemini.qpt.core.Alloc.Circumstance;
 import edu.gemini.qpt.core.Marker.Severity;
 import edu.gemini.qpt.core.Variant.Flag;
+import edu.gemini.qpt.core.util.*;
 import edu.gemini.qpt.shared.sp.Group;
 import edu.gemini.qpt.core.util.LttsServicesClient;
 import edu.gemini.qpt.core.util.MarkerManager;
 import edu.gemini.qpt.shared.sp.Obs;
 import edu.gemini.qpt.shared.util.TimeUtils;
+import edu.gemini.shared.util.immutable.*;
 import edu.gemini.spModel.obscomp.SPGroup.GroupType;
 
 /**
@@ -27,6 +33,17 @@ public class LimitsListener extends MarkerModelListener<Variant> {
 	@SuppressWarnings("unused")
 	private static final Logger LOGGER = Logger.getLogger(LimitsListener.class.getName());
 
+	private static final DateTimeFormatter dateFormat(TimeZone zone) {
+		return DateTimeFormatter.ofPattern("HH:mm").withZone(zone.toZoneId());
+	};
+
+	private static final String formatInterval(TimeZone zone, Interval interval) {
+		final Function<Long, String> f =
+			ms -> dateFormat(zone).format(Instant.ofEpochMilli(ms));
+
+		return String.format("(%s, %s)", f.apply(interval.getStart()), f.apply(interval.getEnd()));
+	}
+
     /**
      * LGS obs have a 40 deg limit by default
      */
@@ -37,6 +54,8 @@ public class LimitsListener extends MarkerModelListener<Variant> {
 		Variant v = (Variant) evt.getSource();
 		MarkerManager mm = v.getSchedule().getMarkerManager();
 		mm.clearMarkers(this, v);
+
+		final TimeZone zone = v.getSchedule().getSite().timezone();
 
 		for (Alloc a: v.getAllocs()) {
 
@@ -67,13 +86,32 @@ public class LimitsListener extends MarkerModelListener<Variant> {
 				mm.addMarker(false, this, Severity.Warning, String.format("Tracking: target reaches %1.2f\u00B0.", maxElevation), v, a);
 			}
 
+			// Function to create markers that report solver intervals.
+			final BiFunction<String, String, Option<Marker>> createSolverMarker =
+					(cacheName, prefix) -> {
+						final Map<Obs, Union<Interval>> c = v.getSchedule().getCache(cacheName);
+						final Union<Interval>          ws = c.getOrDefault(a.getObs(), new Union<>());
+						final ImList<Interval>         is = DefaultImList.create(ws.getIntervals());
+
+						if (is.isEmpty()) {
+							return None.instance();
+						} else {
+							final String m = is.map(i -> formatInterval(zone, i)).mkString(prefix + " ", ", ", ".");
+							return new Some<>(new Marker(false, this, Severity.Notice, m, v, a));
+						}
+					};
+
 			// Timing window
 			int percentVisible = (int) (100 * a.getMean(Circumstance.TIMING_WINDOW_OPEN, false));
 			switch (percentVisible) {
 			case 100: // good!
-				if (a.getObs().getTimingWindows().size() > 0)
+				if (a.getObs().getTimingWindows().size() > 0) {
 					mm.addMarker(true, this, Severity.Info, "Timing constraint is met.", v, a);
-				break;
+
+					// Report timing windows
+					createSolverMarker.apply(Variant.TIMING_UNION_CACHE, "Must be in the timing window").foreach(m -> mm.addMarker(m));
+				}
+                break;
 			case 0:
 				mm.addMarker(false, this, Severity.Error, "Timing constraint is violated for entire scheduled visit.", v, a);
 				break;
@@ -83,12 +121,13 @@ public class LimitsListener extends MarkerModelListener<Variant> {
 
 			}
 
-            // LCH Laser Shutter Warnings
+			// LCH Laser Shutter Warnings
             String msg = LttsServicesClient.getInstance().getShutterWindowWarningMessage(a);
             if (msg != null) {
                 mm.addMarker(false, this, Severity.Warning, msg, v, a);
             }
 
+            // Elevation Constraint
             // Elevation Constraint
 			final Double maxAirmass = a.getMax(Circumstance.AIRMASS, false);
 			final Double minAirmass = a.getMin(Circumstance.AIRMASS, false);
@@ -125,21 +164,28 @@ public class LimitsListener extends MarkerModelListener<Variant> {
 
 			case AIRMASS:
 
-				double airmassLimitMin = a.getObs().getElevationConstraintMin();
-				double airmassLimitMax = a.getObs().getElevationConstraintMax();
+				final double airmassLimitMin = a.getObs().getElevationConstraintMin();
+				final double airmassLimitMax = a.getObs().getElevationConstraintMax();
 
+				Option<Marker> airmassMarker = None.instance();
 				if (maxAirmass > airmassLimitMax) {
-					mm.addMarker(false, this, Severity.Error, String.format("Airmass constraint violated (%1.2f > %1.2f).", maxAirmass, airmassLimitMax), v, a);
+					airmassMarker = new Some<>(new Marker(false, this, Severity.Error, String.format("Airmass constraint violated (%1.2f > %1.2f).", maxAirmass, airmassLimitMax), v, a));
 				} else if (maxAirmass > airmassLimitMax * 0.975) {
-					mm.addMarker(false, this, Severity.Warning, String.format("Observation reaches airmass %1.2f.", maxAirmass), v, a);
+					airmassMarker = new Some<>(new Marker(false, this, Severity.Warning, String.format("Observation reaches airmass %1.2f.", maxAirmass), v, a));
 				}
 				if (airmassLimitMin > 1.0) {
 					if (minAirmass < airmassLimitMin) {
-						mm.addMarker(false, this, Severity.Error, String.format("Airmass constraint violated (%1.2f < %1.2f).", minAirmass, airmassLimitMin), v, a);
+						airmassMarker = new Some<>(new Marker(false, this, Severity.Error, String.format("Airmass constraint violated (%1.2f < %1.2f).", minAirmass, airmassLimitMin), v, a));
 					} else if (maxAirmass < airmassLimitMin * 1.025) {
-						mm.addMarker(false, this, Severity.Warning, String.format("Observation reaches airmass %1.2f.", maxAirmass), v, a);
+						airmassMarker = new Some<>(new Marker(false, this, Severity.Warning, String.format("Observation reaches airmass %1.2f.", maxAirmass), v, a));
 					}
 				}
+
+				airmassMarker.orElse(() -> {
+					final String m = String.format("Must be observed in the airmass range %1.2f - %1.2f", airmassLimitMin, airmassLimitMax);
+					return createSolverMarker.apply(Variant.VISIBLE_UNION_CACHE, m);
+				}).foreach(m -> mm.addMarker(m));
+
 				break;
 
 
@@ -153,20 +199,25 @@ public class LimitsListener extends MarkerModelListener<Variant> {
 				double minDelta = Math.abs(haLimitMin - minHA);
 				double maxDelta = Math.abs(haLimitMax - maxHA);
 
+				Option<Marker> haMarker = None.instance();
 				if (minHA < haLimitMin) {
-					mm.addMarker(false, this, Severity.Error, String.format("Hour angle constraint violated (%s < %s).", TimeUtils.hoursToHHMMSS(minHA), TimeUtils.hoursToHHMMSS(haLimitMin)), v, a);
+					haMarker = new Some<>(new Marker(false, this, Severity.Error, String.format("Hour angle constraint violated (%s < %s).", TimeUtils.hoursToHHMMSS(minHA), TimeUtils.hoursToHHMMSS(haLimitMin)), v, a));
 				} else if (minDelta < 1. / 15.) {
-					mm.addMarker(false, this, Severity.Warning, "Target comes within 1\u00B0 of lower HA constraint.", v, a);
+					haMarker = new Some<>(new Marker(false, this, Severity.Warning, "Target comes within 1\u00B0 of lower HA constraint.", v, a));
 				}
 
 				if (maxHA > haLimitMax) {
-					mm.addMarker(false, this, Severity.Error, String.format("Hour angle constraint violated (%s > %s).", TimeUtils.hoursToHHMMSS(maxHA), TimeUtils.hoursToHHMMSS(haLimitMax)), v, a);
+					haMarker = new Some<>(new Marker(false, this, Severity.Error, String.format("Hour angle constraint violated (%s > %s).", TimeUtils.hoursToHHMMSS(maxHA), TimeUtils.hoursToHHMMSS(haLimitMax)), v, a));
 				} else if (maxDelta < 1. / 15.) {
-					mm.addMarker(false, this, Severity.Warning, "Target comes within 1\u00B0 of upper HA constraint.", v, a);
+					haMarker = new Some<>(new Marker(false, this, Severity.Warning, "Target comes within 1\u00B0 of upper HA constraint.", v, a));
 				}
 
-				break;
+				haMarker.orElse(() -> {
+					final String m = String.format("Must be observed in the hour angle range %1.2f - %1.2f", haLimitMin, haLimitMax);
+					return createSolverMarker.apply(Variant.VISIBLE_UNION_CACHE, m);
+				}).foreach(m -> mm.addMarker(m));
 
+				break;
 			}
 
 			// Lunar proximity warnings
@@ -178,9 +229,14 @@ public class LimitsListener extends MarkerModelListener<Variant> {
 			}
 
 			// Sky brightness warnings
+
 			Double sb = a.getMin(Circumstance.TOTAL_SKY_BRIGHTNESS, false);
-			if (sb != null && !a.getObs().getConditions().containsSkyBrightness(sb))
+
+			if (sb != null && !a.getObs().getConditions().containsSkyBrightness(sb)) {
 				mm.addMarker(false, this, Severity.Error, "Sky brightness constraint violated.", v, a);
+			} else  {
+				createSolverMarker.apply(Variant.DARK_UNION_CACHE, "Background constraints are met between").foreach(m -> mm.addMarker(m));
+			}
 
 
 			// Variant Obs Flags
--- a/bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/Variant.java
+++ b/bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/Variant.java
@@ -70,6 +70,11 @@ public final class Variant extends BaseMutableBean implements PioSerializable, C
 	private Map<Obs, Union<Interval>> constrainedUnionCache;
 	private Map<Obs, Union<Interval>> timingUnionCache;
 
+	// Cache names
+	public static final String DARK_UNION_CACHE = "darkUnionCache";
+	public static final String VISIBLE_UNION_CACHE = "visibleUnionCache";
+	public static final String TIMING_UNION_CACHE = "timingUnionCache";
+
 	// The list of all groups encountered in the plan, by first appearance.
 	private List<Group> groups = new ArrayList<>();
 	private boolean flagUpdatesEnabled = true;


bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
```diff
@@ -78,7 +88,28 @@ public void propertyChange(PropertyChangeEvent evt) {
 
 			}
 
-            // LCH Laser Shutter Warnings
+			// Target timing window
+			final TimingWindowSolver timingWindowSolver = new TimingWindowSolver(a.getObs());
+			final Iterator<Interval> timingWindowIntervals = timingWindowSolver.solve(tbn.getStartTime(), tbn.getEndTime()).iterator();
+
+			if (timingWindowIntervals.hasNext()) {
+			    final Interval timingWindowInterval = timingWindowIntervals.next();
```
jdnavarro: These `next`s are provisional until we decide what to do with multiple intervals of constraints.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
```diff
@@ -135,6 +169,24 @@ public void propertyChange(PropertyChangeEvent evt) {
 						mm.addMarker(false, this, Severity.Warning, String.format("Observation reaches airmass %1.2f.", maxAirmass), v, a);
 					}
 				}
+
+				if (elevationIntervals.hasNext()) {
```
jdnavarro: These `hasNext` conditions should always be `True` but I left it just in case after getting so many null exceptions.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
```diff
@@ -36,6 +40,12 @@ public void propertyChange(PropertyChangeEvent evt) {
 		MarkerManager mm = v.getSchedule().getMarkerManager();
 		mm.clearMarkers(this, v);
 
+		final TimeZone zone = v.getSchedule().getSite().timezone();
+		final DateTimeFormatter dateFormat = DateTimeFormatter.ofPattern("HH:mm") .withZone(zone.toZoneId());
+
+		final Site site = v.getSchedule().getSite();
+		final TwilightBoundedNight tbn = new TwilightBoundedNight(TwilightBoundType.OFFICIAL, v.getSchedule().getMiniModel().getTimestamp(), site);
```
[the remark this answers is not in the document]
swalker2m: I don't understand the model well enough to point you in the right direction but it feels like this should be known. We're making a schedule for a specific night so that should be obtainable from somewhere, I would think.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
swalker2m: I would say list them all: (t0 - t1), (t2 - t3),  ...  Usually it will be just one but if there are multiple instances they should probably be displayed.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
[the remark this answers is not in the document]
swalker2m: @tpolecat Could you take a look at this PR?  Are the night bounds stored somewhere that we can get to?  What would you suggest?

It seems okay to just calculate them if not but my understanding was that official twilight is too bright and that we used nautical twilight bounds.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
swalker2m: Perhaps if the interval starts and/or ends at the twilight bounds it should say "twilight" instead of the actual time?

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
swalker2m: > These hasNext conditions should always be True since if you are in a non NONE case, it means there is at least one elevation constraint.

It may be there there is a constraint but the night may not have any time at which the constraint is met?  If so the `elevationIntervals` would be empty.

tpolecat:
I don't think I understand what you're trying to do here. QPT already respects timing windows.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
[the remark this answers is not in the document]
tpolecat: The timestamp marks when the schedule was constructed so that's not what you want.

A schedule can have multiple blocks, although right now there is exactly one, which is the observing night. So what you probably want to do is examine all the blocks. If there are none then the schedule is totally empty and you should be able to deal with that.

However I can't really figure out what they're asking you to do. What is a "constraint interval"? Is it distinct from a timing window?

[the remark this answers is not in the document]
andrewwstephens: The qpt is good about warning you when a constraint isn't  met, i.e. when you *can't* schedule an observation.  The idea here was to give the time range when you *can* schedule an observation.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
```diff
@@ -36,6 +41,17 @@ public void propertyChange(PropertyChangeEvent evt) {
 		MarkerManager mm = v.getSchedule().getMarkerManager();
 		mm.clearMarkers(this, v);
 
+		final TimeZone zone = v.getSchedule().getSite().timezone();
+		final DateTimeFormatter dateFormat = DateTimeFormatter.ofPattern("HH:mm") .withZone(zone.toZoneId());
+
+		final Site site = v.getSchedule().getSite();
+
+		SortedSet<Block> blocks = v.getSchedule().getBlocks();
+		Interval night = null;
```
swalker2m: It's a tiny wee detail, but `blocks` could be marked `final` here as well.  Also, we could use a Java `Optional` or else our own `edu.gemini.shared.util.immutable.Option` to avoid a nullable variable.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
```diff
@@ -66,9 +82,45 @@ public void propertyChange(PropertyChangeEvent evt) {
...
+                    final Iterator<Interval> timingWindowIntervals = timingWindowSolver.solve(night).iterator();
+
+                    if (timingWindowIntervals.hasNext()) {
+                    	StringBuilder sb = new StringBuilder();
+						final Interval timingWindowInterval0 = timingWindowIntervals.next();
+						sb.append(
+							String.format(
+								"Must be in the timing window (%s, %s)",
+								dateFormat.format(Instant.ofEpochMilli(timingWindowInterval0.getStart())),
+								dateFormat.format(Instant.ofEpochMilli(timingWindowInterval0.getEnd()))
```
swalker2m: Since we format an `Interval` into a `String` "(t0 - t1)" in several places, it's probably worth extracting that code into a static method.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
```diff
@@ -66,9 +82,45 @@ public void propertyChange(PropertyChangeEvent evt) {
...
 			switch (percentVisible) {
 			case 100: // good!
-				if (a.getObs().getTimingWindows().size() > 0)
+				if (a.getObs().getTimingWindows().size() > 0 && night != null)
 					mm.addMarker(true, this, Severity.Info, "Timing constraint is met.", v, a);
-				break;
+
+                    // Report timing windows
+                    final TimingWindowSolver timingWindowSolver = new TimingWindowSolver(a.getObs());
+                    final Iterator<Interval> timingWindowIntervals = timingWindowSolver.solve(night).iterator();
```
swalker2m: If `percentVisible` is `100` here, what are we doing again?  The idea is to report on other times besides the current `Alloc` where we could also place this observation?

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
```diff
@@ -66,9 +82,45 @@ public void propertyChange(PropertyChangeEvent evt) {
...
+
+                    if (timingWindowIntervals.hasNext()) {
```
swalker2m: If you're willing to wander into the seamy underbelly, you might get some mileage out of the `edu.gemini.shared.util.immutable` stuff.

```java
final ImList<Interval> timingWindowIntervals = DefaultImList.create(timingWindowSolver.solve(night));
if (timingWindowIntervals.nonEmpty()) {
    final String msg = timingWindowIntervals.map(i -> formatInterval(i)).mkString("Must be in the timing window ", ", ", ".");
    mm.addMarker(false, this, Severity.Notice, msg, v, a);
}
``` 

assuming you extract the code that does `Interval => String` in a method called `formatInterval`. (See next comment)

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
```diff
@@ -160,8 +238,25 @@ public void propertyChange(PropertyChangeEvent evt) {
 					mm.addMarker(false, this, Severity.Warning, "Target comes within 1\u00B0 of upper HA constraint.", v, a);
 				}
 
-				break;
+				if (minHA < haLimitMin && maxHA > haLimitMax && elevationIntervals != null && elevationIntervals.hasNext()) {
```
swalker2m: Hmm, aren't these conditions already accounted for above in the errors and warnings.  That is, shouldn't you only add this information if `minHA >= haLimitMin && maxHA <= haLimitMax`?

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
jdnavarro: I already attempted to use `ImOption` but got into a mess with the switch cases, but this approach looks like very easy to implement.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
jdnavarro: The idea, if I understood correctly, is that if all constraints are met we give a notice message showing where they are so the user can rearrange allocations without having to find out manually and writing down the constraint ranges.

If any constraints is not met then we get an error right away so we don't need to show the constraints info, it's already on the error. The timing window error is kind of weird because anything that is not `100` visible is already an error, but the error reported is an overall percentage of all allocations, not of the specific allocation not meeting the constraint.

jdnavarro:
I'm not satisfied with the use of `ImOption<night>` but I'm not sure how to properly use it without changing everything.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
```diff
@@ -93,6 +139,10 @@ public void propertyChange(PropertyChangeEvent evt) {
 			final Double maxAirmass = a.getMax(Circumstance.AIRMASS, false);
 			final Double minAirmass = a.getMin(Circumstance.AIRMASS, false);
 
+			final ElevationConstraintSolver elevationSolver = ElevationConstraintSolver.forObs(site, a.getObs());
+
+            final Option<Interval> elevationInterval = night.map(n -> elevationSolver.solve(n).getIntervals().first());
```
swalker2m: Is it guaranteed that the `elevationSolver.solve(n).getIntervals()` will be nonempty?  Otherwise `.first()` will throw an exception.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
```diff
@@ -178,9 +262,32 @@ public void propertyChange(PropertyChangeEvent evt) {
...
+			Iterator<Interval> skyBackgroundIntervals = null;
+			if (!night.isEmpty() && sb != null) {
+				final Obs obs = a.getObs();
+				final SkyBackgroundSolver skyBackgroundSolver = new SkyBackgroundSolver(site, obs::getCoords, sb);
+				skyBackgroundIntervals = skyBackgroundSolver.solve(night.getValue()).iterator();
+			}
+
+			if (sb != null && !a.getObs().getConditions().containsSkyBrightness(sb)) {
 				mm.addMarker(false, this, Severity.Error, "Sky brightness constraint violated.", v, a);
+			} else if (skyBackgroundIntervals.hasNext()) {
```
swalker2m: It seems like `skyBackgroundIntervals` could be `null` here?

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
```diff
@@ -140,6 +190,24 @@ public void propertyChange(PropertyChangeEvent evt) {
 						mm.addMarker(false, this, Severity.Warning, String.format("Observation reaches airmass %1.2f.", maxAirmass), v, a);
 					}
 				}
+
+				if (maxAirmass < airmassLimitMax && minAirmass > airmassLimitMin && !elevationInterval.isEmpty()) {
```
swalker2m: You can use `.foreach()` as in `elevationInterval.foreach(i -> {if (maxAirmass ...) ...} )` but i dunno if it is much better really.  I think this is fine.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
```diff
@@ -38,6 +63,17 @@ public void propertyChange(PropertyChangeEvent evt) {
...
+		final TimeZone zone = v.getSchedule().getSite().timezone();
+
+		final Site site = v.getSchedule().getSite();
+
+		final SortedSet<Block> blocks = v.getSchedule().getBlocks();
+		Interval nnight = null;
+		if (!blocks.isEmpty()) {
+			nnight = blocks.first().getInterval();
+		}
+		final Option<Interval> night = ImOption.apply(nnight);
```
swalker2m: If you wanted, you could avoid the `nnight` with the `? :` operator thingy:

```
final Option<Interval> night = ImOption.apply(blocks.isEmpty() ? null : blocks.first().getInterval());
```

This seems fine though.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
```diff
@@ -71,9 +100,19 @@ public void propertyChange(PropertyChangeEvent evt) {
...
+				if (a.getObs().getTimingWindows().size() > 0 && !night.isEmpty())
 					mm.addMarker(true, this, Severity.Info, "Timing constraint is met.", v, a);
-				break;
+
+                    // Report timing windows
+                    final TimingWindowSolver timingWindowSolver = new TimingWindowSolver(a.getObs());
+                    final ImList<Interval> timingWindowIntervals = DefaultImList.create(timingWindowSolver.solve(night.getValue()).getIntervals());
+
+                    if (!timingWindowIntervals.nonEmpty()) {
+						final String msg = timingWindowIntervals.map(i -> formatInterval(zone, i)).mkString("Must be in the timing window ", ", ", ".");
```
swalker2m: `!timingWindowIntervals.nonEmpty()` == `timingWindowIntervals.isEmpty()` in which case the `msg` if added would always be "Must be in the timing window ."?  Am I reading it wrong?

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
```diff
@@ -178,9 +261,38 @@ public void propertyChange(PropertyChangeEvent evt) {
...
+				final SkyBackgroundSolver skyBackgroundSolver = new SkyBackgroundSolver(site, obs::getCoords, sb);
+				final SortedSet<Interval> intervals = skyBackgroundSolver.solve(n).getIntervals();
+				if (intervals.isEmpty()) {
+					return None.instance();
+				} else {
+					final Interval interval = intervals.first();
+					if (interval.getStart() >= n.getStart() && interval.getEnd() <= n.getEnd()) {
+						return None.instance();
+					}
+					return new Some<>(intervals.first());
```
swalker2m: Is the idea that the `skyBackgroundInterval` be empty if the solver finds an interval that falls inside of the night? Could you add a comment or something that explains what this block does and what it will be used for?

jdnavarro: That was a last moment hack. For some reason the default sky background interval is larger than the night interval (concretely the end is different, the start is exactly the same).

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
```diff
@@ -27,6 +35,19 @@
...
+	private static final DateTimeFormatter dateFormat(TimeZone zone) {
+		return DateTimeFormatter.ofPattern("HH:mm").withZone(zone.toZoneId());
+	};
+
+	private static final String formatInterval(TimeZone zone, Interval interval) {
+
+		return String.format(
+			"(%s, %s)",
+			DateTimeFormatter.ofPattern("HH:mm").withZone(zone.toZoneId()).format(Instant.ofEpochMilli(interval.getStart())),
+			DateTimeFormatter.ofPattern("HH:mm").withZone(zone.toZoneId()).format(Instant.ofEpochMilli(interval.getEnd()))
```
swalker2m: Did you mean `dateFormat(zone).format(Instant.ofEpochMilli(interval.getStart()))` etc?

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
```diff
@@ -178,9 +261,39 @@ public void propertyChange(PropertyChangeEvent evt) {
...
+				final SortedSet<Interval> intervals = skyBackgroundSolver.solve(n).getIntervals();
+				if (intervals.isEmpty()) {
+					return None.instance();
+				} else {
+					final Interval interval = intervals.first();
+					// For some reason the default background interval is larger than the night interval
+					if (interval.getStart() >= n.getStart() && interval.getEnd() <= n.getEnd()) {
+						return None.instance();
+					}
+					return new Some<>(intervals.first());
```
swalker2m: My understanding was that the solver would clip the intervals based on what is passed to `solve`, but that's not happening?  At any rate, won't it often be the case that the interval will fall inside of the night?  That is, `interval.getStart() >= n.getStart() && interval.getEnd() <= n.getEnd()` seems like the _usual_ case?

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
```diff
@@ -178,9 +261,39 @@ public void propertyChange(PropertyChangeEvent evt) {
...
 				mm.addMarker(false, this, Severity.Error, "Sky brightness constraint violated.", v, a);
+			} else if (!skyBackgroundInterval.isEmpty()) {
+				mm.addMarker(
+					false,
+					this,
+					Severity.Notice,
+					String.format(
+						"Must be observed in the sky background range %s - %s",
+						dateFormat(zone).format(Instant.ofEpochMilli(skyBackgroundInterval.getValue().getStart())),
+						dateFormat(zone).format(Instant.ofEpochMilli(skyBackgroundInterval.getValue().getEnd()))
```
swalker2m: Wouldn't you want to use your `formatInterval` method here as well?  Is it just that parens would be used?  That seems like it would be okay too.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
tpolecat: Yes it seems like the solver should always clip the intervals. Why are we only looking at the first such interval? If the moon rises and sets there might be two usable intervals, seems like.

Calculating this stuff is expensive and it's already been done at least once in the `Variant` to determine whether the obs is schedulable in the first place, so it seems like we could cache that computation and rely on it rather than recomputing here. This is fine for now but this class is getting messy and needs some cleanup eventually.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
sraaphorst: I think lines 38-40 should be removed entirely, and instead, make `formatInterval` as follows:

````
private static final String formatInterval(TimeZone zone, Interval interval) {
    final Function<Long, String> f =
            ms -> DateTimeFormatter.ofPattern("HH:mm").withZone(zone.toZoneId()).format(Instant.ofEpochMilli(ms));
    return String.format("(%s, %s)", f.apply(interval.getStart()), f.apply(interval.getEnd()));
}
````

That way, you avoid instantiating and initializing the `DateTimeFormatter` twice, get rid of the otherwise useless function, and instead use a lambda.

File: bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java [hunk 18]
```diff
@@ -83,7 +122,7 @@ public void propertyChange(PropertyChangeEvent evt) {
 
 			}
 
-            // LCH Laser Shutter Warnings
+			// LCH Laser Shutter Warnings
```
sraaphorst: Something funky happened with the comment indentation here.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
sraaphorst: By "drop it all together", what exactly do you mean? Omit it from 2018B due to the complexity at this point, or just go with the flawed result?

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
sraaphorst: It seems like what we would really like, instead of my above solution, is something along the line of:

```
final String formatInterval(final TimeZone zone, final Interval interval, final String sep = " - ") {
    final Function<Long, String> f = ms ->
        DateTimeFormatter.ofPattern("HH:mm").withZone(zone.toZoneId()).format(Instant.ofEpochMilli(ms));
   return f.apply(interval.getBegin()) + set + f.apply(interval.getEnd());
```
All  other bells and whistles (such as surrounding parentheses) can be done outside in the calling code of this method.


Then  L289 - 29 changes to:
```
"Must be observed in the sky background rage " + formatInterval(zone, skyBackgroundInterval.getValue());
```
Although using the naive exception-throwing `SkyBackground.getValue` makes me uncomfortable still.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
jdnavarro: @sraaphorst I meant dropping just the sky background check, but I'll try to fix it.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
```diff
@@ -38,6 +58,12 @@ public void propertyChange(PropertyChangeEvent evt) {
 		MarkerManager mm = v.getSchedule().getMarkerManager();
 		mm.clearMarkers(this, v);
 
+		final TimeZone zone = v.getSchedule().getSite().timezone();
+
+		final Site site = v.getSchedule().getSite();
+
+		final SortedSet<Block> blocks = v.getSchedule().getBlocks();
```
swalker2m: I don't think we use the `blocks` anywhere now, correct?  For that matter, maybe `site` isn't used anywhere now either?

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
```diff
@@ -67,13 +93,32 @@ public void propertyChange(PropertyChangeEvent evt) {
...
 			int percentVisible = (int) (100 * a.getMean(Circumstance.TIMING_WINDOW_OPEN, false));
 			switch (percentVisible) {
 			case 100: // good!
-				if (a.getObs().getTimingWindows().size() > 0)
+				if (a.getObs().getTimingWindows().size() > 0) {
 					mm.addMarker(true, this, Severity.Info, "Timing constraint is met.", v, a);
-				break;
+
+					// Report timing windows
+					createSolverMarker.apply("timingUnionCache", "Must be in the timing window").foreach(m -> mm.addMarker(m));
```
swalker2m: We should define these cache names like `"timingUnionCache"` somewhere as we discussed.  Maybe in `Variant` since that's where they are created?

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
```diff
@@ -153,20 +205,25 @@ public void propertyChange(PropertyChangeEvent evt) {
...
+					haMarker = new Some<>(new Marker(false, this, Severity.Error, String.format("Hour angle constraint violated (%s > %s).", TimeUtils.hoursToHHMMSS(maxHA), TimeUtils.hoursToHHMMSS(haLimitMax)), v, a));
 				} else if (maxDelta < 1. / 15.) {
-					mm.addMarker(false, this, Severity.Warning, "Target comes within 1\u00B0 of upper HA constraint.", v, a);
+					haMarker = new Some<>(new Marker(false, this, Severity.Warning, "Target comes within 1\u00B0 of upper HA constraint.", v, a));
 				}
 
-				break;
+				haMarker.orElse(() -> {
+					final String m = String.format("Must be observed in the hour angle range %1.2f - %1.2f", haLimitMin, haLimitMax);
+					return createSolverMarker.apply("VisibleUnionCache", m);
```
swalker2m: Here we have a capital `V` in `VisibleUnionCache` which underscores the need to put these names in constants.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
tpolecat: QPT is formatted with tabs so we should probably do a PR to reformat everything.

tpolecat: It's my fault. I used to think it was a good idea.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
tpolecat: Yes, these should be constants on `Variant`.

tpolecat:
This looks kosher thus far, but yes I also questioned the requirement at the outset, because if they're getting all these warnings they will want to find the intersection where all (or some, which will introduce UI complications) are satisfied.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
jdnavarro: There was already a mix of spaces of tabs on that file, so I used spaces for new editions.

I didn't want to change everything to spaces in this PR because it would be difficult to review, but we should create a separate PR just for replacing all the tabs.

PR #1475, opened 2018-04-20
REL-3279: ITC full exposure time for GMOS N&S

swalker2m:
This PR is another attempt at REL-3279.  Unfortunately it doesn't improve much upon [the previous version](https://github.com/gemini-hlsw/ocs/pull/1460/files), but does at least complain if it finds unexpected values in the sequence or if required values are missing instead of using defaults.

After discussion with Bryan and Arturo, we decided that I would work on building out sequence test cases in May as time permits in order to focus on getting more amnesty tasks into the test release now.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/config2/config.scala
+++ b/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/config2/config.scala
@@ -1,0 +1,70 @@
+package edu.gemini.spModel.config2
+
+import scala.reflect.ClassTag
+
+import scalaz._
+import Scalaz._
+
+/** A bit of syntax to simplify working with Config.
+  */
+trait ConfigSyntax {
+  implicit final def configSyntax(c: Config): ConfigOps =
+    new ConfigOps(c)
+}
+
+object ConfigSyntax extends ConfigSyntax
+
+final class ConfigOps(val c: Config) extends AnyVal {
+  private def unsafe[A](e: String \/ A): A =
+    e.fold(s => sys.error(s), identity)
+
+
+  def extractOption[A](k: ItemKey)(implicit ev: ClassTag[A]): String \/ Option[A] =
+    Option(c.getItemValue(k)).fold(Option.empty[A].right[String]) { a =>
+      \/.fromTryCatchNonFatal {
+        Some(ev.runtimeClass.cast(a).asInstanceOf[A])
+      }.leftMap(_ => s"Expected '$k' to have type ${ev.runtimeClass.getName}")
+    }
+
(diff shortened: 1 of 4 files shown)


bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/config2/config.scala
```diff
@@ -0,0 +1,70 @@
...
+/** A bit of syntax to simplify working with Config.
+  */
+trait ConfigSyntax {
+  implicit final def configSyntax(c: Config): ConfigOps =
+    new ConfigOps(c)
+}
+
+object ConfigSyntax extends ConfigSyntax
+
+final class ConfigOps(val c: Config) extends AnyVal {
```
cquiroz: Don't we have a utility like this already on `itc`? It seems vaguely familiar

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/config2/config.scala
```diff
@@ -0,0 +1,70 @@
...
+  private def unsafe[A](e: String \/ A): A =
+    e.fold(s => sys.error(s), identity)
+
+
+  def extractOption[A](k: ItemKey)(implicit ev: ClassTag[A]): String \/ Option[A] =
+    Option(c.getItemValue(k)).fold(Option.empty[A].right[String]) { a =>
+      \/.fromTryCatchNonFatal {
+        Some(ev.runtimeClass.cast(a).asInstanceOf[A])
+      }.leftMap(_ => s"Expected '$k' to have type ${ev.runtimeClass.getName}")
+    }
```
swalker2m: There's a similar `ConfigExtractor` in the ITC code, but we really need this type of thing in multiple places so having a version in `pot` makes sense I believe.  Ideally `ConfigExtractor` would be updated to delegate to this version or something.  What is the seqexec doing to get typed values out of the sequence?

There's probably a correct way to do the casting, and I'd be curious to know what it is.  After searching a bit I couldn't come up with anything better than what Florian did in `ConfigExtractor`.

cquiroz: i think we took Florian's version and copied it but it has evolved ever since. We could use your version to reduce duplication

swalker2m:
@tpolecat I agree it isn't insanely great.  I'd be quite happy to pass it to you if you've got a better idea for how to accomplish this in the context of the old codebase.

tpolecat:
Oh, no, I think it looks great. I'm just sad that you're having to deal with the old code.

PR #1476, opened 2018-04-20
REL-2660: ODB browser now writes HorizonsDesignation.show string out.

sraaphorst:
Previously, the ODB browser would push out the query string, which did not provide as much information. Instead, now, the ODB browser populates the HORIZONS information for queries with the object name encoded by `HorizonsDesignation.show`, which is then converted to a query by LTTS to be consistent with OCS queries.


cquiroz:
Where does the `browser-api...jar` comes from?

cquiroz:
Where does the `browser-api...jar` comes from?

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
```diff
@@ -81,7 +81,7 @@ class LchQueryFunctor(queryType: LchQueryFunctor.QueryType,
             n <- target.getNonSiderealTarget
             h <- n.horizonsDesignation
           } {
-            setHorizonsObjectId(h.queryString)
```
swalker2m: The change seems trivial enough but I don't know how this is used.  Presumably something will parse the formatted horizons designation and use it to query horizons?  Is that done?

sraaphorst:
@cquiroz The `browser-api...jar` comes from LTTS. It is created by XSD files. The ODB browser in the OCS creates the XML output by populating these, which the LTTS uses.

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
sraaphorst: Yes, we have the `HorizonsDesignation` class, which will parse the output of `show` using `read` into a query string that can be used to uniquely identify a Horizons object.

sraaphorst: (And that query string will be used in a query URL to Horizons to get the necessary data.)

sraaphorst:
On the LTTS side of things, we now read the Horizons ID field as:

`final Option<HorizonsDesignation> hd = ImOption.apply(t.getHorizonsObjectId()).flatMap(h -> ImOption.fromScalaOpt(HorizonsDesignation.read(h)));`

tpolecat:
It would be nice to have a test that demonstrates the problem you're addressing.

tpolecat:
But if you're confident and the fix will be verified during the test release then this is probably ok.

sraaphorst:
I'm not sure the best way to test this short of the local tests I've done, since it works across apps... the LTTS reads this info and constructs Horizon query URLs.

Frankly, upon reviewing this, I was initially confused about REL-2660, since it seemed like things should work. Then I realized that I think I indirectly fixed REL-2660 when I was fixing REL-3128; however, using `show` and `read` instead of the `queryString` directly seems like a better idea to me, just in case the `queryString` changes formats (as Horizons seems to enjoy mixing things up on us).

PR #1478, opened 2018-04-22
REL-1577: Revert Action

swalker2m:
Adds a "revert" option to VCS.  "Revert" isn't exactly what it does but is a close enough approximation. Revert will fetch the current version of a program from the remote database and use it to replace the local version, wiping out any changes that may have been made.  It's like a checkout but for programs that already exist.  Of course, if updates have been stored remotely since the user last did a sync, the "revert" pulls those in as well.

Here is the new menu item, which is only active for programs that were fetched remotely and have been edited:
revertchanges

Here is the dialog that it displays for confirmation:
confirmrevert

### DIFF
--- a/bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs2/VcsServer.scala
+++ b/bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs2/VcsServer.scala
@@ -100,6 +100,14 @@ class VcsServer(odb: IDBDatabaseService) { vs =>
     }
   }
 
+  /** Replaces the given program in the database. */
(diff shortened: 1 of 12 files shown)


bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs2/VcsServer.scala
```diff
@@ -100,6 +100,23 @@ class VcsServer(odb: IDBDatabaseService) { vs =>
     }
   }
 
+  /** Replaces the program in the database whose key and id match program the
+    * given program with program p.
```
sraaphorst: Minor issue, but we can ignore it, I suppose:

"match program the given program"

PR #1480, opened 2018-04-27
REL-1502: multi-select in open dialog

swalker2m:
Updates the OT Open dialog to allow multi-interval selection, changing the list selection model from `SINGLE_SELECTION` to `MULTIPLE_INTERVAL_SELECTION`.

It also enforces a total storage size limit and prevents checking out programs that would take the total storage over the size limit.  This wasn't explicitly requested but I worried about the consequences of attempting to check out the entire ODB.

The previous single select model would generate distinct warnings when deleting programs depending upon whether the program is not present in the remote database, has changes that haven't been committed, or is in sync.  Updating this for multi-select required a bit of simplification but nevertheless is the grimmest part of the PR.

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/viewer/open/OpenDialog.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/viewer/open/OpenDialog.scala
@@ -8,6 +8,7 @@ import edu.gemini.shared.util.VersionComparison
 import edu.gemini.shared.util.VersionComparison.{Conflicting, Newer}
 import edu.gemini.shared.util.immutable.ApplyOp
 import edu.gemini.sp.vcs.reg.VcsRegistrar
+import edu.gemini.sp.vcs2.TryVcs
 import edu.gemini.sp.vcs2.VcsAction._
 import edu.gemini.sp.vcs2.VcsFailure
 import edu.gemini.spModel.core.{Peer, SPProgramID}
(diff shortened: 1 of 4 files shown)


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/viewer/open/OpenDialog.scala
```diff
@@ -110,21 +112,27 @@ class OpenDialog private(db: IDBDatabaseService, auth: KeyChain, vcs: VcsRegistr
     override def apply(t: Option[(Peer, Key)]): Unit = Contents.ProgTable.refresh(full = true)
   })
 
-  var selection: Option[ISPProgram] = None
+  var selection: List[ISPProgram] = Nil
   val filter = new DBProgramChooserFilter(DBProgramChooserFilter.Mode.localAndRemote)
 
+  val MaxM = 50
+  val MaxK = MaxM * 1000
```
sraaphorst: Why 1000 instead of 1024, out of curiosity?

swalker2m: For me, it was continuity with what was there before.  It does seem like it should be 1024.  In the end it doesn't matter too much because we are using the size of the serialized file to give us an idea of the size of how much memory the program takes when loaded.  That's pretty sketchy anyway and the 50M maximum is only a guess for when the OT might start getting bogged down.

PR #1482, opened 2018-04-30
REL-3421: Prevent ITC OutOfMemoryError with oversized user-supplied SEDs

andrewwstephens:
If a user supplies a SED which covers too broad a range in wavelength the ITC will resample it from start to finish and end up with an array which is too large for memory and it will silently crash, providing no feedback to the user about what is wrong.  The SED that was submitted by a user extends from 944 - 2,000,000 nm and the resampled SED has 99,952,807 points.

The solution implemented here is to extract only the wavelength range required by the instrument configuration, ignoring the unnecessary ends.  This required adding yet another implementation of DefaultSampledSpectrum which takes the desired wavelength range and the target redshift.

During testing, I noticed that some redshifts (e.g. 0.3) would truncate the input SED by one element when the redshift was applied.  Thus there is a +2 in the calculation of the number of points to compensate for this and allow for any integer truncation.

### DIFF
--- a/bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/base/DefaultSampledSpectrum.java
+++ b/bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/base/DefaultSampledSpectrum.java
@@ -9,10 +9,10 @@ package edu.gemini.itc.base;
  * This interface provides functionality for defining and manipulating a
  * SampledSpectrum.
  * Units are not stored for either axis so the client must know from context.
- * The SampledSpectrum plays the rold of Element in a visitor pattern.
+ * The SampledSpectrum plays the role of Element in a visitor pattern.
  * This pattern is used to separate operations from the SampledSpectrum
  * elements.
- * The SampledSpectrum plays the rold of Element in a visitor pattern.
+ * The SampledSpectrum plays the role of Element in a visitor pattern.
  * This pattern is used to separate operations from the elements.
  * Because of this separation, Concrete Elements must offer enough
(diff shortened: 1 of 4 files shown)


bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/base/DefaultSampledSpectrum.java
```diff
@@ -54,6 +53,36 @@ public DefaultSampledSpectrum(ArraySpectrum sp, double xInterval) {
...
+    public DefaultSampledSpectrum(ArraySpectrum sp, double xStart, double xEnd, double xInterval, double z) {
+
+        if ( (1+z) * sp.getStart() > xStart || (1+z) * sp.getEnd() < xEnd ) {
+            throw new IllegalArgumentException(
+                    String.format("Redshifted SED (%.1f - %.1f nm) does not cover range of instrument configuration (%.1f - %.1f nm).",
+                            (1+z)*sp.getStart(), (1+z)*sp.getEnd(), xStart, xEnd));
+        }
+
+        // The SED will be redshifted later so resample over range / (1+z):
+        xStart /= 1+z;
```
cquiroz: Should we check for z begin -1?

andrewwstephens: That's a good question.  Large negative redshifts are not terribly physical (it implies movement *toward* us) so I'm not too worried if it fails.  Let me know if you think its worthwhile; it would be easy to throw an exception.

PR #1483, opened 2018-04-30
REL-1577: Add Revert button to OpenDialog

swalker2m:
Adds a "Revert" button to the "Open" dialog.  The button is active if you select at least one program that exists in a remote database and that has pending local changes.  If pressed it will discard the local changes and fetch the remote version of the program.

As usual, I feel a bit sheepish about this PR.  The "Open" dialog has been abused a bit already and this doesn't improve the situation.  Ideally the "Open" dialog would be refactored into a generic program chooser with the ability to checkout remote programs.  It would be configured for distinct purposes: opening programs, deleting programs, or reverting changes.  That would hide features that don't apply to the other modes and simplify the UI.  This would be a lot more work though and isn't what we agreed to do in REL-1577.

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/viewer/open/OpenDialog.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/viewer/open/OpenDialog.scala
@@ -9,8 +9,8 @@ import edu.gemini.shared.util.VersionComparison.{Conflicting, Newer}
 import edu.gemini.shared.util.immutable.ApplyOp
 import edu.gemini.sp.vcs.reg.VcsRegistrar
 import edu.gemini.sp.vcs2.TryVcs
+import edu.gemini.sp.vcs2._
 import edu.gemini.sp.vcs2.VcsAction._
-import edu.gemini.sp.vcs2.VcsFailure
 import edu.gemini.spModel.core.{Peer, SPProgramID}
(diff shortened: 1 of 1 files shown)


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/viewer/open/OpenDialog.scala
```diff
@@ -202,6 +216,55 @@ class OpenDialog private(db: IDBDatabaseService, auth: KeyChain, vcs: VcsRegistr
...
+              Contents.ProgTable.refresh(full = false)
+              updateStorage()
+            }
+            handleFailure(fs)
+        }
+      }
+
+      if (total > 0) {
+        val confirm = Dialog.showConfirmation(dialog.Contents, warning, "Confirm Revert", Dialog.Options.YesNo, Dialog.Message.Warning)
+        if (confirm == Dialog.Result.Ok) {
```
sraaphorst: It feels weird at this point seeing `==` in Scala code, but I guess trying to define `Eq` on `scala.Enumeration.Value` would be overkill and still not type-safe.

swalker2m: Yeah, I just figured it wasn't worth the trouble especially since this code is supposedly slated for retirement in the next few years.

PR #1485, opened 2018-05-03
REL-397 add marker when higher-scoring obs is available

tpolecat:
This adds a QC-only warning to a scheduled visit if there exists a higher-scoring unscheduled visit. I'm not convinced that this is going to be useful because the score isn't all that trustworthy, but I think that's another issue to deal with.

### DIFF
--- a/bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
+++ b/bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
@@ -1,10 +1,11 @@
 package edu.gemini.qpt.core.listeners;
 
 import java.beans.PropertyChangeEvent;
-import java.util.List;
+import java.util.*;
+import java.util.function.Predicate;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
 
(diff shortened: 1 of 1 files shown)


tpolecat:
Here is an example. 
image

tpolecat:
The hard part was doing the indentation with tabs to match the original source file.

cquiroz:
Wouldn't you rather convert the tabs to spaces? Regardless, this looks good

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
```diff
@@ -238,3 +243,51 @@ protected MarkerManager getMarkerManager(Variant t) {
...
+		// observations with a better score than `a.getObs`
+		final Predicate<Obs> higherScoring = otherObs -> {
+			final double myScore    = v.getScore(a.getObs());
+			final double otherScore = v.getScore(otherObs);
+			return otherScore > myScore;
+		};
+
+		// all of the above, ordered by cost to compute
+		final Predicate<Obs> all =
+			unscheduled.and(nearby).and(higherScoring);
```
sraaphorst: I did not know about `Predicate.and` and `Predicate.or`. Neat. (Well, by Java standards.)

PR #1487, opened 2018-05-04
REL-3279: test case for full n&s exposure time

swalker2m:
This PR follows up on #1475, adding a test case that was missing.


swalker2m:
> Curious though, are these too twitchy to use property-based tests?

Actually that probably would have been better in the end.  I just followed the path of least resistance and started with an existing sequence test and modified it to work for N&S values.

tpolecat:
Makes sense. Anyway 👍

PR #1488, opened 2018-05-04
Feature/REL-3222 - Update GSAOI ITC web form

andrewwstephens:
This PR includes a minor update to the GSAOI ITC web form, removing some obsolete text after the implementation of the automatic Strehl lookup, and fixing the "(more info)" links which were pointing to more info about NIRI.  I also fixed a few background links which were still using http instead of https (per REL-3332).


andrewwstephens:
Please do.  Thanks!

PR #1501, opened 2018-05-15
REL-1714-C Report intersection of constraints

jdnavarro:
This is yet another iteration on the new constraint messages for the QPT.

This is still WIP, but I'm sending it anyway to make sure I'm on the right path. This is what I was planning to do next.

- [x] Restore error messages for altitude constraints. Get rid of the `Option` code for altitude constraints.
- [x] Don't show message when intersection spans the whole night.
- [x] Figure out how to deal with the timezone.

Notice that now we are reporting the message even when there is an error. We think this is alright since 1. there are not as many messages as when we were reporting messages for each constraint, and 2. whenever there is an error we expect the user to correct it right away.

### DIFF
--- a/bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
+++ b/bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
@@ -3,8 +3,6 @@ package edu.gemini.qpt.core.listeners;
 import java.beans.PropertyChangeEvent;
 import java.util.*;
 import java.util.function.*;
-import java.time.Instant;
-import java.time.format.DateTimeFormatter;
 import java.util.TimeZone;
 import java.util.logging.Logger;
 import java.util.stream.Collectors;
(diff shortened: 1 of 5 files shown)


bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
```diff
@@ -127,7 +109,28 @@ public void propertyChange(PropertyChangeEvent evt) {
...
 
-            // Elevation Constraint
+			// Function to create markers that report solver intervals.
+			final BiFunction<String, String, Option<Marker>> createSolverMarker =
+					(cacheName, prefix) -> {
+						final Map<Obs, Union<Interval>> c = v.getSchedule().getCache(cacheName);
+						final Union<Interval>          ws = c.getOrDefault(a.getObs(), new Union<>());
+						final ImList<Interval>         is = DefaultImList.create(ws.getIntervals()).filter(
+							// Remove intervals that span the whole night.
+							i -> i.getStart() <= v.getSchedule().getStart() && i.getEnd() >= v.getSchedule().getEnd()
```
swalker2m: I believe this will do the exact opposite of the comment and remove intervals that don't span the entire night.

tpolecat:
As I noted in Stride:

> I believe the viewer that cares about the property should register for change events (via `PreferenceManager.addPropertyChangeListener`) and then on update it can call `refresh()` on itself to re-draw. The menu item shouldn’t do anything other than change the preference. Anyone who cares should listen.

jdnavarro:
> I believe the viewer that cares about the property should register for change events ...

I can't figure out what's the property that changes when switching the time zone. There should be one because other boxes than `Problems` get refreshed right after the change.

tpolecat:
Ah, sorry. You can do `TimePreference.BOX.addPropertyChangeListener(...)`.

jdnavarro:
>`TimePreference.BOX.addPropertyChangeListener`

Oh, I was trying to add the listener to the Problems Controller. That must be it.

PR #1502, opened 2018-05-16
REL-3111 Use new HORIZONS lookup logic in PIT

tpolecat:
This makes a number of changes to the PIT's target lookup logic.

- Removes the "add target" text boxes in the obs/target views. That UI was insufficient to capture the information needed to do a lookup (approved by Andy and Bryan).
- Changes the nonsidereal lookups to use the same back-end logic as the OT.
- Updates the synchronous lookup popups, etc., to match the OT.
- Removes the old HORIZONS and NED code, and the target lookup robot which tried to resolve names in the background.

The remaining SIMBAD code is over-general and should probably be swapped out for what's used by the OT, but that code depends on POT for some reason and I didn't want to untangle it on this PR. In any case the SIMBAD stuff works ok and is out of scope for this issue.

Probably the only way to verify this PR is to pull my branch and build the PIT locally. Sorry.


tpolecat:
There is some copy/paste here but it's hard to avoid because there's no way to abstract over `HS2` (`EitherT[IO, ...]`) and `Task`. This is a lot nicer in cats.

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/editor/HorizonsLookup.scala
```diff
@@ -0,0 +1,144 @@
...
+            .toArray[Object],
+          null)
+      }.map(_.asInstanceOf[Wrapper].unwrap)
+    }
+
+  /** Pop up a dialog for disambiguating many results. */
+  private def handleManyResults(rs: List[Row[_ <: HorizonsDesignation]]): HS2[Unit] =
+    HS2.delay {
+      case class Wrapper(unwrap: Row[_ <: HorizonsDesignation]) {
+        override def toString = unwrap.name + " - " + unwrap.a.des
```
cquiroz: string interpolation?

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/editor/HorizonsLookup.scala
```diff
@@ -0,0 +1,144 @@
...
+        ephemeris.data.toAscList.map {
+          case (t, cd) => EphemerisElement(cd, None, t) // no magnitude info
+        },
+        CoordinatesEpoch.J_2000
+      )
+    }
+
+  /** Run a thunk on the EDT. */
+  private def onEDT(f: => Unit): HS2[Unit] =
+    HS2.delay(Swing.onEDT(f))
```
cquiroz: nice

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/editor/SimbadLookup.scala
```diff
@@ -0,0 +1,100 @@
...
+      Simbad.find(query)(r => k(r.right))
+    } >>= {
+
+      // Failure cases
+      case Offline     => fail("SIMBAD is offline or unreachable.")
+      case NotFound(d) => fail("No results found.")
+      case Error(t)    => Task.fail(t) // roll into top-level handler
+
+      // Success case will always be Nil on the right for a VOTableCatalog.
+      // This is a bad design but ¯\_(ツ)_/¯
```
cquiroz: Why's that?

tpolecat: The types need to be more precise. It's basically `Ior[Nel[A], Nel[B]]` so there are three cases to consider but they're all folded into one here, with no non-empty constraint. But it works and improving it isn't in the scope of this issue.

PR #1503, opened 2018-05-17
REL-3421: fix GNIRS XD wavelength range

andrewwstephens:
This fixes the import of user-supplied SEDs for GNIRS XD.

### DIFF
--- a/bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/base/SEDFactory.java
+++ b/bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/base/SEDFactory.java
@@ -143,10 +143,22 @@ public final class SEDFactory {
         } else if (sdp.distribution() instanceof UserDefinedSpectrum) {
             final UserDefinedSpectrum userDefined = (UserDefinedSpectrum) sdp.distribution();
             final MagnitudeBand band = sdp.normBand();
-            // The user-supplied SED must cover the range of the instrument configuration and the normalization band.
+            Double inst_start_wave = instrument.getObservingStart();
+            Double inst_end_wave = instrument.getObservingEnd();
+
+            // GNIRS overrides getObservingStart and getObservingEnd for each order,
+            // so use the full GNIRS XD wavelength range:
+            if (instrument instanceof Gnirs) {
+                if (((Gnirs) instrument).XDisp_IsUsed()) {
+                    inst_start_wave = 750.;
+                    inst_end_wave = 2600.;
+                }
+            }
+
+            // The user-supplied SED must cover the range of the instrument configuration AND the normalization band:
             temp = getUserSED(userDefined,
-                    Math.min(instrument.getObservingStart(), band.start().toNanometers()),
-                    Math.max(instrument.getObservingEnd(), band.end().toNanometers()),
+                    Math.min(inst_start_wave, band.start().toNanometers()),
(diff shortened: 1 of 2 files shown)


bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/base/SEDFactory.java
```diff
@@ -143,10 +143,22 @@ private static VisitableSampledSpectrum getSED(final SourceDefinition sdp, final
...
-            // The user-supplied SED must cover the range of the instrument configuration and the normalization band.
+            Double inst_start_wave = instrument.getObservingStart();
+            Double inst_end_wave = instrument.getObservingEnd();
+
+            // GNIRS overrides getObservingStart and getObservingEnd for each order,
+            // so use the full GNIRS XD wavelength range:
+            if (instrument instanceof Gnirs) {
+                if (((Gnirs) instrument).XDisp_IsUsed()) {
+                    inst_start_wave = 750.;
+                    inst_end_wave = 2600.;
```
tpolecat: N.B. the `123.` syntax for doubles is legal Java, but it's weird. `123.0` is more idiomatic.

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/base/SEDFactory.java
```diff
@@ -143,10 +143,22 @@ private static VisitableSampledSpectrum getSED(final SourceDefinition sdp, final
         } else if (sdp.distribution() instanceof UserDefinedSpectrum) {
             final UserDefinedSpectrum userDefined = (UserDefinedSpectrum) sdp.distribution();
             final MagnitudeBand band = sdp.normBand();
-            // The user-supplied SED must cover the range of the instrument configuration and the normalization band.
+            Double inst_start_wave = instrument.getObservingStart();
```
cquiroz: I've the impression that you could do this overriding `getObservingStart` in `Gnirs`

tpolecat: Good point. It's not clear whether this special case always applies to GNIRS or only for this particular computation.

andrewwstephens: The context is that previously we parsed and imported the entire user-supplied SED, but if the uploaded file covered a silly range it would trigger an OutOfMemoryError without giving any feedback to the user.  The implemented solution is to only import the data that are needed for the calculation.  This requires knowing the full wavelength range of the instrument configuration.  That is normally returned by instrument.getObservingStart/End.  However, some instruments override that method to use different algorithms for determining the range of wavelengths they are sensitive to (and that's fine), but GNIRS overrides those methods to return the wavelength range of each cross-dispersed order.  That is correct for most uses, but in this particular case, we need the *full* range over all orders.  This is defined by a single component (the XD filter), so I'm fine hard-coding it here if everybody else is. :-)

tpolecat: 👍  I think what you have here makes sense.

cquiroz: 👍

PR #1511, opened 2018-05-24
REL-1608: Time Accounting Updates

swalker2m:
> In the beginning, Developer created the science program model and the ODB. Now the model was formless and overly generic, and sequence execution time was not tracked, and the spirit of Developer moved upon its API.
>
> And Developer said, "Let there be time accounting," and there was time accounting.  Developer saw the time accounting, that it was bad, and He left it that way anyway.

This PR brings the first update to time accounting since it was initially introduced about 20 years ago.  The primary goal is to no longer charge for visits unless they produce at least one dataset that isn't marked `FAIL` or `USABLE`.  Along the way we also needed to come up with a way to update the algorithm without changing past results, _even if they were computed incorrectly_.

A bit of background:

* The term "visit" comes up a lot in time accounting.  It corresponds to a sequence of events and associated datasets produced by an attempt to observe.  An observation may require several visits, possibly on different nights, to complete.  Time accounting is performed per visit and summed to produce results for the entire observation.

* Time accounting calculations are performed on demand by examining the events in a visit.  Nothing is stored so an update to the algorithm can impact past results.

* The goal of time accounting is to determine how much of the total time associated with executing an observation should be charged to the program, how much should be charged to the partner, and how much is not charged.

* In general we don't charge for anything that happens during the day and we don't charge for datasets that don't pass QA.

* Some types of observes correspond to calibrations which are charged to the partner as a whole instead of to the program itself.  This is determined by the observe's `ChargeClass`.

* A staff user can apply manual corrections to the calculations if they determine that the default rules are not adequate for a given situation.

When I started looking though the old code I saw it could be greatly simplified using `Interval` and `Union` from the QPT (or rather the copied versions of them in `skycalc`).  Unfortunately a simplified algorithm revealed a number of bugs in the original code so I decided to leave the "primordial" time accounting algorithm alone.  Again, we cannot fix those bugs since it would change past results that have possibly had manual corrections applied.  Instead I added a new 2018B+ version that is only used when the first event of a visit happens in 2018B.


bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/obsrecord/PrimordialVisitCalculator.java
```diff
@@ -0,0 +1,253 @@
...
+ *
+ * 3. If the first event should happen to be a start dataset event then the
+ * first dataset isn't charged.
+ *
+ * Nevertheless, we cannot change time accounting for past observations so none
+ * of these bugs will be fixed in the PrimordialVisitCalculator below.  The
+ * code is essentially unchanged but instead simply lifted from `PrivateVisit`
+ * and given a calc method with fewer dependencies.
+ */
+enum PrimordialVisitCalculator {
```
swalker2m: I'm ashamed, but I have to admit it was (a younger version of) me who wrote this code.  I'd hoped to delete it and replace it with a simplified version in `VisitCalculator` but I found so many bugs in it that we cannot fix without changing past results that I decided to leave it alone.  Keeping in mind that this isn't really new code, it might be best to skip it and focus on `VisitCalculator` instead.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/obsrecord/PrivateVisit.java
```diff
@@ -88,191 +88,8 @@ Site divineSite() {
...
-        while (curNight.getStartTime() < e2.obsNight.getStartTime()) {
-            Night darkNight = TwilightBoundedNight.forObservingNight(TwilightBoundType.NAUTICAL, curNight);
-            nighttime += darkNight.getTotalTime();
-            curNight   = new ObservingNight(site, curNight.getEndTime());
-        }
-        inv.nightTime = nighttime;
-    }
-
-    VisitTimes getTimeCharges(ObsQaRecord qa,
-                              ConfigStore store) {
```
swalker2m: All of this is just the original version that was moved into `PrimordialVisitCalculator`.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/obsrecord/VisitTimes.java
```diff
@@ -57,4 +72,41 @@ ObsTimeCharges getTimeCharges(ChargeClass mainChargeClass) {
...
+        return getTotalTime() - getClassifiedTime(ChargeClass.NONCHARGED);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(_unclassifiedTime, _classifiedTimes);
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
swalker2m: I added `equals` and `toString` for testing and debugging.  `hashCode` came along for the ride.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/obsrecord/TimeAccounting.scala
```diff
@@ -0,0 +1,50 @@
...
+
+    // Find the visit calculator that applies to the events.  We use the start
+    // time of the first event for the visit to select the corresponding
+    // calculator that is used for the rest of the sequence regardless of when
+    // it ends.
+    val calculator = for {
+      h <- ve.sorted.headOption
+      s <- h.site
+      c <- VisitCalculator.all.find(_.validAt(s).isBefore(h.instant))
+    } yield c
```
swalker2m: This is where we keep old calculations using the old time accounting rules and allow new rules to be applied for new sequences.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/obsrecord/VisitCalculator.scala
```diff
@@ -0,0 +1,148 @@
...
+   * result is that visits without a passing dataset are not charged at all.  It
+   * also simplifies and fixes a number of bugs with the original algorithm:
+   *
+   * <ul>
+   *   <li>Charges for the portion of a dataset after start and before overlap</li>
+   *   <li>Handles start/end dataset pairs with intervening events correctly</li>
+   *   <li>Doesn't drop time if the first event is start dataset</li>
+   * </ul>
+   */
+  case object Update2018B extends VisitCalculator {
```
swalker2m: Actually, now that I think about it.  2018B is coming right up.  This won't go into effect until 2019A.  Testing will be difficult.  I may need to set the start time to 2018B anyway so they can evaluate the impact it would have using 2018B data during the November testing period.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/obsrecord/PrimordialVisitCalculator.java
cquiroz: I'll trust you and skip it 😉

tpolecat:
> When I started looking though the old code I saw it could be greatly simplified using `Interval` and `Union` from the QPT

FYI spire has `Interval` and `IntervalSet` which we should use in ocs3/gem when this comes up.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/obsrecord/TimeAccounting.scala
tpolecat: It seems like we might want to return something like `Either[Reason, VisitTimes]` so we can give a reason why we can't compute it (missing site, no applicable calculator), rather than returning an empty `VisitTimes` in that case. I kind of assume it's this way because that's how it has always worked, in which case ¯\\\_(ツ)\_/¯

PR #1515, opened 2018-05-30
REL-2822: bulk scheduling block update

swalker2m:
This task adds a menu item to the OT's "Edit" menu, allowing the user to update the scheduling block for multiple observations in a single operation.  I believe the intention is to use this with related observations, like an acquisition and the associated science observation(s) but it will work with any selected observations.

menu

The difficulty was handling the ephemeris update after the new scheduling block is selected.  I found a lot of work that went into doing this for the single-observation case in `ParallacticAngleControls` and `EphemerisUpdater`.  I adapted that work, extracting it from the parallactic angle stuff and making it work with multiple observations.

I also renamed the `ParallacticAngleDialog` to `SchedulingBlockDialog` and added a bit of code to support this case.  In particular since the menu item just brings up the calendar and time box, I also included a "Set To:" menu in the bottom left corner since the REL task specifically mentions needing that.

dialog


bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/syntax/sp/Node.scala
```diff
@@ -0,0 +1,28 @@
+package edu.gemini.spModel.syntax.sp
+
+import edu.gemini.pot.sp.ISPNode
+
+import scalaz._
+import Scalaz._
+import scalaz.effect.IO
+
+
+final class NodeOps(val self: ISPNode) {
```
swalker2m: This was just pulled out of `ParallacticAngleControls` into the syntax package.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
```diff
@@ -334,6 +336,18 @@ private boolean isBagsUpdate(final ISPObsComponent oldObsComp, final TargetObsCo
...
+        // Okay if the scheduling block changed then it isn't just a bags update.
+        // Of course we don't know whether the scheduling block changed because
+        // that is kept in the observation node itself.  We can however look at
+        // the ephemeris for each nonsidereal target.
+        // Sorry, this is a terrible hack to make bulk scheduling block updates
+        // be considered enough of a change to refresh the target environment
+        // editor when it is selected..
+        if (!getNonSiderealTargets(oldTOC).equals(getNonSiderealTargets(newTOC))) {
+            return false;
+        }
```
swalker2m: This is the worst part of the PR.  The issue was that the OT would be fooled into not updating the target editor for the last target component that was selected.  It would show the old guide star which is likely no longer workable with the new coordinates after a scheduling block change.  Fixing it required clicking on some other target node and then coming back to the same node, which was confusing.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/EphemerisUpdater.scala
```diff
@@ -68,63 +68,122 @@ object EphemerisUpdater {
...
+        uc.updateAction(m).withResultLogging(Log).run.map(_.leftMap(_ => ()).merge)
+      }
+
     val ui = UI(c)
-    refreshEphemerides(obsN, start, ui).withResultLogging(Log).run.map {
-      case -\/(e) => IO.ioUnit
-      case \/-(a) => (a.withResultLogging(Log) ensuring ui.hide).run.void
+    (bulkLookup(obsNs, start, ui).withResultLogging(Log) ensuring ui.hide).run.map {
+      case -\/(_) => (obsN: ISPObservation) => IO.ioUnit
+      case \/-(m) => setEphemerides(_, m)
```
swalker2m: Here the behavior differs a bit from before in that the glass pane is dismissed after the lengthy horizons lookup but before the actual observation nodes are updated.  The observation node updates are quick though so I deemed this to be an unimportant detail.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/parallacticangle/ParallacticAngleControls.scala
```diff
@@ -237,54 +243,17 @@ class ParallacticAngleControls(isPaUi: Boolean) extends GridBagPanel with Publis
...
-          val up: IO[Unit] =
-            for {
-              _ <- updateSchedBlock
-              _ <- completion
-            } yield ()
-
-          // Here we bracket the update to disable/enable events and grab/release the program lock
-          ispObs.silentAndLocked(up)
-
+    } SchedulingBlockUpdate.runWithCallback(
```
swalker2m: I took the code that followed the `for` in `ParallacticAngleControls` and moved it to a `SchedulingBlockUpdate` so it could be reused by the bulk update menu item.  It also needed to work with multiple observations of course.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/schedulingBlock/SchedulingBlockDialog.scala
```diff
@@ -0,0 +1,391 @@
...
+  observation: ISPObservation,
+  site:        Option[Site],
+  mode:        Mode
+) extends Dialog {
+
+  private object state {
+    var schedulingBlock: Option[SchedulingBlock] =
+      observation.getDataObject.asInstanceOf[SPObservation].getSchedulingBlock.asScalaOpt
+
+    var timeZone: TimeZone =
```
cquiroz: Any reason to use the old `TimeZone` rather than the new one from `java.time`?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/schedulingBlock/SchedulingBlockDialog.scala
```diff
@@ -0,0 +1,391 @@
...
+
+      object setToPanel extends FlowPanel {
+
+        hGap = 0
+
+        object button extends RadioButton("Set To:")
+
+        // Set the number of minutes of duration, converting from ms.
+        val initialMs     = state.schedulingBlock.flatMap(_.duration.toOption).getOrElse(0L)
+        val durationField = new NumberField(Some(initialMs / 60000.0), allowEmpty = false) {
```
cquiroz: Maybe this conversion can be done with `java.time` apis?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/EphemerisUpdater.scala
swalker2m: I should say that since usually they'll be setting the scheduling block for multiple observations of the *same* nonsidereal target i thought it was important not to make them pay the price of waiting for the ephemeris lookup multiple times.  So the `bulkLookup` extracts all the nonsidereal target (horizons id, site) pairs and does the lookup just once for each one.

PR #1517, opened 2018-06-01
REL-3431: Commented out the culprits in HS2Spec.

sraaphorst:
This comments out the two test cases in `HS2Spec` that are failing, thus causing all Travis builds to fail and preventing PR merging.

@tpolecat explains why these cases are failing and why it is not a matter of concern:

"We handle that asteroid format but no longer have any way to get horizons to generate it because all the applicable targets have been reclassified as comets, as far as I can tell."


sraaphorst:
I'm good either way. Let's delete them.

tpolecat:
I would prefer to at least leave a comment saying "there's no test for format 6 because all known examples have been reclassified as comets".

sraaphorst:
Alright: comment added. 👍

tpolecat:
👍

PR #1518, opened 2018-06-05
REL-3397: Add GHOST asterisms to target table and make them editable

sraaphorst:
This very long PR comprises the code to:

1. Allow sky positions to be displayed and selected in the target table (which required a fundamental reworking of rows and selections as they were previously limited to `SPTarget` and `IndexedGuideGroup`);

2. Allow all GHOST asterisms to be displayed in the target table;

3. Add a checkbox to the target table UI to couple / decouple the GHOST base position from the IFU viewing a target;

4. Create an editor for sky positions that is a generalization of the coordinate editor for targets;

5. Change GHOST asterisms to use `SPCoordinates` for their sky positions, as we need these to be immutable, so now they are just a fairly trivial wrapper around `Coordinates`;

6. Make a `SPSkyObject` superclass of `SPTarget` and `SPCoordinates` so that I could generalize a number of editors and constructs;

7. Make the TPE draw all sky positions (one bug remaining: see below).

There are two outstanding issues of which I'm aware:

1. If the GHOST base position is the same as a target position, you can only select the base position in the table: this is because the selection mechanism compares by reference. If we want to change this, we will have to give some thought as to what we want to do.

2. If the dual target mode is selected for GHOST and the base position is set to be an interpolation between the two science targets, for some reason, this is not yet being drawn on the TPE.

I will post some screencaps / possible video tomorrow: right now, I want to see if Travis has any complaints and get this out there as promised before the end of the day.


sraaphorst:
I should point out that we're not even sure where we want to merge this yet: into the release branch or into develop.

I'm quite glad to see that I haven't broken any test cases!

swalker2m:
> Change GHOST asterisms to use SPCoordinates for their sky positions, as we need these to be immutable, so now they are just a fairly trivial wrapper around Coordinates

`SPCoordinates` are mutable right?  That is `SPCoordinates` are to `Coordinates` as `SPTarget` is to `Target`.

sraaphorst:
@swalker2m Yes, exactly that. `SPCoordinates` is just a mutable wrapper around a `var` of `Coordinates` in the same way that `SPTarget` is just a mutable wrapper around a `var` of `Target`.

I based `SPCoordinates` specifically off of `SPTarget` and then extracted common methods to `SPSkyObject` so that we could use a lot of the classes using `SPTarget` to handle both `SPTarget` and `SPCoordinates` via parameterization `T <: SPSkyObject`, which has worked well and saved a lot of code duplication.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostAsterism.scala
```diff
@@ -178,13 +182,13 @@ object GhostAsterism {
...
   }
 
-  case class SingleTarget(target: GhostTarget, override val base: Option[Coordinates]) extends StandardResolution
-  case class DualTarget(target1: GhostTarget, target2: GhostTarget, override val base: Option[Coordinates]) extends StandardResolution
-  case class TargetPlusSky(target: GhostTarget, sky: Coordinates, override val base: Option[Coordinates]) extends StandardResolution
-  case class SkyPlusTarget(sky: Coordinates, target: GhostTarget, override val base: Option[Coordinates]) extends StandardResolution
+  case class SingleTarget(target: GhostTarget, override val base: Option[SPCoordinates]) extends StandardResolution
+  case class DualTarget(target1: GhostTarget, target2: GhostTarget, override val base: Option[SPCoordinates]) extends StandardResolution
+  case class TargetPlusSky(target: GhostTarget, sky: SPCoordinates, override val base: Option[SPCoordinates]) extends StandardResolution
+  case class SkyPlusTarget(sky: SPCoordinates, target: GhostTarget, override val base: Option[SPCoordinates]) extends StandardResolution
```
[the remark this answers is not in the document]
swalker2m: The good news is also the bad news: `Asterism` was always mutable:

```
  // N.B. most members must be defs because `t` is mutable.
  final case class Single(t: SPTarget) extends Asterism {
    override def allSpTargets = NonEmptyList(t) // def because Nel isn't serializable
    override def allTargets = NonEmptyList(t.getTarget)
    override def basePosition(time: Option[Instant]) = t.getCoordinates(time.map(_.toEpochMilli))
    override def copyWithClonedTargets() = Single(t.clone)
    override def basePositionProperMotion = Target.pm.get(t.getTarget)
    override def asterismType: AsterismType = AsterismType.Single
  }
```
also `final case class GhostTarget(spTarget: SPTarget ...`

@sraaphorst will have all this fresh in his mind, but I believe it's impossible to get the OT and TPE working without rewriting them otherwise.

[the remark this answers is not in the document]
tpolecat: Oh ok. I didn't look closely enough. All is "well"  👍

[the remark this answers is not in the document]
sraaphorst: As Shane said, `GhostAsterism` was always mutable since it has one or more `GhostTarget`s, each of which contain an `SPTarget`.

With regards to the `SPCoordinates`, the problem here was that we had two choices:

1. Use immutable `Coordinates` to represent sky positions, but then every time the RA / Dec of one of these changed in any way, we'd have to create a new `Asterism` and set it in the `TargetEnvironment`. This would have been a nightmare in terms of the target table and the TPE, since they rely on using `WatchablePos` with mutability to coordinate everything.

2. The easier alternative that fit in with how we're handling `SPTarget`s was to make a mutable `SPCoordinates` wrapper around `Coordinates` (kind of like how `SPTarget` wraps the three types of `Target`s). Then `SPCoordinates` could fit exactly into the TPE and target table using the existing `WatchablePos` / `TelescopePosWatcher` infrastructure.

By extracting the common `SPSkyObject` from `SPTarget` and `SPCoordinates`, I was able to reuse a lot of existing code that formerly would only work with `SPTarget` by parameterizing it `[T <: SPSkyObject]`.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostAsterism.scala
```diff
@@ -123,11 +120,18 @@ object GhostAsterism {
       case SkyPlusTarget(_,t,_) => NonEmptyList(t.spTarget)
     }
 
+    override def allSpCoordinates: List[SPCoordinates] = this match {
+      case SingleTarget(_,b)    => b.toList
+      case DualTarget(_,_,b)    => b.toList // The base will be handled individually
+      case TargetPlusSky(_,s,b) => b.toList ++ List(s)
```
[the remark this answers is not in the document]
sraaphorst: Thanks for looking over this and for the suggestions! I'll make your recommended changes.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/TargetEnvironmentDiff.java
```diff
@@ -62,22 +62,42 @@ private static TargetEnvironmentDiff primaryGuideGroupExtraction(TargetEnvironme
...
+        }
+
+        if (oldEnv != null) {
+            newSet.removeAll(oldTargets.toList());
+            newSetC.removeAll(oldCoords.toList());
+        }
+        if (newEnv != null) {
+            oldSet.removeAll(newTargets.toList());
+            oldSetC.removeAll(newCoords.toList());
+        }
```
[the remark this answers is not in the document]
sraaphorst: Excellent. I can abstract all four constructions down to one method.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/TargetEnvironment.java
```diff
@@ -232,17 +234,22 @@ public boolean containsTarget(SPTarget target) {
     }
 
     /**
-     * Gets a list of all the guide groups in this environment
+     * Get a list of all the SPCoordinates objects in this environment, in the
+     * order:
```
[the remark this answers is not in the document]
sraaphorst: Looking over the uses of this and of `Asterism.allSPCoordinatesJava()`, I can see that it doesn't matter. I think I thought at some point that it would matter, but it turned out that it was unimportant. I'll fix the comment to reflect this.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostAsterism.scala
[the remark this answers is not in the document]
tpolecat: Ok, I understand. I stopped at this point when reviewing but I'll pick up there after lunch.

bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/general/GeneralRule.java
```diff
@@ -547,9 +547,9 @@ public IP2Problems check(final ObservationElements elements)  {
             final ISPTemplateFolder templateFolderNode = elements.getProgramNode().getTemplateFolder();
             if (templateFolderNode != null) {
 
-                final SPTarget obsTarget = getObsTarget(elements);
-                if (obsTarget != null) {
-
+                // TODO:GHOST Do we have to do anything for SPCoordinates here as well?
+                final Asterism a = getObsAsterism(elements);
+                a.allSpTargetsJava().foreach(obsTarget -> {
```
swalker2m: I'm guessing the answer to your question is no. This rule is about not slipping in new targets that weren't approved and that should be accomplished with the real science targets.

It does bring up the problem of matching this to the P1 model. We don't have the asterism concept in P1 yet so we really can't do this check. It would require two TemplateParameters for each dual target, each with its own conditions and observation time. Since it doesn't make sense maybe we should skip this check for now for anything but single target asterism.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/SPCoordinates.scala
```diff
@@ -0,0 +1,108 @@
...
+    ps.addParamSet(TargetParamSetCodecs.CoordinatesParamSetCodec.encode(CoordinatesName, coordinates))
+    ps
+  }
+
+  /** Re-initialize these SPCoordinates from the given paramset. */
+  def setParamSet(ps: ParamSet): Unit = {
+    if (ps != null) {
+      val tps = ps.getParamSet(CoordinatesName)
+      if (tps != null) {
+        val c = TargetParamSetCodecs.CoordinatesParamSetCodec.decode(tps).toOption.get
```
tpolecat: `.toOption.foreach(setCoordinates)`?

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/SPCoordinates.scala
```diff
@@ -0,0 +1,108 @@
...
+  override def setRaDecDegrees(ra: Double, dec: Double): Unit =
+    Coordinates.fromDegrees(ra, dec).foreach(setCoordinates)
+
+  override def getRaDegrees(time: GOLong): GODouble = {
+    val d: JDouble = coordinates.ra.toDegrees
+    Some(d).asGeminiOpt
+  }
+
+  override def getRaHours(time: GOLong): GODouble = {
+    val d: JDouble = coordinates.ra.toDegrees
```
tpolecat: grim

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/TargetParamSetCodecs.scala
```diff
@@ -16,6 +16,9 @@ object TargetParamSetCodecs {
       .withParam("ra",  Coordinates.ra)
       .withParam("dec", Coordinates.dec)
 
+  implicit val SPCoordinatesParamSetCodec: ParamSetCodec[SPCoordinates] =
+    CoordinatesParamSetCodec.xmap(new SPCoordinates(_), _.getCoordinates)
```
tpolecat: 🔥

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
```diff
@@ -156,26 +161,110 @@ private void updateSchedulingBlockEnabledState(final boolean enabled) {
...
+                if (ghta.base().isDefined()) return new Right<>(ghta.base().get());
+                else return new Left<>(ghta.target().spTarget());
+            case GhostHighResolutionTargetPlusSky:
+                final GhostAsterism.HighResolutionTargetPlusSky ghtsa = (GhostAsterism.HighResolutionTargetPlusSky) a;
+                if (ghtsa.base().isDefined()) return new Right<>(ghtsa.base().get());
+                else return new Left<>(ghtsa.target().spTarget());
+        }
+
+        // Make the compiler shut up.
+        return new Right<>(new SPCoordinates());
```
tpolecat: I prefer to throw an exception when something unpossible happens.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
```diff
@@ -156,26 +161,110 @@ private void updateSchedulingBlockEnabledState(final boolean enabled) {
...
+    }
+
+    /**
+     * Auxiliary method to determine if the current selection is base target.
+     */
+    private boolean selectionIsBaseTarget() {
+        final ISPObsComponent node = getContextTargetObsComp();
+        if (TargetSelection.getIndex(node).forall(i -> i != 0))
+            return false;
```
tpolecat: The thing at index zero is the base?

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
```diff
@@ -222,22 +311,39 @@ private boolean selectionIsGuideTarget() {
...
-        _curSelection = ImOption.apply(t).map(Left::new);
+    private boolean selectionIsCoordinates() {
+        return selectedCoordinates().isDefined();
     }
 
     /**
-     * Auxiliary method to set the selection to the specified group.
+     * Auxiliary method to determine if the base coordinates are explicitly defined.
      */
+    private boolean baseCoordinatesAreDefined() {
```
tpolecat: This seems easy to misunderstand ... `isGhostAsterismWithExplicitlyDefinedBaseCoordinates`?

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
```diff
@@ -893,67 +1078,66 @@ private void setSelectionFromNode() {
...
-                for (GuideGroup group : env.getGroups()) {
-                    for (GuideProbeTargets gt : group) {
-                        if (gt.getTargets().contains(target)) {
-                            group = group.put(gt.update(OptionsList.UpdateOps.append(newTarget)));
-                            duplicated = true;
-                            break;
-                        }
+        final Option<IndexedGuideGroup> iggOpt =
+                TargetSelection.getIndexedGuideGroupForNode(dataObject.getTargetEnvironment(), obsComponent);
+        targetOpt.foreach(target -> {
```
swalker2m: The target vs. group cases used to be in an `if / then/ else` statement.  Now each will be executed if their respective `Option`s are defined.  Is it the case that if one is defined the other definitely won't be?

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
```diff
@@ -971,18 +1155,79 @@ private void pasteSelectedPosition(final ISPObsComponent obsComponent, final Tar
...
             _w.positionTable.updatePrimaryStar();
         }
     };
 
+    private final ActionListener linkedBaseListener = new ActionListener() {
+        private scala.Option<SPCoordinates> coords(final GhostAsterism ga, final boolean linked) {
+                if (linked)
+                    return ImOption.scalaNone();
+                return ImOption.toScalaOpt(
+                        ImOption.fromScalaOpt(ga.basePosition(ImOption.scalaNone()))
```
swalker2m: Is `None` the thing to use for `when`?  Should it use the scheduling block instead?

sraaphorst: Yes, I was passing `None` for `when`. I didn't know we had scheduling block information available at that point, but if we do, I'll pass that in. (Not that it matters: I think it's ignored for sky positions anyway.)

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
sraaphorst: If more than one is defined, then something has definitely gone very wrong in the OT.

I have no opposition to switching to and `if` block if you think it's safer, but realistically, if more of one of these executes, we should bork because the internal state is messed up.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
sraaphorst: 👍

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
sraaphorst: Yes. It will always be the case that the thing at position 0 will be the base. If the base position for GHOST is linked to a science target, then both the base position and the science target are technically the base but we display them both and they both change together.

A checkbox allow us to uncouple / unlink them.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
sraaphorst: Agree. That's verbose, but much clearer.

bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/general/GeneralRule.java
```diff
@@ -582,20 +582,17 @@ public IP2Problems check(final ObservationElements elements)  {
...
-        private SPTarget getObsTarget(final ObservationElements elements)  {
+        private Asterism getObsAsterism(final ObservationElements elements)  {
             final TargetObsComp targetEnv = elements.getTargetObsComp().getOrNull();
             if (targetEnv != null) {
                 final ObsClass obsClass = ObsClassService.lookupObsClass(elements.getObservationNode());
                 if (obsClass == ObsClass.SCIENCE) {
-                    // TODO:ASTERISM: this needs to handle multiple targets … also why only sidereal?
-                    final SPTarget target = targetEnv.getArbitraryTargetFromAsterism();
-                    if (target.isSidereal())
-                        return target;
```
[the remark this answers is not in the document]
sraaphorst: I'll restore the original functionality of only sidereal targets, along with the comment. I can't claim to know the answer so probably best to leave it as it was.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/TargetEnvironmentDiff.java
```diff
@@ -62,22 +55,29 @@ private static TargetEnvironmentDiff primaryGuideGroupExtraction(TargetEnvironme
     private final Collection<SPTarget> removedTargets;
     private final Collection<SPTarget> addedTargets;
 
-    private TargetEnvironmentDiff(TargetEnvironment oldEnv, TargetEnvironment newEnv,
-                                  ImList<SPTarget> oldTargets, ImList<SPTarget> newTargets) {
+    private final Collection<SPCoordinates> removedCoordinates;
+    private final Collection<SPCoordinates> addedCoordinates;
+
+    // Return an ImList<T> containing the elements of list1 \ list2.
```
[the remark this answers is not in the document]
sraaphorst: Ahhh... it's a slash and not a vertical bar: it just looks ambiguous in italics. When I learned set theory, we wrote set difference as `S \ T`, but since this isn't clear, I'll change to the minus symbol.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/obsComp/TargetObsComp.java
```diff
@@ -75,6 +78,7 @@ public Object clone() {
         toc.targetEnv = targetEnv.cloneTargets();
         toc.prop      = toc.new PcePropagator();
         toc.targetEnv.getTargets().foreach(target -> target.addWatcher(toc.prop));
+        toc.targetEnv.getCoordinates().foreach(coords -> coords.addWatcher(toc.prop));
```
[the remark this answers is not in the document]
sraaphorst: The watchers were probably one of the most complicated things to get right.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/obsComp/TargetSelection.java
```diff
@@ -172,24 +250,58 @@ public int getIndex() {
...
-    public static Option<Integer> indexOfGuideGroupByIndex(final TargetEnvironment env,
+    private static Option<Integer> indexOfGuideGroupByIndex(final TargetEnvironment env,
                                                            final int guideGroupIndex) {
         final ImList<GuideGroupSelection> selections = GuideGroupSelection.toGuideGroupSelections(env);
         return selections.getOption(guideGroupIndex).map(GuideGroupSelection::getIndex);
     }
 
+    private static Option<Integer> indexOfCoordinates(final TargetEnvironment env, final SPCoordinates coords) {
+        return Selection.toSelections(env).find(sel -> sel.getCoordinates().exists(coords::equals)).map(sel -> sel.index).orElse(new Some<>(0));
+    }
```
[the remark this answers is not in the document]
sraaphorst: Agreed. I'm amazed it survived this.

Also, the files are so long and complex that IDEA can barely handle navigating them. I could type 10 keystrokes and have to wait a couple seconds for the results to appear. Given it's a Java IDE, that seems insane.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostAsterism.scala
```diff
@@ -194,81 +198,101 @@ object GhostAsterism {
...
-                                  override val base: Option[Coordinates]) extends GhostAsterism {
-
+  sealed abstract class HighResolution(target: GhostTarget) extends GhostAsterism {
     override def allSpTargets: NonEmptyList[SPTarget] =
-      NonEmptyList(ghostTarget.spTarget)
+      NonEmptyList(target.spTarget)
+
+    override def allSpCoordinates: List[SPCoordinates] = this match {
+      case HighResolutionTarget(_,b)          => b.toList
+      case HighResolutionTargetPlusSky(_,s,b) => b.toList ++ List(s)
```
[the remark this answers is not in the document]
sraaphorst: 👍

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostAsterism.scala
```diff
@@ -290,7 +314,17 @@ object GhostAsterism {
...
+
+  def createEmptyHRTargetPlusSkyAsterism: Asterism = {
+    HighResolution.emptyHRTargetPlusSky.copyWithClonedTargets
+  }
+
+  // Names of the IFUs.
+  val SRIFU1: String = "SRIFU1"
+  val SRIFU2: String = "SRIFU2"
+  val HRIFU1: String = "HRIFU1"
+  val HRIFU2: String = "HRIFU2"
```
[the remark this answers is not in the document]
sraaphorst: I think we sufficiently cleared this up in our meeting with Steve, right?

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/SPCoordinates.scala
sraaphorst: 👍

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostAsterism.scala
```diff
@@ -154,10 +158,10 @@ object GhostAsterism {
 
     // Sky coords are immutable, so we don't need to copy them.
```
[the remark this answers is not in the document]
sraaphorst: Oops. Forgot to remove that comment. 😊

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
```diff
@@ -156,26 +161,110 @@ private void updateSchedulingBlockEnabledState(final boolean enabled) {
...
         final TargetEnvironment env = getDataObject().getTargetEnvironment();
-        return selectedTarget().exists((target) -> target == env.getArbitraryTargetFromAsterism());
+        final Asterism a            = env.getAsterism();
+
+        switch (a.asterismType()) {
+            case Single:
+                return new Left<>(((Asterism.Single) a).t());
+            case GhostSingleTarget:
+                final GhostAsterism.SingleTarget gsa = (GhostAsterism.SingleTarget) a;
+                if (gsa.base().isDefined()) return new Right<>(gsa.base().get());
```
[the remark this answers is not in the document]
swalker2m: Maybe we should just commit to fixing any issues that come up for November and put it in `develop`?  Otherwise we'd run into issues with having multiple branches that we have to keep in sync.  I guess that does mean that the Ghost testing run in September comes with a bunch of other unrelated and untested changes though.  @arturog8m what do you think?

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
swalker2m: > realistically, if more of one of these executes, we should bork because the internal state is messed up.
Perhaps we should make that explicit and throw an exception in that case?

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
```diff
@@ -324,7 +327,7 @@ private boolean selectionIsCoordinates() {
     /**
      * Auxiliary method to determine if the base coordinates are explicitly defined.
      */
-    private boolean baseCoordinatesAreDefined() {
+    private boolean isGhostAsterismWithExplicitlyDefinedBaseCoordinates() {
```
swalker2m: Sweet.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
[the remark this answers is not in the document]
sraaphorst: I'd prefer this, if everyone agrees. Given the scope of these changes, trying to merge at a later date is likely to cause serious merge conflicts.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
sraaphorst: Sounds good. I'll change that.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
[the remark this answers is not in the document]
swalker2m: > I'd prefer this, if everyone agrees. Given the scope of these changes, trying to merge at a later date is likely to cause serious merge conflicts.

It needs to go into `develop` regardless so I don't see a reason not to do that now.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
sraaphorst: (Actually, this is a very good idea, because during development, I reached a point where I was creating things incorrectly, and I did have more than one thing set per node.)

I have added a `checkTargetSelectionConsistency` method that checks to make sure 0 or 1 things are selected, and otherwise throws a `RuntimeException`. I call it before any `TargetSelection` access as we have here. (It'll be in my next commit.)

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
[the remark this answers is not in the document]
sraaphorst: The `ImEither` approach is much cleaner. Thanks for that recommendation. This particularly hideous block of code with all the `isDefined()`...`get()` was not sitting well with me.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
```diff
@@ -251,16 +250,34 @@ private boolean selectionIsBaseTarget() {
...
+     * Ensure that there is at most one selection for this node in TargetSelection.
+     */
+    static void checkTargetSelectionConsistency(final TargetEnvironment env,
+                                                final ISPObsComponent node) {
+        final Option<SPTarget> tOpt = TargetSelection.getTargetForNode(env, node);
+        final Option<SPCoordinates> cOpt = TargetSelection.getCoordinatesForNode(env, node);
+        final Option<IndexedGuideGroup> gOpt = TargetSelection.getIndexedGuideGroupForNode(env, node);
+        if ((tOpt.isDefined() && (cOpt.isDefined() || gOpt.isDefined()))
+                || (cOpt.isDefined() && gOpt.isDefined()))
+            throw new RuntimeException("Illegal node state: more than one selection recorded for node");
```
swalker2m: Perhaps we are missing a way to define the selection such that the unpossible cases cannot be represented?  An ADT with a case for none, target, coordinate, or group.  Then if you have a selection you know it is just one of those four cases.  You could add a fold method to handle logic that applies to each individual case.

I don't want to prolong this task any more though, and this at least will give us a warning in those cases where we check.

sraaphorst: When I initially started working on this, I began with:
```
Option<ImEither<ImEither<Coodinates, SPTarget>, IndexedGuideGroup>>
```
but that just seemed ridiculous and was far too hard to work with, so I began to implement a class, `OneOfThree` that was isomorphic to a nested `ImEither`, and was going to make:
```
Option<OneOfThree<Coordinates, SPTarget, IndexedGuideGroup>>
```
to represent the four possible selection types. Again, that just struck me as being too much, even though it would have captured the constraints in the type better than what I ended up with. (I was inspired by how Rob used `fold` over `Target` subclasses in `Target`, as you mention here.)

In the end, this seemed easiest and I was afraid of mucking with the code, so I went with it, although it is a shortcoming in that safety is not ensured by type.

sraaphorst: (`OneOfThree<R,S,T>` would of course require `First<R,S,T>`, `Second<R,S,T>`, and `Third<R,S,T>`.)

swalker2m: Let's just go with this as is.  It's much better than it was before regardless. 👍

sraaphorst: Excellent. I pushed my last PR, which makes the target table work as we discussed (the base is disabled if it is linked to another target, and when attempting to select the base, the actual target serving as the base is selected), and Travis just said yes, so I'm going to take the terrifying step and merge this.

PR #1520, opened 2018-06-20
REL-3435: GHOST now can be loaded by the TCC

sraaphorst:
This implements the code to let the TCC load GHOST observations from the session queue, as well as fix some minor comment issues and cleanup along the way.

Here are two screenshots of GHOST observations in the TCC.

The first is a single target standard resolution GHOST observation of M62:
screen shot 2018-06-20 at 3 45 26 pm

The second is a dual target observation, in which case, the base position is not a named target, and thus we call it `Sky Position`:
screen shot 2018-06-20 at 3 45 49 pm

### DIFF
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/TargetEnvironment.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/TargetEnvironment.java
@@ -1,11 +1,13 @@
 package edu.gemini.spModel.target.env;
 
 import edu.gemini.shared.util.immutable.*;
+import edu.gemini.spModel.gemini.ghost.GhostAsterism;
 import edu.gemini.spModel.guide.GuideProbe;
 import edu.gemini.spModel.pio.ParamSet;
 import edu.gemini.spModel.pio.Pio;
 import edu.gemini.spModel.pio.PioFactory;
 import edu.gemini.spModel.target.SPCoordinates;
+import edu.gemini.spModel.target.SPSkyObject;
 import edu.gemini.spModel.target.SPTarget;
 import edu.gemini.spModel.target.SPTargetPio;
 
@@ -89,10 +91,25 @@ public final class TargetEnvironment implements Serializable, TargetContainer {
     }
 
     /** Returns an arbitrary target from the asterism. This method will be removed for 18A. */
+    public SPSkyObject getPrimaryTargetFromAsterism() {
+        if (asterism instanceof GhostAsterism) {
+            final GhostAsterism ga = (GhostAsterism) asterism;
+            if (ga.base().isDefined())
+                return ga.base().get();
+            else if (ga instanceof GhostAsterism.DualTarget) {
+                final Option<SPCoordinates> c =
+                        ImOption.fromScalaOpt(ga.basePosition(ImOption.scalaNone())).map(SPCoordinates::new);
+                if (c.isDefined())
+                    return c.getValue();
+            }
+        }
+        return asterism.allSpTargets().head();
+    }
+
     // TODO:ASTERISM
     @Deprecated
     public SPTarget getArbitraryTargetFromAsterism() {
-      return getAsterism().allSpTargets().head();
+        return getAsterism().allSpTargets().head();
     }
 
     /**
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/TargetEnvironment.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/TargetEnvironment.java
@@ -90,8 +90,8 @@ public final class TargetEnvironment implements Serializable, TargetContainer {
         this.user     = user;
     }
 
-    /** Returns an arbitrary target from the asterism. This method will be removed for 18A. */
-    public SPSkyObject getPrimaryTargetFromAsterism() {
+    /** Returns the sky object from the asterism that dictates the slew position. */
+    public SPSkyObject getSlewPositionObjectFromAsterism() {
         if (asterism instanceof GhostAsterism) {
             final GhostAsterism ga = (GhostAsterism) asterism;
             if (ga.base().isDefined())
--- a/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/SPCoordinates.scala
+++ b/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/SPCoordinates.scala
@@ -98,9 +98,15 @@ final class SPCoordinates(var coordinates: Coordinates) extends SPSkyObject {
 
   override def getDecString(time: GOLong): GOption[String] =
     Some(coordinates.dec.toAngle.formatHMS).asGeminiOpt
+
+  override val getName: String = SPCoordinates.Name
+
+  override def setName(name: String): Unit =
+    throw new UnsupportedOperationException("Cannot set name of coordinates")
 }
 
 object SPCoordinates {
   val ParamSetName = "spCoordinates"
   val CoordinatesName   = "coordinates"
+  val Name = "Sky Position"
 }
--- a/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/SPCoordinates.scala
+++ b/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/SPCoordinates.scala
@@ -100,13 +100,10 @@ final class SPCoordinates(var coordinates: Coordinates) extends SPSkyObject {
     Some(coordinates.dec.toAngle.formatHMS).asGeminiOpt
 
   override val getName: String = SPCoordinates.Name
-
-  override def setName(name: String): Unit =
-    throw new UnsupportedOperationException("Cannot set name of coordinates")
 }
 
 object SPCoordinates {
   val ParamSetName = "spCoordinates"
   val CoordinatesName   = "coordinates"
-  val Name = "Sky Position"
+  val Name = "Sky"
 }
--- a/bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/GhostSupport.java
+++ b/bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/GhostSupport.java
@@ -12,6 +12,13 @@ public class GhostSupport implements ITccInstrumentSupport {
         this.oe = oe;
     }
 
+    /**
+     * Factor for creating a new GHOST Instrument Support.
+     *
+     * @param oe The current ObservationEnvironment
+     * @return A new instance
+     * @throws NullPointerException returned if the ObservationEnvironment is null.
+     */
     static public ITccInstrumentSupport create(final ObservationEnvironment oe) throws NullPointerException {
         return new GhostSupport(oe);
     }
@@ -34,7 +41,7 @@ public class GhostSupport implements ITccInstrumentSupport {
 
     @Override
     public String getTccConfigInstrumentOrigin() {
-        return null;
+        return "ghost";
     }
 
     @Override
--- a/bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/GhostSupport.java
+++ b/bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/GhostSupport.java
@@ -9,6 +9,8 @@ final public class GhostSupport implements ITccInstrumentSupport {
 
     private GhostSupport(final ObservationEnvironment oe) {
         Objects.requireNonNull(oe, "Observation environment can not be null");
+        if (!(oe.getInstrument() instanceof Ghost))
+            throw new RuntimeException("Incompatible instrument in ObservationEnvironment.");
         this.oe = oe;
     }
 
--- a/bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/GhostSupport.java
+++ b/bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/GhostSupport.java
@@ -4,8 +4,8 @@ import edu.gemini.spModel.gemini.ghost.Ghost;
 
 import java.util.Objects;
 
-public class GhostSupport implements ITccInstrumentSupport {
-    private ObservationEnvironment oe;
+final public class GhostSupport implements ITccInstrumentSupport {
+    final private ObservationEnvironment oe;
 
     private GhostSupport(final ObservationEnvironment oe) {
         Objects.requireNonNull(oe, "Observation environment can not be null");
--- a/bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/GhostSupport.java
+++ b/bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/GhostSupport.java
@@ -1,0 +1,54 @@
+package edu.gemini.wdba.tcc;
+
+import edu.gemini.spModel.gemini.ghost.Ghost;
+
+import java.util.Objects;
+
+public class GhostSupport implements ITccInstrumentSupport {
+    private ObservationEnvironment oe;
+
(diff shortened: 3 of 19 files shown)


bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/ObservationEnvironment.java
```diff
@@ -126,7 +129,18 @@ public String getObservationTitle() {
     }
 
     public String getBasePositionName() {
-        return _targetEnv.getArbitraryTargetFromAsterism().getName();
+        final Asterism a = _targetEnv.getAsterism();
+
+        // A dual asterism always has a sky position, and an asterism with an
+        // overridden base has a sky position.
+        if (a instanceof GhostAsterism) {
```
cquiroz: Do you need the external check? it seems only the internal check is needed? or maybe collapse it into a single condition

bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/TccFieldConfig.java
```diff
@@ -146,7 +155,7 @@ private void addGuideGroup(GuideProbeTargets gt) throws WdbaGlueException {
     // private method to log and throw and exception
     private void _logAbort(String message, Exception ex) throws WdbaGlueException {
         //LOG.error(message);
-        throw new WdbaGlueException(message, (ex != null) ? ex : null);
+        throw new WdbaGlueException(message, ex);
```
cquiroz: nice

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/TargetEnvironment.java
```diff
@@ -89,10 +91,25 @@ public static TargetEnvironment create(SPTarget base, GuideEnvironment guide, Im
     }
 
     /** Returns an arbitrary target from the asterism. This method will be removed for 18A. */
+    public SPSkyObject getPrimaryTargetFromAsterism() {
```
swalker2m: The comment needs to be reworded, I think.  It isn't clear to me what the "primary" target means though.  Is it supposed to be the slew position?  Also, it can be an `SPCoordinate` so it isn't technically necessarily a target in the `Target` ADT sense?

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/TargetEnvironment.java
```diff
@@ -89,10 +91,25 @@ public static TargetEnvironment create(SPTarget base, GuideEnvironment guide, Im
...
+        if (asterism instanceof GhostAsterism) {
+            final GhostAsterism ga = (GhostAsterism) asterism;
+            if (ga.base().isDefined())
+                return ga.base().get();
+            else if (ga instanceof GhostAsterism.DualTarget) {
+                final Option<SPCoordinates> c =
+                        ImOption.fromScalaOpt(ga.basePosition(ImOption.scalaNone())).map(SPCoordinates::new);
```
swalker2m: Looking at it here, without doubling back to check the definition, it isn't clear to me what the difference is between `base` and `basePosition`.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/SPCoordinates.scala
```diff
@@ -98,9 +98,15 @@ final class SPCoordinates(var coordinates: Coordinates) extends SPSkyObject {
 
   override def getDecString(time: GOLong): GOption[String] =
     Some(coordinates.dec.toAngle.formatHMS).asGeminiOpt
+
+  override val getName: String = SPCoordinates.Name
+
+  override def setName(name: String): Unit =
+    throw new UnsupportedOperationException("Cannot set name of coordinates")
```
swalker2m: Perhaps `setName` shouldn't be in the `SPSkyObject` interface if it doesn't apply?

bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/GhostSupport.java
```diff
@@ -0,0 +1,61 @@
+package edu.gemini.wdba.tcc;
+
+import edu.gemini.spModel.gemini.ghost.Ghost;
+
+import java.util.Objects;
+
+public class GhostSupport implements ITccInstrumentSupport {
```
swalker2m: I suppose it might as well be `final`.

bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/GhostSupport.java
```diff
@@ -0,0 +1,61 @@
...
+    private ObservationEnvironment oe;
```
swalker2m: Here too, might as well be `final`.

bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/GhostSupport.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    @Override
+    public String getWavelength() {
+        return null;
+    }
+
+    @Override
+    public String getPositionAngle() {
+        Ghost inst = (Ghost) oe.getInstrument();
```
swalker2m: We should probably check this assumption (that the instrument is `Ghost`) when we create the `GhostSupport`. Amazingly `ObservationEnvironment` is immutable so it can just be checked once there in construction.

bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/NIRISupport.java
```diff
@@ -30,7 +30,7 @@ private NIRISupport(ObservationEnvironment oe) throws NullPointerException {
     }
 
     /**
-     * Factory for creating a new TReCS Instrument Support.
+     * Factory for creating a new NIRI Instrument Support.
```
swalker2m: Thanks for fixing all these!

bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/TargetGroupConfig.java
```diff
@@ -21,27 +22,29 @@
...
-    private TargetGroupConfig(final String name, final ImList<SPTarget> targets, final Option<SPTarget> primaryTarget) {
+    private TargetGroupConfig(final String name, final ImList<SPTarget> targets, final Option<SPSkyObject> primaryTarget) {
         super(name);
 
         addAttribute(TYPE, TYPE_VALUE);
 
-        primaryTarget.map(t -> t.getName())
-                .filter(n -> !"".equals(n))
+        primaryTarget.map(SPSkyObject::getName)
+                .filter(n -> !n.isEmpty())
```
swalker2m: The previous version would handle a `null` name.  Is that unpossible now?

bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/TccFieldConfig.java
```diff
@@ -98,11 +100,18 @@ private static boolean isEmpty(String name) {
...
         // Add the target itself.
-        SPTarget base = env.getArbitraryTargetFromAsterism();
-        if (isEmpty(base.getName())) {
-            base.setName(TccNames.BASE);
+        SPSkyObject so = env.getPrimaryTargetFromAsterism();
+
+        if (so instanceof  SPTarget) {
+            final SPTarget spt = (SPTarget) so;
+            if (isEmpty(spt.getName()))
+                spt.setName(TccNames.BASE);
```
swalker2m: I see this was here before too, so I'm not suggesting we change it necessarily, but mutating the target name seems like a weird thing to do here.

bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/GhostSupport.java
```diff
@@ -0,0 +1,61 @@
...
+     * @return A new instance
+     * @throws NullPointerException returned if the ObservationEnvironment is null.
+     */
+    static public ITccInstrumentSupport create(final ObservationEnvironment oe) throws NullPointerException {
+        return new GhostSupport(oe);
+    }
+
+    @Override
+    public String getWavelength() {
+        return null;
```
swalker2m: I guess I don't know what else this would be without changing the signature to `Option<String>` (btw, `String`!), but are callers set up to expect this?

bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/GhostSupport.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    /**
+     * Factor for creating a new GHOST Instrument Support.
+     *
+     * @param oe The current ObservationEnvironment
+     * @return A new instance
+     * @throws NullPointerException returned if the ObservationEnvironment is null.
+     */
+    static public ITccInstrumentSupport create(final ObservationEnvironment oe) throws NullPointerException {
```
swalker2m: It's not often you see an unchecked exception in the signature.  I suppose the other `FooSupport` instances do the same thing.

bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/GhostSupport.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    @Override
+    public String getTccConfigInstrumentOrigin() {
+        return "ghost";
+    }
+
+    @Override
+    public String getFixedRotatorConfigName() {
+        return null;
```
swalker2m: Same question about `null` here: are callers expecting that?

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/TargetEnvironment.java
sraaphorst: Agreed. I'm going to release a PR next getting rid of the `getArbitraryTargetFromAsterism` next. In the single target case, it was the only science target, whereas now I believe we should be using it as the slew position. I'll change this here now, though, to `getSlewPositionObjectFromAsterism`?

bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/GhostSupport.java
sraaphorst: I have a bad habit of forgetting that. 😊

bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/TargetGroupConfig.java
sraaphorst: Hmmm. Good point. I'm not certain. I'll check, and change it back if I'm not sure.

sraaphorst: I don't think it will ever be `null`, but let's not play with 🔥 .

bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/TccFieldConfig.java
sraaphorst: Agree. I think this may have significance to the TCC?

bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/GhostSupport.java
sraaphorst: There is special handling for this potentially being `null` in the one class that calls this method; that being said, I don't think that that's the behaviour we want, so maybe we should set it to the empty string? I'm hoping we'll know more about what this should be soon.

bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/GhostSupport.java
sraaphorst: Yes, I thought this was weird, too, but I copied this from the other `...Support` classes (as clearly others did, since most instruments were labeled as TReCS in their comments 😆 .)

bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/GhostSupport.java
sraaphorst: Many of the other instruments return `null` here.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/TargetEnvironment.java
swalker2m: Seems good.

bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/ObservationEnvironment.java
sraaphorst: It would have to be added to the internal check to ensure the conversion to `GhostAsterism` wouldn't fail; that's not problematic, but I thought it delineated the fact that the next `if` check was only applicable to `GhostAsterism`s better.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/TargetEnvironment.java
swalker2m: Maybe `getOverriddenBase` for the `Option[SPCoordinates]` one?

sraaphorst:
In any case, we are getting an error in the TCC when the base position is a sky position. Grah.

sraaphorst:
Okay, all errors have now been resolved. We are only passing in the base position, and no other information about the science positions as per my discussion with Steve.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/TargetEnvironment.java
sraaphorst: Do you mean changing this in `GhostAsterism`? I'm a bit confused.

bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/GhostSupport.java
sraaphorst: Now done in constructor, since `ObservationEnvironment` is immutable.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/TargetEnvironment.java
swalker2m: Yes, just a renaming in `GhostAsterism` that makes the difference clearer.  One is an explicitly defined "override" base position and may not be set at all.  The other is an algorithm that gives you a base position regardless.  So maybe `overriddenBase` vs `calculatedBase`?  I dunno.  Good names don't occur to me.  `getBase` vs `getBasePosition` doesn't clear to me though.

PR #1521, opened 2018-06-22
REL-3437: Eliminated TargetEnvironment.getArbitraryTargetFromAsterism

sraaphorst:
As a temporary measure before the implementation of asterisms with more than one target (due to GHOST), a method, `TargetEnvironment.getArbitraryTargetFromAsterism` was implemented to simply return the first (and in the case of all previous instruments, only) science target.

Now, with GHOST, it was necessary to eliminate this method and ensure that the code properly handles asterisms with multiple targets.

A major part of this change was implementing, for `ITCPanel`, a target selection combo box to select from the targets in an asterism. Here are the results (i.e. the Target section inserted above the Conditions section):
screen shot 2018-06-22 at 5 29 32 pm


bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/TargetEnvironment.java
```diff
@@ -68,11 +68,26 @@ public static TargetEnvironment create(Asterism asterism) {
      *
      * @param user list of user targets
      */
-    @Deprecated
-    public static TargetEnvironment create(SPTarget base, GuideEnvironment guide, ImList<UserTarget> user) {
+    public static TargetEnvironment create(final SPTarget base, final GuideEnvironment guide, final ImList<UserTarget> user) {
```
sraaphorst: I could see no good reason to keep this deprecated, since the other method, `TargetEnvironment.create(SPTarget)` was not, and they seem like good convenience methods.

sraaphorst: I could see no good reason to keep this deprecated, given that the other convenience method, `TargetEnvironment.create(SPTarget)` was not.

sraaphorst:
@swalker2m I forgot to eliminate the method confusion between `base` and `basePosition` here, and while I like the idea of making `base` into `overriddenBase`, I'm not sure we should change `basePosition` to `calculatedBase`. This is a method inherited from `Asterism`, which would mean changing `Asterism.basePosition` to `Asterism.calculatedBase`.

If you're good with that, I'll go ahead and do the refactor renaming, but I thought I'd just ask for your opinion on it first.

swalker2m:
> If you're good with that, I'll go ahead and do the refactor renaming

Sounds good to me, thanks.

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
```diff
@@ -59,43 +59,58 @@ class LchQueryFunctor(queryType: LchQueryFunctor.QueryType,
...
+
+        case c: SPCoordinates =>
+          // For now, we pass off sky coordinates as a nonsidereal target to
+          // avoid having to significantly modify LTTS.
+          Some(new NonSidereal() {
+            setName(c.getName)
+            setType(targetType.orNull)
+          })
+
+        case _ => None
```
swalker2m: Should we throw an exception here perhaps?  There are no other kinds of `SPSkyObject` (?) but if we did introduce one, we'd want to handle it and an exception would make it immediately obvious what is going on.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
```diff
@@ -1375,17 +1376,20 @@ else if (contents instanceof TableGroupSelection) {
...
+        if (!(base instanceof SPTarget))
+            return;
+        final SPTarget baseTarget = (SPTarget) base;
 
         final GuideEnvironment   genv = env.getGuideEnvironment();
-        final ImList<UserTarget> user = env.getUserTargets().append(new UserTarget(UserTarget.Type.other, base));
+        final ImList<UserTarget> user = env.getUserTargets().append(new UserTarget(UserTarget.Type.other, baseTarget));
 
-        final TargetEnvironment newEnv = TargetEnvironment.create(target, genv, user);
+        final TargetEnvironment newEnv = TargetEnvironment.create(baseTarget, genv, user);
```
swalker2m: How does this work when the `Asterism` is a `GhostAsterism` since this `create` method will return an environment with a `Single` target asterism?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/TargetFeedback.scala
```diff
@@ -146,9 +146,13 @@ object TargetGuidingFeedback {
...
-    if (target == env.getArbitraryTargetFromAsterism) baseAnalysis(ctx, mt)
-    else guideProbe(env, target).fold(List.empty[AnalysisRow]) { vgp =>
-      guideStarAnalysis(ctx, mt, vgp, target).toList
+    env.getSlewPositionObjectFromAsterism match {
+      case base: SPTarget if target == base =>
+        baseAnalysis(ctx, mt)
+      case _ =>
+        guideProbe(env, target).fold(List.empty[AnalysisRow]) { vgp =>
+          guideStarAnalysis(ctx, mt, vgp, target).toList
+        }
```
swalker2m: I'm surprised that here we don't show the `baseAnalysis` for the two targets in the Ghost dual target mode case.  I would guess that the base analysis would be what you want to see unless you're parked on a guide star.

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
sraaphorst: You're right: the only two subclasses of `SPSkyObject` are `SPCoordinates` and `SPTarget`.

I think an exception is a good idea. I'll go ahead and change that.

sraaphorst: Oh, wait... it just occurred to me: there is one target type that we're not handling here, namely `TooTarget`, so it would throw exceptions for that. Is that what we would want?

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
sraaphorst: Very good point. That slipped by me entirely. I think that that would result in an incorrect state. I'm not really sure about this morphing, given the dependency of GHOST base on the target(s). Maybe we should just disallow it for GHOST?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/TargetFeedback.scala
sraaphorst: I'll give this a little bit of thought.

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
swalker2m: I'm not sure what should happen in that case.  It depends on how this is used but I'm not familiar with LCH.  We could make it explicit though and add a case for ToO targets that returns `None` (to match the previous behavior) and then an exception for anything unexpected.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
swalker2m: Yeah it seems like this would require a lot more thought for asterisms in general.  It's also a pretty dubious feature, imho.  Perhaps we can hide the target type change combo altogether for now unless the asterism is `Single` target.

swalker2m:
> > If you're good with that, I'll go ahead and do the refactor renaming
Sounds good to me, thanks.

> Sounds good to me, thanks.

I suppose I should be more clear about what I'm signing-off on.  I think you're right that we should probably leave `basePosition` name alone and just change the overridden base position accessor name.

sraaphorst:
Re: the renaming. Yes: we leave `basePosition` alone, and rename the previously confusing `base` in the `GhostAsterism` hierarchy as `overriddenBase`.

bundle/jsky.app.ot/src/main/java/jsky/app/ot/gemini/editor/targetComponent/EdCompTargetList.java
sraaphorst: Sure. I think that sounds like the best option.

bundle/edu.gemini.lchquery.servlet/src/main/scala/edu/gemini/lchquery/servlet/LchQueryFunctor.scala
sraaphorst: I'm going to do exactly that: `None` for `TooTarget` and throw an exception for the case that should never happen.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/editor/targetComponent/TargetFeedback.scala
sraaphorst: This is definitely an issue. I changed it so that no guiding feedback would be produced for any GHOST asterism targets, but we have the problem where if you click on a sky position or a group name node in the tree and you're missing a guide star, the warning isn't displayed.

I'll create a separate JIRA task for this since I think it falls outside the scope of this one.

PR #1523, opened 2018-06-27
REL-1714: Include contraint type in intersection message

jdnavarro:


bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
```diff
@@ -97,26 +97,44 @@ public void propertyChange(PropertyChangeEvent evt) {
...
-			// Function to create markers that report solver intervals.
-			final BiFunction<String, String, Option<Marker>> createSolverMarker =
-				(cacheName, prefix) -> {
+            final Supplier<Union<Interval>> wholeNight = () -> new Union<>(new Interval(v.getSchedule().getStart(), v.getSchedule().getEnd()));
+
+			final Predicate<String> contributes =
+				(cacheName) -> {
 					final Map<Obs, Union<Interval>> c = v.getSchedule().getCache(cacheName);
-					final Union<Interval>       valid = c.getOrDefault(a.getObs(), new Union<>());
+					final Union<Interval> u = c.getOrDefault(a.getObs(), new Union<>());
```
swalker2m: I wonder if the default should be the whole night here instead of the empty `Union`.  I believe the cache should have an entry for every observation so it's probably irrelevant.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
```diff
@@ -97,26 +97,44 @@ public void propertyChange(PropertyChangeEvent evt) {
...
-						return new Some<>(new Marker(false, this, Severity.Notice, prefix, new Some<>(valid), v, a));
+					    final List<String> attrs = new ArrayList<>();
+						if (contributes.test(Variant.VISIBLE_UNION_CACHE)) { attrs.add("airmass"); }
+					    if (contributes.test(Variant.DARK_UNION_CACHE))    { attrs.add("background"); }
+						if (contributes.test(Variant.TIMING_UNION_CACHE))  { attrs.add("timing window"); }
+
+                        final String s = attrs.isEmpty() ?
+							"" : DefaultImList.create(attrs).mkString(" (", ", ", ")");
+
+						return new Some<>(new Marker(false, this, Severity.Notice, prefix + s, new Some<>(valid), v, a));
```
swalker2m: This will leave the attributes as part of the prefix instead of appended after the formatted time, but to achieve that we'd need to create a template message such that the timestamp can appear anywhere in the template.

PR #1525, opened 2018-07-04
REL-3396: Added code to populate static sequence config for GHOST.

sraaphorst:
This adds the code to `GhostCB` to populate GHOST with the target information that it needs as per the requirements.

Here is a screenshot of the static configuration with the added fields under the "instrument" section. Note that @swalker2m and I decided to pass in target information in both HMS and degree format: this could change at a later date.

ghostseq


cquiroz:
Will you decide the format of the coordinates based on the seqexec?

PR #1526, opened 2018-07-04
REL-3447: Upping max memory allocation.

sraaphorst:
Increasing the max memory allocation for the QPT and OT as the garbage collector generated an `OutOfMemoryError` for the program GS-2018A-Q-419.

### DIFF
--- a/app/itc/build.sbt
+++ b/app/itc/build.sbt
@@ -103,7 +103,6 @@ def itc(version: Version) = AppConfig(
   distribution = List(TestDistro),
   vmargs = List(
     "-Xmx1024M",
-    "-XX:MaxPermSize=196M",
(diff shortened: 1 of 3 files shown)


cquiroz:
Could you test if these options can be used with the jvm version we use;
    "-J-XX:+ExitOnOutOfMemoryError",
    "-J-XX:+CrashOnOutOfMemoryError"

sraaphorst:
Absolutely. Will do!

sraaphorst:
It appears not. They look like they were introduced in 1.8.0_92, and we're locked into 1.8.0_73 because of the eObsLog.

cquiroz:
ok, never mind

PR #1527, opened 2018-07-04
REL-3447: Upping max memory allocation.

sraaphorst:
Increasing the max memory allocation for the QPT and OT as the garbage collector generated an `OutOfMemoryError` for the program GS-2018A-Q-419.


swalker2m:
Was it out of perm gen space or heap? I think the exception will specify which. It seemed like a pretty big jump in perm. 

Did it generate that exception with just that program downloaded or were there others in the database?

cquiroz:
I think only the heap memory mas increased by the patch

PR #1531, opened 2018-07-13
QPT cherry pickings

jdnavarro:
Both PRs were accepted in #1529 and #1530. The tabs to spaces conversion was also backported to avoid merge conflicts in the other commit.


jdnavarro:
I'm keeping the commits separate when merging in case we need to take out just the tabs easily if that gives us problems.

PR #1532, opened 2018-07-17
REL-3452: parse mask name and use it as lookup key

swalker2m:
In the initial QPT+ICTD release I misinterpreted the columns of the ICTD `MOS` table.  I had thought that an observation's custom mask name should be matched to the `MOS` table's `ODF` field to find the mask's availability.  In reality the `ODF` field turns out to be informational and the `MOS` table should instead by keyed by its corresponding program id + `RunningNo` columns.

This PR updates the code to ignore the `MOS` table `ODF` column altogether.  Instead it parses a GMOS observation's custom mask field to obtain a `CustomMaskKey`, identifying the program id and running number for the mask entry in the `MOS` table.

### DIFF
--- a/bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/Variant.java
+++ b/bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/Variant.java
@@ -1,5 +1,6 @@
 package edu.gemini.qpt.core;
 
+import edu.gemini.ictd.CustomMaskKey;
 import edu.gemini.qpt.core.Marker.Severity;
 import edu.gemini.qpt.core.util.*;
 import edu.gemini.qpt.core.util.Interval.Overlap;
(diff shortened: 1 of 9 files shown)


bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/Variant.java
```diff
@@ -689,15 +689,20 @@ private void updateObsFlags() {
...
+                    // possible and use it to lookup the Availability.
+                    final Availability a =
+                        ImOption.fromScalaOpt(CustomMaskKey.parse(m))
+                            .map(k -> owner.maskAvailability(k))
+                            .getOrElse(Availability.Missing);
+
+                    if (a == Availability.SummitCabinet) {
+                        maskFlags.add(Flag.MASK_IN_CABINET);
+                    } else if (a != Availability.Installed) {
+                        maskFlags.add(Flag.MASK_UNAVAILABLE);
```
cquiroz: No flags for `Missing`?

swalker2m: Right, from the point of view of the QPT, a mask is either installed, not installed but in the summit cabinet, or else simply unavailable.  There is more detail in the ICTD app if anyone is curious.

PR #1535, opened 2018-07-24
REL-1714: Change night boundaries to civil twilight

swalker2m:
This PR changes the QPT twilight boundaries to the civil, 6 degree definition.  This allows `SBAny` observations to be scheduled earlier in the night without impacting other sky background constraints. 

To ensure consistency, I centralized the twilight bound computation in a single place.  I'm not particularly pleased with how it came out though.


swalker2m:
I found a few places where nautical twilight is used:

1) Time accounting.  I don't think that should change but I can check if this flies.

2) OT Elevation plot.  I would guess that needs to be made consistent at some point, but I don't know if it warrants an OT patch before December.  I can check on this as well.

3) ToO "Scheduling Report".   Same as above.  The scheduling report doesn't do a SB check though so for most targets nautical is probably necessary.  Or actually, it should be updated to use a SB solver like the QPT does anyway.

At any rate, I'm waiting for feedback on whether this idea is going to be acceptable to the QCs before continuing.

swalker2m:
This is a modest improvement, in that it centralizes the definition of twilight for the QPT.  I'll go ahead and merge it but based on feedback I believe that I'll need to make a subsequent PR returning to nautical twilight and the old 2.0 airmass limit.

PR #1536, opened 2018-07-25
REL-1714: Report constraint contributing to intersection boundary

jdnavarro:
This replaces the contributing constraints in the QPT intersection messages with a letter indicating which constraint was responsible for a given constraint limit. For example: "Notice: Must be observed between: 19:55(A) - 22:18(T)."

### DIFF
--- a/bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/Marker.java
+++ b/bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/Marker.java
@@ -1,12 +1,7 @@
 package edu.gemini.qpt.core;
 
-import edu.gemini.qpt.core.util.Interval;
-import edu.gemini.qpt.core.util.Union;
 import edu.gemini.qpt.ui.util.TimePreference;
-import edu.gemini.shared.util.immutable.DefaultImList;
-import edu.gemini.shared.util.immutable.ImList;
-import edu.gemini.shared.util.immutable.None;
-import edu.gemini.shared.util.immutable.Option;
+import edu.gemini.shared.util.immutable.*;
 import edu.gemini.spModel.core.Site;
 
 /**
@@ -24,19 +19,19 @@ public class Marker implements Comparable<Marker> {
     private final String text;
     private final Severity severity;
     private final boolean qcOnly;
-    private final Option<Union<Interval>> union;
+    private final ImList<Long> timestamps;
 
-    public Marker(boolean qcOnly, Object owner, Severity severity, String text, Option<Union<Interval>> union, Object... path) {
+    public Marker(boolean qcOnly, Object owner, Severity severity, String text, ImList<Long> timestamps, Object... path ) {
         this.qcOnly = qcOnly;
         this.owner = owner;
         this.path = path;
         this.text = text;
         this.severity = severity;
-        this.union = union;
+        this.timestamps = timestamps;
     }
 
     public Marker(boolean qcOnly, Object owner, Severity severity, String text, Object... path) {
-        this(qcOnly, owner, severity, text, None.instance(), path);
+        this(qcOnly, owner, severity, text, ImCollections.emptyList(), path);
     }
 
     public boolean isQcOnly() {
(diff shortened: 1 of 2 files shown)


bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
```diff
@@ -103,14 +100,29 @@ public void propertyChange(PropertyChangeEvent evt) {
...
+            final BiFunction<Function<Interval, Long>, String, Predicate<Long>> matchWith =
+                (f, cacheName) ->
+                    ts -> {
+                        final Map<Obs, Union<Interval>> c = v.getSchedule().getCache(cacheName);
+                        final Union<Interval> u = c.getOrDefault(a.getObs(), new Union<>());
+
+                        for (Interval i : u) {
+                            if (f.apply(i) == ts) { return true; }
+                        }
+                        return false;
```
swalker2m: We can get rid of the `for` statement and early return using the `Stream` API:

```java
return u.getIntervals().stream().anyMatch(i -> f.apply(i) == ts);
```

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/Marker.java
```diff
@@ -59,11 +62,11 @@ public String getText() {
...
 
-    private static final String formatInterval(Site site, Interval interval, TimePreference p) {
+    private static final String formatInterval(Site site, String formatter, Interval interval, TimePreference p) {
         return String.format(
-                "%s - %s",
-                p.format(site, interval.getStart(), "HH:mm"),
-                p.format(site, interval.getEnd(), "HH:mm")
+            formatter,
+            p.format(site, interval.getStart(),"HH:mm"),
+            p.format(site, interval.getEnd(), "HH:mm")
```
swalker2m: How do we know that the formatter is expecting two `String`s?

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/Marker.java
```diff
@@ -24,9 +21,15 @@
...
-    public Marker(boolean qcOnly, Object owner, Severity severity, String text, Option<Union<Interval>> union, Object... path) {
+    private final Option<Pair<ImList<String>, Union<Interval>>> union;
+
+    public Marker(boolean qcOnly,
+                  Object owner,
+                  Severity severity,
+                  String text,
+                  Option<Pair<ImList<String>, Union<Interval>>> union,
+                  Object... path
+    ) {
```
swalker2m: Getting the timestamps formatted in the message is tricky.  I was thinking that the `String text` argument would potentially be a format string itself with made up placeholder variables for each timestamp that we need to format.  Then the next argument would be a `List<Long>` or `ImList<Long>` of matching timestamps to format.  Nothing would guarantee that the variables in `text` would be the same number as the elements in the list of course, but I believe it would be clearer what is going to happen.

So for example, the `text` might be `"Must be observed between %t(A) - %t(sunrise)."` and we'd expect two corresponding `Long` arguments in the given `ImList`.  So basically fold the format `text` over the `List<Long>` replacing the first `%t` if any with the current `Long`, formatted as `"HH:mm"` in the right timezone of course, at each step.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/Marker.java
```diff
@@ -95,7 +105,7 @@ public Severity getSeverity() {
         return severity;
     }
 
-    public Option<Union<Interval>> getUnion() { return union; }
+    public Option<Pair<ImList<String>, Union<Interval>>> getUnion() { return union; }
```
swalker2m: I doubt this is used anywhere?  It isn't really just a `Union` anymore regardless.  We might just get rid of it.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
```diff
@@ -103,14 +100,26 @@ public void propertyChange(PropertyChangeEvent evt) {
...
+                        final Map<Obs, Union<Interval>> c = v.getSchedule().getCache(cacheName);
+                        final Union<Interval> u = c.getOrDefault(a.getObs(), new Union<>());
+
+                        return u.getIntervals().stream().anyMatch(i -> f.apply(i) == ts);
+                    };
+
+            final BiFunction<Function<Interval, Long>, Interval, Option<String>> explainLimit =
+                (f, i) -> {
+                    if (matchWith.apply(f, Variant.VISIBLE_UNION_CACHE).test(f.apply(i))) {
+                        return new Some<>("%s(A)");
```
swalker2m: Another issue that I just happened to spot is that they requested an "A" for airmass limits and an "H" for hour angle. It all boils down to elevation constraints and the `VISIBLE_UNION_CACHE`, but we should try to show the appropriate letter for the observation.  We know the constraint that should be used from the observation via `obs.getElevationConstraintType()`.  So for example:

```java
final String flag;
switch (obs.getElevationConstraintType()) {
    case HOUR_ANGLE:
        flag = "H";
        break;
    default:
        flag = "A";
        break;
}
```

The hard part will be shuffling around the arguments to `explainLimit`.  Let's update the formatting of the timestamps and then look into this issue.

PR #1548, opened 2018-07-31
REL-3290: Implemented empty observation in PIT.

sraaphorst:
This redoes some work that was originally intended to be in the 2018B PIT, but that we felt hadn't had adequate testing to ensure it was working exactly as expected.

The idea here is that normally, when the PIT is launched, the Observations section contains no entries.

Science requested in REL-3290 that this be changed, and the Observations section begin with an "empty observation" to serve as a guide for users so that they better understood how to create the necessary data.

Here is a before picture:
pit_original

Here is an after picture. As you can see, there is now an empty observation:
screen shot 2018-07-31 at 11 58 55 am


cquiroz:
This was reviewed last semester and was delayed to make time for testing. I think now is a good time to integrate in

PR #1549, opened 2018-07-31
REL-3453: Made Altair LGS unavailable for 2019A

sraaphorst:
The laser at GN is not working and thus will not be available for 2019A proposals.

This PR:

1. Removes the LGS options from Altair;

2. Issues a warning if upconverting with an observation requiring Altair LGS; and

3. Produces an error (that disallows submission) if the proposal contains an observation with Altair LGS.

Here is a screenshot that shows the Altair node in the resource selection:
screen shot 2018-07-31 at 3 48 54 pm

### DIFF
--- a/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/transform/UpConverter.scala
+++ b/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/transform/UpConverter.scala
@@ -145,7 +145,12 @@ case class LastStepConverter(semester: Semester) extends SemesterConverter {
  * This converter will upgrade to 2019A.
  */
 case object SemesterConverter2018BTo2019A extends SemesterConverter {
-  override val transformers = Nil
+  // REL-3453: Altair LGS not available for 2019A.
+  val altairLgsNotAvailable: TransformFunction = {
+    case a @ <altair>{ns@_*}</altair> if (a \\ "lgs").nonEmpty =>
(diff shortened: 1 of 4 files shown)


bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/transform/UpConverter.scala
```diff
@@ -206,7 +211,7 @@ case object SemesterConverter2018ATo2018B extends SemesterConverter {
 
   val (texesRemover, texesRemoverMessage) = removeBlueprint("texes", "Texes")
 
-  override val transformers = List(gmosnAltairRemover, dssiGNToAlopeke, phoenixSite, texesRemover)
+  override val transformers = List(dssiGNToAlopeke, phoenixSite, texesRemover)
```
cquiroz: Why is this removed? Because Altair won't be available for any instrument rather that only for gmosN?

sraaphorst: Yes, exactly that. It seems like a pointless check at this point.

I was hemming and hawing about removing it entirely since it only applied to that one semester and probably won't be applicable moving forward.

cquiroz: Fair enough

PR #1551, opened 2018-08-02
REL-1714: Support for migrating pre-civil twilight schedules

swalker2m:
When opening an old schedule document, the block interval is set to nautical twilight.  Because the block interval trims all the cache results (such as the `VISIBLE_UNION_CACHE`) we mistakenly report availability ranges limited by nautical twilight.

This PR adjusts block boundaries to civil twilight upon import if they were set to nautical twilight.

### DIFF
--- a/bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/ScheduleIO.java
+++ b/bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/ScheduleIO.java
@@ -40,7 +40,8 @@ public class ScheduleIO {
     static final int VERSION_104 = 104;   // 1.0.4
     static final int VERSION_1030 = 1030; // 1.0.30 (new infrastructure, Dec 2013)
     static final int VERSION_1031 = 1031; // 1.0.31 (additional facilities (enums) as part of QV, Jan 2014)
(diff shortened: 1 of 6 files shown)


bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/ScheduleIO.java
```diff
@@ -40,7 +40,8 @@
     static final int VERSION_104 = 104;   // 1.0.4
     static final int VERSION_1030 = 1030; // 1.0.30 (new infrastructure, Dec 2013)
     static final int VERSION_1031 = 1031; // 1.0.31 (additional facilities (enums) as part of QV, Jan 2014)
-    static final int VERSION_CURRENT = VERSION_1031;
+    static final int VERSION_1032 = 1032; // 18B.1.1.5 (switch to civil twilight)
```
swalker2m: These used to be related to the QPT version number, but now it is just an integer.

swalker2m:
Don't be put off by the +4,524 line diff.  It's almost all test XML files.

swalker2m:
> How come this PR is first submitted against the release branch instead of develop. Did I miss it?

No, I was working on it in the release branch so I could test it with the production database.  I'll put in a matching PR for the develop branch.

PR #1553, opened 2018-08-03
REL-3413: Update GPI magnitude checks in PIT

sraaphorst:
The faintness level acceptable by GPI in the PIT is now affected by the conditions (image quality and cloud cover). This incorporates the faintness level adjustments as per the JIRA task, and prioritizes problem messages as per discussions with Frederik and Bryan.

(Note: it is possible, with the new updates, for a target to be simultaneously "too bright" and "too dark" if certain conditions are chosen: hence why the error messages are prioritized. This was discussed with science and despite the oddity of it, is the desired requirement.)

### DIFF
--- a/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/package.scala
+++ b/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/package.scala
@@ -38,6 +38,7 @@ package object immutable {
   type CloudCover = M.CloudCover
   object CloudCover extends EnumObject[M.CloudCover] {
     val BEST = M.CloudCover.cc50
+    val CC70 = M.CloudCover.cc70
     val CC80 = M.CloudCover.cc80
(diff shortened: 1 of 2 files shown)


bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/package.scala
```diff
@@ -38,6 +38,7 @@ package object immutable {
   type CloudCover = M.CloudCover
   object CloudCover extends EnumObject[M.CloudCover] {
     val BEST = M.CloudCover.cc50
+    val CC70 = M.CloudCover.cc70
```
cquiroz: Are these selectable on the instrument selection UI?

sraaphorst: They are selectable in the conditions UI; we just didn't have them specifically linked to names because we weren't using them for anything previously.

PR #1557, opened 2018-08-05
REL-3470 ITC output improvements

andrewwstephens:
This PR fixes the problem reported in REL-3470 (wrong units displayed for power-law) and makes a few other minor improvements.

1. Put the correct units on the power-law SED output
2. Label extended sources according to whether they are Gaussian or Uniform surface brightness
3. Add the Gaussian FHWM to the output
4. General label wording improvements


bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/MagnitudeSystem.scala
```diff
@@ -20,8 +20,8 @@ sealed abstract class SurfaceBrightness(val name: String) extends BrightnessUnit
   */
 object MagnitudeSystem {
 
-  case object Vega                extends MagnitudeSystem("Vega")
-  case object AB                  extends MagnitudeSystem("AB")
+  case object Vega                extends MagnitudeSystem("Vega mag")
```
cquiroz: I thin this used when displaying the Mag systems in the OT. is this change desired?

sraaphorst:
It looks additionally like this change is causing `VoTableParserSpec.scala` to break due to the name changes.

andrewwstephens:
@cquiroz and @sraaphorst - thanks for pointing that out!  While I do think updating the magnitude labels is a good thing, I have reverted them for now so that this PR can (hopefully) go through.  I left the modified surface brightness labels since I *think* they are only used in the ITC.

PR #1559, opened 2018-08-08
REL-3111: Cutting back the number of ephemerides entries to 200.

sraaphorst:
This is in response to Bryan's note on REL-3111 saying there were too many ephemeris data entries during PIT lookups (where we had approximately four): Rob and Bryan decided to cut the number back to 200 entries.

### DIFF
--- a/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/editor/HorizonsLookup.scala
+++ b/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/editor/HorizonsLookup.scala
@@ -111,9 +111,8 @@ final class HorizonsLookup(editor: TargetEditor, site: Site, when: Long) {
       val d1 = new Date(n.getStartTime)
       val d2 = new Date(n.getEndTime)
       for {
-        e1 <- HorizonsService2.lookupEphemerisWithPadding(d, site, 1000, s)
-        e2 <- HorizonsService2.lookupEphemeris(d, site, d1, d2, 300)
-      } yield Ephemeris(site, e1.data.union(e2.data))
+        e1 <- HorizonsService2.lookupEphemerisWithPadding(d, site, 200, s)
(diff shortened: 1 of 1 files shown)


bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/editor/HorizonsLookup.scala
```diff
@@ -111,9 +111,8 @@ final class HorizonsLookup(editor: TargetEditor, site: Site, when: Long) {
       val d1 = new Date(n.getStartTime)
       val d2 = new Date(n.getEndTime)
```
tpolecat: You no longer need `n`, `d1`, or `d2` … can you remove these?

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/editor/HorizonsLookup.scala
```diff
@@ -111,9 +111,8 @@ final class HorizonsLookup(editor: TargetEditor, site: Site, when: Long) {
...
       for {
-        e1 <- HorizonsService2.lookupEphemerisWithPadding(d, site, 1000, s)
-        e2 <- HorizonsService2.lookupEphemeris(d, site, d1, d2, 300)
-      } yield Ephemeris(site, e1.data.union(e2.data))
+        e1 <- HorizonsService2.lookupEphemerisWithPadding(d, site, 200, s)
+      } yield Ephemeris(site, e1.data)
```
tpolecat: This `for`-comp can just be `.map` now, and `e1` can just be `e`.

sraaphorst: You're right. I'll go ahead and do that. (Sorry... I rushed through this to get to another bug.)

PR #1562, opened 2018-08-13
PIT REL-3290: Extend template observations to band 3

sraaphorst:
In the original change I made for REL-3290, I implemented a "template" observation, i.e. an empty observation so that the fields needed to be filled by people filing proposals was more user-friendly:

screen shot 2018-07-31 at 11 58 55 am

However, it was brought to my attention during testing that the template did not show up for band 3 observations. This PR rectifies that. Note that a bit of ugly hackery is involved, as we need `Observation` to always have a band 1/2 observation and a band 3 observation (although, if empty, these are eliminated before submission).

Here is the final result:

screen shot 2018-08-13 at 2 29 55 pm

screen shot 2018-08-13 at 2 30 16 pm

### DIFF
--- a/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/view/obs/ObsListView.scala
+++ b/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/view/obs/ObsListView.scala
@@ -281,7 +281,7 @@ class ObsListView(shellAdvisor:ShellAdvisor, band:Band) extends BorderPanel with
       this.viewer.getTable.setDragEnabled(canEdit)
     }
 
-    // Repaint everything when the catalog or guidstar state changes
+    // Repaint everything when the catalog or guide star state changes
     AgsRobot.addListener {_:Any => refresh()}
     GsaRobot.addListener {_:Any => refresh()}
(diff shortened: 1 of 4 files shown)


bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/view/obs/ObsListView.scala
```diff
@@ -322,7 +322,7 @@ class ObsListView(shellAdvisor:ShellAdvisor, band:Band) extends BorderPanel with
             case n  => Proposal.targets.set(p0, ts.updated(n, newT))
           }
         }
-
+
```
cquiroz: Why the empty spaces?

sraaphorst: I have no idea... I guess that's what my editor wanted. (I'm using IntelliJ IDEA Pro.)
Looks like I have invoked the wrath of Travis, so in the process of fixing that, I'll remove the spaces as well.

sraaphorst:
Travis takes forever...

PR #1563, opened 2018-08-21
WIP REL-361 Cron job for expiring timing windows

jdnavarro:
This cron job obtains the basic data structure that will be used to figure out when to send emails notifying about timing window expirations.

There are some potential edge cases that we still need to discuss with the scientists to make sure it's worth going ahead but, assuming we get the green light, these are the remaining tasks needed for merged.

- [ ] Filter programs with the timing windows about to expire.
- [ ] Format email body with the right parameters.
- [x] Keep state of whether an email was already sent or not. (There is a workaround to avoid this)
- [ ] Logic to actually send the email.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/obscomp/SPSiteQuality.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/obscomp/SPSiteQuality.java
@@ -105,7 +105,8 @@ public class SPSiteQuality extends AbstractDataObject implements PropertyProvide
         }
 
         public Option<Instant> getEnd() {
-            return this.repeat == -1 ? None.instance() : new Some<>(Instant.ofEpochMilli(this.start + this.repeat * this.period + this.duration));
+            return repeat == REPEAT_FOREVER || duration == WINDOW_REMAINS_OPEN_FOREVER ?
+                None.instance() : new Some<>(Instant.ofEpochMilli(start + repeat * period + duration));
         }
 
         ParamSet getParamSet(final PioFactory factory) {
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/obscomp/SPSiteQuality.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/obscomp/SPSiteQuality.java
@@ -21,6 +21,7 @@ import edu.gemini.spModel.type.SpTypeUtil;
 
 import java.beans.PropertyDescriptor;
 import java.io.Serializable;
+import java.time.Instant;
 import java.util.*;
 import java.util.function.Supplier;
 
@@ -101,6 +102,10 @@ public class SPSiteQuality extends AbstractDataObject implements PropertyProvide
 
         public long getStart() {
             return start;
+        }
+
+        public Option<Instant> getEnd() {
+            return this.repeat == -1 ? None.instance() : new Some<>(Instant.ofEpochMilli(this.start + this.repeat * this.period + this.duration));
         }
 
         ParamSet getParamSet(final PioFactory factory) {
--- a/bundle/edu.gemini.spdb.reports.collection/src/main/scala/edu/gemini/dbTools/timingwindowcheck/TimingWindowCheckCron.scala
+++ b/bundle/edu.gemini.spdb.reports.collection/src/main/scala/edu/gemini/dbTools/timingwindowcheck/TimingWindowCheckCron.scala
@@ -1,0 +1,34 @@
+package edu.gemini.dbTools.timingwindowcheck
+
+import org.osgi.framework.BundleContext
+import java.io.File
+import java.security.Principal
+import java.time.Instant
+import java.util.logging.{Level, Logger}
+
+import scalaz._
+
+object TimingWindowCheckCron {
+
+  case class ActionLogger(logger: Logger) {
+    def log(level: Level, msg: String): Action[Unit] =
+      Action.delay(logger.log(level, msg))
+  }
+
+  /** Cron job entry point.  See edu.gemini.spdb.cron.osgi.Activator. */
+  def run(ctx: BundleContext)(tmpDir: File, logger: Logger, env: java.util.Map[String, String], user: java.util.Set[Principal]): Unit = {
+
+    val action = for {
+      env  <- TimingWindowCheckEnv.fromBundleContext(ctx)
+      now  <- Action.delay(Instant.now)
+      pids <- TimingWindowFunctor.query(env.odb, user)
+      // _ <- sendEmails...
+    } yield ()
+
+    action.run.unsafePerformIO() match {
+      case -\/(ex) => logger.log(Level.WARNING, "Error executing TimingWindowCheckCron", ex)
+      case \/-(_)  => logger.info("TimingWindowCheckCron complete")
+    }
+  }
+
+}
--- a/bundle/edu.gemini.spdb.reports.collection/src/main/scala/edu/gemini/dbTools/timingwindowcheck/TimingWindowCheckCron.scala
+++ b/bundle/edu.gemini.spdb.reports.collection/src/main/scala/edu/gemini/dbTools/timingwindowcheck/TimingWindowCheckCron.scala
@@ -6,7 +6,12 @@ import java.security.Principal
 import java.time.Instant
 import java.util.logging.{Level, Logger}
 
+import edu.gemini.skycalc.{Interval, TwilightBoundType, TwilightBoundedNight, Union}
 import scalaz._
+import Scalaz._
+import edu.gemini.dbTools.mailer.ProgramAddresses
+import edu.gemini.pot.spdb.IDBDatabaseService
+import edu.gemini.spModel.core.Site
 
 object TimingWindowCheckCron {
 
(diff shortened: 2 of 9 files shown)


bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/obscomp/SPSiteQuality.java
```diff
@@ -103,6 +104,10 @@ public long getStart() {
             return start;
         }
 
+        public Option<Instant> getEnd() {
+            return this.repeat == -1 ? None.instance() : new Some<>(Instant.ofEpochMilli(this.start + this.repeat * this.period + this.duration));
+        }
```
swalker2m: A couple notes on this:

* Compare the repeat count to `TimingWindow.REPEAT_FOREVER` instead of directly `-1` so that it is clearer what is going on
* If the `duration == TimingWindow.WINDOW_REMAINS_OPEN_FOREVER` (which is just `-1`), then there will also be no end

bundle/edu.gemini.spdb.reports.collection/src/main/scala/edu/gemini/dbTools/timingwindowcheck/TimingWindowCheckMailer.scala
```diff
@@ -0,0 +1,35 @@
...
+
+sealed abstract class TimingWindowCheckMailer(mailer: Mailer) {
+
+  def notifyPendingCheck(
+    pid:   SPProgramID,
+    to:    ProgramAddresses
+  ): Action[Unit] = {
+
+    val subject   =
+      s"${pid.stringValue} Timing Window Still Pending"
```
swalker2m: Probably we'll want something like `"Timing window  closing in ${pid.stringValue}"`.  Science may be able to suggest better wording.

bundle/edu.gemini.spdb.reports.collection/src/main/scala/edu/gemini/dbTools/timingwindowcheck/TimingWindowFunctor.scala
```diff
@@ -0,0 +1,65 @@
...
+private class TimingWindowFunctor extends DBAbstractQueryFunctor {
+  import TimingWindowFunctor.{ isActive, isScience }
+
+  val results: Buffer[(SPProgramID, List[(SPObservationID, Instant)])] = Buffer.empty
+
+  override def execute(db: IDBDatabaseService, node: ISPNode, principals: JSet[Principal]): Unit = {
+
+    def findEnd(o: ISPObservation): Option[(SPObservationID, Instant)] = {
+      // TODO: Will the null case of getDataObject make it an empty list?
+      val tws: List[TimingWindow] = SPTreeUtil.findObsCondNode(o).getDataObject.asInstanceOf[SPSiteQuality].getTimingWindows.asScala.toList
```
swalker2m: You should be prepared for a `null` result from `findObsCondNode(o)`.  The model and the OT allow you to delete the observing conditions / site quality node.  If it is `null` then an empty `List` makes sense.  Same for a `null` data object.  That's less likely and the ODB and OT will probably explode anyway in that case but you can handle `null` data object if you want to be complete.

bundle/edu.gemini.spdb.reports.collection/src/main/scala/edu/gemini/dbTools/timingwindowcheck/TimingWindowFunctor.scala
```diff
@@ -0,0 +1,65 @@
...
+      tws.flatMap(_.getEnd.asScala).maximumBy(_.toEpochMilli).map(tw => (o.getObservationID, tw))
+    }
+
+    node match {
+      case p: ISPProgram => if (isActive(p) && isScience(p)) results += ((p.getProgramID, p.getAllObservations.asScala.toList.flatMap(findEnd)))
```
swalker2m: This relies on an implicit conversion from `Option` to `List` that we should probably make explicit with `Option.toList` as in `.flatMap(o => findEnd(o).toList)`.  Also, I think there will be a good number of programs that have no observations with a finite timing window.  In other words, the `List` associated with the program id is often going to be empty.  I think we should skip those results since we're just interested in finding observations with a finite timing window.

bundle/edu.gemini.spdb.reports.collection/src/main/scala/edu/gemini/dbTools/timingwindowcheck/package.scala
```diff
@@ -0,0 +1,39 @@
...
+
+import scalaz._
+import Scalaz._
+import edu.gemini.pot.sp.SPObservationID
+import edu.gemini.spModel.core.SPProgramID
+import scalaz.effect.IO
+
+package object timingwindowcheck {
+
+  type Action[A] = EitherT[IO, Throwable, A]
```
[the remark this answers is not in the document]
swalker2m: I believe I did it this way, with a transformer, for the other email job but you're right.  It's probably not necessary.

bundle/edu.gemini.spdb.reports.collection/src/main/scala/edu/gemini/dbTools/timingwindowcheck/TimingWindowCheckMailer.scala
```diff
@@ -0,0 +1,35 @@
...
+import edu.gemini.spModel.core.{ Site, SPProgramID }
+
+import scalaz._
+import Scalaz._
+import scalaz.effect.IO
+
+sealed abstract class TimingWindowCheckMailer(mailer: Mailer) {
+
+  def notifyPendingCheck(
+    pid:   SPProgramID,
```
swalker2m: When it comes time to write the body of the email, we'll also want to know the list of observations whose window closing is coming up.

bundle/edu.gemini.spdb.reports.collection/src/main/scala/edu/gemini/dbTools/timingwindowcheck/TimingWindowCheckCron.scala
```diff
@@ -15,14 +20,56 @@ object TimingWindowCheckCron {
...
+          }
+        }
+    }.void
+
   /** Cron job entry point.  See edu.gemini.spdb.cron.osgi.Activator. */
   def run(ctx: BundleContext)(tmpDir: File, logger: Logger, env: java.util.Map[String, String], user: java.util.Set[Principal]): Unit = {
 
+    def getCheckUnion(now: Instant, site: Site): Union[Interval] = {
+
+      val night = TwilightBoundedNight.forInstant(TwilightBoundType.NAUTICAL, now, site)
```
swalker2m: I think if this is called during the night, it will get you the night you want.  On the other hand if called during the day, it will get you the coming night.

bundle/edu.gemini.spdb.reports.collection/src/main/scala/edu/gemini/dbTools/timingwindowcheck/TimingWindowFunctor.scala
```diff
@@ -33,7 +33,7 @@ object TimingWindowFunctor {
 
   private def isActive(p: ISPProgram): Boolean = {
     val obj = p.getDataObject.asInstanceOf[SPProgram]
-    obj.isActive && !obj.isCompleted
+    if (obj == null) { false } else obj.isActive && !obj.isCompleted
```
swalker2m: You can combine a `null` check and a type check in a `match`:

```scala
Option(p.getDataObject) match {
  case Some(obj: SPProgram) => obj.isActive && !obj.isCompleted // && obj.optInTimingWindowWarnings
  case _                    => false
}

File: bundle/edu.gemini.spdb.reports.collection/src/main/scala/edu/gemini/dbTools/timingwindowcheck/TimingWindowFunctor.scala [hunk 9]
```diff
@@ -47,7 +47,7 @@ object TimingWindowFunctor {
 private class TimingWindowFunctor extends DBAbstractQueryFunctor {
   import TimingWindowFunctor.{ isActive, isScience }
 
-  val results: Buffer[(SPProgramID, List[(SPObservationID, Instant)])] = Buffer.empty
+  var results: Buffer[(SPProgramID, List[(SPObservationID, Instant)])] = Buffer.empty
```
swalker2m: The idea with the `var` was to change to an immutable data type like `List`.  If you stick with `Buffer` you don't need `var`.

bundle/edu.gemini.spdb.reports.collection/src/main/scala/edu/gemini/dbTools/timingwindowcheck/TimingWindowCheckMailer.scala
```diff
@@ -1,17 +1,20 @@
...
 import scalaz.effect.IO
 
 sealed abstract class TimingWindowCheckMailer(mailer: Mailer) {
 
   def notifyPendingCheck(
-    pid:   SPProgramID,
-    to:    ProgramAddresses
+    pid: SPProgramID,
+    to:  ProgramAddresses,
+    tws: NonEmptyList[(SPObservationID, Instant)]
```
swalker2m: I don't think we need the actual `Instant` here.  We just want to mention the observations that have expired timing windows.

bundle/edu.gemini.spdb.reports.collection/src/main/scala/edu/gemini/dbTools/timingwindowcheck/package.scala
```diff
@@ -0,0 +1,39 @@
...
+
+    def delay[A](a: => A): Action[A] =
+      EitherT.right(IO(a))
+
+    def fail[A](failMessage: => String): Action[A] =
+      EitherT.left(IO(new RuntimeException(failMessage)))
+
+  }
+
+  type TimingWindows = Map[SPProgramID, List[(SPObservationID, Instant)]]
```
swalker2m: This name is a bit misleading because we don't have timing windows here but rather the instant that the last timing window expires.  Also, do we need a `Map` here?  I don't think we ever do any lookup by program id?

PR #1566, opened 2018-08-24
Fixes for TPE image handling.

tpolecat:
These are some fixes to bugs I found when prototyping a mosaic service for 2MASS images.

I tried to open a Montage-generated FITS image and immediately ran into destruction.

- `ImageDisplayControlFrame` and `ImageDisplayControl` need to be abstract … the constructor does some downcasting that only works if you're using the one and only subclass `TpeImageDisplayFrame`. So I made them abstract and moved the non-functioning `main` method.
- `FITSDataDouble` was downcasting to a JAI class that has since appeared in `java.awt`, I guess. Changing it to use the AWT class instead made it work. No idea.
- We couldn't deal with FITS images with double-precision data so I copy/pasted the handler for `Float` to deal with it. Seems to work.

**Late addition:** I also noticed that every image request was getting sent twice, once for the headers (to determine the content type and encoding) and again to fetch the actual data. So I fixed that.

### DIFF
--- a/bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/image/ImageCatalogClient.scala
+++ b/bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/image/ImageCatalogClient.scala
@@ -160,7 +160,7 @@ object ImageCatalogClient {
     */
   def downloadImageToFile(cacheDir: Path, url: URL, query: ImageSearchQuery): Task[ImageInFile] = {
 
-    case class ConnectionDescriptor(contentType: Option[String], contentEncoding: Option[String]) {
+    case class ConnectionDescriptor(contentType: Option[String], contentEncoding: Option[String], stream: InputStream) {
       def extension: String = (contentEncoding, contentType) match {
           case (Some("x-gzip"), _)                                                              => "fits.gz"
           // REL-2776 At some places on the sky DSS returns an error, the HTTP error code is ok but the body contains no image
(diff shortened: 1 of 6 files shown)


bundle/jsky.app.ot/src/main/java/jsky/image/operator/CutLevelOpImage.java
```diff
@@ -154,8 +147,19 @@ protected void accumulateStatistics(String name, Raster source, Object ar) {
                 }
                 break;
 
+            case DataBuffer.TYPE_DOUBLE:
```
swalker2m: I think you can just add this below the `TYPE_FLOAT` case instead of repeating it, unless you think it is clearer this way?

```java
case DataBuffer.TYPE_FLOAT:
case DataBuffer.TYPE_DOUBLE:
    {
        ...
    }

Comment (swalker2m) [reply bundle/jsky.app.ot/src/main/java/jsky/image/operator/CutLevelOpImage.java]: Never mind, I guess it does call a different method in the end.  I didn't see that at first.

Comment (tpolecat) [reply bundle/jsky.app.ot/src/main/java/jsky/image/operator/CutLevelOpImage.java]: Yeah … primitives, yay!

**Comment by tpolecat** [comment]
> - `FITSDataDouble` was downcasting to a JAI class that has since appeared in `java.awt`, I guess. Changing it to use the AWT class instead made it work. No idea.

I'm a little worried about changing one downcast to another since there's no obvious way to predict what's JAI is going to return for various image types. I'm going to leave this open and look a little more closely, or at least add some tests.

**Comment by tpolecat** [comment]
This was part of REL-1093 which Bryan doesn't think is worth pursuing right now, so I'm closing this.

**Comment by tpolecat** [comment]
Re-opening this. I found another bug (see the description above). I exercised this a lot and I think it's ok.

File: bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/image/ImageCatalogClient.scala [hunk 2]
```diff
@@ -179,29 +179,14 @@ object ImageCatalogClient {
...
     }
 
     /**
       * Downloads the url and writes it to a file
       */
-    def writeToTempFile(file: File): Task[Unit] = Task.delay {
-      // Simple old fashioned download
-      val out = new FileOutputStream(file)
-      try {
-        val in = url.openStream()
```
tpolecat: This line was opening a new connection even though the caller already has an open one.

PR #1569, opened 2018-08-28
REL-1958: ICTD / QV Integration

swalker2m:
This PR adds minimal ICTD integration with the QV tool.  In particular, it adds a column to the observation table listing that indicates whether each observation's requested instrument components (and custom mask if any) are installed.  It also adds a somewhat redundant option to highlight observation rows according to the availability of their components.

Because the QV tool is distributed with the OT, I was hesitant to allow the OT to directly query the ICTD. To do so would have meant distributing an ICTD user and password with the OT. Instead I created a service that runs in the ODB and receives queries on behalf of the QV forwarding the results back to the OT.  Note that the QPT was written to directly query the ICTD so I also refactored it to use the new `IctdService` just as the OT/QV now do.

### DIFF
--- a/bundle/edu.gemini.qpt.shared/src/main/java/edu/gemini/qpt/shared/util/Ictd.java
+++ b/bundle/edu.gemini.qpt.shared/src/main/java/edu/gemini/qpt/shared/util/Ictd.java
@@ -17,18 +17,15 @@ import edu.gemini.spModel.pio.PioFactory;
 import edu.gemini.util.security.auth.keychain.KeyChain;
 import edu.gemini.util.trpc.client.TrpcClient$;
 
-import java.io.IOException;
 import java.lang.reflect.UndeclaredThrowableException;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.logging.Level;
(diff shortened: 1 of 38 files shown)


bundle/edu.gemini.qpt.shared/src/main/java/edu/gemini/qpt/shared/util/Ictd.java
```diff
@@ -1,80 +1,61 @@
...
 
     /**
      * Constructs an ICTD instance from query results, if possible. Returns a
      * Right<String, Ictd> if so, and a Left<String, Ictd> if not.
      */
-    public static ImEither<String, Ictd> query(SiteConfig dbs, Site s) {
-        final IctdDatabase.Configuration c = dbs.configFor(s);
+    public static ImEither<String, IctdSummary> query(KeyChain kc, Peer peer, Site s) {
+
+        final ClassLoader cl = Thread.currentThread().getContextClassLoader();
```
cquiroz: Long time i haven't seen class loaders 😄

swalker2m:
Note there are corresponding updates to `OcsCredentials.scala` from svn.

PR #1570, opened 2018-08-29
REL-1958: show which configuration items are missing

swalker2m:
This PR implements an additional feature request for REL-1958: ICTD / QV integration.  Namely, it shows *which* instrument components in an observation configuration are not installed in a tooltip popup.

### DIFF
--- a/bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/table/ObservationTable.scala
+++ b/bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/table/ObservationTable.scala
@@ -3,8 +3,8 @@ package edu.gemini.qv.plugin.table
 import java.awt.Color
 import java.awt.event.{AdjustmentEvent, AdjustmentListener}
 import java.util.regex.PatternSyntaxException
-import javax.swing.table.{TableCellRenderer, DefaultTableCellRenderer, TableColumn, TableRowSorter}
-import javax.swing.{JTable, BorderFactory, RowFilter, SwingConstants}
+import javax.swing.table.{DefaultTableCellRenderer, TableColumn, TableRowSorter}
+import javax.swing._
 
 import edu.gemini.pot.sp.SPObservationID
 import edu.gemini.qv.plugin.data._
@@ -20,6 +20,7 @@ import scala.None
 import scala.collection.mutable
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.Future
+import scala.swing.Action
 import scala.swing.GridBagPanel.Fill._
 import scala.swing.ScrollPane.BarPolicy._
 import scala.swing._
@@ -185,10 +186,8 @@ class ObservationTable(ctx: QvContext) extends GridBagPanel { ui =>
     peer.setDefaultRenderer(classOf[DecValue], DecValueRenderer)
     peer.setDefaultRenderer(classOf[TimeValue], TimeValueRenderer)
 
-    // idk why this doesn't work ...
-    //    peer.setDefaultRenderer(classOf[InstallationState], InstallationStateRenderer)
     peer.setDefaultRenderer(InstallationState.AllInstalled.getClass,     InstallationStateRenderer)
-    peer.setDefaultRenderer(InstallationState.SomeNotInstalled.getClass, InstallationStateRenderer)
+    peer.setDefaultRenderer(classOf[InstallationState.SomeNotInstalled], InstallationStateRenderer)
     peer.setDefaultRenderer(InstallationState.Unknown.getClass,          InstallationStateRenderer)
 
     peer.setComponentPopupMenu(popup.peer)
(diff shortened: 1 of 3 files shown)


bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/table/ObservationTable.scala
```diff
@@ -185,10 +186,8 @@ class ObservationTable(ctx: QvContext) extends GridBagPanel { ui =>
     peer.setDefaultRenderer(classOf[DecValue], DecValueRenderer)
     peer.setDefaultRenderer(classOf[TimeValue], TimeValueRenderer)
 
-    // idk why this doesn't work ...
-    //    peer.setDefaultRenderer(classOf[InstallationState], InstallationStateRenderer)
     peer.setDefaultRenderer(InstallationState.AllInstalled.getClass,     InstallationStateRenderer)
-    peer.setDefaultRenderer(InstallationState.SomeNotInstalled.getClass, InstallationStateRenderer)
+    peer.setDefaultRenderer(classOf[InstallationState.SomeNotInstalled], InstallationStateRenderer)
```
cquiroz: So, there is a difference

bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/table/ObservationTable.scala
```diff
@@ -322,18 +321,37 @@ class ObservationTable(ctx: QvContext) extends GridBagPanel { ui =>
...
           })
         }
 
+        // Set an informative tooltip message for observations with missing config..
+        comp.asInstanceOf[JComponent].setToolTipText(is match {
+          case SomeNotInstalled(ms) =>
+            s"""<html>
+              |<b>Missing configuration:</b>
+              |
+              |${ms.list.toList.mkString("<ul><li>", "</li><li>", "</li></ul>")}
```
cquiroz: cool!

bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/table/ObservationTable.scala
swalker2m: I don't think I fully understand all this but:

* `classOf` expects a type so you cannot get the `classOf[AllInstalled]` because `AllInstalled` is an `object`. 

```
scala> classOf[AllInstalled]
<console>:17: error: not found: type AllInstalled
       classOf[AllInstalled]
               ^
```

* Now, I thought its type would be `AllInstalled.type` 

```
scala> val a: AllInstalled.type = AllInstalled
a: edu.gemini.qv.plugin.table.InstallationState.AllInstalled.type = AllInstalled

scala> val b = AllInstalled
b: edu.gemini.qv.plugin.table.InstallationState.AllInstalled.type = AllInstalled
```

* but `classOf[AllInstalled.type]` doesn't work.  I'm not sure why.

```
scala> classOf[AllInstalled.type]
<console>:17: error: class type required but edu.gemini.qv.plugin.table.InstallationState.AllInstalled.type found
       classOf[AllInstalled.type]
```

* On the other hand, you can get the class of the singleton object, and that works fine:
```
scala> AllInstalled.getClass
res7: Class[_ <: edu.gemini.qv.plugin.table.InstallationState.AllInstalled.type] = class edu.gemini.qv.plugin.table.InstallationState$AllInstalled$
```

* `SomeNotInstalled.getClass` I assume is the class of the companion object because `SomeNotInstalled` is now a `case class`.  That isn't what we want.  We store instances of `SomeNotInstalled` and need the associated renderer for those, not the companion.

```
scala> SomeNotInstalled.getClass
res6: Class[_ <: edu.gemini.qv.plugin.table.InstallationState.SomeNotInstalled.type] = class edu.gemini.qv.plugin.table.InstallationState$SomeNotInstalled$
```

* `SomeNotInstalled` is also a type though so `classOf[SomeNotInstalled]` obtains the associated `Class` that we want.

```
scala> classOf[SomeNotInstalled]
res5: Class[edu.gemini.qv.plugin.table.InstallationState.SomeNotInstalled] = class edu.gemini.qv.plugin.table.InstallationState$SomeNotInstalled
```

* `peer.setDefaultRenderer(classOf[InstallationState], ...)` doesn't work for all three instances as I initially assumed.  If I look at `AllInstalled.getClass.getSuperclass` though I see it is just `java.lang.Object` and not `InstallationState`.  I'm guessing if I made `InstallationState` an `abstract class` instead of a `trait` then this *would* work.

```
scala> AllInstalled.getClass.getSuperclass
res8: Class[?0] forSome { type ?0 >: ?0; type ?0 <: edu.gemini.qv.plugin.table.InstallationState.AllInstalled.type } = class java.lang.Object
```

tpolecat: There's a good article [here](https://typelevel.org/blog/2017/02/13/more-types-than-classes.html) that talks about types and classes. I didn't realize singletons types weren't class types. Anyway it's nice that we don't have to do this very often because it's terrible.

PR #1571, opened 2018-08-31
Feature/rel 3188

andrewwstephens:
This trivial PR modifies @olesjasmirnova 's work to include the coadds in the ITC output page.

There is still a bug in that the "Intermediate S/N for one exposure" is always zero.
This comes from line 91 of CalculatablePrinter.java
where singleSNRatio() is defined on line 91 of ImagingS2NCalculation.java
where coadds appears to be undefined.
I am leaving this bug to the professionals.


sraaphorst:
Hi Andy,

Looking at this now. It appears that in `ImagingS2NCalculation`, `coadds` is declared but never set to anything: because of this, Java initializes it to 0, which explains why the method:
```
public double singleSNRatio() {
        return Math.sqrt(coadds) * signal / noise;
    }
```
returns 0. Do you know what this should be set to, or where it comes from?

I'll poke around and see if I can figure it out.

sraaphorst:
It seems like the `ObservationDetails`  passed into the constructor for `ImagingS2NCalculation` has a `coadds` method that returns `Option[Int]` based on the `CalculationMethod` that is being used, so I suspect this will be an easy enough fix if we grab the `coadds` from there and set them in the `ImagingS2NCalculation` constructor.

BTW, `ObservationDetails`  is defined in the file `Parameters.java`, L102.

The `coadds` parameter in `ObservationDetails` comes from earlier classes in `Parameters.java`, namely `ImagingS2N`, `ImagingInt`, and `SpectroscopyS2N`.

My only question for you, then, is if the coadds parameter is set to `None` in one of these three (which seems to happen occasionally, from what I can tell, only in the test `Fixtures`, so we might be alright), what should we take for its value? Zero? One?

andrewwstephens:
Hi Sebastian,
I used odp.calculationMethod().coaddsOrElse(1) to ge tthe number of coadds.
-Andy

sraaphorst:
Hi Andy,

I suspect that adding this line to the end of the `ImagingS2NCalculation` constructor will fix the problem:

```coadds = obs.coadds().getOrElse(() -> 1);```

If you like, I can try  that and make another ITC test release?

andrewwstephens:
Either way is fine with me.
Are these equivalent?
obs.calculationMethod().coaddsOrElse(1)
obs.coadds().getOrElse(() -> 1)

sraaphorst:
I just checked, and yes, they are equivalent! I'm not sure why, then, that we have the coadds in `ObservationDetails` - or really, any methods at all - since they all defer to the `CalculationMethod`, which we can just grab from the `ObservationDetails` like you're doing.

So yes, I would go with your code instead... I'll see if I can check out this branch, make the change, and then  set up a test release on itcdev. I'll let you know when I'm done.

sraaphorst:
I see you made the change. Thanks 👍 . I'll go ahead and do the build.

PR #1572, opened 2018-09-07
REL-2719: indicate guide quality in TPE candidate plot

swalker2m:
This is a collaborative effort with @cquiroz to change the way that guide star candidates are plotted in the TPE.  The idea is to indicate, using green, orange, and red colors for the guide star circle, the quality of the candidate regardless of whether it is in range of the guide probe.   In particular:

* Green Circle - Delivers requested IQ or possible degradation 
* Orange Circle - Definite degradation
* Red Circle - Possibly unusable 
* Red X - Definitely unusable

(Though I wonder what is the point of plotting unusable options.  Perhaps they should just be filtered out?)

after

Because the usability icon no longer incorporates reachability, the Catalog Query Tool has also grown a new column to indicate reachability.

catalogtable

(Though it bothers me that the pie chart icon color is always green instead of matching the TPE plot color.)

### DIFF
--- a/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsAnalysis.scala
+++ b/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsAnalysis.scala
@@ -156,7 +156,11 @@ object AgsAnalysis {
     else magnitudeAnalysis(ctx, mt, guideProbe, guideStar)
   }
 
-  private def magnitudeAnalysis(ctx: ObsContext, mt: MagnitudeTable, guideProbe: ValidatableGuideProbe, guideStar: SiderealTarget): Option[AgsAnalysis] = {
+  /**
+   * Analysis of the suitability of the magnitude of the given guide star
+   * regardless of its reachability.
+   */
+  def magnitudeAnalysis(ctx: ObsContext, mt: MagnitudeTable, guideProbe: ValidatableGuideProbe, guideStar: SiderealTarget): Option[AgsAnalysis] = {
     import AgsGuideQuality._
     import GuideSpeed._
 
--- a/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsAnalysis.scala
+++ b/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsAnalysis.scala
@@ -211,3 +211,19 @@ object AgsAnalysis {
     }
   }
 }
+
+sealed trait GuideInFOV
+case object InsideFOV extends GuideInFOV
+case object OutsideFOV extends GuideInFOV
+
+object GuideInFOV {
+  implicit val order: Order[GuideInFOV] = Order.order {
+    case (InsideFOV, OutsideFOV) => Ordering.GT
+    case (OutsideFOV, InsideFOV) => Ordering.LT
+    case _                       => Ordering.EQ
+  }
+
+  implicit val ordering: scala.Ordering[GuideInFOV] = order.toScalaOrdering
+
+  val All: List[GuideInFOV] = List(InsideFOV, OutsideFOV)
+}
--- a/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsAnalysis.scala
+++ b/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsAnalysis.scala
@@ -61,41 +61,41 @@ sealed trait AgsAnalysis {
 }
 
 object AgsAnalysis {
-  case class NoGuideStarForProbe(guideProbe: GuideProbe) extends AgsAnalysis {
+  final case class NoGuideStarForProbe(guideProbe: GuideProbe) extends AgsAnalysis {
     override def message(withProbe: Boolean): String = {
       val p = if (withProbe) s"${guideProbe.getKey} " else ""
       s"No ${p}guide star selected."
     }
   }
 
-  case class NoGuideStarForGroup(guideGroup: GuideProbeGroup) extends AgsAnalysis {
+  final case class NoGuideStarForGroup(guideGroup: GuideProbeGroup) extends AgsAnalysis {
     override def message(withProbe: Boolean): String =
       s"No ${guideGroup.getKey} guide star selected."
   }
 
-  case class MagnitudeTooFaint(guideProbe: GuideProbe, target: SiderealTarget) extends AgsAnalysis {
+  final case class MagnitudeTooFaint(guideProbe: GuideProbe, target: SiderealTarget) extends AgsAnalysis {
     override def message(withProbe: Boolean): String = {
       val p = if (withProbe) s"use ${guideProbe.getKey}" else "guide"
       s"Cannot $p with the star in these conditions, even using the slowest guide speed."
     }
   }
 
-  case class MagnitudeTooBright(guideProbe: GuideProbe, target: SiderealTarget) extends AgsAnalysis {
+  final case class MagnitudeTooBright(guideProbe: GuideProbe, target: SiderealTarget) extends AgsAnalysis {
     override def message(withProbe: Boolean): String = {
       val p = if (withProbe) s"${guideProbe.getKey} g" else "G"
       s"${p}uide star is too bright to guide."
     }
   }
 
-  case class NotReachable(guideProbe: GuideProbe, target: SiderealTarget) extends AgsAnalysis {
+  final case class NotReachable(guideProbe: GuideProbe, target: SiderealTarget) extends AgsAnalysis {
     override def message(withProbe: Boolean): String = {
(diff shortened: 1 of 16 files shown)


tpolecat:
Yeah, under what circumstances would you try a probably/definitely unusable guidestar? What do you think @andrewwstephens ?

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsAnalysis.scala
```diff
@@ -210,4 +214,20 @@ object AgsAnalysis {
       analysisOpt.getOrElse(NoMagnitudeForBand(guideProbe, guideStar))
     }
   }
-}
\ No newline at end of file
+}
+
+sealed trait GuideInFOV
+case object InsideFOV extends GuideInFOV
+case object OutsideFOV extends GuideInFOV
```
tpolecat: You might consider putting these on the companion so you could say `GuideInFOV.Inside` and `GuideInFOV.Outside`. Extremely minor but I'm liking that style better these days.

bundle/jsky.app.ot/src/main/scala/edu/gemini/catalog/ui/QueryResultsFrame.scala
```diff
@@ -69,12 +72,27 @@ object QueryResultsFrame extends Frame with PreferredSizeFrame {
     }
   }
 
+  private object InFOVRenderer {
+    val inFovIcon: ImageIcon = Resources.getIcon("bullet/bullet_green.png")
+    val outFovIcon: ImageIcon = Resources.getIcon("bullet/bullet_red.png")
+    val icons: Map[GuideInFOV, ImageIcon] = Map(InsideFOV -> inFovIcon, OutsideFOV -> outFovIcon)
```
tpolecat: Any reason to not have a [total] method here? Do we ever do anything other than `.get`?

bundle/jsky.app.ot/src/main/scala/edu/gemini/catalog/ui/QueryResultsFrame.scala
```diff
@@ -87,11 +105,54 @@ object QueryResultsFrame extends Frame with PreferredSizeFrame {
...
+
+                if (updated) {
+                  val info = ObservationInfo(c, mt)
+                  val tm   = TargetsModel(info.some, q.base, q.radiusConstraint, curModel.targets)
+                  updateResultsModel(tm)
+
+                  // Update the plotter's table list so that the selection link
+                  // between the table rows and the plot continues to work.
+                  iw.plotter.unplotAll()
+                  iw.plotter.plot(TableQueryResultAdapter(tm))
```
tpolecat: TableQueryResultAdapter™

bundle/jsky.app.ot/src/main/scala/edu/gemini/catalog/ui/QueryResultsFrame.scala
```diff
@@ -382,8 +448,10 @@ object QueryResultsFrame extends Frame with PreferredSizeFrame {
           }
       }
     }
-    lazy val catalogBox = new ComboBox(List[CatalogName](UCAC4, PPMXL)) with TextRenderer[CatalogName] {
-      override def text(a: CatalogName) = ~Option(a).map(_.displayName)
+    lazy val catalogBox: ComboBox[CatalogName] with TextRenderer[CatalogName] {
+      def text(a: CatalogName): String
+    } = new ComboBox(List[CatalogName](UCAC4, PPMXL)) with TextRenderer[CatalogName] {
+      override def text(a: CatalogName): String = ~Option(a).map(_.displayName)
```
tpolecat: I prefer `Option(a).foldMap(_.displayName)` in cases like this.

bundle/jsky.app.ot/src/main/scala/edu/gemini/catalog/ui/QueryResultsFrame.scala
```diff
@@ -457,10 +531,12 @@ object QueryResultsFrame extends Frame with PreferredSizeFrame {
...
       }
 
-      override def text(a: SPSiteQuality.SkyBackground) = a.displayValue()
+      override def text(a: SPSiteQuality.SkyBackground): String = a.displayValue()
     }
 
-    lazy val ccBox = new ComboBox(List(SPSiteQuality.CloudCover.values().filter(!_.isObsolete): _*)) with TextRenderer[SPSiteQuality.CloudCover] {
+    lazy val ccBox: ComboBox[SPSiteQuality.CloudCover] with TextRenderer[SPSiteQuality.CloudCover] {
+      def text(a: SPSiteQuality.CloudCover): String
+    } = new ComboBox(List(SPSiteQuality.CloudCover.values().filter(!_.isObsolete): _*)) with TextRenderer[SPSiteQuality.CloudCover] {
```
tpolecat: `.filterNot`?

bundle/jsky.app.ot/src/main/scala/edu/gemini/catalog/ui/QueryResultsFrame.scala
```diff
@@ -335,8 +397,10 @@ object QueryResultsFrame extends Frame with PreferredSizeFrame {
         }
     }
 
-    lazy val objectName = new TextField("") with SelectOnFocus {
-      val foregroundColor = UIManager.getColor("TextField.foreground")
+    lazy val objectName: TextField with SelectOnFocus {
+      def resetName(n: String): Unit
+    } = new TextField("") with SelectOnFocus {
```
tpolecat: All these refinements are slightly sketchy because it's so easy to crush them down to non-refined types, but I guess if it works here it's fine. The alternative would be to declare a bunch of single-use traits I guess.

bundle/jsky.app.ot/src/main/scala/edu/gemini/catalog/ui/package.scala
```diff
@@ -198,111 +198,123 @@ abstract class CatalogNavigatorColumn[T >: Null: Manifest] {
...
-  def ordering = implicitly[scala.math.Ordering[AgsGuideQuality]]
+  def ordering: scala.Ordering[AgsGuideQuality] = implicitly[scala.math.Ordering[AgsGuideQuality]]
 }
 
+case class InFOVColumn(info: Option[ObservationInfo], title: String) extends CatalogNavigatorColumn[GuideInFOV] {
+  val gf: SiderealTarget @?> GuideInFOV = PLens{ t => GuidingQualityColumn.target2FOV(info, t).map(p => Store(_ => sys.error("Not in use"), p)) }
+
+  override val lens: Target @?> GuideInFOV = PLens(_.fold(
+    PLens.nil.run,
+    gf.run _ andThen (_.map(_.map(x => x: Target))),
```
tpolecat: Depending on the scalaz version I think you can say `_.widen[Target]` here.

bundle/jsky.app.ot/src/main/scala/edu/gemini/catalog/ui/TpePlotter.scala
```diff
@@ -82,46 +82,87 @@ object adapters {
...
   }
 
-  case object PossibleIqDegradationSymbol extends GuideQualitySymbol {
-    val quality = PossibleIqDegradation.some
+  case object PossibleIqDegradationSymbolOutFOV extends GuideQualitySymbol {
+    val quality: Option[AgsGuideQuality] = PossibleIqDegradation.some
+    val inFOV: Option[GuideInFOV] = OutsideFOV.some
     setFg(Color.green)
-    setShape(TablePlotSymbol.ELLIPSE)
+    setShape(TablePlotSymbol.CIRCLE)
```
cquiroz: I noted it is very hard to tell `CIRCLE` from `ELLIPSE` on the TPE

andrewwstephens:
:+1:  This looks good to me.

@tpolecat - Good question.  There is probably a better way (there usually is), but the main reason to show "unusable" stars is that the guide stars you can use are limited by the conditions specified, but those are not necessarily the *actual* conditions, and we (staff) don't like to modify the PI-defined constraints.  For example, if you are the observer selecting a guide star at the average parallactic angle you may be able to use one of the "unusable" stars if the conditions are better than what is listed in the program.  Also, if we didn't mark all the stars, you wouldn't know if a star is unmarked because it is too faint or because it is missing from the catalog (which does happen).

swalker2m:
Added a few commits to address PR feedback, and [one](https://github.com/gemini-hlsw/ocs/pull/1572/commits/3a242040728f791d6f37c8d845f911d641d17749) to stop clipping the plotted stars to the image bounds.  If you move the base position a bit then the probe range often falls just outside of the image bounds.  That would lead to not showing many candidates that appear in the table.  Since the point of this update to use the manual catalog selection tool to pick alternative guide stars that require a small adjustment, it seemed remiss to not plot all the options.

Testing in a somewhat crowded field of 364 candidates around M1 didn't reveal any performance issue.

offimage

PR #1575, opened 2018-10-02
REL-2718: plot AGS candidates

swalker2m:
This PR updates the "catalog" TPE feature to show AGS catalog results until and unless the manual guide star search tool is visible, in which case manual search results are plotted.  It updates `TpeCatalogFeature` to handle management of unplotting and plotting so that the enabled state of the feature is always respected.  Toward that end, the manual catalog tool "Unplot" / "Plot" button is removed.  To unplot the user instead deselects the catalog feature just as would be done for any other feature.  Because this requires working with Scala, I ported `TpeCatalogFeature` from Java.

The key difficulty has to do with the way that image features and plot symbols are drawn.  Displaying the plot symbols requires a call to the image widget canvas `repaint` which triggers a call to draw all the features which plots candidates which generates a `repaint`, etc.  To break this cycle, a sketchy hash of the "plot state" is maintained and used to avoid continuing when there is nothing new to display.

This is embarrassing.  Other suggestions are welcome.  We can also just shelve these changes.


tpolecat:
No.

tpolecat:
🐔

sraaphorst:
I'm so, so very sorry that this got shuffled off to you from my plate. I'm sure it's worse than anything left over by 9:00 PM at Luby's.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/tpe/feat/TpeCatalogFeature.scala
```diff
@@ -0,0 +1,125 @@
...
+    p: TablePlotter,
+    state: Option[PlotState]
+  ): Unit =
+    for {
+      c <- w.getObsContext.asScalaOpt
+      s <- AgsRegistrar.currentStrategy(c)
+      m  = ProbeLimitsTable.loadOrThrow()
+      q <- s.catalogQueries(c, m).headOption
+    } q match {
+      case cs: ConeSearchCatalogQuery =>
```
tpolecat: This may read better if you fold the pattern-match into the for-comp above; i.e.
```scala
cs <- s.catalogQueries(c, m).headOption.collect { case cs: ConeSearchCatalogQuery => cs }
```

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/tpe/feat/TpeCatalogFeature.scala
```diff
@@ -0,0 +1,125 @@
...
+    AgsHash.hash(c, when)
+  }
+
+  /** PlotState contains all relevant information for the catalog plot. */
+  private final case class PlotState(
+    visible: Boolean,
+    agsHash: Int,
+    manual:  Boolean,
+    targets: Option[TargetsModel]
+  )
```
tpolecat: I don't know if it would help, but this could be an ADT with a `NoPlot` case, rather than having `Option[PlotState]` everywhere. Possibly more cases … are the four variables here really independent?

sraaphorst:
Even Travis said no.

swalker2m:
Thanks for the feedback.  I refactored `PlotState` a bit and, though it doesn't address the underlying sketchiness, it is a bit clearer what is going on.  I can promise that it _often_ works, bordering on _usually_ works perhaps.  In other words, it's about par for the course for the OT.

PR #1578, opened 2018-10-26
SEQNG-776: add planned time to seqexec sequence data

swalker2m:
Updates the `SeqExecService` used by the seqexec to replace the sequence with `PlannedTime`.  `PlannedTime` contains the sequence in addition to the time estimations requested by this task.  You can extract the sequence from the `PlannedTime` instance with `plannedTime.sequence`.


cquiroz:
Will this compile with scala 2.12?

swalker2m:
So this should work, but I'm not sure we should do it.  It will complicate migration to the new sequence and new codebase because we don't have planned time estimates there.  Can we push back on this task?

cquiroz:
Is this change required to implement estimation of sequence time on the new seqexec?

tpolecat:
> Will this compile with scala 2.12?

Yes, the `left`/`right` projections on `Either` still work.

swalker2m:
I'm going to go ahead and shelve this PR since we agreed it would delay migration to the new ODB.

PR #1581, opened 2018-11-05
REL-3519: make twilight calculation consistent

swalker2m:
This PR makes a small but important change to the way that the twilight bounds are calculated for the purpose of adding markers explaining observability limits.  Instead of repeatedly calculating the twilight bounds based on the time of individual allocations, we make it consistent with the schedule caches themselves using the schedule start time as a whole.

This fixes an issue in which an allocation well off of the night into the morning (when adjusted from a previous schedule and used as a template) was triggering an exception.

### DIFF
--- a/bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
+++ b/bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
@@ -8,6 +8,7 @@ import java.util.stream.Collectors;
 
 import edu.gemini.qpt.core.Alloc;
 import edu.gemini.qpt.core.Marker;
+import edu.gemini.qpt.core.Schedule;
 import edu.gemini.qpt.core.Variant;
 import edu.gemini.qpt.core.Alloc.Circumstance;
 import edu.gemini.qpt.core.Marker.Severity;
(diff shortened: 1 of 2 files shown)


bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
```diff
@@ -40,10 +41,15 @@
...
 
     public void propertyChange(PropertyChangeEvent evt) {
-        Variant v = (Variant) evt.getSource();
-        MarkerManager mm = v.getSchedule().getMarkerManager();
+        final Variant  v               = (Variant) evt.getSource();
+        final Schedule schedule        = v.getSchedule();
+        final MarkerManager mm         = schedule.getMarkerManager();
         mm.clearMarkers(this, v);
 
+        final TwilightBoundedNight tbn = Twilight.forTime(schedule.getStart(), schedule.getSite());
```
swalker2m: Moving this line up out of the big loop over allocations and using `schedule.getStart()` is the important bit of this PR.

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/Ephemeris.scala
```diff
@@ -30,7 +30,7 @@ final class Ephemeris(val site: Site, val compressedData: Deflated[List[(Long, F
     if (_task != null) _task.cancel()
     _task = new TimerTask {
       def run(): Unit = {
-        logger.info("Discarding inflated ephemeris.")
+        logger.fine("Discarding inflated ephemeris.")
```
swalker2m: Lowering the priority of these log messages is unrelated but very helpful for debugging.

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/listeners/LimitsListener.java
swalker2m: It is now consistent with the [cache entry calculations](https://github.com/gemini-hlsw/ocs/blob/develop/bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/Variant.java#L774).

PR #1586, opened 2018-11-09
REL-3536: Added ri filter for GMOS-N.

sraaphorst:
We have a new filter for GMOS-N: this just adds it.

It's taking the place of `DS920_G0312`, but I have been told not to obsolete it.


cquiroz:
does this go to the PIT too? And the seqexec?

cquiroz:
Could you add a ticket to add this to the new seqexec

sraaphorst:
Will do! Here is the task for what I assume is the old seqexec:
http://jira.gemini.edu:8080/browse/REL-3537

cquiroz:
both old and new

sraaphorst:
Since there was already a task for the old one, I just added a task for the new one. You should get a notification soon, since I tagged you in it.

PR #1587, opened 2018-11-09
REL-3438: Increase OT local DB storage limit.

sraaphorst:
This increases the limit on the local OT storage from 50 MB to 100 MB as requested.

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/viewer/open/OpenDialog.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/viewer/open/OpenDialog.scala
@@ -115,7 +115,7 @@ class OpenDialog private(db: IDBDatabaseService, auth: KeyChain, vcs: VcsRegistr
   var selection: List[ISPProgram] = Nil
   val filter = new DBProgramChooserFilter(DBProgramChooserFilter.Mode.localAndRemote)
 
-  val MaxM = 50
+  val MaxM = 100
   val MaxK = MaxM * 1000
 
   def usedBytes: Long =
--- a/build.sbt
+++ b/build.sbt
@@ -4,7 +4,7 @@ name := "ocs"
 
 organization in Global := "edu.gemini.ocs"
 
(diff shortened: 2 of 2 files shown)


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/viewer/open/OpenDialog.scala
```diff
@@ -115,7 +115,7 @@ class OpenDialog private(db: IDBDatabaseService, auth: KeyChain, vcs: VcsRegistr
   var selection: List[ISPProgram] = Nil
   val filter = new DBProgramChooserFilter(DBProgramChooserFilter.Mode.localAndRemote)
 
-  val MaxM = 50
+  val MaxM = 100
```
cquiroz: This seems just for the UI? where is the size enforced?

sraaphorst: If you look at L367 of `OpenDialog`, you can see that the limit is enforced and programs are only opened if there is enough room in `MaxK`, which is calculated from `MaxM`.

sraaphorst: Ooops. I changed the version. Grrrr. This perpetual 2016A versioning is annoying.

cquiroz: Ok, good to have UI and side effects mixed 😄

swalker2m: How so? The UI shows and enforces a limit on how many programs you can have locally. It’s related to keeping the used heap memory down to a reasonable size but it isn’t the same thing as the command line max heap parameter.

swalker2m: Btw, the “size” isn’t even heap memory per se but the size of the sum of the corresponding serialized program files.

PR #1588, opened 2018-11-10
REL-3544: Add GMOS-N ri filter to the ITC

andrewwstephens:
1. Updated the case of the filter in GmosNorthType.java to be consistent with other filters (and moved it up in the list to be with the other broad-band filters).
2. Added the new filter option in ITCgmos.html
3. Added transmission profile of new filter.
The new filter should now work with the ITC in both the OT and web interfaces.


sraaphorst:
FWIW, I'll chime in and say that I don't like the name `ri_G0349` being used in an enum. The convention for constants and enum values in Java is to use uppercase letters and not lowercase letters: hence, my choice to name the filter as I did.

Names that start with a lowercase letter are typically methods or member variables, and not enums or constants.

Here is the convention on the matter:

https://docs.oracle.com/javase/tutorial/java/javaOO/enum.html

Furthermore, searching for "Java enum naming conventions" will reaffirm this.

That's why I think we should endeavour to be consistent with naming and follow the guidelines and recommendations (e.g. the same reason I don't name a method `GetTimingWindows` instead of `getTimingWindows`,, which would just look awkward), although I'm aware that we have deviations (as you can see in the GMOS filter list, which should - by convention - be capitalized, and is why I chose the name that I did instead of the name to which it was changed.)

cquiroz:
That is a good point, thanks for pointing that up

PR #1590, opened 2018-11-12
REL-3519: check usages of viewer.getModel()

swalker2m:
`edu.gemini.ui.gface.GViewer` maintains a nullable `model` field.  It was implemented years before the `Option` concept was fully understood in the Java world and has caused problems, particularly for drag and drop, in the QPT.  Ideally we would refactor `model` to wrap it in an `Option` and in fact I started down that path.  Because the change was beginning to grow beyond what I'm confortable updating during the testing period I punted though and just started doing `null` checks. 

This PR looks more daunting than it actually is because the editor has trimmed trailing whitespace.  If we could ignore whitespace the changes view would be much smaller.

### DIFF
--- a/bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/ui/view/candidate/CandidateObsController.java
+++ b/bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/ui/view/candidate/CandidateObsController.java
@@ -3,30 +3,42 @@ package edu.gemini.qpt.ui.view.candidate;
 import java.beans.PropertyChangeEvent;
 import java.beans.PropertyChangeListener;
 
+import java.util.logging.Logger;
+
 import edu.gemini.qpt.core.Schedule;
 import edu.gemini.qpt.shared.sp.MiniModel;
 import edu.gemini.qpt.shared.sp.Obs;
+import edu.gemini.shared.util.immutable.ImOption;
 import edu.gemini.ui.gface.GTableController;
 import edu.gemini.ui.gface.GViewer;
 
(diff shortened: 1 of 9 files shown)


bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/ui/view/candidate/CandidateObsController.java
```diff
@@ -3,30 +3,42 @@
...
         case P: return obs.getPriority();
-        case RA: return obs.getRa(viewer.getModel().getMiddlePoint());
+        case RA:
+            final Double ra  = ImOption.apply(viewer.getModel())
+                                       .map(s -> obs.getRa(s.getMiddlePoint()))
+                                       .getOrNull();
+            if (ra == null) {
+                LOG.warning("Prevented attempt to get middle point for a null schedule.  Returning null RA.");
+            }
+            return ra;
```
swalker2m: You could complain that we're just putting off the issue for further downstream where it is less clear where the `null` comes from, but the consumer of these values seems prepared to handle `null` results.  See the `Score` `subElement` below for example.

swalker2m:
> If we could ignore whitespace the changes view would be much smaller.

Diff settings, hide whitespace shrinks it to +83 -25 .

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/ui/view/candidate/CandidateObsController.java
cquiroz: I won't complain 😄

PR #1591, opened 2018-11-13
REL-3519: warn if plan contains out of range allocs

swalker2m:
This PR updates the QPT to validate allocation time against the schedule blocks.  In particular, if any allocs are found that do not at least start or end within the block range, a warning is displayed listing the problem observation(s) and the plan is not opened.

There is an underlying bug that is not being addressed yet in that we do not know how to create plans with allocs out of the time range.  Instead it merely detects this case quickly instead of attempting to open plans with inappropriate allocs and blocking for long periods while it tries to calculate the sun path over decades.

### DIFF
--- a/bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/Schedule.java
+++ b/bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/Schedule.java
@@ -16,6 +16,7 @@ import edu.gemini.skycalc.TwilightBoundedNight;
 import static edu.gemini.skycalc.TwilightBoundType.CIVIL;
 import static edu.gemini.skycalc.TwilightBoundType.NAUTICAL;
 import edu.gemini.shared.util.immutable.DefaultImList;
+import edu.gemini.shared.util.immutable.ImList;
 import edu.gemini.shared.util.immutable.ImOption;
 import edu.gemini.shared.util.immutable.Option;
 import edu.gemini.spModel.core.Site;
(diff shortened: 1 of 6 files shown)


bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/Schedule.java
```diff
@@ -225,6 +227,74 @@ public Schedule(MiniModel model, ParamSet params, int upgradingFrom) {
...
+
+        if (outOfRangeObsList.nonEmpty()) {
+            // Trim duplicates.
+            final ImList<String> os = DefaultImList.create(new TreeSet<>(outOfRangeObsList.toList()));
+
+            // Show the problem in the exception message, which ends up being
+            // displayed to the user from the open dialogs.
+            throw new IOException(
+                os.take(10).mkString(
+                    "Some observations are out of the schedule's time range:\n",
```
cquiroz: will you display this message?

swalker2m: Yeah, it is [displayed in the open dialog](https://github.com/gemini-hlsw/ocs/blob/develop/bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/ui/action/AbstractOpenAction.java#L118-L120).

tpolecat:
Would it make sense to discard these allocs and open the remainder of the plan? I worry about people losing their work.

tpolecat:
Actually the fact that we could get a serialized plan with a ludicrously out-of-bounds alloc might be a hint that the problem is in the serialization code … otherwise the QPT would have hung before the user had the chance to save the file, seems like … ?

PR #1592, opened 2018-11-13
REL-3526: set default wavelength for visitor instruments

swalker2m:
When creating a blueprint for visitor instruments, the changes in this PR will set the wavelength according to the instrument name.  Since we don't have a type for the various visitor instruments and yet we need to distinguish between Alopeke and DSSI to set the default wavelength, we just check the name set from the PIT.

### DIFF
--- a/bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/template/factory/impl/visitor/VisitorBase.scala
+++ b/bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/template/factory/impl/visitor/VisitorBase.scala
@@ -9,12 +9,25 @@ trait VisitorBase extends GroupInitializer[SpVisitorBlueprint] with TemplateDsl 
   val program = "VISITOR INSTRUMENT PHASE I/II MAPPING BPS"
   val seqConfigCompType = VisitorInstrument.SP_TYPE
 
+  val AlopekeWavelength: Double = 0.674
+  val DssiWavelength: Double    = 0.700
+
+  private val WavelengthMapping: List[(String, Double)] =
+    List(
+      "alopeke" -> AlopekeWavelength,
+      "dssi"    -> DssiWavelength
+    )
+
+
   implicit def pimpInst(obs: ISPObservation) = new {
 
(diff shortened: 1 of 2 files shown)


bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/template/factory/impl/visitor/VisitorBase.scala
```diff
@@ -9,12 +9,25 @@ trait VisitorBase extends GroupInitializer[SpVisitorBlueprint] with TemplateDsl
   val program = "VISITOR INSTRUMENT PHASE I/II MAPPING BPS"
   val seqConfigCompType = VisitorInstrument.SP_TYPE
 
+  val AlopekeWavelength: Double = 0.674
+  val DssiWavelength: Double    = 0.700
+
+  private val WavelengthMapping: List[(String, Double)] =
+    List(
```
cquiroz: This seems a bit fragile. Do we have a list of these on spModel?

swalker2m: Neither the visitor instrument possibilities nor default wavelengths are stored anywhere that I'm aware of, but if you can point me to it I'll definitely use it.  I could also move this to the `VisitorInstrument` class I suppose but it is going to be fragile regardless until we create instrument types for these.

cquiroz: i guess you're right, maybe it is a bit better in `VistiorInstrument` rather than in the template

sraaphorst: I'm just wondering where the lowercase names come from.

sraaphorst: Never mind: I saw in your PR that they come from the PIT. I'm just having a bit of a time convincing myself that the names come in this format, but provided you've tested it, this looks about the best we can do right now.

PR #1600, opened 2018-11-16
REL-3519: widen the valid allocation time range in QPT schedule files

swalker2m:
TIL that QCs will deliberately "schedule" observations in the day time as alternatives. When we open a queue plan with allocations during the day surrounding the schedule night, they shouldn't be flagged as incorrect. This PR widens the valid range to include the surrounding day time.


sraaphorst:
Where were there two PRs, out of curiosity?

swalker2m:
> Why were there two PRs, out of curiosity?

I mistakenly closed it and then after rebasing I couldn't reopen.

PR #1604, opened 2018-11-20
REL-1985: fix NullPointerException when there are no email recipients

swalker2m:
`MimeMessage` returns `null` when asked for recipients if there are no destination addresses, even though the return type is an array which could easily be empty.  This PR hacks around the problem and makes a few other mailer adjustments.

### DIFF
--- a/bundle/edu.gemini.spdb.reports.collection/src/main/scala/edu/gemini/dbTools/mailer/Mailer.scala
+++ b/bundle/edu.gemini.spdb.reports.collection/src/main/scala/edu/gemini/dbTools/mailer/Mailer.scala
@@ -61,14 +61,14 @@ object Mailer {
     }
   }
 
-  private def write(m: MimeMessage): IO[Unit] = {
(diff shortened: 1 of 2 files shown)


bundle/edu.gemini.spdb.reports.collection/src/main/scala/edu/gemini/dbTools/mailer/Mailer.scala
```diff
@@ -32,53 +45,52 @@ sealed abstract class Mailer(val site: Site, val smtpHost: String) {
...
 
-  def send(m: MimeMessage): IO[Unit]
+  // Left abstract so that the various types of mailer (production, test, dev)
+  // can take the appropriate action.
+  protected def send(m: SafeMimeMessage): IO[Unit]
 
+  protected final def sendAsync(m: SafeMimeMessage): IO[Unit] = IO {
+    import Mailer.Log
+
+    // OCSADV-85: use a future here; it can block for a long time if the mail server is down
```
cquiroz: IS this still valid?

PR #1606, opened 2018-11-22
REL-2718: untangle AGS and GeMS plots

swalker2m:
The GeMS manual guide star search dialog plots guide stars on the TPE image widget.  Independently, the `TpeCatalogFeature` plots AGS guide stars.  When viewing the GeMS manual guide star search, *both* were being plotted using conflicting symbols and breaking the two-way selection link between tables and the symbols in the plot.

This is a modest attempt to improve the situation by placing the GeMS plotting under the control of the `TpeCatalogFeature` just as for the AGS and normal manual guide star search.  It adds a fourth plot case, `GemsPlot`, in addition to the existing `NoPlot`, `AgsPlot`, and normal `ManualPlot`.

At this point, one of the four options will always be plotted -- or not plotted in the case of `NoPlot`.  However which is selected will depend upon which, if any, manual search tool is open.  This can lead to confusion if both the catalog query tool and the gems manual search is open.  All this PR does is to respect the catalog feature toggle button and stop the TPE from displaying multiple plots at once and breaking the link between plot symbol selection and tables.

### DIFF
--- a/bundle/jsky.app.ot/src/main/java/jsky/app/ot/tpe/TpeImageWidget.java
+++ b/bundle/jsky.app.ot/src/main/java/jsky/app/ot/tpe/TpeImageWidget.java
@@ -80,7 +80,7 @@ public class TpeImageWidget extends CatalogImageDisplay implements MouseInputLis
     private boolean _baseOutOfView = false;
 
     // Dialog for GeMS manual guide star selection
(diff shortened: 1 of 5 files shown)


swalker2m:
It may seem incredible, but this represents almost two days of effort trying to align all the planets to make the GeMS plot work at least as well as it was originally intended to.

cquiroz:
To me 2 days sounds a very short time for this

PR #1610, opened 2018-11-28
REL-3552: Added adaptive optics back to GMOS-N.

sraaphorst:
Forgot this part when restoring Altair...


bryanmiller:
Please undo this since we will not be offering GMOS-N with Altair. Leaving GMOS-N out of the instrument list in REL-3552 was not an oversight. Thanks.

PR #1611, opened 2018-11-28
REL-3552: Restore code from earlier semester that removes Altair from GMOS-N

sraaphorst:
As LGS isn't possible with GMOS-N for 2019A, we add code to remove Altair nodes from GMOS-N observations, as we did in REL-3363.


bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/transform/UpConverter.scala
```diff
@@ -145,6 +145,24 @@ case class LastStepConverter(semester: Semester) extends SemesterConverter {
  * This converter will upgrade to 2019A.
  */
 case object SemesterConverter2018BTo2019A extends SemesterConverter {
+    // REL-3552: No GMOS-N + Altair offered for 2019A. Remove Altair from component.
```
cquiroz: Do we have a specific test case for this?

sraaphorst: We have a few that check this. I just fixed them, as this makes the problem count increase by 1.

PR #1619, opened 2019-02-05
REL-3612 Properly save/restore HalfWavePlateAngle property for GPI

cquiroz:
GPI Half wave plate angle not stored

### DIFF
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gpi/Gpi.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gpi/Gpi.java
@@ -1120,6 +1120,7 @@ public class Gpi extends SPInstObsComp implements PropertyProvider, GuideProbeCo
     public static final boolean DEFAULT_ASTROMETRIC_FIELD = false;
     public static final double DEFAULT_EXPOSURE_TIME = 1.49; // OT-54
     public static final int MAX_EXPOSURE_TIME = 60; // OT-78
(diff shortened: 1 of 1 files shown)


bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gpi/Gpi.java
```diff
@@ -2031,6 +2032,7 @@ public ParamSet getParamSet(PioFactory factory) {
         }
 
         Pio.addParam(factory, paramSet, DISPERSER_PROP, getDisperser().name());
+        Pio.addParam(factory, paramSet, HALF_WAVE_PLATE_ANGLE_VALUE_PROP, String.valueOf(getHalfWavePlateAngle()));
```
swalker2m: This seems fine for `ocs` code, but I wonder if we shouldn't use the [same mechanism](https://github.com/gemini-hlsw/ocs/blob/develop/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/obscomp/SPInstObsComp.java#L40-L45) that is used to format the position angle (i.e., `NumberFormat`)?

swalker2m: No, mostly just whether we should set it to two decimal places as we do for position angle.

cquiroz: Ah I see. I think in practice we don't allow decimals but let me check

PR #1620, opened 2019-02-11
REL-3591: Updated PIT icons.

sraaphorst:


cquiroz:
i tried to see the files but they don't come up right. at least the icns is empty and the ico is cut in half

sraaphorst:
Weird. Thanks for pointing this out. I'll try again.

PR #1625, opened 2019-02-12
REL-3454: Add support for Zorro at GS to replace DSSI.

sraaphorst:
Zorro is a copy of 'Alopeke that is replacing DSSI at GS just as 'Alopeke replaced DSSI at GN.

Thus, the implementation was straightforward, following the implementation that was done for 2018B when 'Alopeke replaced DSSI. Note that this also covers REL-3607, i.e. adding initial implementation for Zorro to the OT, since it was easier to just do them both at once.

Here are two screenshots of a 2017B DSSI GS program being opened in the 2019B PIT.

screen shot 2019-02-12 at 4 06 20 pm

screen shot 2019-02-12 at 4 06 32 pm

### DIFF
--- a/bundle/edu.gemini.ags.client.impl/src/main/scala/edu/gemini/ags/client/impl/QueryArgs.scala
+++ b/bundle/edu.gemini.ags.client.impl/src/main/scala/edu/gemini/ags/client/impl/QueryArgs.scala
@@ -56,6 +56,7 @@ object QueryArgs {
     case g: GracesBlueprint     => Right(Instrument.GmosNorth.id) // REL-3288: GRACES uses GMOS-N OI.
     case g: TexesBlueprint      => Right(Instrument.Nifs.id)      // REL-1062
     case g: VisitorBlueprint    => Right(Instrument.Niri.id)      // REL-1090
+    case g: ZorroBlueprint      => Right(Instrument.Nifs.id)      // REL-3454: treat Zorro like Alopeke and DSSI.
     case g: GeminiBlueprintBase => Right(g.instrument.id)
     case _ => Left("Not a Gemini Instrument")
   }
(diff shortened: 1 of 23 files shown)


bundle/edu.gemini.ags.client.impl/src/main/scala/edu/gemini/ags/client/impl/QueryArgs.scala
```diff
@@ -56,6 +56,7 @@ object QueryArgs {
     case g: GracesBlueprint     => Right(Instrument.GmosNorth.id) // REL-3288: GRACES uses GMOS-N OI.
     case g: TexesBlueprint      => Right(Instrument.Nifs.id)      // REL-1062
     case g: VisitorBlueprint    => Right(Instrument.Niri.id)      // REL-1090
+    case g: ZorroBlueprint      => Right(Instrument.Nifs.id)      // REL-3454: treat Zorro like Alopeke and DSSI.
```
cquiroz: Why is that we treat them like `NIFS`? i probably did this myself 😄

cquiroz:
Did you  test the pdf output?

bundle/edu.gemini.ags.client.impl/src/main/scala/edu/gemini/ags/client/impl/QueryArgs.scala
sraaphorst: It's something about P1 (or possibly P2?) not being properly configurable with them now, so they default by request of science to be the same as `Nifs`.

sraaphorst:
@cquiroz Yup... checked the PDF output and it looks correct. I can attach if you want to see.

cquiroz:
image

PR #1626, opened 2019-02-13
REL-3609: addendum

swalker2m:
Changes an error message when the exposure time cannot be parsed as an integer and raises the maximum exposure time to 1200 seconds.


swalker2m:
@sraaphorst  Yeah, unfortunately we'll need another test release after this is merged.  We don't need the migration code to run but I don't know how easy it is to skip that step.

sraaphorst:
Actually, that should be very easy to do. It looks like Travis finally chugged through it all. Let me know when you're ready for me to do a build.

PR #1627, opened 2019-02-13
REL-3606: Readded Texes to the PIT.

sraaphorst:
Readding Texes as per REL-3606. The other parts of 3606 (i.e. show Zorro and hide DSSI) were taken care of by the PR for REL-3454.


cquiroz:
Do you need some upconversion? Perhaps from some remote semester where texas was present?

sraaphorst:
No, I don't think so. I just removed the code that removed Texes in `UpConverter`, and I tried importing a Texes program from 2017A, and it worked just fine.

PR #1630, opened 2019-02-14
REL-3621: Added new GMOS-N ri filter to PIT and import.

sraaphorst:


cquiroz:
Is it replacing some existing filter? Do we need to transition old proposals tu use this filter?

sraaphorst:
No, it's totally new.

PR #1631, opened 2019-02-19
REL-3625: time accounting rule update

swalker2m:
Changes the time accounting rules to always charge for visitor instrument observations regardless of whether there are any chargeable datasets in a visit.  This change is to be retroactively applied to the start of 2019A (which is why the existing 2019A rule could be simply updated).

Time accounting has heretofore been instrument agnostic, so the bulk of this update is passing the instrument down through multiple layers of API to the `VisitCalculator` where it is needed.  Otherwise there is a new `InstrumentService` that calculates the `Instrument` associated with an `ISPObservation` (if any) and an update to the `SPObsCache` to cache the calculation.

### DIFF
--- a/bundle/edu.gemini.auxfile.workflow/build.sbt
+++ b/bundle/edu.gemini.auxfile.workflow/build.sbt
@@ -29,3 +29,5 @@ OsgiKeys.exportPackage := Seq(
   "edu.gemini.auxfile.client",
   "edu.gemini.auxfile.server",
   "edu.gemini.auxfile.copier")
+
+parallelExecution in Test := false
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/obs/SPObsCache.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/obs/SPObsCache.java
@@ -185,7 +185,7 @@ public final class SPObsCache implements ISPEventMonitor, ISPCloneable, Serializ
     private Option<TargetCalculator> _targetCalculator;
 
     // The observation's instrument, if any.
-    private Option<Instrument> _instrument;
+    private Option<Instrument> _instrument = ImOption.empty();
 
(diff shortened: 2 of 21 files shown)


bundle/edu.gemini.auxfile.workflow/build.sbt
```diff
@@ -29,3 +29,5 @@ OsgiKeys.exportPackage := Seq(
   "edu.gemini.auxfile.client",
   "edu.gemini.auxfile.server",
   "edu.gemini.auxfile.copier")
+
+parallelExecution in Test := false
```
swalker2m: This is completely unrelated, but (I think) works around spurious issues with test cases in this bundle that cannot be executed in parallel because they manipulate a shared `static` field in `edu.gemini.auxfile.server.file.FileManager`.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/obs/SPObsCache.java
```diff
@@ -49,6 +50,16 @@ public static void setTargetCalculator(ISPObservation obs, Option<TargetCalculat
         setObsCache(obs, cache);
     }
 
+    public static Option<Instrument> getInstrument(ISPObservation obs) {
+        return ImOption.apply(getObsCache(obs)).flatMap(c -> c.getInstrument());
+    }
+
+    public static void setInstrument(ISPObservation obs, Option<Instrument> instrument) {
```
tpolecat: Can you remind me the circumstances under which the obs cache is invalidated?

swalker2m: [Whenever anything inside of an observation tree changes](https://github.com/gemini-hlsw/ocs/blob/develop/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/obs/SPObsCache.java#L317-L325).

PR #1636, opened 2019-02-21
REL-3627: fix duplicate on import

swalker2m:
Fixes an issue with the "Create duplicate" option when trying to import from XML a program without a program id that already exists in the database.  We cannot assume that an `ISPProgram` has a program id.

### DIFF
--- a/bundle/edu.gemini.spModel.io/src/main/scala/edu/gemini/spModel/io/SpImportService.scala
+++ b/bundle/edu.gemini.spModel.io/src/main/scala/edu/gemini/spModel/io/SpImportService.scala
@@ -44,7 +44,8 @@ object SpImportService {
         }
       }
 
-      val cp = db.getFactory.copyWithNewKeys(im, POTUtil.getUnusedProgramId(im.getProgramID, db))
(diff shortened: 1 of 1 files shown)


bundle/edu.gemini.spModel.io/src/main/scala/edu/gemini/spModel/io/SpImportService.scala
```diff
@@ -44,7 +44,8 @@ object SpImportService {
         }
       }
 
-      val cp = db.getFactory.copyWithNewKeys(im, POTUtil.getUnusedProgramId(im.getProgramID, db))
+      val pid = Option(im.getProgramID).map(POTUtil.getUnusedProgramId(_, db)).orNull
```
sraaphorst: I'm a bit confused about the `orNull` here. I guess we pass `null` into `db.getFactory.copyWithNewKeys` if we have no ID?

swalker2m: Correct, and so it just doesn't assign an ID in that case.  It's suboptimal but this is code that dates back to the mid-1990s.

PR #1640, opened 2019-03-11
REL-3569: Fixed issue with base morphing.

sraaphorst:
Forgot to move this change into the release branch.


swalker2m:
Are we planning another patch release then?

PR #1641, opened 2019-03-13
REL-3636: AcqCam, Phoenix and Texes exceptions

swalker2m:
Adds Acquisition Camera, Phoenix and Texes to the list of instruments which should be an exception to the new Time Accounting rule of REL-1608.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/syntax/sp/Instrument.scala
+++ b/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/syntax/sp/Instrument.scala
@@ -1,0 +1,35 @@
+package edu.gemini.spModel.syntax.sp
+
+import edu.gemini.pot.sp.Instrument
+import edu.gemini.pot.sp.Instrument._
+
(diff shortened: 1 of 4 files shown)


bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/syntax/sp/Instrument.scala
```diff
@@ -0,0 +1,35 @@
...
+
+final class InstrumentOps(val self: Instrument) extends AnyVal {
+
+  /**
+   * Is this a "visitor" instrument.  In other words, is it fully integrated
+   * with the observing system and DHS?
+   */
+  def isVisitor: Boolean =
+    self match {
+      case AcquisitionCamera | Phoenix | Texes | Visitor => true
```
cquiroz: I guess we don't have Zorro, Alopeke, DSSI right? They are just Visitors

swalker2m: Right, there is no specific component for those.

PR #1643, opened 2019-03-14
REL-3628: time accounting update for GPI acquisitions

swalker2m:
Updates the 2019A time accounting rule to always charge for GPI acquisitions, since they often do not produce datasets.  The bulk of the work was adding the observation `ObsClass` as a parameter since it is needed to make the determination.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/obsrecord/PrivateVisitList.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/obsrecord/PrivateVisitList.java
@@ -10,6 +10,7 @@ import edu.gemini.spModel.core.Site;
 import edu.gemini.spModel.event.ExecEvent;
 import edu.gemini.spModel.event.ObsExecEvent;
 import edu.gemini.spModel.event.StartVisitEvent;
+import edu.gemini.spModel.obsclass.ObsClass;
 import edu.gemini.spModel.time.ChargeClass;
 import edu.gemini.spModel.time.ObsTimeCharges;
 
@@ -22,7 +23,7 @@ import java.util.*;
  */
 final class PrivateVisitList implements Serializable {
     private static final long serialVersionUID = -2686488059242714341L;
-    
+
     private List<PrivateVisit> _visits;
 
     PrivateVisitList() {
@@ -114,6 +115,7 @@ final class PrivateVisitList implements Serializable {
 
     ObsTimeCharges getTimeCharges(
         Option<Instrument> instrument,
+        ObsClass           oc,
         ChargeClass        mainChargeClass,
         ObsQaRecord        qa,
         ConfigStore        store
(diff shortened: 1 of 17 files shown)


bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/obsrecord/VisitCalculator.scala
```diff
@@ -129,9 +135,14 @@ private[obsrecord] object VisitCalculator {
...
-        dsets.exists { case (lab, _) => qa(lab).isChargeable }
+        dsets.exists { case (lab, _) => datasetQa(lab).isChargeable }
 
-      if (isVisitor || hasChargeableDataset) normalCharges
+      // REL-3628: many GPI acquisitions don't produce datasets so charge for
+      //           them regardless
+      def isGpiAcquisition: Boolean =
+        instrument.exists(_ == Instrument.Gpi) && (obsClass == ObsClass.ACQ)
+
+      if (isVisitor || hasChargeableDataset || isGpiAcquisition) normalCharges
```
swalker2m: This is the important bit where we add a special case for GPI acquisition.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/obsrecord/VisitCalculator.scala
```diff
@@ -129,9 +135,14 @@ private[obsrecord] object VisitCalculator {
         instrument.exists(_.isVisitor)
 
       def hasChargeableDataset: Boolean =
-        dsets.exists { case (lab, _) => qa(lab).isChargeable }
+        dsets.exists { case (lab, _) => datasetQa(lab).isChargeable }
 
-      if (isVisitor || hasChargeableDataset) normalCharges
+      // REL-3628: many GPI acquisitions don't produce datasets so charge for
+      //           them regardless
+      def isGpiAcquisition: Boolean =
```
cquiroz: I suppose this is due to the A&G steps that don't produce a dataset and are of class Acquisiton

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/obsrecord/PrivateVisitList.java
```diff
@@ -22,7 +23,7 @@
  */
 final class PrivateVisitList implements Serializable {
     private static final long serialVersionUID = -2686488059242714341L;
-    
+
```
swalker2m: I verified that serialization won't be an issue which stands to reason since the same APIs were previously updated to pass in the instrument.

cquiroz:
This has a historical reason. GPI needs an "acquisition" step to run the align & calib procedure. We don't want to produce a data set here as it is not needed and would slow down the process
Given how big this change would be on the ocs and the seqexec we opted to use an acquisiton observation with one step for this

The new seqexec should be able in the near future to remove this need
Of course we didn't consider time accounting back then

PR #1646, opened 2019-03-17
REL-2897: Update default QVis options

andrewwstephens:
Updates to the default options in the OT Queue Visualization plugin.


tpolecat:
It's an OT plugin so it's only distributed with the OT as far as I know.

sraaphorst:
@tpolecat Ahhh... for some reason, I thought at one time that it was a standalone  app, but I think you're right. We should decide if we want this in the upcoming patch release. @andrewwstephens?

andrewwstephens:
Will there be a test version before the upcoming patch release?  It would be good if the QCs could sign off on these changes.

PR #1649, opened 2019-04-10
AGS JSON Servlet

swalker2m:
Creates an `ocs` codebase servlet application for executing AGS requests.  The application runs the old AGS estimation servlet used by the PIT and a new `JsonServlet` that accepts a JSON post and returns a JSON result describing the corresponding guide star choices.  Along the way it moves some of the reusable JSON codec code from `itc` into a new bundle `edu.gemini.json`.

Meant to be called programmatically, but an example request document:

```json
{
  "site": "GS",
  "coordinates": {
    "ra":  "0:00:00",
    "dec": "0:00:00"
  },
  "targetType": "sidereal",
  "conditions": {
    "cc": "PERCENT_50",
    "iq": "PERCENT_20",
    "sb": "PERCENT_50"
  },
  "posAngle": 150.0,
  "instrument": {
    "gmosSouth": {
      "fpu":        "LONGSLIT_2",
      "constraint": "FIXED_180"
    }
  },
  "offsets": [
    {
      "p": 0.00028,
      "q": 0.00028
    }
  ] 
}
```

sent via `curl http://localhost:9123/json-ags -d @data.json` returns

```json
{
  "posAngle" : 150.0,
  "assignments" : [
    {
      "probe" : "GMOS OIWFS",
      "star" : {
        "coordinates" : {
          "ra" : "00:00:06.159",
          "dec" : "-00:02:47.04"
        },
        "name" : "450-000005",
        "magnitudes" : [
          {
            "value" : 14.27,
            "band" : "UC",
            "error" : null,
            "system" : "Vega"
          },
          {
            "value" : 12.526,
            "band" : "J",
            "error" : 0.02,
            "system" : "Vega"
          },
          {
            "value" : 11.935,
            "band" : "H",
            "error" : 0.03,
            "system" : "Vega"
          },
          {
            "value" : 11.674,
            "band" : "K",
            "error" : 0.02,
            "system" : "Vega"
          }
        ],
        "properMotion" : {
          "deltaRA" : 55.7,
          "deltaDec" : -7.5,
          "epoch" : 2000.0
        },
        "parallax" : null,
        "redshift" : null
      }
    }
  ]
}
```


bundle/edu.gemini.ags.servlet/src/main/scala/edu/gemini/ags/servlet/JsonServlet.scala
```diff
@@ -0,0 +1,105 @@
...
+    val src  = Source.fromInputStream(req.getInputStream, enc)
+    val json = try src.mkString finally src.close
+
+    // The AGS API uses `Future` but we map it to a `Task`.
+    val task: \/[String, Task[Option[AgsStrategy.Selection]]] =
+      for {
+        agsReq     <- \/.fromEither(Parse.decodeEither[AgsRequest](json))
+        obsContext <- agsReq.toContext
+        strategy   <- AgsRegistrar.currentStrategy(obsContext) \/> "Could not determine AGS strategy."
+      } yield Task.delay(strategy.select(obsContext, magTable)(executionContext)).flatMap(fut => toTask(fut))
```
swalker2m: I recognize that this switching to `Task` business may be dumb or not implemented correctly.  Please advise if it should be changed.

bundle/edu.gemini.ags.servlet/src/main/scala/edu/gemini/ags/servlet/json/SiderealTargetCodec.scala
```diff
@@ -0,0 +1,68 @@
...
+      "band",
+      "error",
+      "system"
+    )
+
+  implicit val SiderealTargetEncoder: EncodeJson[SiderealTarget] =
+    EncodeJson((s: SiderealTarget) =>
+      ("magnitudes"   := s.magnitudes         ) ->:
+      ("parallax"     := s.parallax.map(_.mas)) ->:
+      ("redshift"     := s.redshift.map(_.z)  ) ->:
```
swalker2m: Perhaps redshift and parallax are never extracted from catalog results?  If so I'll just remove them.

sraaphorst:
I guess it would make the most sense to run this on `hbfauxodb-lv1` aka `gnauxodb`, since that's the "AGS server" as the PIT was hitting the production ODB (can't remember which one), IIRC.

swalker2m:
@sraaphorst That's a good point about the existing estimation servlet app.  Where is that specified?  I should remove it in favor of the estimation servlet in this PR. It runs with fewer bundles and doesn't start an actual ODB instance.

cquiroz:
THere is an `ags` app that contains the estimation servlet. I agree it maybe better to run it there

bundle/edu.gemini.ags.servlet/src/main/scala/edu/gemini/ags/servlet/JsonServlet.scala
```diff
@@ -0,0 +1,105 @@
...
+
+    result match {
+      case -\/((code, msg)) =>
+        res.sendError(code, msg)
+
+      case \/-(sel)         =>
+        res.setStatus(SC_OK)
+        res.setContentType("text/json; charset=UTF-8")
+        val writer = res.getWriter
+        writer.write(sel.asJson.spaces2)
```
cquiroz: I guess it would be slightly more efficient to compact the spaces but then it would be less readable. though that is easy to fix with e.g. `jq`

sraaphorst:
@swalker2m Actually, right now, we're just lazily running the entire ODB on the servers I mentioned. It was sort of a last-minute decision when making the, IIRC, 2018A PIT, and you had suggested simplifying it down to just the necessary bundles, but we never did that, so this is the perfect opportunity to do so, I'd think.

cquiroz:
Ah right @sraaphorst thanks for reminding me, so we should rather do as you suggest I guess

tpolecat:
Do we need to run it on-site in order to talk to our catalog server, or could we run on Heroku? I set up ITC [thus](https://github.com/gemini-hlsw/itc-server) … the advantage being we can scale it up easily (which doesn't help with load on the catalog server but might be helpful for compute-heavy bits).

swalker2m:
> Do we need to run it on-site in order to talk to our catalog server, or could we run on Heroku?

The AGS runs in the OT so it must be available offsite as well.  I'll try to copy the ITC example.

tpolecat:
You'll need to create a new heroku app, and you can do it without creating a remote repository on Heroku since it will be running in a docker container. `heroku create --no-remote <name>`

cquiroz:
The gemini catalogues are at the summits and I'm not sure if they are reachable from the outside. the other catalogues are external

bundle/edu.gemini.ags.servlet/src/main/scala/edu/gemini/ags/servlet/JsonServlet.scala
tpolecat: This looks correct (`toTask` below, specifically) although the pool sizing doesn't really protect the catalog server if we end up scaling this service up. Maybe there's no need to do so.

bundle/edu.gemini.ags.servlet/src/main/scala/edu/gemini/ags/servlet/JsonServlet.scala
tpolecat: It's really nice to be able to read the output so I'm ok with `spaces2` for now.

PR #1653, opened 2019-04-21
REL-3620: Add GNIRS Hartmann masks and focus

andrewwstephens:
This adds an engineering item to the GNIRS iterator to select one of the Hartmann masks, and another engineering item to manually set the focus.  This seems to work as expected from the UI, but please check that it makes sense.  One complication is that the Hartmann masks are *not* in a separate mechanism, but rather installed in filter wheel #1.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gnirs/GNIRSParams.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gnirs/GNIRSParams.java
@@ -4,6 +4,7 @@ import edu.gemini.spModel.config2.ItemKey;
 import edu.gemini.spModel.data.SuggestibleString;
 import static edu.gemini.spModel.seqcomp.SeqConfigNames.INSTRUMENT_KEY;
 import edu.gemini.spModel.type.*;
+import edu.gemini.shared.util.immutable.*;
 
 import java.beans.PropertyEditorManager;
 import java.beans.PropertyEditorSupport;
@@ -1020,12 +1021,12 @@ public class GNIRSParams {
     public enum HartmannMask implements DisplayableSpType, SequenceableSpType {
 
         OUT("Out"),
-        LeftMask("Left"),
-        RightMask("Right"),
+        LEFT_MASK("Left"),
+        RIGHT_MASK("Right"),
         ;
 
-        public static HartmannMask DEFAULT = OUT;
-        private String _displayValue;
+        public static final HartmannMask DEFAULT = OUT;
+        private final String _displayValue;
 
         HartmannMask(String displayValue) {
             _displayValue = displayValue;
@@ -1094,24 +1095,28 @@ public class GNIRSParams {
         public String toString() {
             return _displayValue;
         }
-
     }
 
-    public static class Focus extends SuggestibleString {
-        public Focus() {
+    public static final class Focus extends SuggestibleString {
+
+        public Focus(String value) {
             super(FocusSuggestion.class);
-            setStringValue(FocusSuggestion.DEFAULT.displayValue());
+            setStringValue(value);
+        }
+
+        public Focus() {
+            this(FocusSuggestion.DEFAULT.displayValue());
+        }
+
+        public Focus copy() {
+            return new Focus(getStringValue());
         }
     }
 
-    public static class FocusEditor extends PropertyEditorSupport {
-        public Object getValue() {
-            Focus f = (Focus) super.getValue();
-            if (f == null) return null;
+    public static final class FocusEditor extends PropertyEditorSupport {
 
-            Focus res = new Focus();
-            res.setStringValue(f.getStringValue());
-            return res;
+        public Object getValue() {
+            return ImOption.apply((Focus) super.getValue()).map(f -> f.copy()).getOrNull();
         }
 
         public void setValue(Object value) {
@@ -1125,18 +1130,11 @@ public class GNIRSParams {
         }
 
         public String getAsText() {
-            Focus val = (Focus) getValue();
-            if (val == null) return null;
-            return val.getStringValue();
+            return ImOption.apply((Focus) getValue()).map(f -> f.getStringValue()).getOrNull();
         }
 
         public void setAsText(String string) throws IllegalArgumentException {
-            Focus val = (Focus) super.getValue();
-            if (val == null) {
-                val = new Focus();
-                super.setValue(val);
-            }
-            val.setStringValue(string);
+            setValue(new Focus(string));
         }
     }
     
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gnirs/GNIRSParams.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gnirs/GNIRSParams.java
@@ -1014,6 +1014,132 @@ public class GNIRSParams {
         }
     }
 
+    /**
+     * Hartmann Masks
+     */
+    public enum HartmannMask implements DisplayableSpType, SequenceableSpType {
+
+        OUT("Out"),
+        LeftMask("Left"),
+        RightMask("Right"),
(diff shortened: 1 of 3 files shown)


cquiroz:
Note this PR has some effects on the web seqexec. Nothing major but we have to remember to update it once merged

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gnirs/GNIRSParams.java
```diff
@@ -1014,6 +1014,132 @@ static public AcquisitionMirror getAcquisitionMirror(String name, AcquisitionMir
...
+    }
+
+    public static class FocusEditor extends PropertyEditorSupport {
+        public Object getValue() {
+            Focus f = (Focus) super.getValue();
+            if (f == null) return null;
+
+            Focus res = new Focus();
+            res.setStringValue(f.getStringValue());
+            return res;
```
swalker2m: I can see that this is taken from the way that `WavelengthEditor` works.  In fact it pains me to admit I'm originally responsible for this `SuggestibleString` business and the fact that it is so mutable and `null`ish. I wouldn't do it this way these days though.

At any rate, a slight improvement might be to add a `copy` method to `Focus`:

```java
public static final class Focus extends SuggestibleString {
    public Focus(String value) {
        super(FocusSuggestion.class);
        setStringValue(value);
    }

    public Focus() {
        this(FocusSuggestion.DEFAULT.displayValue());
    }

    public Focus copy() {
        return new Focus(getStringValue());        
    }
}
```

and then 

```java
import edu.gemini.shared.util.immutable.*;

public Object getValue() {
    return ImOption.apply((Focus) super.getValue()).map(f -> f.copy()).getOrNull();
}
```

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gnirs/InstGNIRS.java
```diff
@@ -364,6 +371,41 @@ public void setOverrideAcqObsWavelength(boolean newValue) {
...
+    /**
+     * Return the (read-only, static) value for focus.
+     */
+    public Focus getFocus() {
+        return new Focus();
+    }
+
+    public void setFocus(Focus focus) {
+        // Required for reflection
+    }
```
swalker2m: I'm not sure about this part.  When you say "read-only, static" you just mean that it is included in the instrument component but isn't edited there?  It does appear, and is editable, in the GNIRS iterator though?

I worry because there is machinery around syncing the first step of the iterator with the static component.  With this implementation it will not be in sync which may cause problems.  I realize there is an outstanding task to remove that syncing but for now perhaps we should just add a `Focus` property and get/set it like everything else.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gnirs/InstGNIRS.java
```diff
@@ -155,6 +161,7 @@ private static PropertyDescriptor initProp(String propName, boolean query, boole
     private Decker _decker = Decker.DEFAULT;
     private Filter _filter = Filter.DEFAULT;
     private WellDepth _wellDepth = WellDepth.DEFAULT;
+    private HartmannMask _hartmannMask = HartmannMask.DEFAULT;
```
swalker2m: I believe this property along with `Focus` should be included in `getParamSet` and `setParamSet` as well.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gnirs/GNIRSParams.java
```diff
@@ -1014,6 +1014,132 @@ static public AcquisitionMirror getAcquisitionMirror(String name, AcquisitionMir
...
+            return _displayValue;
+        }
+
+        public String toString() {
+            return _displayValue;
+        }
+
+    }
+
+    public static class Focus extends SuggestibleString {
```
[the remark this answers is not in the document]
andrewwstephens: The focus will be an integer between -150000 to +180000.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gnirs/GNIRSParams.java
```diff
@@ -1014,6 +1014,132 @@ static public AcquisitionMirror getAcquisitionMirror(String name, AcquisitionMir
...
+     */
+    public enum HartmannMask implements DisplayableSpType, SequenceableSpType {
+
+        OUT("Out"),
+        LeftMask("Left"),
+        RightMask("Right"),
+        ;
+
+        public static HartmannMask DEFAULT = OUT;
+        private String _displayValue;
```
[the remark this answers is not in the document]
andrewwstephens: Sounds good.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gnirs/GNIRSParams.java
andrewwstephens: Thanks!

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gnirs/InstGNIRS.java
andrewwstephens: It wasn't clear why some properties are included but others are not.  I copied the focus bits from NIRI where they are not included.  :-\

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gnirs/InstGNIRS.java
andrewwstephens: I admit that I copied this verbatim from the NIRI focus editor. The only thing that is a little strange is that it is not possible to edit the first value (it immediately reverts to 'best focus'), perhaps because of the synchronization that you described, which is a minor annoyance but not a hindrance to using this for engineering.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gnirs/GNIRSParams.java
```diff
@@ -1014,6 +1015,129 @@ static public AcquisitionMirror getAcquisitionMirror(String name, AcquisitionMir
...
+            setStringValue(value);
+        }
+
+        public Focus() {
+            this(FocusSuggestion.DEFAULT.displayValue());
+        }
+
+        public Focus copy() {
+            return new Focus(getStringValue());
+        }
```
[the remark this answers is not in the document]
andrewwstephens: I don't think we need to get fancy in this engineering component, but I see what you mean and how it could be implemented.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gnirs/InstGNIRS.java
andrewwstephens: Thanks! That works perfectly!

andrewwstephens:
GNIRS has 2 filter wheels, and AFAIK the OT does not know which filters are in which wheels.  The desired behavior is that if the Hartmann Mask is `OUT` then both FW1 and FW2 are configured normally as they are now.  If one of the Hartmann masks is selected then FW1 is moved to the requested Hartmann mask and FW2 is configured normally.  However, if the user requests a filter in wheel 1 *and* a Hartmann mask the seqexec should throw an error saying that the requested configuration is not possible.  I am open to alternative solutions.

swalker2m:
We could model that so it reflects reality and we should do so in the upgrades project. I just wanted to be sure that it was clear what to do with the information in the seqexec and it sounds like it is.

PR #1656, opened 2019-04-28
REL-1371: added line to find fwhm for uniform sources 

leashma:
Simple change to use user-supplied FWHM for a uniform source, and get the AO corrected FWHM just like REL-1352. So the output should not be FWHM = 0.0, but actually calculate the AO-corrected one.


bundle/edu.gemini.itc.web/src/test/scala/edu/gemini/itc/web/baseline/BaselineTest.scala
```diff
@@ -26,7 +26,7 @@ import org.junit.{Ignore, Test}
  */
 class BaselineTest {
 
-  @Ignore
+  //@Ignore
```
sraaphorst: This should just be commented out to create the `baseline.txt` file: once you've done that, you should uncomment it.

leashma: I started working on a new feature, and those changes were included in this commit. I think I need to cancel this pull request and start from scratch.

PR #1658, opened 2019-04-30
REL-3603: Missing ToO Target Type in OT Templates menu

sraaphorst:
The ToO target type was missing in the OT templates menu. As is visible in this screen cap, this resulted in ToO targets being labeled as non-sidereal in the template editor:

Screen Shot 2019-04-30 at 12 49 08 PM

When the templates were instantiated, they were still created as having target type ToO; however, it was not possible to create new targets with target type ToO in the template editor.

Now, in the case that a program is a ToO program, we add the target type ToO to the combo box, and ToO targets display the proper type in the template editor:

Screen Shot 2019-04-30 at 2 32 36 PM

Screen Shot 2019-04-30 at 2 32 47 PM

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/template/TemplateParametersEditor.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/template/TemplateParametersEditor.scala
@@ -38,7 +38,7 @@ object TemplateParametersEditor {
     if (s.size == 1) s.headOption else None
   }
 
-  trait Initializable {
+  sealed trait Initializable {
     def init(ps: Iterable[TemplateParameters]): Unit
   }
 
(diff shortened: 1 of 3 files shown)


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/editor/template/TemplateParametersEditor.scala
```diff
@@ -65,7 +65,7 @@ object TemplateParametersEditor {
     }
   }
 
-  abstract class ColumnPanel(hGap: Int = HGap, rowAnchor: GridBagPanel.Anchor.Value = East) extends GridBagPanel with Initializable {
+  sealed abstract class ColumnPanel(hGap: Int = HGap, rowAnchor: GridBagPanel.Anchor.Value = East) extends GridBagPanel with Initializable {
```
cquiroz: Do you need `sealed` if the trait is already `sealed`?

sraaphorst: No, probably not. I may have gotten a little `seal`-happy. I'll go ahead and remove it since it's redundant.

PR #1659, opened 2019-04-30
REL-3661: Updating F2 blueprint template rules.

sraaphorst:


sraaphorst:
@cquiroz let me know that more needed to be done for this PR, but I'm running into some stumbling blocks and I was hoping @tpolecat or @swalker2m could chime in.

It looks like some of this algorithm from the old `.txt` file, namely:

```
    FOR {21,22,23,24,25,26,27}:	        SET DISPERSER FROM PI
        SET CONDITIONS FROM PI	            Put FILTERS from PI into F2 ITERATOR


     FOR {24,25}:                              # MOS science	    FOR {21,22,23,24,25,26,28,29}:
        SET "Custom MDF" = G(N/S)YYYYS(Q/C/DD/SV/LP/FT)XXX-NN 	        SET CONDITIONS FROM PI
            where: 	
             (N/S) is the site 	    FOR {24,25,26,30}:                              # MOS science
            YYYYS is the semester, e.g. 2015A 	        SET "Custom MDF" = G(N/S)YYYYS(Q/C/DD/SV/LP/FT)XXX-NN 
            (Q/C/DD/SV/LP/FT) is the program type 	            where: 
            XXX is the program number, e.g. 001, or 012, or 123 	            (N/S) is the site 
            NN should be the string "NN" since the mask number is unknown
```

doesn't appear to be implemented in the old code. I'm not quite sure, as a result, what we want to do in the new code. Is this handled somewhere?

I see that the `GNIRS_BP.txt` also mentions `SET CONDITIONS FROM PI` for GNIRS spectroscopy mode, but never does anything with this in `GnirsSpectroscopy` either, so I'm going to assume for now that nothing needs to be done for this part, and the rest, I believe, I've implemented properly.

PR #1661, opened 2019-04-30
REL-3667: Change the number of QVis repeated windows to N+1

andrewwstephens:
Timing windows may be defined with a number of "Repeats". The OT elevation plot and the QPT interpret this to mean there are Nrepeat + 1 windows, i.e. if a window repeats once there are two total.  This change updates QVis to have the same interpretation.


bundle/edu.gemini.qv.plugin/src/main/scala/edu/gemini/qv/plugin/util/ConstraintsCache.scala
```diff
@@ -252,7 +252,7 @@ case class TimingWindowConstraint(windows: Seq[TimingWindow]) extends Constraint
           } else tw.getRepeat match {
             case REPEAT_NEVER => Solution(Seq(new Interval(tw.getStart, tw.getStart + tw.getDuration)))           // single window
             case REPEAT_FOREVER => Solution(createIntervals(tw.getStart, interval.end, Int.MaxValue, tw, Seq()))  // repeat forever, use interval end as limiter
-            case n => Solution(createIntervals(tw.getStart, interval.end, n, tw, Seq()))                          // repeat n times
+            case n => Solution(createIntervals(tw.getStart, interval.end, n+1, tw, Seq()))                        // repeat n times
```
arturog8m: Since this hasn't been merged yet, could you update the comment as well to indicate why it's n+1 now?

andrewwstephens: Absolutely; thanks for the suggestion.

PR #1662, opened 2019-05-02
REL-3669: Trying Trusty dist, which supports OpenJDK 8.

sraaphorst:
This is a test to fix our Travis configuration.


sraaphorst:
I'm forcing this in since it works and everyone else is away.

PR #1665, opened 2019-05-03
REL-3661: Updating F2 blueprint rules according to new rules.

sraaphorst:
This will involve a few PRs.


sraaphorst:
There are (I believe) two issues with this PR, which I suspect can be traced back to the XML file, although I am not certain.

For F2 longslit, we should have the note:
```
"Use the same PA for science target and telluric"
```

and probably also the new note:
```
"Detector readout modes"
```

but neither of these is appearing. I have spend half a day trying to figure out what is going on, so I am going to merge this for the test release, and we can fix it later.

PR #1669, opened 2019-05-17
REL-3661: F2 template updates

swalker2m:
This PR fixes a few issues with skeleton creation and template instantiation for F2.

* All notes should now be included
* The order of observations should be what is expected
* Conditions should be copied in when instantiating

Inexplicably, we sort the observations by library id so I had to renumber the ids in the XML and p-code.

I removed the explicit conditions from the XML because they were preventing the template conditions from being used.

It doesn't include the logic for setting the custom mask name since it relies on having the program id, which isn't available.  I left a bit of code commented out that may do the trick if add a parameter to initialize to pass in the program id.


tpolecat:
I hate to ask, but can you add stuff to the F2 testcase to verify these changes? It's really hard to be confident in this stuff because of the idiotic way I wrote it.

swalker2m:
Many of the changes were in the XML itself, but I added tests to check that the notes and F2 MOS observations are added.

PR #1670, opened 2019-05-17
Feature/rel 3643 - Replace ITC "not background limited" warning with "readnoise limited" warning

andrewwstephens:
This is a minor update addressing tester feedback, expanding the warning message and adding the check to F2 imaging results.  I don't like how it prints an extra blank line when the warning is not triggered, but I don't see an obvious way around it...


bundle/edu.gemini.itc.web/src/main/java/edu/gemini/itc/web/html/Flamingos2Printer.java
```diff
@@ -110,6 +110,8 @@ private void writeImagingOutput(final ImagingResult result, final ItcImagingResu
         _println(CalculatablePrinter.getTextResult(result.iqCalc()));
         _println(CalculatablePrinter.getTextResult(result.is2nCalc(), result.observation()));
 
+        _println(CalculatablePrinter.WarnIfReadNoiseLimited(result.is2nCalc()));
+
```
swalker2m: >  I don't like how it prints an extra blank line when the warning is not triggered, but I don't see an obvious way around it...

I don't know much about ITC or how this usually looks, but I suppose you could just call `_print` instead of `_println`?  Maybe adding another `\n` to the warning when present if you want to preserve the two newlines for that case.

A less hacky way to do it would be to change the warning to return an `Option<String>` instead of a `String`.

```java
import edu.gemini.shared.util.immutable.*;

public static Option<String> ReadNoiseLimitedWarning(final ImagingS2NCalculatable s2n) {
    return (s2n.getVarReadout() ...) ?
        ImOption.apply("Warning: observation is read noise limited ...") :
        ImOption.<String>empty();
}
```
Then
```java
CalculatablePrinter.ReadNoiseLimitedWarning(result.is2nCalc()).foreach(s -> _println(s));
```

swalker2m:
These updates look great, but I'm not sure what to do about the "baseline.txt" file conflict.

PR #1674, opened 2019-05-27
REL-3661: Updated F2 blueprint for MOS.

sraaphorst:
One more update was required for the F2 MOS blueprint. It has been tested with an F2 MOS proposal using the skeleton ODB import and shows a correct result.


sraaphorst:
Turns out my loaner laptop doesn't have VPN installed, so I'll do the test release tomorrow morning.

PR #1676, opened 2019-06-18
Turn off reports by default in g[ns]odbtest

swalker2m:
The reports service does not run reliably in `g[ns]odbtest`.  This is apparently related to the polling of EPICS weather channels in the test environment and results in pegging the CPU and the inability to turn off either the reports bundle itself or the `spdb` process as a whole.  It would be nice to find out why the EPICS channels cannot be monitored in the test environment, but in lieu of that we can also just skip it for the test servers.  That's what this PR accomplishes.  Specifically it

* runs the reports bundle at level 60
* sets the default startlevel to 50 in the `g[ns]odbtest` configurations

Yes, this is a hack.

### DIFF
--- a/app/spdb/build.sbt
+++ b/app/spdb/build.sbt
@@ -427,7 +427,8 @@ def odbtest(version: Version) = AppConfig(
     "edu.gemini.dbTools.tcs.ephemeris.directory" -> "/home/software/ugemini/ephemerides",
     "edu.gemini.smartgcal.svnRootUrl"            -> "http://source.gemini.edu/gcal/branches/development/calibrations",
     "edu.gemini.spdb.dir"                        -> "/home/software/ugemini/spdb/spdb.active",
-    "edu.gemini.util.trpc.name"                  -> "Gemini ODB (Test)"
+    "edu.gemini.util.trpc.name"                  -> "Gemini ODB (Test)",
+    "org.osgi.framework.startlevel.beginning"    -> "50"
   )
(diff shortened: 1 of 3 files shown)


sraaphorst:
Excellent. Thanks for doing this: it should save a lot of time and hassle.

app/spdb/build.sbt
```diff
@@ -427,7 +427,8 @@ def odbtest(version: Version) = AppConfig(
     "edu.gemini.dbTools.tcs.ephemeris.directory" -> "/home/software/ugemini/ephemerides",
     "edu.gemini.smartgcal.svnRootUrl"            -> "http://source.gemini.edu/gcal/branches/development/calibrations",
     "edu.gemini.spdb.dir"                        -> "/home/software/ugemini/spdb/spdb.active",
-    "edu.gemini.util.trpc.name"                  -> "Gemini ODB (Test)"
+    "edu.gemini.util.trpc.name"                  -> "Gemini ODB (Test)",
+    "org.osgi.framework.startlevel.beginning"    -> "50"
```
sraaphorst: Isn't the default startlevel 50? Why did you explicitly add it as such?

swalker2m: The startlevel default is 99 everywhere else.  I dropped it to 50 for the test machines, which corresponds to the normal gemini bundles.  That leaves the reports bundle and the dataman off by default on the test machines.  In production and in development, reports runs by default but dataman still won't.

PR #1678, opened 2019-06-27
REL-3698: more logging

swalker2m:
REL-3698 concerns an end dataset event that was apparently received but not processed.  It isn't clear from the logs what happened.  This PR simply tightens a few loose screws, removes some unused code, and adds more logging that might help diagnose the issue or at least narrow the search space if it comes up again.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/event/ExecEvent.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/event/ExecEvent.java
@@ -7,6 +7,8 @@ import edu.gemini.spModel.pio.PioParseException;
 import edu.gemini.shared.util.GeminiRuntimeException;
 
 import java.io.Serializable;
+import java.time.Instant;
+import static java.time.format.DateTimeFormatter.ISO_INSTANT;
 import java.util.Comparator;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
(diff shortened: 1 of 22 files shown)


bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/event/ExecEvent.java
```diff
@@ -148,4 +150,20 @@ public int hashCode() {
      * Returns a human-readable name for the event.
      */
     public abstract String getName();
+
+    @Override
+    public String toString() {
+        final String subProps          = toStringProperties();
+        final String formattedSubProps = "".equals(subProps) ? "" : ", " + subProps;
```
cquiroz: neat

bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/session/AbstractSessionEventConsumer.java
```diff
@@ -69,21 +66,22 @@ public AbstractSessionEventConsumer(ISessionEventProducer ssp) {
...
-    public void stop() {
-        _done = true;
-    }
+    public abstract String getName();
 
     // Runnable
     public void run() {
         // Remove from queue and run
-        while (!_done) {
+        while (true) {
```
cquiroz: Is this correct? we run this forever?

swalker2m: Apparently.  There were no usages of `stop` and anyway it wasn't implemented correctly.

PR #1680, opened 2019-07-09
REL-3674: take slew (mostly) into account

swalker2m:
An update to time accounting rules to take slew observations into account.  Some background:

* Acquisition observations are separate from science observations because reasons (I can elaborate if necessary)
* The typical pattern is to use the science observation to slew to a target in one visit, switch to the acquisition observation, then switch back and complete the science observation in a separate visit
* REL-1608 introduced a change to not charge when a visit produces no chargeable datasets

Taken together this means that slew time in the initial visit was going uncharged.  Unfortunately we have no way of knowing that the science and acquisition observations are related and that there were no intervening observations in the middle of this { slew, acquire, science } sequence.  We would need that information to correctly charge or not for visits.

This PR represents a compromise selected by science to mostly account for slew observations. In particular, when calculating the charges for a visit we will now charge for visits that:

* contain a slew but otherwise are not chargeable (no chargeable datasets)
* are followed by a visit that does not contain a slew but does have chargeable time


swalker2m:
For your convenience:
image

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/obsrecord/VisitCalculator.scala
```diff
@@ -167,8 +183,112 @@ private[obsrecord] object VisitCalculator {
...
+
+  /**
+   * Finds the `VisitCalculator` that corresponds to the given visit.
+   */
+  def lookup(e: VisitEvents): VisitCalculator =
+    (for {
+      h <- e.sorted.headOption
+      s <- h.site
+      c <- all.find(_.validAt(s).isBefore(h.instant))
+    } yield c).getOrElse(all.head)
```
tpolecat: Would it be a hassle to make `all` a NEL while we're mucking around in here?

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/obsrecord/VisitCalculator.scala
```diff
@@ -167,8 +183,112 @@ private[obsrecord] object VisitCalculator {
...
+    instrument: Option[Instrument],
+    obsClass:   ObsClass,
+    datasetQa:  DatasetLabel => DatasetQaState,
+    datasetOc:  DatasetLabel => ObsClass
+  ): List[VisitTimes] =
+
+    visits.foldRight(List.empty[(VisitTimes, VisitEvents)]) { (visit, result) =>
+      val times = lookup(visit).calc(visit, instrument, obsClass, datasetQa, datasetOc, result.headOption)
+      (times, visit) :: result
+    }.unzip._1
```
tpolecat: This is very clever.

tpolecat:
I thought we weren't really paying attention to execution times anymore, as long as everyone gets their allocation at ITAC. Is that not actually true?

swalker2m:
> I thought we weren't really paying attention to execution times anymore, as long as everyone gets their allocation at ITAC. Is that not actually true?

Good question.  @andrewwstephens would know more but I think that though they aren't sweating over every second anymore they do want to make sure that we're not going way over allocation on any given program.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/obsrecord/VisitCalculator.scala
swalker2m: Good point. It's only used in one place so it was easy to switch.

andrewwstephens:
@tpolecat - @swalker2m is correct.  We don't bother correcting small errors (e.g. 10 minutes lost due to an instrument fault) we do fix gross errors so that PI's get the time they were promised.  In this case each instance is small (typically <10 minutes) but it is happening **every** time we acquire a target which adds up fast for programs with lots of targets.

PR #1692, opened 2019-07-30
REL-3701: Removed Texes and Phoenix from 2020A.

sraaphorst:
As per the REL task instructions...

### DIFF
--- a/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
+++ b/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
@@ -83,6 +83,7 @@ class ProblemRobot(s: ShellAdvisor) extends Robot {
           TacProblems(p, s).all ++
           Semester2018BProblems(p, s).all ++
           Semester2019BProblems(p, s).all ++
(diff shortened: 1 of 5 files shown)


bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
```diff
@@ -711,6 +712,26 @@ object TimeProblems {
   }
 }
 
+case class Semester2020AProblems(p: Proposal, s: ShellAdvisor) {
```
cquiroz: i don't recall this very well but do we need to keep `SemesterXXXXProblems` of previous semesters?

sraaphorst: I introduced the `SemesterXXXXXProblems` just to segregate things to make them easier to know when they were introduced, IIRC, but you are right: looking at them, we do not need the older ones. I'll go ahead and remove them.

sraaphorst: I got rid of the two outdated `SemesterXXXXXProblems` and graduated `dssiObsolete` to a permanent problem.

PR #1693, opened 2019-08-02
REL-3714: Add IGRINS to the PIT

sraaphorst:
This incredibly frustrating PR is finally done. IGRINS is the first instrument we have that has no configuration whatsoever, which necessitated determining how to add it since it seemed a configuration UI was expected for each instrument and the type of `Root.apply` makes no sense whatsoever.

Then there was a strange bug which arose from a small error in `Proposal.xsd`, but now it appears that everything is working correctly and all the specs pass. Here are some fancy pics of things in action:

Screen Shot 2019-08-01 at 12 38 50 PM

Screen Shot 2019-08-01 at 12 39 01 PM


sraaphorst:
Note that this also handles the OT / ODB work, and I corrected a small issue I came across in the `ZorroBlueprintSpec` where I had the `'` from `'Alopeke` accidentally in front of `Zorro`.

sraaphorst:
This will definitely need a squash and merge. 😠

cquiroz:
This looks good. However being hard on the PIT makes me worry how hard will it be on ITAC? Any considerations about that?

sraaphorst:
@cquiroz I'll see if I can get it working witht ITAC on Monday... just this once, though. 😈

sraaphorst:
@cquiroz I checked this and it looks easy enough to do. Seems like pretty much a straightforward copy of Alopeke and Zorro, but without the mode.

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/dtree/Root.scala
```diff
@@ -8,7 +8,7 @@ class Root(sem:Semester) extends SingleSelectNode[Semester, Instrument, Any](sem
   var title = "Select Instrument"
   var description = s"The following instruments are available for semester ${sem.year}${sem.half}. See the Gemini website for information on instrument capabilities and configuration options."
 
-  def choices = List(Alopeke, Flamingos2, GmosNorth, GmosSouth, Gnirs, Gpi, Graces, Gsaoi, Nifs, Niri, Visitor, Zorro)
+  def choices = List(Alopeke, Flamingos2, GmosNorth, GmosSouth, Gnirs, Gpi, Graces, Gsaoi, Igrins, Nifs, Niri, Visitor, Zorro)
```
swalker2m: It was already this way before, I know, but let's go ahead and add the type to public methods?

```scala
def choices: List[Instrument] =
  List(Alopeke, ...)
```

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/dtree/Root.scala
```diff
@@ -8,7 +8,7 @@ class Root(sem:Semester) extends SingleSelectNode[Semester, Instrument, Any](sem
...
 
   def apply(i:Instrument) = i match {
```
swalker2m: Here the type is maybe something like
```scala
Either[SelectNode[Instrument, _, _, GeminiBlueprintBase], GeminiBlueprintBase]
```

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/IgrinsBlueprint.scala
```diff
@@ -0,0 +1,31 @@
...
+import edu.gemini.model.p1.mutable.IgrinsBlueprintChoice
+import edu.gemini.model.p1.{mutable => M}
+
+object IgrinsBlueprint {
+  def apply(m: M.IgrinsBlueprint): IgrinsBlueprint = new IgrinsBlueprint(m)
+}
+
+case class IgrinsBlueprint() extends GeminiBlueprintBase {
+  override val name: String = "IGRINS"
+  override val visitor = true
```
swalker2m: `: Boolean`

(I know, it's obvious here, but I think it's a good idea to document the expected type of all public methods.)

sraaphorst:
Agreed with the typing of public methods. I'll get that taken care of.

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/dtree/Root.scala
sraaphorst: Trying both `SelectNode` and `Node` doesn't seem to work, either. I get, for each instrument, something akin to:
```
Error:(14, 41) type mismatch;
 found   : edu.gemini.model.p1.dtree.inst.Alopeke.ModeNode
 required: edu.gemini.model.p1.dtree.SelectNode[edu.gemini.model.p1.immutable.Instrument, _, _, edu.gemini.model.p1.immutable.GeminiBlueprintBase]
    case Alopeke    => Left(inst.Alopeke())
```

I'll take another look at it tomorrow.

sraaphorst: Note this is for:
```
  def apply(i:Instrument): Either[SelectNode[Instrument, _, _, GeminiBlueprintBase], GeminiBlueprintBase] = i match {
     case Alopeke    => Left(inst.Alopeke())
     ...
```

swalker2m: It looks like `Alopeke.ModeNode` is a `SingleSelectNode` where the instrument parameter is just `Unit` and not really an `Instrument`.  Don't worry I don't think you should spend too much time on this since it has always been this way.

PR #1698, opened 2019-08-13
Updated mac icon

cquiroz:
For some reason the previous mac icon didn't work when packaging. This icon was test building a dmg and installing it

image


sraaphorst:
The sites I used to use to generate the icons have become notoriously unreliable. What did you use to convert the PNG to the icns?

cquiroz:
I used this one
https://cloudconvert.com/png-to-icns

sraaphorst:
I'm pretty sure that that's the one I used to use, which started generating garbled output for icns. Maybe they finally fixed it? It was definitely the most convenient of the ones I was able to find online.

I ended up downloading an app to make icns files, but I guess that that didn't work properly, although it seemed to.

cquiroz:
I think I did the conversion for this years PR but don't recall where I did it

PR #1703, opened 2019-08-30
REL-3676: add Gaia adapter

swalker2m:
This installment wedges in a Gaia catalog adapter, completing the abuse of the original design.

### DIFF
--- a/bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/api/CatalogQuery.scala
+++ b/bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/api/CatalogQuery.scala
@@ -38,6 +38,27 @@ object CatalogName {
     override val  supportedBands = List(MagnitudeBand._g, MagnitudeBand._r, MagnitudeBand._i, MagnitudeBand.B, MagnitudeBand.V, MagnitudeBand.UC, MagnitudeBand.J, MagnitudeBand.H, MagnitudeBand.K)
   }
 
+  case object Gaia extends CatalogName("gaia", "GAIA @ ESA") {
+
+    // Gaia bands (why isn't this a Set?)
+    override val supportedBands: List[MagnitudeBand] =
+      List(
+        MagnitudeBand.V,
+        MagnitudeBand.R,
+        MagnitudeBand.I,
+        MagnitudeBand._r,
+        MagnitudeBand._i,
+        MagnitudeBand._g,
+        MagnitudeBand.K,
+        MagnitudeBand.H,
+        MagnitudeBand.J
+      )
+
+    override val rBand: MagnitudeBand =
+      MagnitudeBand.R
+
+  }
+
   case object TWOMASS_PSC extends CatalogName("twomass_psc", "TwoMass PSC @ Gemini")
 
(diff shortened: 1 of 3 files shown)


bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableParser.scala
```diff
@@ -288,6 +288,102 @@ object CatalogAdapter {
...
+        Conversion(MagnitudeBand._r,  0.128790, -0.24662, 0.027464, 0.049465,  0.20, 2.00),
+        Conversion(MagnitudeBand._i,  0.296760, -0.64728, 0.101410, 0.000000,  0.00, 4.50),
+        Conversion(MagnitudeBand._g, -0.135180,  0.46245, 0.251710, 0.021349, -0.50, 2.00),
+        Conversion(MagnitudeBand.K,   0.188500, -2.09200, 0.134500, 0.000000,  0.25, 5.50),
+        Conversion(MagnitudeBand.H,   0.162100, -1.96800, 0.132800, 0.000000,  0.25, 5.00),
+        Conversion(MagnitudeBand.J,   0.018830, -0.39400, 0.078930, 0.000000, -0.50, 5.50)
+      )
+
+    override def isMagnitudeField(v: (FieldId, String)): Boolean =
+      sys.error("unused")
```
swalker2m: Gaia magnitudes are all derived from a single value, so none of these methods are used.  I think this is telling us that catalogs are different enough that we should just refactor this into making each catalog implementation distinct instead of trying to use adapters.

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/api/CatalogQuery.scala
```diff
@@ -38,6 +38,27 @@ object CatalogName {
     override val  supportedBands = List(MagnitudeBand._g, MagnitudeBand._r, MagnitudeBand._i, MagnitudeBand.B, MagnitudeBand.V, MagnitudeBand.UC, MagnitudeBand.J, MagnitudeBand.H, MagnitudeBand.K)
   }
 
+  case object Gaia extends CatalogName("gaia", "GAIA @ ESA") {
+
+    // Gaia bands (why isn't this a Set?)
```
cquiroz: Feel free to make it a `Set`

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableParser.scala
```diff
@@ -288,6 +288,102 @@ object CatalogAdapter {
...
+
+      type Try[A] = CatalogProblem \/ A
+
+      def doubleValue(f: FieldId): OptionT[Try, Double] =
+        OptionT[Try, Double](
+          entries
+            .get(f)
+            .filter(_.nonEmpty)
+            .traverseU(CatalogAdapter.parseDoubleValue(f.ucd, _))
+        )
```
swalker2m: There are admittedly a few similar parsing methods floating around the `VoTableParser`.

PR #1704, opened 2019-09-02
REL-3676: add Gaia to VoTableClient

swalker2m:
The next REL-3676 installment adds Gaia to `VoTableClient` and updates the manual catalog UI to include Gaia in the list of options.

The bulk of this PR is the VO Table schema 1.3 version (which is needed for the Gaia results) but doesn't need to be reviewed.  It also adds a `VersionToken` class to work with the schema versions.

### DIFF
--- a/bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/api/CatalogQuery.scala
+++ b/bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/api/CatalogQuery.scala
@@ -1,6 +1,6 @@
 package edu.gemini.catalog.api
 
-import edu.gemini.spModel.core.{MagnitudeBand, Coordinates, SiderealTarget}
+import edu.gemini.spModel.core.{MagnitudeBand, Coordinates, SiderealTarget, VersionToken}
 
 import scalaz._
 import Scalaz._
@@ -18,9 +18,17 @@ case class MagnitudeQueryFilter(mc: MagnitudeConstraints) extends QueryResultsFi
 }
 
 sealed abstract class CatalogName(val id: String, val displayName: String) extends Product with Serializable {
-  def supportedBands: List[MagnitudeBand] = Nil
+
+  def supportedBands: List[MagnitudeBand] =
+    Nil
+
   // Indicates what is the band used when a generic R band is required
-  def rBand: MagnitudeBand = MagnitudeBand.UC
+  def rBand: MagnitudeBand =
+    MagnitudeBand.UC
+
+  def voTableVersion: VersionToken =
+    VersionToken.unsafeFromIntegers(1, 2)
+
 }
 
 object CatalogName {
@@ -56,6 +64,9 @@ object CatalogName {
 
     override val rBand: MagnitudeBand =
       MagnitudeBand.R
+
+    override val voTableVersion: VersionToken =
+      VersionToken.unsafeFromIntegers(1, 3)
 
   }
 
--- a/bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableClient.scala
+++ b/bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableClient.scala
@@ -4,7 +4,7 @@ import java.net.{URL, UnknownHostException}
 import java.util.concurrent.atomic.AtomicInteger
 import java.util.logging.Logger
 
-import edu.gemini.catalog.api.{RadiusConstraint, NameCatalogQuery, ConeSearchCatalogQuery, CatalogQuery}
+import edu.gemini.catalog.api.{CatalogName, RadiusConstraint, NameCatalogQuery, ConeSearchCatalogQuery, CatalogQuery}
 import edu.gemini.spModel.core.Angle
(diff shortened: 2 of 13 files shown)


bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/api/CatalogQuery.scala
```diff
@@ -18,9 +18,17 @@ case class MagnitudeQueryFilter(mc: MagnitudeConstraints) extends QueryResultsFi
...
+  def supportedBands: List[MagnitudeBand] =
+    Nil
+
   // Indicates what is the band used when a generic R band is required
-  def rBand: MagnitudeBand = MagnitudeBand.UC
+  def rBand: MagnitudeBand =
+    MagnitudeBand.UC
+
+  def voTableVersion: VersionToken =
+    VersionToken.unsafeFromIntegers(1, 2)
```
swalker2m: The associated VO Table version.  This is actually included in the query results and we should take it from there instead at some point.  In the meantime I needed a way to refer to the 1.3 schema for Gaia.

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableClient.scala
```diff
@@ -190,6 +190,54 @@ case object ConeSearchBackend extends CachedBackend with RemoteCallBackend {
   override def queryUrl(e: SearchKey): String = s"${e.url}/cgi-bin/conesearch.py"
 }
 
+case object GaiaBackend extends CachedBackend with RemoteCallBackend {
```
swalker2m: The substantive part of the PR is the `GaiaBackend`, which despite the name is charged with forming the query to send to the Gaia server.

None of the "backend" options are really unit tested, presumably because that would require making remote calls. I'm not sure what to do for testing other than monkey testing.

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableClient.scala
```diff
@@ -232,10 +280,20 @@ trait VoTableClient {
...
+      case CatalogName.SIMBAD => SimbadNameBackend
+      case _                  => ConeSearchBackend
+    }
+
   /**
    * Do a query for targets, it returns a list of targets and possible problems found
    */
-  def catalog(query: CatalogQuery, backend: VoTableBackend = ConeSearchBackend)(ec: ExecutionContext): Future[QueryResult] = {
+  def catalog(query: CatalogQuery, explicitBackend: Option[VoTableBackend])(ec: ExecutionContext): Future[QueryResult] = {
+    val backend = explicitBackend.getOrElse(defaultBackend(query.catalog))
```
swalker2m: I changed the `catalog` to make the "backend" default according to the catalog instead of always being `ConeSearchBackend`.

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableParser.scala
```diff
@@ -31,20 +31,19 @@ object VoTableParser extends VoTableParser {
...
-  val xsd = "/votable-1.2.xsd"
+  private def validate(catalogName: CatalogName, xmlText: String): CatalogProblem \/ Unit =
+    \/.fromTryCatchNonFatal {
+      import javax.xml.transform.stream.StreamSource
+      import javax.xml.validation.SchemaFactory
 
-  private def validate(xmlText: String): Throwable \/ String = \/.fromTryCatchNonFatal {
-    import javax.xml.transform.stream.StreamSource
-    import javax.xml.validation.SchemaFactory
+      val xsd        = s"/votable-${catalogName.voTableVersion.format}.xsd"
```
swalker2m: The version should probably come from the XML itself, at some point.

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/VersionToken.scala
```diff
@@ -0,0 +1,73 @@
+package edu.gemini.spModel.core
+
+import scalaz._
+import Scalaz._
+
+/**
+ * Simple integer version token (e.g., 1.2.3) used to distinguish VO Table
+ * versions.
+ */
+sealed abstract case class VersionToken private (toNel: NonEmptyList[Int]) {
```
swalker2m: Perhaps this is overkill.  I just needed a way to specify a version number like "1.2.3".

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableClient.scala
cquiroz: If I recall this, it was possible to do some testing using local files

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableParser.scala
cquiroz: 👍

PR #1706, opened 2019-09-03
Update squants

cquiroz:
This should help remove the issues upstream when updating sbt to 1.3.0


cquiroz:
Fixed the issue with ITC, some tests depended on the hashes and that changed on squants
I updated squants to 1.5.0 in the meanwhile

tpolecat:
> some tests depended on the hashes
 
O_o

PR #1712, opened 2019-10-18
REL-3729: Removed emailConf.xml from OODB TOO code and into ocs credentials

sraaphorst:
TOO email exploder addresses were formerly configured directly in the OODB.

This made testing the TOO email notification system unwieldly, as testing often requires sending out many emails. Having these emails sent to the exploders is confusing to those people who are listed on them.

Instead, to test TOO email notification, we direct the emails sent by the system to individuals within Gemini. This eliminates the exploder confusion, but previously had to be done by hand, necessitating finicky editing of `emailConf.xml` when alternating between testing and production.

This PR removes this inconvenience by determining whether the running version is in production mode or in testing mode. If in testing mode, the system is configured to send TOO emails to individuals. If in production mode, the system functions normally and sends TOO emails to all individuals on the exploders.

**NOTE:** This requires updating the OCS credentials from our internal SVN:

http://source.gemini.edu/software/ocs-credentials

and relinking them by performing a `link.sh <ocs repo location>`.


bundle/edu.gemini.oodb.too.window/src/main/java/edu/gemini/too/email/osgi/OsgiEmailConfig.java
```diff
@@ -74,7 +68,12 @@ public Site getSite() {
     }
 
     public URL getEmailConfig() {
-        return getClass().getClassLoader().getResource("/resources/emailConf.xml");
+        // If we are testing, use the testing email configurations.
+        // If we are in production, use the production email configurations.
+        if (Version.current.isTest())
+            return getClass().getClassLoader().getResource("resources/emailConf-test.xml");
```
swalker2m: I think this will need a leading "/".

How about something like

```java
final String test = Version.current.isTest() ? "-test" : "";
return getClass().getClassLoader().getResource("/resources/emailConf" + test + ".xml");
```

bundle/edu.gemini.oodb.too.window/src/main/java/edu/gemini/too/email/osgi/OsgiEmailConfig.java
```diff
@@ -74,7 +68,12 @@ public Site getSite() {
...
+        else
+            return getClass().getClassLoader().getResource("/resources/emailConf.xml");
```
cquiroz: Wouldn't we need `emailConf.xml` for the distributions?

bundle/edu.gemini.oodb.too.window/src/main/java/edu/gemini/too/email/osgi/OsgiEmailConfig.java
sraaphorst: Yes, I think you're right.

bundle/edu.gemini.oodb.too.window/src/main/java/edu/gemini/too/email/osgi/OsgiEmailConfig.java
swalker2m: Before it was just one file taken from the classpath.  The only change is to add another one, with testing emails that we will use if the version is a test version.  I may be missing something though?

sraaphorst: I don't think so? Previously there was only one for all distributions, and this just splits it into two depending on whether the test flag is for the current `Version`.

PR #1715, opened 2019-10-23
add montage image servers

tpolecat:
This adds the Montage version of the 2MASS image server, which create custom tiled images that fill the TPE field better than the existing servers. Here is an example looking at M3. It's a little movie actually, if you click on it.

Note that the image in the example has been cached locally. The first time it was accessed it took about 30 seconds, because the mosaic process has to do a lot of work aligning/scaling/rotating and matching contrast levels. The server caches tiles (so nearby positions will resolve more quickly) as well as the final mosaic image (so other people looking at the observation should receive the image quickly).

This is work-in-progress (need to move the montage server over to Gemini, it's running on my private Heroku account right now).

### DIFF
--- a/bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/image/ImageCatalog.scala
+++ b/bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/image/ImageCatalog.scala
@@ -40,9 +40,18 @@ case object TwoMassHId extends CatalogId {
 case object TwoMassKId extends CatalogId {
   val filePrefix = "2massK"
 }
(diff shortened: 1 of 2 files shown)


cquiroz:
Why do we still have the non-montage version? To have a faster alternative?

tpolecat:
Yeah, the normal 2MASS servers are much faster because you just pull whatever tile contains the target and you're done. I think most of the time it's probably ok since it has worked like this forever. Montage is just for cases where you need a 2MASS image that covers the entire TPE field.

tpolecat:
Ok this is working now. The [gsp-montage](https://github.com/gemini-hlsw/gsp-montage) service is working and auto-deploying properly.

PR #1716, opened 2019-10-24
Scheduling Blocks

swalker2m:
This PR contains a couple of updates to the way that scheduling blocks are handled:

1. When importing old programs it adds a default scheduling block to observations that do not have one.
2. It makes the scheduling block editor always available in the target environment form.

The context for these changes is that GeMS AGS is failing for older observations that have no scheduling block.  This happens because in theory you cannot know the base position coordinates without first knowing *when* you're asking for them. "In theory" because in reality we don't apply proper motion updates to sidereal targets anyway. Nevertheless, without base position coordinates there's no way to know whether potential guide stars are in range.

We create a default scheduling block for new observations anyway, so this seems like the natural progression.  At this point, I don't see any reason why the scheduling block should remain optional in the model but I didn't want to incorporate that refactoring.

I made the scheduling block editable since it is disconcerting to see a setting that you cannot edit and no clear way to enable the widgets.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/obscomp/SPProgram.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/obscomp/SPProgram.java
@@ -45,7 +45,7 @@ public class SPProgram extends AbstractDataObject implements ISPStaffOnlyFieldPr
     // for serialization
     private static final long serialVersionUID = 4L;
 
-    public static final String VERSION = "2018A-1";
+    public static final String VERSION = "2020A-1";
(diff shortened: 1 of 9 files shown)


cquiroz:
Code looks great. I wonder if this would affect the seqexec. I guess no since you didn't remove the Option but perhaps we need to upgrade the libs anyway

swalker2m:
> I wonder if this would affect the seqexec.

It isn't included in the sequence itself so I wouldn't expect it to.  Note this is part of 2020A.

The scheduling block is used for computing where a target is at a given time, and for calculating the average parallactic angle over a specific time period.

PR #1718, opened 2019-10-25
REL-3743-B: fix test case which would have failed in the future

swalker2m:
This PR updates #1716 to future-proof a test case.  The way that `DefaultSchedulingBlock` works is that it

1. Extracts the semester from a program id
2. Computes the midpoint of said semester
3. Compares it to the midpoint of the current semester
4. Chooses the bigger of the two for the scheduling block start

```scala
  def forPid(pid: SPProgramID): SchedulingBlock =
    SchedulingBlock(
      getPidSemesterCenter(pid).foldLeft(getCurrentSemesterCenter(Site.GS))(_ max _)
    )
```

The test case ignored that fact and would begin failing post-2020A without this update.

### DIFF
--- a/bundle/edu.gemini.spModel.io/src/test/scala/edu/gemini/spModel/io/impl/migration/to2020A/SchedulingBlockTest.scala
+++ b/bundle/edu.gemini.spModel.io/src/test/scala/edu/gemini/spModel/io/impl/migration/to2020A/SchedulingBlockTest.scala
@@ -2,16 +2,28 @@ package edu.gemini.spModel.io.impl.migration.to2020A
 
 import edu.gemini.pot.sp.{ISPObservation, ISPProgram}
 import edu.gemini.shared.util.immutable.ScalaConverters._
+import edu.gemini.spModel.core.{Semester, Site}
 import edu.gemini.spModel.io.impl.migration.MigrationTest
 import edu.gemini.spModel.obs.{SchedulingBlock, SPObservation}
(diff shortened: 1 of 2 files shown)


bundle/edu.gemini.spModel.io/src/test/scala/edu/gemini/spModel/io/impl/migration/to2020A/SchedulingBlockTest.scala
```diff
@@ -2,17 +2,29 @@ package edu.gemini.spModel.io.impl.migration.to2020A
...
 class SchedulingBlockTest extends Specification with MigrationTest {
 
   val ProgramName: String =
     "GS-2020A-Q-501.xml"
 
+  val sem2020A: Semester =
+    new Semester(2020, Semester.Half.A)
+
+  def currentSemester: Semester =
+    new Semester(Site.GS, System.currentTimeMillis)
```
swalker2m: I realize this isn't referentially transparent, should be done in `IO`, etc. but I think this is fine for a small test case.  Technically it can fail if you're unlucky and run the test right at the semester switchover time at GS but that also doesn't seem like it is worth sweating over.

cquiroz: 👍

PR #1722, opened 2019-10-29
re-use JAXB context

tpolecat:
re-using the JAXB context for subsequent P1 Proposal deserialization speeds things up significantly. Loading up the 19B proposals went from 30s to < 2s.


cquiroz:
That is a good insight? Is there any reason why you'd need a different context each time? Is it stateful perhaps?

tpolecat:
The context is stateless but the marshallers it creates are stateful and can't be used concurrently.

PR #1729, opened 2019-11-22
GHOSTSW-9: Divide asterism types into resolution modes

sraaphorst:
In the GHOST instrument configuration node, the user chooses the asterism type to be observed: this determines the behaviour of the target node component in terms of what is allowed and what is expected.

Originally, we had a single drop-down list of all the available asterism modes. It was decided that this should be refined into three resolution modes - standard resolution, high resolution, and precision radial velocity - and then have sub-modes that determine the asterism type.

In addition, when the user changes resolution modes, we attempt to pick the asterism type closest in the new resolution mode to the old resolution mode and perform a conversion on the target environment so that no information is lost.

Here is what the original looked like:

Screen Shot 2019-11-22 at 9 52 36 AM

Here are some screen shots of the new layout:

Screen Shot 2019-11-22 at 9 46 37 AM
Screen Shot 2019-11-22 at 9 46 27 AM
Screen Shot 2019-11-22 at 9 46 10 AM

### DIFF
--- a/bundle/jsky.app.ot/src/main/resources/resources/conf/UIInfo.xml
+++ b/bundle/jsky.app.ot/src/main/resources/resources/conf/UIInfo.xml
@@ -390,7 +390,7 @@
 
     <UIInfo site="GS"
             dataObject="edu.gemini.spModel.gemini.ghost.Ghost"
-            name="GHOST instrument"
+            name="GHOST Instrument"
             type="instrument"
             imageKey="component.gif"
             shortDescription="The GHOST instrument is configured with this component."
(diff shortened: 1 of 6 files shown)


sraaphorst:
I added `Fill.Horizontal` to the `Constraints` on the `ComboBox`es so that they are now evenly sized:

Screen Shot 2019-11-22 at 10 03 13 AM

bundle/jsky.app.ot/src/main/resources/resources/conf/UIInfo.xml
```diff
@@ -390,7 +390,7 @@
 
     <UIInfo site="GS"
             dataObject="edu.gemini.spModel.gemini.ghost.Ghost"
-            name="GHOST instrument"
+            name="GHOST Instrument"
```
cquiroz: Are we still using these xml files?

sraaphorst:
The AGS parts were already in place and aren't affected by these changes.

PR #1730, opened 2019-11-22
GHOSTSW-8: Added ISS port.

sraaphorst:
Added the ISS port tab to the GHOST instrument editor.
We include this in a tab because we will have multiple tabs forthcoming for more information / configuration.

Screen Shot 2019-11-22 at 12 35 32 PM

### DIFF
--- a/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/Ghost.scala
+++ b/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/Ghost.scala
@@ -11,10 +11,11 @@ import edu.gemini.spModel.data.property.{PropertyProvider, PropertySupport}
 import edu.gemini.spModel.gemini.init.{ComponentNodeInitializer, ObservationNI}
 import edu.gemini.spModel.obs.SPObservation
 import edu.gemini.spModel.obscomp.{InstConfigInfo, InstConstants, SPInstObsComp}
-import edu.gemini.spModel.pio.{ParamSet, PioFactory}
+import edu.gemini.spModel.pio.{ParamSet, Pio, PioFactory}
 import edu.gemini.spModel.seqcomp.SeqConfigNames
 import edu.gemini.spModel.target.env.TargetEnvironment
 import edu.gemini.spModel.target.obsComp.TargetObsComp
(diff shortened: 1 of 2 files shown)


cquiroz:
Can the instrument be really in either port?
I'm not sure what's the use of port in the OT (this is not particular to GHOST) since the port is the in A&G and can be different than what's on the OT

sraaphorst:
Yes, GHOST can be in either port. This was confirmed with Steve and was requested to be configurable via the OT. Setting the port in the OT, I believe, sets the position of the science area in the TPE so it affects AGS and user configuration of targets?

cquiroz:
Ok, that makes sense. It is a pity that this setting isn't connected with the reality of where the instrument is at the time of the observation.
OTOH it has been that way for ever

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/Ghost.scala
```diff
@@ -44,19 +45,37 @@ final class Ghost extends SPInstObsComp(GhostMixin.SP_TYPE) with PropertyProvide
...
     sc.putParameter(DefaultParameter.getInstance(Ghost.POS_ANGLE_PROP, getPosAngle))
+    sc.putParameter(DefaultParameter.getInstance(Ghost.PORT_PROP, getIssPort))
     sc
   }
+
+  /**
+   * ISS Port
+   */
+  private var port: IssPort = IssPort.DEFAULT
+  println("***** ISSPORT: " + port.displayValue())
```
cquiroz: println?

tpolecat:
totally gridbag https://www.youtube.com/watch?v=UuLaxbFKAcc

cquiroz:
`MigLayout` is the future of swing

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/Ghost.scala
sraaphorst: LOL ooops. Thanks for catching that!

cquiroz:
I have a special respect for anybody who can use `GridBagLayout`

PR #1731, opened 2019-11-22
REL-3757 - Fix error importing skeleton using GMOS NS 0.25" slit

andrewwstephens:
O -> 0


cquiroz:
Hmm, ITAC's not going to like this a lot

bundle/edu.gemini.model.p1/src/main/xjb/GmosN.xjb
```diff
@@ -162,7 +162,7 @@
         <!-- GmosNFpuNs => GmosNorthType.FPUnitNorth -->
         <jxb:bindings node="./xsd:simpleType[@name='GmosNFpuNs']/xsd:restriction">
             <jxb:bindings node="./xsd:enumeration[@value='0.25 arcsec slit']">
-                <jxb:typesafeEnumMember name="NS_O"/>
+                <jxb:typesafeEnumMember name="NS_0"/>
```
cquiroz: I see, this would break the skeleton, amazing we haven't hit this earlier

PR #1734, opened 2019-12-23
Add hardened runtime option for codesigning

cquiroz:
This PR adds an option to the OSX signing stage to enable the "hardened runtime" option. It still needs testing


swalker2m:
Seems fine. Does this work w/o updating the build machine?

cquiroz:
yes it requires a new build machine

swalker2m:
I think this is fine but maybe we should have the build machine updated before we merge it?

cquiroz:
I guess I should close this one

PR #1738, opened 2020-01-17
GHOST merge: PR requests

sraaphorst:
Additional PR requests from the GHOST merge to develop.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/calunit/CalUnitParams.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/calunit/CalUnitParams.java
@@ -123,7 +123,7 @@ public final class CalUnitParams {
             return Arrays.stream(ar).map(Lamp::getLamp).collect(Collectors.toList());
         }
 
-        public static String write(final TreeSet<Lamp> lamps) {
+        public static String write(final SortedSet<Lamp> lamps) {
             return show(lamps, Lamp::name);
         }
     }
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/calunit/CalUnitParams.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/calunit/CalUnitParams.java
@@ -6,10 +6,7 @@ import edu.gemini.spModel.type.ObsoletableSpType;
 import edu.gemini.spModel.type.SequenceableSpType;
 import edu.gemini.spModel.type.SpTypeUtil;
 
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
 import java.util.stream.Collectors;
 
 public final class CalUnitParams {
@@ -126,7 +123,7 @@ public final class CalUnitParams {
             return Arrays.stream(ar).map(Lamp::getLamp).collect(Collectors.toList());
         }
 
-        public static String write(final Set<Lamp> lamps) {
+        public static String write(final TreeSet<Lamp> lamps) {
             return show(lamps, Lamp::name);
         }
     }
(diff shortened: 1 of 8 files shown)


bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/seqcomp/GhostSeqRepeatFlatObs.java
```diff
@@ -65,7 +67,7 @@ public int getLampCount() {
         return lamps.size();
     }
 
-    public Set<Lamp> getLamps() {
+    public TreeSet<Lamp> getLamps() {
```
cquiroz: Why not return the more generic type?

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/calunit/CalUnitParams.java
```diff
@@ -126,7 +123,7 @@ public static boolean containsArc(Collection<Lamp> lamps) {
             return Arrays.stream(ar).map(Lamp::getLamp).collect(Collectors.toList());
         }
 
-        public static String write(final Set<Lamp> lamps) {
+        public static String write(final TreeSet<Lamp> lamps) {
```
swalker2m: Perhaps just `SortedSet`?

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/AsterismTypeConverters.scala
```diff
@@ -29,8 +29,5 @@ object AsterismTypeConverters {
       (GhostHigh, GhostHighResolutionTargetPlusSky, GhostStandard) -> GhostTargetPlusSky,
       (GhostHigh, GhostHighResolutionTargetPlusSky, GhostPRV) -> GhostHighResolutionTarget,
 
-      (GhostPRV, GhostHighResolutionTarget, GhostStandard) -> GhostSingleTarget,
-
-      // This oddly happens sometimes?
-      (GhostStandard, GhostHighResolutionTarget, GhostHigh) -> GhostHighResolutionTarget)
+      (GhostPRV, GhostHighResolutionTarget, GhostStandard) -> GhostSingleTarget)
```
swalker2m: Is the map complete? What would happen if we try to look up a conversion that isn't defined?

swalker2m: Perhaps it should be something like:

```scala
val asterismTypeConverters: ((ResolutionMode, AsterismType, ResolutionMode)) => Option[Asterism Type] =
  Map(
    (GhostHigh, GhostHighResolutionTargetPlusSky, GhostStandard) -> GhostTargetPlusSky,
    ...
  ).lift
```
so that it is obvious some mappings may not be defined?

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/seqcomp/GhostSeqRepeatFlatObs.java
sraaphorst: I have no objection returning `Set`, but you mentioned in the last PR that order might be important (and it is stored and returned as `TreeSet`), so I figured I would carry through the order preservation.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/calunit/CalUnitParams.java
sraaphorst: Sure.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/seqcomp/GhostSeqRepeatFlatObs.java
sraaphorst: (I was lifting this code from `SeqRepeatFlatObs`, which uses `TreeSet`.)

PR #1739, opened 2020-01-20
GHOSTSW-18: Proper motion coordinates now sent in sequence for targets.

sraaphorst:
GHOST requires target coordinates sent with proper motion taken into effect.
Previously, we sent regular coordinates. Now we use the proper motion calculator that was implemented from OCS3 to send the proper motion coordinates for targets.

Here, in the target editor, you can see the coordinates from the catalog for M32 along with the proper motion information.

Screen Shot 2020-01-20 at 4 13 54 PM

Here, in the static configuration, you can see the proper-motion adjusted coordinates for M32 being sent:

Screen Shot 2020-01-20 at 4 14 19 PM

### DIFF
--- a/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostCB.scala
+++ b/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostCB.scala
@@ -1,19 +1,22 @@
 package edu.gemini.spModel.gemini.ghost
 
+import java.time.Instant
 import java.util.{Map => JMap}
 
 import edu.gemini.pot.sp.ISPObsComponent
 import edu.gemini.spModel.config.AbstractObsComponentCB
 import edu.gemini.spModel.data.config._
 import edu.gemini.spModel.obscomp.InstConstants
-import edu.gemini.spModel.target.SPSkyObject
+import edu.gemini.spModel.target.{SPCoordinates, SPSkyObject, SPTarget}
 import edu.gemini.spModel.target.obsComp.TargetObsComp
 
 import scala.collection.JavaConverters._
+
 import scalaz._
 import Scalaz._
 
-/** Configuration builder for GHOST.
+/**
+  * Configuration builder for GHOST.
   */
 final class GhostCB(obsComp: ISPObsComponent) extends AbstractObsComponentCB(obsComp) {
   @transient private var sysConfig: Option[ISysConfig] = None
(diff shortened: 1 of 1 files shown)


bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostCB.scala
```diff
@@ -49,12 +52,18 @@ final class GhostCB(obsComp: ISPObsComponent) extends AbstractObsComponentCB(obs
...
-        config.putParameter(systemName, StringParameter.getInstance(raHMS,  c.ra.formatHMS))
-        config.putParameter(systemName, StringParameter.getInstance(decDMS, c.dec.formatDMS))
+                     raHMS: String, decDMS: String) : Unit = {
+        val coords = so match {
+          case c: SPCoordinates => c.getCoordinates.some
+          case t: SPTarget => t.getProperMotion.map(_.calculateAt(t.getTarget, Instant.now())).orElse(t.getCoordinates(None))
+        }
+        coords.foreach { c =>
+          name.foreach(n => config.putParameter(systemName, StringParameter.getInstance(n, so.getName)))
+          config.putParameter(systemName, StringParameter.getInstance(raDeg, c.ra.formatDegrees))
```
cquiroz: From the UI it looks like this contains the degree symbol? Is that correct? That could cause troubles downstream. I'd expect this to be sent as a double rather than a string

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostCB.scala
```diff
@@ -49,12 +52,18 @@ final class GhostCB(obsComp: ISPObsComponent) extends AbstractObsComponentCB(obs
...
-                     raHMS: String, decDMS: String) : Unit = so.getCoordinates(None).foreach { c =>
-        name.foreach(n => config.putParameter(systemName, StringParameter.getInstance(n, so.getName)))
-        config.putParameter(systemName, StringParameter.getInstance(raDeg,  c.ra.formatDegrees))
-        config.putParameter(systemName, StringParameter.getInstance(decDeg, c.dec.formatDegrees))
-        config.putParameter(systemName, StringParameter.getInstance(raHMS,  c.ra.formatHMS))
-        config.putParameter(systemName, StringParameter.getInstance(decDMS, c.dec.formatDMS))
+                     raHMS: String, decDMS: String) : Unit = {
+        val coords = so match {
+          case c: SPCoordinates => c.getCoordinates.some
+          case t: SPTarget => t.getProperMotion.map(_.calculateAt(t.getTarget, Instant.now())).orElse(t.getCoordinates(None))
```
swalker2m: If we hooked it up to `Target`, which we shouldn't do since it will change existing behavior elsewhere, this would be just `t.coords(instant)`.  As is, it makes sense.

I suppose we could add the time to the parameter list because you sometimes call this for multiple targets and it's better to push this up as much as we easily can anyway.  It will make *no difference* at the rate that these things go but nevertheless ...

It feels odd to not wrap this in `IO` because as is, this isn't referentially transparent, but I guess not in this codebase.

sraaphorst:
I changed from `StringParameter`s to` DefaultParameter`s for the angles in `GhostCB` when building the sequence, and now we use `Double` to represent the angles in degrees, as per discussion on Slack.

PR #1740, opened 2020-01-21
PIT-3768: Update icons for 2020B.

sraaphorst:


cquiroz:
If I recall we also update a png inside the PIT

sraaphorst:
@cquiroz Yes, thank you for catching that. I knew we changed for the OT but forgot about the PIT. I copied the new PNG in and will make a note so I don't forget next time.

PR #1743, opened 2020-01-23
REL-3769: Stupid erasure.

sraaphorst:
Of course, due to erasure, every blueprint was triggering the message that GPI is not offered in 2020B. I think this is the proper way to fix it: it appears to be working now.

### DIFF
--- a/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
+++ b/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
@@ -720,10 +720,10 @@ object TimeProblems {
 }
 
 object ResourceIssues {
-  def resourceNotOfferedCheck[B <: BlueprintBase](p: Proposal, s: ShellAdvisor, sem: Semester, name: Option[String] = None): List[Problem] = for {
+  def resourceNotOfferedCheck[B <: BlueprintBase : Manifest](p: Proposal, s: ShellAdvisor, sem: Semester, name: Option[String] = None): List[Problem] = for {
     o <- p.observations
     b <- o.blueprint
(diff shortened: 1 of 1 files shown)


bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
```diff
@@ -720,10 +720,10 @@ object TimeProblems {
 }
 
 object ResourceIssues {
-  def resourceNotOfferedCheck[B <: BlueprintBase](p: Proposal, s: ShellAdvisor, sem: Semester, name: Option[String] = None): List[Problem] = for {
+  def resourceNotOfferedCheck[B <: BlueprintBase : Manifest](p: Proposal, s: ShellAdvisor, sem: Semester, name: Option[String] = None): List[Problem] = for {
```
cquiroz: Isn't `Manifest` kind of deprecated, preferring `TypeTag` instead?

tpolecat: Hm, `b.isInstanceOf[B]` should have been causing a loud warning.

```
scala> def foo[A](x: Any): Boolean = x.isInstanceOf[A] 
<console>:11: warning: abstract type A is unchecked since it is eliminated by erasure
       def foo[A](x: Any): Boolean = x.isInstanceOf[A]
                                                   ^
foo: [A](x: Any)Boolean

scala> 
```

This is all terrible but :+1:. If this codebase had a future I would suggest we figure out a way to do this more safely.

@cquiroz in this case I think a class check is probably fine. With a `TypeTag` you could use a type pattern and do an exact match but you really only need that if `B` is a parameterized type.

PR #1744, opened 2020-01-23
REL-3772: Do not include investigator information in FT proposals.

sraaphorst:
As per the requirements, investigator information is no longer displayed in the generated PDFs for fast turnaround programs.

NOAO with investigators:
[ft-investigators.pdf](https://github.com/gemini-hlsw/ocs/files/4104498/ft-investigators.pdf)

NOAO without investigators:
[ft-no-investigators.pdf](https://github.com/gemini-hlsw/ocs/files/4104500/ft-no-investigators.pdf)

### DIFF
--- a/bundle/edu.gemini.model.p1.pdf/src/main/scala/edu/gemini/model/p1/pdf/P1PDF.scala
+++ b/bundle/edu.gemini.model.p1.pdf/src/main/scala/edu/gemini/model/p1/pdf/P1PDF.scala
@@ -3,9 +3,11 @@ package edu.gemini.model.p1.pdf
 import scala.util.Try
 import xml.{Node, XML}
 import java.io._
+
 import javax.xml.transform.URIResolver
 import javax.xml.transform.stream.StreamSource
 import edu.gemini.util.pdf.PDF
+
 import io.Source
 
 /**
@@ -84,7 +86,7 @@ object P1PDF {
   /** Gets a list with all templates that are currently available. */
   def templates = List(GeminiDefault, GeminiDefaultNoInvestigatorsList, GeminiDefaultListAtTheEnd, AU, CL, NOAO, NOAOListAtTheEnd, NOAONoInvestigatorsList)
 
-  def templatesMap = templatesList.toMap
+  def templatesMap: Map[String, Template] = templatesList.toMap
 
   def templatesList = List(
     "ar"     -> GeminiDefault,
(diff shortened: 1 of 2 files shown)


bundle/edu.gemini.model.p1.pdf/src/main/scala/edu/gemini/model/p1/pdf/P1PDF.scala
```diff
@@ -106,19 +108,29 @@ object P1PDF {
...
 
   /**
    * Creates a pdf from a given xml element and template and writes the resulting pdf file to the output folder.
    * This method also merges the attached pdf file to the end of the resulting pdf.
    */
   def createFromNode (xml: Node, template: Template, pdfFile: File, workingDir:Option[File] = None) {
+    // REL-3772: Do not display PI information for FT proposals.
+    val workingTemplate = {
+      val isFT: Boolean = (xml \ "proposalClass" \\ "fastTurnaround").theSeq.nonEmpty
+      if (isFT) template.copy(investigatorsList = InvestigatorsListOption.NoList) else template
```
cquiroz: Hmm, I thought we'd update the xslt and remove the page altogether, though this seems to be a valid alternative. Does the whole page disappear or is it just empty? Does that make any difference?

sraaphorst: We only want it for fast turnaround, and I figured since there was a `Template` class with a specific option for investigator information, it was probably the best way to go since the code was already in place to omit them.

I'm trying to upload an example but github is complaining about not being able to process the file.

sraaphorst: It allowed me to add the files to the description up above in case you want to take a look.

sraaphorst: (The last page is blank because I just use a blank PDF for the PDF attachment.)

sraaphorst: Also, another benefit: we don't have to edit multiple XSLT files, which we'd have to do if we went that route. (Wouldn't we have to change the one for each format?)

sraaphorst:
Excellent! Thanks! Next: test release!

PR #1745, opened 2020-01-27
REL-3772: Remove investigator information from FT backend PDFS2, take 2.

sraaphorst:
The investigator information for FT should only be removed from the p1-monitor. This restores the default behaviour for the PIT and p1pdfmaker and removes it p1-monitor.


bundle/edu.gemini.model.p1.pdf/src/main/scala/edu/gemini/model/p1/pdf/P1PDF.scala
```diff
@@ -121,16 +121,11 @@ object P1PDF {
    * This method also merges the attached pdf file to the end of the resulting pdf.
    */
   def createFromNode (xml: Node, template: Template, pdfFile: File, workingDir:Option[File] = None) {
-    // REL-3772: Do not display PI information for FT proposals.
```
sraaphorst: This is the revert for PIT and p1pdfmaker.

PR #1748, opened 2020-01-30
GHOSTSW-22: Added Science area.

sraaphorst:
Added the GHOST science area FOV, which is a 7.5 arcmin circle centred at the base.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostScienceAreaGeometry.scala
+++ b/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostScienceAreaGeometry.scala
@@ -1,0 +1,19 @@
+package edu.gemini.spModel.gemini.ghost
+
+import java.awt.Shape
+import java.awt.geom.Ellipse2D
+
+import edu.gemini.spModel.inst.ScienceAreaGeometry
+import edu.gemini.spModel.obs.context.ObsContext
+import scalaz._
+import Scalaz._
+
+object GhostScienceAreaGeometry extends ScienceAreaGeometry {
+  private val size = 7.5 * 60 // arcmin
+  private val radius = size / 2
+
+  /** Create the shape for the science area based on the instrument configuration.
+   * This shape is not adjusted for position angle or offsets. */
+  override def unadjustedGeometry(ctx: ObsContext): Option[Shape] =
+    new Ellipse2D.Double(-radius, -radius, size, size).some
+}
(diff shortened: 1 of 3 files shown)


bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostScienceAreaGeometry.scala
```diff
@@ -0,0 +1,19 @@
...
+import Scalaz._
+
+object GhostScienceAreaGeometry extends ScienceAreaGeometry {
+  private val size = 7.5 * 60 // arcmin
+  private val radius = size / 2
+
+  /** Create the shape for the science area based on the instrument configuration.
+   * This shape is not adjusted for position angle or offsets. */
+  override def unadjustedGeometry(ctx: ObsContext): Option[Shape] =
+    new Ellipse2D.Double(-radius, -radius, size, size).some
```
cquiroz: Is this the only use of `scalaz`? Maybe we should reduce our reliance on scalaz and do `Some(...)` instead

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostScienceAreaGeometry.scala
```diff
@@ -0,0 +1,19 @@
...
+import java.awt.Shape
+import java.awt.geom.Ellipse2D
+
+import edu.gemini.spModel.inst.ScienceAreaGeometry
+import edu.gemini.spModel.obs.context.ObsContext
+import scalaz._
+import Scalaz._
+
+object GhostScienceAreaGeometry extends ScienceAreaGeometry {
+  private val size = 7.5 * 60 // arcmin
```
cquiroz: Can this be defined with a richer type?

swalker2m: When we re-do this for GPP we should be using an `Angle` or something but, to be consistent with everything else in the OT/TPE, I think this is probably fine here.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostScienceAreaGeometry.scala
swalker2m: I guess I'm not too concerned about it because `scalaz` is scattered throughout the codebase already.

sraaphorst: I'm indifferent either way. I don't really like Scalaz so I'm find with `Some(_)` instead of `.some`, but yeah, it is all over the code base as is.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostScienceAreaGeometry.scala
sraaphorst: I had thought about using `Angle`, but this is how we did it for GMOS and F2, so I just followed suit.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostScienceAreaGeometry.scala
cquiroz: I guess I'm being influenced by the need to port these to newer versions of scala java :D Feel free to ignore

PR #1754, opened 2020-02-06
Fix pit launcher hardcoded test property

cquiroz:
The PIT Launcher hardcodes the test property which makes sense for development but it is bad when using it for distribution using ocs-osx

### DIFF
--- a/bundle/edu.gemini.pit.launcher/src/main/scala/edu/gemini/pit/launcher/PITLauncher.scala
+++ b/bundle/edu.gemini.pit.launcher/src/main/scala/edu/gemini/pit/launcher/PITLauncher.scala
@@ -17,10 +17,10 @@ object PITLauncher extends App {
   val locale = Locale.getDefault
   Locale.setDefault(Locale.ENGLISH)
 
-  // Check first on the env or else default to true
+  // Check first on the env or else default to false
   Option(System.getProperty("edu.gemini.pit.test")) match {
-    case None => System.setProperty("edu.gemini.pit.test", "true")
-    case _ => ()
+    case None => System.setProperty("edu.gemini.pit.test", "false")
+    case _    => ()
   }
   // Need to set this manually, as we are not inside OSGi
   val version = s"${Semester.current.year}.2.1"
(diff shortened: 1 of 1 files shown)


bundle/edu.gemini.pit.launcher/src/main/scala/edu/gemini/pit/launcher/PITLauncher.scala
```diff
@@ -17,8 +17,11 @@ object PITLauncher extends App {
   val locale = Locale.getDefault
   Locale.setDefault(Locale.ENGLISH)
 
-  // We normally want to be in test mode in development
-  System.setProperty("edu.gemini.pit.test", "true")
+  // Check first on the env or else default to true
+  Option(System.getProperty("edu.gemini.pit.test")) match {
+    case None => System.setProperty("edu.gemini.pit.test", "true")
+    case _ => ()
+  }
```
tpolecat: In `AppMode.scala` we're treating undefined as false, so these bits of code contradict each other. The default app distribution does not set the property at all, so with this change it will default to true.

sraaphorst: Yeah, this was acting wonky, and we were purposefully removing it, IIRC, to represent `false`.

cquiroz: I'm not sure I follow

cquiroz: You're right about `AppMode`, OTOH `AppMode` is only used from the Activator, maybe I should remove it altogether

cquiroz: `AppMode` is used also in the `ShellAdvisor`, so yes these changes need to be synchronized

cquiroz:
I flipped the logic on `PITLauncher` to be consistent with `AppMode`

bundle/edu.gemini.pit.launcher/src/main/scala/edu/gemini/pit/launcher/PITLauncher.scala
tpolecat: Confusion. Since it defaults to `false` anyway does this change actually accomplish anything?

swalker2m:
Superseded by #1760.

PR #1755, opened 2020-02-10
Switch actor to future in edu.gemini.pit.catalog

swalker2m:
This PR replaces the use of `actor` with `Future` in `edu.gemini.pit.catalog`. This change is required to enable executing PIT on Java 13, which in turn is required for OS X notarization. I've done this in the most straightforward way that I could think of because the PIT catalog code has been superseded even in the `ocs` codebase, which itself is nearing its end of life.

I've also removed the `&&` combinator since it is unused.

### DIFF
--- a/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/catalog/Catalog.scala
+++ b/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/catalog/Catalog.scala
@@ -1,18 +1,21 @@
 package edu.gemini.pit.catalog
 
-import scala.actors.Actor._
+import scala.concurrent.Future
+import scala.concurrent.ExecutionContext
+import scalaz._
+import Scalaz._
 
 object Catalog {
 
   // An empty catalog that always returns an error
   def empty(msg: String) = new Catalog {
-    def find(id: String)(callback: Callback) {
-      actor {
-        callback.safe(Error(new RuntimeException(msg)))
-      }
-    }
-    override def ||(outer: Catalog): Catalog = outer
-    override def &&(outer: Catalog): Catalog = outer
+
+    override def find(id: String)(implicit ex: ExecutionContext): Future[Result] =
+      Future.successful(Error(new RuntimeException(msg)))
+
+    override def ||(outer: Catalog): Catalog =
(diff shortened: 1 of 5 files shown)


bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/catalog/Catalog.scala
```diff
@@ -1,80 +1,45 @@
...
-      outer.find(id) { case (crB) => accum ! (('b, crB)) }
-      lazy val accum = actor {
-        await {
-          case ('a, a: Result) => await {
-            case ('b, b: Result) =>
-              callback.safe(merge(a, b))
-              exit()
-          }
-        }
+      Future.find(fs) {
```
tpolecat: :+1: I had not noticed this combinator before. I guess we do very little with `Future`.

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/catalog/VOTableCatalog.scala
```diff
@@ -15,28 +16,26 @@ trait VOTableCatalog extends Catalog {
...
-      } catch {
-        case e:IOException =>
-          Log.warning("%s: %s(%s)".format(host, e.getClass.getSimpleName, e.getMessage))
-          f(Offline)
-        case t:Throwable =>
-          Log.log(Level.WARNING, "Unexpected trouble looking up %s on %s.".format(id, host), t)
-          f(Error(t))
-      }
+  override def find(id: String)(implicit ex: ExecutionContext): Future[Result] =
+    Future {
```
tpolecat: In principle the body of this future should be inside a `blocking { ... }` call, which will let the EC know that it might want to perform the work on a blocking pool. This is implemented with a `ThreadLocal` so it's really sketchy, but at least for documentation I wanted to leave this comment in case we run into a starvation problem. I think it's probably ok as-is.

swalker2m: Thanks, it was easy to update regardless so I just did it.

PR #1758, opened 2020-02-12
Update apache http client in edu.gemini.catalog

swalker2m:
[Updating the GAIA url to use https](https://github.com/gemini-hlsw/ocs/pull/1756) seemed to work to fix NGS2 guide star searches, but the OT would still mysteriously fail to get a response from time to time.  Since the only change was to switch `http` to `https` in the GAIA server URL and since our version of `HttpClient` was probably released during the Clinton administration, I decided to update it and see if GAIA searches become more reliable. This required some code changes.

I'm a little hesitant about this update because we still use the old http client in other bundles, but hey OSGi excels at this sort of thing.  Of course it may complicate notarization updates since de-OSGi-ing is necessary.  Nevertheless we need a solution to the GAIA issue now because of an upcoming NGS2 run.


swalker2m:
> Of course it may complicate notarization updates

Actually I suppose not so much because classes in the two libraries don't share the same package names.

PR #1760, opened 2020-02-12
Make PIT use pitVersion in build.sbt to set version including test / production

sraaphorst:
This eliminates the unpredictable `edu.gemini.pit.test` property to determine whether the PIT being built is a test release or a production release, and instead uses the `pitVersion` in the main `build.sbt` as is done with the OT.


tpolecat:
Looks good to me! 👍

PR #1762, opened 2020-02-25
GHOSTSW-24: P2 error when targets / sky positions are outside of science FOV

sraaphorst:
This PR adds a `GhostRule` object that adds checks to make sure that targets and sky positions are within the science FOV. Note that the base can be overridden in GHOST, so this has to be taken into account.

I'm not quite sure I'm obtaining the base position properly since to get the proper base position for GHOST, we have to go through the `Asterism.basePosition` method and I'm not certain what I should be passing into this (`None` or `Instant.now()`?).

Here are some screenshots that demonstrate the P2 errors that come up in various situations.

The first is when we are in standard resolution mode, with SRIFU1 pointing at at a target at the base position, and a sky position out of range of the FOV. (I put a green square around the sky position to make it evident where it is, and you can see the P2 error in the usual place.)

sky position out of range

Here is an image where the base position is interpolated from the dual targets, but the dual targets are too far apart and thus outside the FOV:

Screen Shot 2020-02-25 at 4 51 56 PM

I also took @cquiroz's suggestion from a previous PR and defined the FOV for GHOST in terms `Angle` instead of `double` with `double` representing arcsecs.


bundle/edu.gemini.p2checker/src/main/scala/edu/gemini/p2checker/rules/ghost/GhostRule.scala
```diff
@@ -0,0 +1,46 @@
...
+  object CoordinatesOutOfFOVRule extends IRule {
+    override def check(elems: ObservationElements): IP2Problems = {
+      val problems = new P2Problems()
+
+      // COORDINATES, i.e. sky positions
+      for {
+        toc <- elems.getTargetObsComponentNode.asScalaOpt
+        ctx <- elems.getObsContext.asScalaOpt
+        env <- Option(ctx.getTargets)
+        base <- env.getAsterism.basePosition(None)
```
swalker2m: The time of reference is defined in the `ObsContext` as the scheduling block start.  You can directly get the base position coordinates at the correct time via `ctx.getBaseCoordinates` or you can fish out the `getSchedulingBlockStart`.  In reality for sidereal targets it won't matter because we aren't taking into account proper motion.  Non-sidereal GHOST observing is ??? but nevertheless we should probably use the scheduling block time as intended.

bundle/edu.gemini.p2checker/src/main/scala/edu/gemini/p2checker/rules/ghost/GhostRule.scala
```diff
@@ -0,0 +1,46 @@
...
+      } problems.addError(GhostRule.Prefix + "CoordinatesOutOfRange", String.format(GhostRule.CoordinatesOutOfRange, c.getName), toc)
+
+      // TARGETS
+      for {
+        toc <- elems.getTargetObsComponentNode.asScalaOpt
+        ctx <- elems.getObsContext.asScalaOpt
+        env <- Option(ctx.getTargets)
+        base <- env.getAsterism.basePosition(None) //ctx.getBaseCoordinates.asScalaOpt
+        t <- env.getAsterism.allTargets
+        c <- t.coords(None)
```
swalker2m: Here you'd need to use the `ctx.getSchedulingBlockStart` time.

bundle/edu.gemini.p2checker/src/main/scala/edu/gemini/p2checker/rules/ghost/GhostRule.scala
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  override def check(elems: ObservationElements): IP2Problems = {
+    val probs: P2Problems = new P2Problems()
+    probs.append(CoordinatesOutOfFOVRule.check(elems))
+    probs
+  }
+
+  val Prefix: String = "GhostRule_"
+  val CoordinatesOutOfRange: String = "The coordinates for %s are out of range of the base position."
```
swalker2m: Maybe "out of range *at* the base position"?  Or was this prescribed in a requirement somewhere?  It's the most trivial of quibbles but the base position isn't reaching for anything.

sraaphorst: It definitely wasn't stated specifically, and I think your wording is better, so we'll go with that.

PR #1784, opened 2020-05-08
REL-3813: Version change to get PIT to work.

sraaphorst:
When we updated the certificates, I re-versioned the PIT as 2020B.2.2.1, which broke the PIT. It should have been re-versioned as 2020B.2.1.2, which results in a match with the current version. (Note that re-versioning as 2020B.2.2.2 would have broken things as well since the previous version was broken.)

I spoke with Bryan and this is not a problem since 2020B.2.2.1 was never released, so the versioning remains consistent.


sraaphorst:
New PIT versions for 2020A (1.1.1) and 2020B (2.1.2) released.

PR #1785, opened 2020-05-08
REL-3812: 2020B GNIRS OT template updates

sraaphorst:
Changes to GNIRS template updates as specified in the JIRA task. The XML has not changed and the following has been added:

```
IF CROSS-DISPERSED == SXD OR CROSS-DISPERSED == LXD:
          Change the Position Angle to "Average Parallactic"
```

### DIFF
--- a/bundle/edu.gemini.phase2.skeleton.servlet/src/main/resources/edu/gemini/phase2/template/factory/xml/GNIRS_BP.txt
+++ b/bundle/edu.gemini.phase2.skeleton.servlet/src/main/resources/edu/gemini/phase2/template/factory/xml/GNIRS_BP.txt
@@ -138,3 +138,8 @@ IF CROSS-DISPERSED == SXD OR CROSS-DISPERSED == LXD:
           ADD NOTE "NOTE2: Bad pixels in XD mode" in target Scheduling Group
           ADD NOTE "NOTE3: Differential refraction and XD mode" in target Scheduling Group
 
+###### requests 2020.04.08 Siyi #####
+# change mean parallactic angle as the default set up for XD mode
+IF CROSS-DISPERSED == SXD OR CROSS-DISPERSED == LXD:
+          Change the Position Angle to "Average Parallactic"
+          


swalker2m:
Yes, this seems to be missing the actual update to the code.

swalker2m:
To be clear, there will be a corresponding change to `edu.gemini.phase2.template.factory.impl.gnirs.GnirsSpectroscopy`?

sraaphorst:
Ahhh. Thanks. I thought the txt was the only change that needed to be made. I'll probe further.

bundle/edu.gemini.phase2.skeleton.servlet/src/main/resources/edu/gemini/phase2/template/factory/xml/GNIRS_BP.xml
```diff
@@ -54,8 +55,15 @@ The semicircular portion of the field is below the rectangle. The template scien
           <param name="priority" value="MEDIUM"/>
           <param name="tooOverrideRapid" value="false"/>
           <param name="phase2Status" value="INACTIVE"/>
+          <paramset name="schedulingBlock">
+            <param name="start" value="1588354200000"/>
+            <paramset name="duration">
+              <param name="tag" value="unstated"/>
+            </paramset>
+          </paramset>
```
swalker2m, line 63: I'm concerned about these scheduling blocks being inserted into library observations.  These are supposed to work for any semester.  This may have been an oversight resulting from opening up the library program in the OT and saving it?

bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/template/factory/impl/gnirs/GnirsSpectroscopy.scala
```diff
@@ -259,4 +259,12 @@ case class GnirsSpectroscopy(blueprint:SpGnirsBlueprintSpectroscopy, exampleTarg
     addNote("NOTE2: Bad pixels in XD mode") in TargetGroup
     addNote("NOTE3: Differential refraction and XD mode") in TargetGroup
   }
+
+  // requests 2020.04.08 Siyi
+  // change mean parallactic angle as the default set up for XD mode
+  if (xd == SXD || xd == LXD) {
+    val included = curIncludes(TargetGroup).toSet
+    val all      = Set(5,6,7,8,9,10,11,12,13,14,22,23)
+    forObs((all & included).toSeq: _*)(setPositionAngleConstraint(PosAngleConstraint.PARALLACTIC_ANGLE))
```
swalker2m, line 268: This seems reasonable assuming the `all` set is correct.

sraaphorst: It was confirmed with Andy in the JIRA task.

bundle/edu.gemini.phase2.skeleton.servlet/src/main/resources/edu/gemini/phase2/template/factory/xml/GNIRS_BP.xml
sraaphorst: I brought up your comment with @andrewwstephens and he replied:

"I suggest we just try it and see if the instantiated observations look okay... I see the time is 2020-05-01 17:30:00 so we should notice a problem right away.  We will test it with 20B programs which should get default scheduling dates of ~October"

He suggests merging and seeing if an error crops up in the test release.

swalker2m: I think it would be a good idea to test it out with a proposal yourself anyway before merging if you haven’t yet.

PR #1789, opened 2020-05-14
Include the targets when reading the sequence for the seqexec

cquiroz:
GHOST needs target information sent with the seqexec.
This PR includes the targets and their type on the info the seqexec can get from queued sequences

### DIFF
--- a/bundle/edu.gemini.seqexec.odb/src/main/scala/edu/gemini/seqexec/odb/SeqExecService.scala
+++ b/bundle/edu.gemini.seqexec.odb/src/main/scala/edu/gemini/seqexec/odb/SeqExecService.scala
@@ -106,9 +106,8 @@ object SeqExecService {
     catchingAll {
       ObsContext.create(obs).asScala.headOption.map{te =>
         val guideTargets = te.getTargets().getGuideEnvironment().getTargets().asScala.map(_.getTarget).toList
-        val allTargets = te.getTargets.getTargets.asScala.map(_.getTarget).toList
         val userTargets = te.getTargets.getUserTargets.asScala.map(_.target.getTarget).toList
-        val scienceTargets = allTargets.diff(guideTargets).diff(userTargets)
+        val scienceTargets = te.getTargets.getAsterism.allTargets.toList
         scienceTargets.strengthL(TargetType.ScienceTarget) :::
           guideTargets.strengthL(TargetType.GuideTarget) :::
           userTargets.strengthL(TargetType.UserTarget)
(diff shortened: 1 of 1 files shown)


swalker2m:
I had thought that these were going to be sent as coordinates (i.e., with proper motion applied) as part of the GHOST configuration itself.  @sraaphorst could you comment?  You'll need sky positions, and to know which IFU to assign to which target, etc.  It's all in the GHOST asterisms.  I'm probably missing something.

cquiroz:
It was requested to get the positions and names of all targets to be displayed on the ui
Now that you mention it, are the positions corrected for PM by default?

swalker2m:
> Now that you mention it, are the positions corrected for PM by default?

No, usually all the PM information is passed to the TCC and it is handled there or in the TCS (I don't know the details at that level).  Since we needed it for GHOST and GHOST wasn't going to apply proper motion we are sending corrected coordinates.

Do you need to include in the display which target is assigned to which IFU and the coordinates of sky positions?  That stuff won't be in the information you're extracting.

bundle/edu.gemini.seqexec.odb/src/main/scala/edu/gemini/seqexec/odb/SeqExecService.scala
```diff
@@ -94,6 +102,19 @@ object SeqExecService {
       ConfigBridge.extractSequence(obs, null, IDENTITY_MAP, true)
     }
 
+  private def extractTargets(obs: ISPObservation): TrySeq[List[(TargetType, Target)]] =
+    catchingAll {
+      ObsContext.create(obs).asScala.headOption.map{te =>
+        val guideTargets = te.getTargets().getGuideEnvironment().getTargets().asScala.map(_.getTarget).toList
+        val allTargets = te.getTargets.getTargets.asScala.map(_.getTarget).toList
+        val userTargets = te.getTargets.getUserTargets.asScala.map(_.target.getTarget).toList
+        val scienceTargets = allTargets.diff(guideTargets).diff(userTargets)
```
swalker2m, line 111: Could also be `val scienceTargets = te.getAsterism.allTargets` and avoid the `diff`s.

cquiroz: Do user targets get included in the asterism?

cquiroz:
> > Now that you mention it, are the positions corrected for PM by default?
> 
> No, usually all the PM information is passed to the TCC and it is handled there or in the TCS (I don't know the details at that level). Since we needed it for GHOST and GHOST wasn't going to apply proper motion we are sending corrected coordinates.
Will all targets be corrected, meaning the user and science targets? What time is used for the correction?

> 
> Do you need to include in the display which target is assigned to which IFU and the coordinates of sky positions? That stuff won't be in the information you're extracting.
I wondered about that but it seems they only want the positions and names

bundle/edu.gemini.seqexec.odb/src/main/scala/edu/gemini/seqexec/odb/SeqExecService.scala
swalker2m: No, they are kept separate in the target environment.

swalker2m:
> Will all targets be corrected, meaning the user and science targets? What time is used for the correction?

The targets sent to GHOST in the instrument configuration are just those configured in the asterism.  The coordinates are relative the time when the sequence is created, right @sraaphorst ?

PR #1790, opened 2020-05-18
REL-3745: Change CWFS visualization in the Position Editor

sraaphorst:
Unsure what to do next. @swalker2m ?

### DIFF
--- a/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/Ngs2Strategy.scala
+++ b/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/Ngs2Strategy.scala
@@ -160,34 +160,29 @@ final case class Ngs2Strategy(
       .search(ctx, mt)(ec)
       .map { ts =>
 
-        // REL-3745: here instead of using the actual valid range from
-        // GemsCandidates we'll need to make a new special-purpose validator
-        // with an increased range. So a PatrolField.fromRadiusLimits(0, 1.3')
-        // that is repositioned at each science offset and then the intersection
-        // of these will be the valid range.  For now though, we'll just use the
-        // candidates within the (0, 1') range from GemsCandidates.
-        val pf                     = PatrolField.fromRadiusLimits(JAngle.ANGLE_0DEGREES, JAngle.arcmins(1.3))
+        // REL-3745: The code below handles its own range and magnitude filtering of CWFS candidates instead of
+        // calling GemsCandidates.groupAndValidate. We make a new special-purpose validator using a
+        // PatrolField.fromRadiusLimits(0, 1.3') so that candidates slightly outside of range will be shown alongside
+        // strictly valid candidates. Regardless, the candidate pool is ultimately filtered down to valid targets
+        // for selection. We do this in arcsecs so we can use the default value in CanopusWfs and simply grow the size.
+        val pf                     = PatrolField.fromRadiusLimits(JAngle.ANGLE_0DEGREES,
+                                           JAngle.arcsecs(15.5 + CanopusWfs.RADIUS_ARCSEC))
+        assert(JAngle.arcsecs(15.5 + CanopusWfs.RADIUS_ARCSEC).toArcmins.compareToAngle(JAngle.arcmins(1.3)) == 0)
         val magc                   = GemsVoTableCatalog.cwfsMagnitudeConstraints(ctx)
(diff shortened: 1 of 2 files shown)


swalker2m:
If we're trying to get a release out this week then I think it's too late to start on this task.  I expect it'll be tricky to get this right.

sraaphorst:
Ugh. I'm sorry. I was under the impression that this task was easy? I worked on it for several hours this weekend but then didn't know how to proceed to the next step in the description, i .e. "and then the intersection of these will be the valid range." I believe I have the intersection of the offsets correct, but I'm not sure exactly what we should be checking for the valid range.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/Ngs2Strategy.scala
```diff
@@ -162,12 +166,28 @@ final case class Ngs2Strategy(
...
+        val candidatesWithValidMag = ts.filter(t => magc.searchBands.extract(t).exists(magc.contains))
+
+        // A quick radius filter because the exact match done later is expensive
+        // and in a crowded field it takes a while to exact match them all.
+        val radiusFilter =
+        for {
+          rc <- RadiusLimitCalc.getAgsQueryRadiusLimits(Some(pf), ctx)
+          c0 <- ctx.getBaseCoordinates.asScalaOpt
+          c1 <- c0.toCoreCoordinates.asScalaOpt
+        } yield rc.targetsFilter(c1)
```
swalker2m, line 180: A trivial thing, but maybe indent the `for` expression a couple of spaces?

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/Ngs2Strategy.scala
```diff
@@ -162,12 +166,28 @@ final case class Ngs2Strategy(
         // that is repositioned at each science offset and then the intersection
         // of these will be the valid range.  For now though, we'll just use the
         // candidates within the (0, 1') range from GemsCandidates.
```
swalker2m, line 168: We'll need to update this comment a bit.  Maybe:

> REL-3745: The code below does its own range and magnitude filtering of CWFS candidates instead of just calling `GemsCandidates.groupAndValidate`.  We make a new special-purpose validator using a PatrolField.fromRadiusLimits(0, 1.3') so that candidates slightly outside of range will be shown alongside strictly valid candidates.  Regardless, the candidate pool is ultimately filtered down to valid targets for selection.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/Ngs2Strategy.scala
```diff
@@ -162,12 +166,28 @@ final case class Ngs2Strategy(
...
-
-        val cwfs =
-          GemsCandidates.groupAndValidate(ctx, posAngles(ctx), ts)
-            .foldLeft(List.empty[SiderealTarget]) { case (cwfs, gc) =>
-              gc.cwfsCandidates ++ cwfs
-            }
+        val pf                     = PatrolField.fromRadiusLimits(JAngle.ANGLE_0DEGREES, JAngle.arcmins(1.3))
```
swalker2m, line 169: It will never change, but it might be good to explicitly add 18 arcsec to `CanopusWfs.RADIUS_ARCSEC` (making it public first) for clarity instead of using 1.3' directly.

bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/Ngs2Strategy.scala
```diff
@@ -162,12 +166,28 @@ final case class Ngs2Strategy(
...
+
+        // candidates with valid magnitude that are also within the correct distance,
+        // ignoring position angle
+        val cwfs = radiusFilter.fold(candidatesWithValidMag)(candidatesWithValidMag.filter)
+
+//        val cwfs =
+//          GemsCandidates.groupAndValidate(ctx, posAngles(ctx), ts)
+//            .foldLeft(List.empty[SiderealTarget]) { case (cwfs, gc) =>
+//              gc.cwfsCandidates ++ cwfs
+//            }
```
swalker2m, line 190: We can remove this.

swalker2m (commented):
This looks correct to me.

swalker2m (approved):
Looks good to me.  :+1:

PR #1791, opened 2020-05-18
REL-3818: F2_BP.xml update

sraaphorst:
This is a simple update to F2_BP.xml.


sraaphorst:
There seems to be a failure regarding line 87 in F2BlueprintTest.scala: the F2 MOS notes do not include
```MOS Arcs and flats```
which causes the failure.

PR #1793, opened 2020-05-18
Include user targets in the sequence config

cquiroz:
We need to pass GHOST user targets to the seqexec and while a previous PR was included letting the seqexec query them but it maybe better to include them in the sequence:
Note, the targets are corrected by proper motion like for the science targets
image (5)

### DIFF
--- a/bundle/edu.gemini.horizons.api/src/test/scala/edu/gemini/horizons/server/backend/HS2Spec.scala
+++ b/bundle/edu.gemini.horizons.api/src/test/scala/edu/gemini/horizons/server/backend/HS2Spec.scala
@@ -29,15 +29,15 @@ object HS2Spec extends Specification with ScalaCheck {
       runSearch(Search.Comet("kjhdwekuq")) must_== \/-(Nil)
     }
 
-    // "handle multiple results" in {
-    //   runSearch(Search.Comet("hu")).map(_.take(5)) must_== \/-(List(
-    //     Row(HD.Comet("7P"), "Churyumov-Gerasimenko"),
-    //     Row(HD.Comet("06P"), "Schuster"),
-    //     Row(HD.Comet("30P"), "McNaught-Hughes"),
-    //     Row(HD.Comet("78P"), "Hug-Bell"),
-    //     Row(HD.Comet("/1880 Y1"), "Pechule")
-    //   ))
-    // }
+     "handle multiple results" in {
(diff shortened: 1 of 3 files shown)


cquiroz:
I added a fix for some Horizons failing tests

bundle/edu.gemini.horizons.api/src/test/scala/edu/gemini/horizons/server/backend/HS2Spec.scala
```diff
@@ -77,7 +77,7 @@ object HS2Spec extends Specification with ScalaCheck {
 
     "handle single result (Format 2) 29 Amphitrite" in {
       runSearch(Search.Asteroid("amphitrite")) must_== \/-(List(
-        Row(HD.AsteroidOldStyle(29), "Amphitrite")
+        Row(HD.AsteroidNewStyle("A854 EB"), "Amphitrite")
```
swalker2m, line 80: Thanks.  It's disturbing that there are no stable ids for nonsidereal targets.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostCB.scala
```diff
@@ -89,64 +89,70 @@ final class GhostCB(obsComp: ISPObsComponent) extends AbstractObsComponentCB(obs
...
-                guiding(Ghost.SRIFU2_GUIDING, t)
-              }))
-
-            /** HIGH RESOLUTION
+        foreach{t =>
+          t.getTargetEnvironment.getUserTargets.asScala.zipWithIndex.foreach { case (ut, i) =>
+            val (a, b, c, d, e)= Ghost.userTargetParams(i + 1)
+            coordParam(ut.target, Some(a), b, c, d, e)
+            config.putParameter(systemName, DefaultParameter.getInstance(s"userTarget${i+1}Type", ut.`type`))
+          }
```
swalker2m, line 97: This is really the new part correct?  The rest is just moved down from the previous version?

cquiroz: The diff got a bit strange because the old part was moved down. It seems fine to me as the OT shows the old configuration items as well as the new ones

PR #1794, opened 2020-05-27
Include exposure count values in the sequence for ghost

cquiroz:
GHOST is not exporting the count for the red and blue detector and we need it for the seqexec.
This PR fixes it


cquiroz:
Why does  it show in GMOS?

PR #1798, opened 2020-06-30
REL-3737 - Show the Altair field lens position in the ITC web output

andrewwstephens:
This is Alysha's PR with a new baseline.txt.


cquiroz:
Going back to the question of the baseline. It seems odd that a PR that only changes the UI (html changes) would require a new baseline, or is the baseline created out of html?

andrewwstephens:
I think the baseline captures the data + output HTML.

PR #1801, opened 2020-08-03
REL-3828 - support GT proposals in 21A

tpolecat:
This PR does a few things, all required to support GT proposals.

- [x] Adds a new `GUARANTEED_TIME` member to the `SpecialProposalType` enumeration, which shows up as expected in the PIT.

image

- [x] Adds a mapping from this proposal type to `TimeAcctCategory.GT` in the skeleton service.
- [x] Updates the p1-minor configuration to recognize this new proposal destination.

TODO prior to merging:

- [x] Verify that PDF generation works. - *It does … all special proposals just say "Special" which is kind of eh but it's been that way for a while so I'm not going to change it.*
- [x] Test ODB import. - *Hacked up a proposal XML and it imported the GT time correctly.*

TODO prior to 21A CFP:

- [ ] Add a new back-end server.
- [ ] Redeploy the P1 monitor
- [ ] Create a `pitgt@gemini.edu` email exploder.
- [ ] End to end test.

TODO prior to 21A ITAC:

- [ ] Update ITAC software to recognize `GT` "partner".


app/p1-monitor/build.sbt
```diff
@@ -77,24 +77,24 @@ def staging(ov: Version, pv: Version) = AppConfig(
   distribution = List(Linux32, Linux64)
 ) extending List(common(ov, pv))
 
-// 2020A
-def v2020A(ov: Version, pv: Version) = AppConfig(
-  id = "2020A",
+// 2021A
```
cquiroz, line 80: Often we keep two semester running due to de differences on the dates when they open/close

bundle/edu.gemini.p1monitor/src/main/resources/edu/gemini/p1monitor/config/conf.production-2021A.xml
```diff
@@ -0,0 +1,175 @@
...
+        </translation>
+        <translation>
+            <from>Poor Weather</from>
+            <to>PW</to>
+        </translation>
+        <translation>
+            <from>Directors Time</from>
+            <to>DT</to>
+        </translation>
+        <translation>
```
cquiroz, line 121: We may need a translation from `Guaranteed Time` to `GT` this is used to name the saved proposal files

app/p1-monitor/build.sbt
tpolecat: So it should be 20B and 21A?

cquiroz: Yes, though we'd only need 20B if we ever need to rebuild the previous p1 monitor, which has happened a few times

tpolecat:
Ok @cquiroz I re-did the p1 monitor changes, thanks for the help. Can you have another look?

PR #1804, opened 2020-08-06
REL-3842 - update PIT icons

tpolecat:
I don't really know how to check that they work. They seem ok. They look like this in the Finder. It draws a border around `.ico` files for some reason, that's not part of the image.

image


cquiroz:
I think there is also a png used on the linux version as well, but it is not under app but as a resource in the pit module

sraaphorst:
Yes, I believe that Carlos is right... I'll try to find the spot.

sraaphorst:
It is here:

./bundle/edu.gemini.pit/src/main/resources/edu/gemini/pit/ui/util/icons/pit_icon.png

sraaphorst:
It's 512x512 px, which I think is the size of the PNG that Bryan usually distributes?

tpolecat:
Ah ok thanks.

tpolecat:
Ok done

PR #1811, opened 2020-10-02
REL-3854: Update to run standalone AGS service app

swalker2m:
This allows the `ags` app to work with both http and https at the port used by the PIT.  It required corresponding updates to the svn `ocs-credentials` project.  Once merged, the production `gn` and `gs` configuration distributions need to be tested on the build machine.

### DIFF
--- a/app/ags/build.sbt
+++ b/app/ags/build.sbt
@@ -3,12 +3,16 @@ import OcsKeys._
 import edu.gemini.osgi.tools.Version
 import edu.gemini.osgi.tools.app.{ Configuration => AppConfig, _ }
 import edu.gemini.osgi.tools.app.Configuration.Distribution.{ Test => TestDistro, _ }
-import OcsCredentials.Spdb._
+import OcsCredentials.Ags._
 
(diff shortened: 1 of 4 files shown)


app/ags/build.sbt
```diff
@@ -46,8 +50,10 @@ def common(version: Version) = AppConfig(
     "org.osgi.framework.bootdelegation"                -> "*",
     "org.osgi.framework.startlevel.beginning"          -> "99",
     "org.osgi.framework.storage.clean"                 -> "onFirstInit",
-    "org.osgi.service.http.port"                       -> "9123",
-    "org.osgi.service.http.secure.enabled"             -> "false"
+    "org.osgi.service.http.port"                       -> "8442",
```
cquiroz, line 53: I guess this is the core change, is it?

swalker2m: Along with adding the `pax-web` bundles, linking in the `gemKeystore`, and updating the ocs credentials file.

tpolecat (approved):
grim :+1:

PR #1812, opened 2020-10-06
REL-3857: fix the calculated number of exposures when supplied a s/n and source fraction <1

andrewwstephens:
Another contender for the tiniest pr award.


cquiroz (approved):
LGTM

PR #1813, opened 2020-10-09
REL-3849: add new GMOS grating

swalker2m:


tpolecat (approved):
We should add this to the lucuma enum as well (even though it's unused right now) so we don't forget.

PR #1817, opened 2020-10-16
REL-3860: Iterable, omnipresent DisperserOrder

swalker2m:
This PR makes `DisperserOrder` an iterable property and adds it to the configuration regardless of the observing mode.  In particular it doesn't make sense for imaging steps but will be present in the configuration regardless.  This was done to accommodate observations with mixed imaging and spectroscopy steps that heretofore do not allow setting the disperser order (and don't add it to the sequence) when the first step is imaging.

If there is anything outside of `ocs` proper that infers spectroscopy from the presence of a disperser order property in the sequence, it will break with this change.  In theory, scripts should already be checking the disperser itself to make this determination.  The seqexec itself ignores disperser order and wavelength for imaging steps so it should be okay with these changes.


swalker2m:
@jluhrs This implements the changes we discussed.

jluhrs (approved):
LGTM

PR #1818, opened 2020-10-19
Feature/REL-3775

leashma:
The if/else statement from the old code did not properly print to the html output the desired AO parameter status. I defaulted to print out the guide star separation and field lens positions, since those are more frequently used, and then created the "if" statement in case the user chooses the LGS.


leashma:
Can anyone approve this? I am waiting to submit one more software change.

cquiroz (approved):
LGTM

PR #1819, opened 2020-11-03
REL-3865: Updated GRACES blueprint.

sraaphorst:
No changes need to be made to the import process, it seems.


cquiroz (approved):
LGTM

swalker2m:
* The change is just that notes get added, correct?  No other logic changes?
* I think we should strip the scheduling blocks.

sraaphorst:
@swalker2m According to the JIRA task, that appears to be the case. I'm not sure offhand how to strip the scheduling blocks but I'll take a look. (I'm also not sure why the checks failed here when all test cases passed locally. I'll look into it.)

swalker2m:
> I think we should strip the scheduling blocks.

Ideally when we read the template database we could go through each observation and set the scheduling block to `None`. Alternatively we could just delete it from the XML for this case.

PR #1821, opened 2020-11-11
REL-3869: show visitor instrument name

swalker2m:
Shows the visitor instrument name instead of "Visitor Instrument" in the QPT and published plan.  Sadly all visitor instruments are otherwise treated the same so ICTD availability and filtering in the "Instruments" tab may be confusing.


cquiroz (approved):
LGTM

PR #1822, opened 2020-11-16
REL-3874 - Give an ITC error if the spatial binning is not 1 with GMOS IFU

andrewwstephens:
We do not support binning when using the GMOS IFU, so this gives an error if that happens.
I also updated the inputs to the GMOS baseline tests and expanded the explanation of binning on the web form.


cquiroz (approved):
LGTM

PR #1823, opened 2020-11-17
REL-3867 - Include the background from all IFU elements when summing fibers in the ITC

andrewwstephens:
This sets the ITC aperture slit length based on the number of IFU fibers being summed.


cquiroz (approved):
LGTM

PR #1826, opened 2020-12-09
update telnet ip

swalker2m:
Updates the `osgi.shell.telnet.ip` property for the base-facility ODB.  This had been done in the production configuration for 2020B release, but not updated in GitHub.

> **Be aware** The remote shell does not listen on all interfaces by default; it only listens on the localhost. That is, by default the remote shell is only accessible from the host on which the remote shell is running. To access the system from another host, you have to configure the IP address of the interface to which the remote shell should be attached.


cquiroz (approved):
LGTM

PR #1830, opened 2021-01-19
Fix ssl keystore for PIT submissions

cquiroz:
Fixes a submission error due to using an obsolete keystore for the submission post


bundle/edu.gemini.model.p1.submit/src/main/scala/edu/gemini/model/p1/submit/FutureSubmission.scala
```diff
@@ -67,21 +67,7 @@ case class FutureSubmission(dest: SubmitDestination, url: String, proposal: Prop
...
-    }
   }
-
-}
-
-object FutureSubmission {
-
-  val sslFactory = {
-    // Load keystore from the bundle
-    val in = this.getClass.getResourceAsStream("/phase1.jks")
```
swalker2m, line 79: Do you know the history of this?  Just wondering why we did it this way before.

swalker2m (approved):
Looks good to me. :+1:

bundle/edu.gemini.model.p1.submit/src/main/scala/edu/gemini/model/p1/submit/FutureSubmission.scala
cquiroz: I did it in 2015 when we switched to SSL. I have a vague recollection that it was using an external certificate. Let's encrypt was probably already in use then before being used for other gemini servers

PR #1831, opened 2021-01-20
Use data file for dataset

cquiroz:
Replace user visible `dataset` for `data file` in the PIT and pdfs


bundle/edu.gemini.gsa.query/src/main/scala/edu/gemini/gsa/query/JsonCodecs.scala
```diff
@@ -39,7 +39,7 @@ object JsonCodecs {
     EncodeJson((label: DatasetLabel) => jString(label.toString))
 
   def invalidDatasetLabel(s: String): String =
-    s"Could not parse `$s` as a dataset label"
+    s"Could not parse `$s` as a data file label"
```
cquiroz, line 42: This is not strictly part of the pit but these errors could show up on the PIT as we use the same codecs

swalker2m (approved):
:+1:

PR #1832, opened 2021-01-20
Prepare the PIT for 2021B

cquiroz:
Signed-off-by: Carlos Quiroz <cquiroz@gemini.edu>


bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Semester.scala
```diff
@@ -49,7 +49,7 @@ object Semester {
 
 }
 
-case class Semester(year: Int, half: SemesterOption) {
+sealed case class Semester(year: Int, half: SemesterOption) {
```
swalker2m, line 52: Did you mean `final`?

cquiroz: I guess it should be both. we have defined 2 instances A/B

cquiroz: I got confused with `SemesterOption`, yes it should be `final`

swalker2m (approved):
:+1:

PR #1834, opened 2021-01-25
Capture HorizonsDesignation in PIT

tpolecat:
This change:

- Adds `horizonsDesignation` as an optional attribute for `NonSiderealTarget` in the Phase 1 schema. 
- Captures this value after a successful Horizons lookup in the PIT target editor.
- Preserves the value through skeleton generation.

Note that this change does _not_ attempt to parse the designation from imported ephemeris files. I think this is probably impossible in general because you need to know what kind of search you were doing in order to interpret the header.


cquiroz (approved):
LGTM, maybe worth checking that the PIT can open an older proposal (without the new field) just fine

swalker2m (approved):
Looks good.

tpolecat:
@cquiroz I verified that it works with an old proposal that includes nonsidereal targets.

PR #1835, opened 2021-01-26
REL-3790

cquiroz:
This PR contains several updates related to institution changes related to NOIRLab


bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/transform/UpConverter.scala
```diff
@@ -162,7 +159,64 @@ case class LastStepConverter(semester: Semester) extends SemesterConverter {
  * This converter supports migrating to 2021A.
  */
 case object SemesterConverter2021ATo2021B extends SemesterConverter {
-  override val transformers: List[TransformFunction] = Nil
+  val NOAOTransformMessage = "Proposal assigned to NSF's NOIRLab"
+  val GNTransformMessage = "Proposal assigned to Gemini Observatory/NSF's NOIRLab(North)"
+  val GSTransformMessage = "Proposal assigned to Gemini Observatory/NSF's NOIRLab(South)"
+  val TololoTransformMessage = "Proposal assigned to Gemini Observatory/NSF's NOIRLab(South)"
```
swalker2m, line 165: I'm not sure what this is about, but just wanted to check whether this should say something about Tololo instead of Gemini.

swalker2m (approved):
Looks good. :+1:

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/transform/UpConverter.scala
cquiroz: Oh good catch

PR #1838, opened 2021-01-27
Fix a bug on the NOAO template and setup p1 monitor

cquiroz:
Sets p1 monitor for 2021B and assigns FT to use the noao template

It also fixes a bug on the template when using FT


swalker2m (approved):
Looks good. :+1:

PR #1840, opened 2021-02-04
REL-3899: update GMOS TCC config

swalker2m:
Updates the GMOS TCC configuration names to be uniform (also appending OI and P1 as appropriate).


cquiroz (approved):
LGTM

PR #1841, opened 2021-02-12
retain original query as object name

tpolecat:
Followup for REL-3792.

The PIT currently uses the catalog's object name when creating a new target following a Horizons lookup. This is a feature in cases where the catalog name is more precise, but an annoyance in cases where the catalog name is ambiguous.

So this PR changes the behavior to use the query string (i.e., whatever the user typed in) as the target name.


cquiroz (approved):
LGTM

PR #1843, opened 2021-02-17
[REL-3792] (3) add query string to target editor

tpolecat:
In response to Bryan's concern about target names being ambiguous, this PR

- adds the query string to the editor popup, which brings the behavior mostly into parity with what the OT does.

image

- adds a Query column to the target view.

image

- It also adds the query string to the PDF, when present. Implemented for Gemini ...

image

- and NOAO

image

### DIFF
--- a/bundle/edu.gemini.model.p1/src/main/xsd/Target.xsd
+++ b/bundle/edu.gemini.model.p1/src/main/xsd/Target.xsd
@@ -61,6 +61,7 @@
                 </xsd:sequence>
                 <xsd:attribute name="epoch" type="CoordinatesEpoch" default="J2000"/>
                 <xsd:attribute name="horizonsDesignation" type="xsd:token"/>
+                <xsd:attribute name="horizonsQuery" type="xsd:token"/>
             </xsd:extension>
         </xsd:complexContent>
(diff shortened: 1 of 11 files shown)


bundle/edu.gemini.model.p1/src/main/xsd/Target.xsd
```diff
@@ -61,6 +61,7 @@
                 </xsd:sequence>
                 <xsd:attribute name="epoch" type="CoordinatesEpoch" default="J2000"/>
                 <xsd:attribute name="horizonsDesignation" type="xsd:token"/>
+                <xsd:attribute name="horizonsQuery" type="xsd:token"/>
```
tpolecat, line 64: `horizonsQuery` is optional so it shouldn't require an upconverter. It's derivable from `horizonsDesignation` (and vice-versa) but these conversions are unavailable to the PDF generator (the only place it's used), so instead we have a redundant attribute.

cquiroz (approved):
Looks good to me. did you check with existing proposals without the new element. I just wonder if the UI will showe e.g. `null` in some places

tpolecat:
@cquiroz old proposal looks ok

image

PR #1846, opened 2021-02-25
REL-3861 Fixes a bug when saving a GT proposal

cquiroz:
Fixes a bug when saving a GT proposal without submission responses. It also improves the wording of the ProblemRobot in TAC mode

### DIFF
--- a/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/view/tac/TacView.scala
+++ b/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/view/tac/TacView.scala
@@ -36,7 +36,7 @@ object TacView {
       // no partners and the tab's controls will be disabled. Elsewhere we don't bother with the
       // `gt.sub.specialType` check.
       case gt:SpecialProposalClass if gt.sub.specialType === SpecialProposalType.GUARANTEED_TIME =>
-        List(GuaranteedTimePartner -> gt.sub.response)
+        gt.sub.response.map(r => List(GuaranteedTimePartner -> Some(r))).getOrElse(Nil)
 
       case _                           => Nil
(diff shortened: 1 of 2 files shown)


bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/view/tac/TacView.scala
```diff
@@ -36,7 +36,7 @@ object TacView {
       // no partners and the tab's controls will be disabled. Elsewhere we don't bother with the
       // `gt.sub.specialType` check.
       case gt:SpecialProposalClass if gt.sub.specialType === SpecialProposalType.GUARANTEED_TIME =>
-        List(GuaranteedTimePartner -> gt.sub.response)
+        gt.sub.response.map(r => List(GuaranteedTimePartner -> Some(r))).getOrElse(Nil)
```
swalker2m, line 39: `gt.sub.response.toList.tupleLeft(GuaranteedTimePartner)` might work too

swalker2m (approved):
Looks good to me. :+1: I'm not sure about treating LP and GT as partners but that very much predates this PR.

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/view/tac/TacView.scala
swalker2m: Never mind, I guess the response has to be wrapped in an option inside the list.

cquiroz:
Yeah it maybe nicer that GT and LP are regular partners but that’s a bigger change

tpolecat:
> I'm not sure about treating LP and GT as partners but that very much predates this PR.

ITAC treats them as time buckets that are equivalent to partners, which is probably what drove the approach here. The Phase 2 model doesn't really care as far as I can tell.

PR #1848, opened 2021-03-10
Change the GMOS ITC summed IFU signal plot to be for a single element.

andrewwstephens:
The main change here is to pass the "one-pixel" source fraction to SlitThroughput on L250 of GmosRecipe.java so that the "Signal and SQRT(Background)" plot is for a single pixel.


swalker2m (approved):
I'm not familiar with the ITC but this seems fine.  :+1:  Should I merge it before the first test release?

andrewwstephens:
Thanks for checking.  Yes, please merge so that it can be tested.

PR #1849, opened 2021-03-18
REL-3809: GNIRS IFU options

swalker2m:
Adds GNIRS IFU options according to spec.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gnirs/GNIRSParams.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gnirs/GNIRSParams.java
@@ -156,9 +156,11 @@ public class GNIRSParams {
         SW_8("3.0 arcsec", 3.0, "3.0") {
             @Override public boolean isObsolete() { return true; }
         },
-        IFU("IFU", 4.8, "IFU") {
+        IFU("IFU", 3.15, "IFU") {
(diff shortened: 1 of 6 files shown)


bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gnirs/GNIRSParams.java
```diff
@@ -156,9 +156,11 @@ public double getValue() {
         SW_8("3.0 arcsec", 3.0, "3.0") {
             @Override public boolean isObsolete() { return true; }
         },
-        IFU("IFU", 4.8, "IFU") {
+        IFU("IFU", 3.15, "IFU") {
             @Override public boolean isObsolete() { return true; }
         },
+        LR_IFU("LR-IFU", 3.15, "LR-IFU"),
```
cquiroz, line 162: Will this need a change on the PIT?

cquiroz (approved):
LGTM

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gnirs/GNIRSParams.java
swalker2m: Eventually, and in templates etc.  I think for now though they will just be tested.  At any rate, PIT changes weren't mentioned in the task.

PR #1852, opened 2021-03-23
Horizon fix without updating the model version

cquiroz:
Fixes the horizons lookup without changing the model version number
The title on the UI was hardcoded to include the minor version number 🙈


swalker2m (approved):
:+1:

PR #1853, opened 2021-03-23
Horizon fix without updating the model version

cquiroz:
Fix for the horizons lookup without a model change
The version number on the UI was hardcoded for better usability


swalker2m (approved):
Looks good.

PR #1854, opened 2021-04-05
REL-3872: Don't allow negative parallax

swalker2m:
Just put in on my (technical debt) tab.


cquiroz:
Do we need to filter negative parallaxes from the catalogues?

cquiroz (approved):
LGTM

swalker2m:
> Do we need to filter negative parallaxes from the catalogues?

I think you (?) did that in `VoTableParser`:

```scala
    def parsePlx(plx: Option[String]): CatalogProblem \/ Option[Parallax] =
       parseDoubleVal(plx, VoTableParser.UCD_PLX)(d => Parallax(0.0 max d).right)
```

I just updated it a bit because you can no longer create a `Parallax` with a negative value.

```scala
    def parsePlx(plx: Option[String]): CatalogProblem \/ Option[Parallax] =
       parseDoubleVal(plx, VoTableParser.UCD_PLX)(d => Parallax.fromMas(d).orZero.right)
```

PR #1855, opened 2021-04-07
REL-3916: RolloverPeriod

swalker2m:
The period for which a rollover program is considered valid changed in 2018A (though reports and QPT were never updated to reflect it).  The definition of this period is scattered all over the place so I attempted to define it in a single class, `RolloverPeriod`.


cquiroz (approved):
LGTM

PR #1856, opened 2021-04-07
REL-934: Remove obsolete instrument options

swalker2m:
Removes obsolete instrument items from the QPT filters.  The specific request was to remove the GNIRS 3.0 slit (which is obsolete).  It seemed like all obsolete items should be removed though.  Presumably this means that any observations configured to use an obsolete item will not appear in the candidate set.


swalker2m:
@tpolecat would you take a look at this?  It seems to work fine but I may easily be overlooking something.

PR #1859, opened 2021-04-09
REL-3913: strip surrounding text from external reports

swalker2m:
The public reports generated by the reports server match the look of Gemini's public science pages from 1924, but stand out today.  They also contain out of date information in the text surrounding the tables.  To address these concerns, it was requested that we produce the HTML tables themselves but none of the surrounding markup.  Presumably these will be embedded in something managed by our Drupal site.

Because the velocity templates only differ in the surrounding text (except for section headers), I was able to collapse them into a single template that is shared across Q, DD, FT, LP, and bad weather.  The section header to use is supplied as a function when creating the reports.


swalker2m:
I suggested that it might make more sense to just produce the data itself in the form of, say, JSON but this was preferred.

cquiroz:
Are you suggesting to separate content form presentation?

PR #1860, opened 2021-04-09
REL-3863: controlled pos angle update

swalker2m:
Going well beyond just touching the third rail, this PR would update `BagsManager` to:

* make its update to the target environment and position angle consistent by holding a write lock
* only performing the update if the observation is in the expected state as it was when the AGS calculation was kicked off

This is to address the complaint about the position angle in the OT reverting to previous values when typing quickly.  I could not reproduce this issue but these changes _may_ help and _seem_ like the right thing to do regardless. 

There are also a handful of inconsequential updates made while browsing the code as well.  I'll mark the important bits.

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
@@ -164,7 +164,13 @@ object BagsState {
         if (r.posAngle === ctx.getPositionAngle) hash
         else hashObs(ctx.withPositionAngle(r.posAngle))
       }
-      (IdleState(k, Some(h)), BagsManager.applyAction(k, Some(hash), results))
+
+      (FinalizeUpdateState(k, h),
+        for {
+          _ <- BagsManager.applyAction(k, Some(hash), results)
+          _ <- BagsManager.wakeUpAction(k, 0)
+        } yield ()
+      )
     }
 
     // Failed AGS lookup.  Move to FailedState for a while and ask the timer to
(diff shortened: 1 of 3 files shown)


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -162,7 +164,7 @@ object BagsState {
         if (r.posAngle === ctx.getPositionAngle) hash
         else hashObs(ctx.withPositionAngle(r.posAngle))
       }
-      (IdleState(k, Some(h)), BagsManager.applyAction(k, results))
+      (IdleState(k, Some(h)), BagsManager.applyAction(k, Some(hash), results))
```
swalker2m, line 167: `hash` is the AGS computation made from the relevant bits of an observation such as it was before any updates are applied.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -268,18 +272,23 @@ object BagsManager {
...
-  val blockingExecutionContext = ExecutionContext.fromExecutor(blockingWorker)
+  private val blockingWorker: ScheduledThreadPoolExecutor =
+    new ScheduledThreadPoolExecutor(16, new ThreadFactory() {
+      override def newThread(r: Runnable): Thread = {
+        val t = new Thread(r, "BagsManager - Blocking Query Worker")
+        t.setPriority(Thread.NORM_PRIORITY - 2)
+        t.setDaemon(true)
+        t
+      }
+    })
```
swalker2m, line 288: This is just reformatting.  Sorry.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -526,6 +548,32 @@ object BagsManager {
...
+      }
+
+      // Commit all the changes inside a write lock, but first make sure that
+      // the underlying observation has not changed from what we expected.  If
+      // it has changed, then a new BAGS calculation will be triggered anyway.
+
+      ctx.obsShell.foreach { obs =>
+        obs.getProgramWriteLock()
+        try {
+          if (expectedHash.forall(hashObs(obs).contains)) {
```
swalker2m, line 570: Only apply the results if a newly computed AGS hash matches what it was before (or if we don't supply an expected hash at all).

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
```diff
@@ -526,6 +548,32 @@ object BagsManager {
...
+            commitChanges(ctx, oldEnv, newEnv)
+          }
+        } finally {
+          obs.returnProgramWriteLock()
```
swalker2m, line 574: Make the changes with a write lock and return the lock when finished.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/ags/BagsManager.scala
swalker2m: When the hash no longer matches, it means that the observation has been updated in some way that will impact the AGS calculation.  In turn that means that a new AGS calculation will be kicked off anyway.

cquiroz (approved):
LGTM

PR #1863, opened 2021-04-15
REL-3924: Add GAIA @ Gemini

swalker2m:
Adds an additional catalog option for GAIA at Gemini.  I don't yet have the correct query URL for external users, but it can be tested by internal users.  A followup-up PR will be needed to correct the address.  It does show up in all the right places in the OT though.

### DIFF
--- a/bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableClient.scala
+++ b/bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableClient.scala
@@ -233,7 +233,7 @@ sealed trait GaiaBackend extends CachedBackend with RemoteCallBackend {
     val fields = gaia.allFields.map(_.id).mkString(",")
 
     f"""|SELECT TOP $MaxResultCount $fields
-        |      FROM ${gaia.tableName}
+        |      FROM gaiadr2.gaia_source
         |     WHERE CONTAINS(POINT('ICRS',${gaia.raField.id},${gaia.decField.id}),CIRCLE('ICRS', ${cs.base.ra.toDegrees}%9.8f, ${cs.base.dec.toDegrees}%9.8f, ${cs.radiusConstraint.maxLimit.toDegrees}%9.8f))=1
         |       AND (${gaia.plxField.id} > 0)
         |       AND (${gaia.gMagField.id} BETWEEN $BrightLimit AND $FaintLimit)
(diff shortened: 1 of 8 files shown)


bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableClient.scala
```diff
@@ -260,11 +259,34 @@ case object GaiaBackend extends CachedBackend with RemoteCallBackend {
...
+case object GaiaGeminiBackend extends GaiaBackend {
+
+  // For Java
+  val instance: GaiaBackend = this
+
+  override def gaia: CatalogAdapter.GaiaAdapter =
+    CatalogAdapter.GaiaGemini
+
+  override val catalogUrls: NonEmptyList[URL] =
+    NonEmptyList(new URL("https://gncatalog.gemini.edu/gaia"), new URL("https://gscatalog.gemini.edu/gaia"))
```
swalker2m, line 288: This, at least, will need to be updated.

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableClient.scala
```diff
@@ -207,9 +207,9 @@ case object ConeSearchBackend extends CachedBackend with RemoteCallBackend {
   override def queryUrl(e: SearchKey): String = s"${e.url}/cgi-bin/conesearch.py"
 }
 
-case object GaiaBackend extends CachedBackend with RemoteCallBackend {
-  // For Java
-  val instance = this
+sealed trait GaiaBackend extends CachedBackend with RemoteCallBackend {
+
+  def gaia: CatalogAdapter.GaiaAdapter
```
cquiroz, line 212: I assume this means fields are different between ESA/Gemini?

swalker2m: It opens up the possibility of handling that situation, but it's not expected to be the case.  Instead there are two GAIA backends because of the abstract `catalog: CatalogName` in `CatalogAdapter` which must be defined differently for the two GAIA catalog servers.

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/package.scala
```diff
@@ -0,0 +1,22 @@
...
+package object votable {
+
+  private[votable] implicit class StringOps(s: String) {
+
+    /**
+     * A predicate that determines whether a String is both non-empty and not
+     * "NaN".
+     */
+    def nonEmptyNonNan: Boolean =
+      s.nonEmpty && !"NaN".equals(s)
```
swalker2m, line 18: Hacking around the Gemini server's propensity for adding `NaN` entries to the data.

cquiroz: Neat I love the votable standard

cquiroz (approved):
LGTM

PR #1865, opened 2021-04-26
REL-3927: new GMOS-S filters

swalker2m:
Adds new GMOS-S filters, making them available in the OT and QPT.


cquiroz (approved):

cquiroz:
I suppose we need tasks for these for the PiT and seqexec

swalker2m:
> I suppose we need tasks for these for the PiT and seqexec

REL-3929 for the seqexec.  I'm not sure whether they are available for general use yet though (and so maybe don't belong in the PIT).  Somehow though a 2021B program relies upon them.

PR #1869, opened 2021-05-24
REL-3913: remove the 'r' from rollover semester

swalker2m:
Rollover is now called "persistent" in official documentation so 'r' doesn't correspond.  The request then was to remove 'r' from the rollover semester in reports and leave it at, for example, `[21B]` so that it survives the next name change.


cquiroz (approved):
LGTM

PR #1870, opened 2021-05-24
REL-3809: add IFU deckers

swalker2m:
Adds a couple of new decker options for GNIRS.  By default, decker isn't added to the sequence but is available if the user wants to override the default value.


cquiroz (approved):
LGTM

PR #1873, opened 2021-05-26
REL-3924: switch all the catalogs to VO Table 1.3

swalker2m:
Along with GAIA all the catalogs have been updated in the catalog server.  To make it work we need to match the VO Table version being served or else the output doesn't validate.


cquiroz (approved):
LGTM

PR #1874, opened 2021-06-03
REL-3924: Prepare for 2021B

swalker2m:
We're leaving GAIA @ Gemini and the new catalog server for December.  To facilitate the upgrade at a later date this PR:

* Accepts either VO table version 1.2 or 1.3
* Removes GAIA @ Gemini from the manual catalog search list only

### DIFF
--- a/bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableParser.scala
+++ b/bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableParser.scala
@@ -4,14 +4,19 @@ import java.io.{ByteArrayInputStream, InputStream}
 import edu.gemini.catalog.api.CatalogName
 import edu.gemini.spModel.core._
 
+import java.util.logging.Logger
+
 import scala.collection.immutable
 import scala.io.Source
+import scala.util.matching.Regex
 import scala.xml.XML
 import scala.xml.Node
 import scalaz._
 import Scalaz._
 
 object VoTableParser extends VoTableParser {
+  private val LOG: Logger = Logger.getLogger(this.getClass.getName)
+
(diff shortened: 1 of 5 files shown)


bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableParser.scala
```diff
@@ -32,12 +37,48 @@ object VoTableParser extends VoTableParser {
...
 
-  private def validate(catalogName: CatalogName, xmlText: String): CatalogProblem \/ Unit =
+  private val VoTableHeader: Regex =
+    """<VOTABLE([^>]*)>""".r.unanchored
+
+  private val MinVersion: VersionToken = VersionToken.unsafeFromIntegers(1, 2)
+  private val MaxVersion: VersionToken = VersionToken.unsafeFromIntegers(1, 3)
+
+  private val VersionRegex: Regex =
+    """version="(\d+\.\d+)"""".r.unanchored
```
swalker2m, line 47: Regex parsing is kind of low rent but I didn't want to dig up an old OSGi-safe version of Atto or something.

tpolecat (approved):
:+1: assuming tests pass

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableParser.scala
tpolecat: I want an editor plugin that lets me click on a regex and it takes me to regexper, like [this](https://regexper.com/#version%3D%22%28%5Cd%2B%5C.%5Cd%2B%29%22).

swalker2m: Wow, that's excellent.

PR #1875, opened 2021-07-05
REL-3949: NPE in obslog GUI

swalker2m:
I'm not sure why this has never been an issue before, or has never been reported before.  Nevertheless, working around it seems straightforward.  See suggested changes.


cquiroz (approved):
Great find

PR #1876, opened 2021-07-05
REL-3955: initialization issue in DHS_WRITE_TIME

swalker2m:
Changes a constant to a method call to solve an initialization issue.  In particular, a `static` block was not being initialized before referenced in `SPInstObsComp`.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/obscomp/SPInstObsComp.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/obscomp/SPInstObsComp.java
@@ -436,7 +436,8 @@ public abstract class SPInstObsComp extends AbstractDataObject {
         );
     }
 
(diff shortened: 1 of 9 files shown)


bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/obscomp/SPInstObsComp.java
```diff
@@ -436,7 +436,8 @@ public Instrument getInstrument() {
         );
     }
 
-    public final PlannedTime.CategorizedTime DHS_WRITE_TIME =
-        DHS_WRITE_TIMES.get(getInstrument());
```
swalker2m, line 440: This is the substantive update.

cquiroz: Odd we noted this now, anyway 👍

cquiroz (approved):

PR #1880, opened 2021-08-04
Remove keywords from the p1 model

cquiroz:
This is the first part of REL-3957 where keywords are removed from the p1 model.
I wonder if this may have any unforeseen impact on p2, especially with existing programs


swalker2m (approved):
Looks good.  I don't see any usages of `GsaPhase1Data` keywords, so fingers crossed.  :+1:

PR #1884, opened 2021-08-19
MaroonX

cquiroz:


bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/overheads/Overheads.scala
```diff
@@ -14,8 +14,6 @@ import scala.annotation.tailrec
 sealed trait Overheads {
   // REL-2985 -> REL-2926: acquisitionOverhead and otherOverheadFraction are no longer used.
   def partnerOverheadFraction: Double // fpart
-  def acquisitionOverhead: Time       // acqover
```
cquiroz, line 17: This is a bit scary but these fields are not being used

bundle/edu.gemini.ags.client.impl/src/main/scala/edu/gemini/ags/client/impl/QueryArgs.scala
```diff
@@ -58,6 +58,7 @@ object QueryArgs {
     case g: TexesBlueprint      => Right(Instrument.Nifs.id)      // REL-1062
     case g: VisitorBlueprint    => Right(Instrument.Niri.id)      // REL-1090
     case g: ZorroBlueprint      => Right(Instrument.Nifs.id)      // REL-3454: treat Zorro like NIFS.
+    case g: MaroonXBlueprint    => Right(Instrument.Nifs.id)
```
swalker2m, line 61: Might add a comment here to indicate that this is intentional?

swalker2m (approved):
Looks great to me. :+1:

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/overheads/Overheads.scala
sraaphorst: Nice to see this hot mess being cleaned up.

sraaphorst (approved):

PR #1887, opened 2021-08-24
WIP:  REL-3580 - ITC support for exact IQ and CC

andrewwstephens:
This PR adds the ability to set exact IQ and CC in the ITC web form.  This is done by adding a new accepted value "EXACT" to the list of allowed IQ and CC web form options.  When this option is selected the web form provides values for ExactIQ and ExactCC.  The exactcc value is used to calculate the atmospheric transmission (CloudTransmissionVisitor.java).  The exactiq is treated as a Gaussian source of the specified size (ImageQualityCalculationFactory.java).  This all seems to work as expected, and the usual baseline tests all pass.  However, one of the JsonServletSpec tests fail:

> project bundle_edu_gemini_itc_web
[info] Set current project to edu.gemini.itc.web (in build file:/home/astephens/ocs/ocs/)
> test
...
[info] JsonServletSpec
[info] 
[info] JsonServlet should
[info]   + return SC_BAD_REQUEST with an error message in the case of malformed Json
[info]   + return SC_BAD_REQUEST with an error message in the case of incorrect Json
[error]   x return SC_OK with a valid response for a valid request
[error]    Falsified after 4 passed tests.
[error]    > ARG_0: ObservingConditions(IQ 0,CC90,WV50,SB20,2.6285854169719077E303,-7.275257229517602E-167,-6.8277559092249724E-124)
[error]    > ARG_1: TelescopeDetails(SILVER, Up-looking, PWFS)
[error]    > ARG_2: NiriParameters(BBF_J,WOLLASTON,F32_PV,IMAG_SPEC_3TO5,DEEP,MASK_9,CENTRAL_256,Some(AltairParameters(4.567890646741859E-117,1.2959084932867202E-182,OUT,NGS)))
[error]    The seed is jqj1852W9BLjBWl6BVYInYW6VfR6GBDn1KzzKciekyM=
[error]    
[error]    > 'text/html; charset=UTF-8' != 'text/json; charset=UTF-8' (JsonServletSpec.scala:116)
[error] Actual:   text/[html]; cha...
[error] Expected: text/[json]; cha...

### DIFF
--- a/bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/Parameters.scala
+++ b/bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/Parameters.scala
@@ -10,7 +10,9 @@ final case class ObservingConditions(
                       cc: CloudCover,
                       wv: WaterVapor,
                       sb: SkyBackground,
-                      airmass: Double)
+                      airmass: Double,
+                      exactiq: Double,
+                      exactcc: Double)
 
 // ==== Source definition
 final case class SourceDefinition(
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/obscomp/SPSiteQuality.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/obscomp/SPSiteQuality.java
@@ -304,8 +304,7 @@ public class SPSiteQuality extends AbstractDataObject implements PropertyProvide
             @Override public boolean isObsolete() { return true; }
         },
         ANY(       "Any",       100, -3.0),
-        ;
-
+        EXACT(     "Exact",       0,  0.0);
 
         /** The default CloudCover value **/
         public static CloudCover DEFAULT = ANY;
@@ -366,7 +365,8 @@ public class SPSiteQuality extends AbstractDataObject implements PropertyProvide
         PERCENT_20("20%/Best",  20,  0.5),
         PERCENT_70("70%/Good",  70,  0.0),
         PERCENT_85("85%/Poor",  85, -0.5),
-        ANY(       "Any",      100, -1.0);
+        ANY(       "Any",      100, -1.0),
+        EXACT(     "Exact",      0,  0.0);
 
         /** The default ImageQuality value **/
         public static ImageQuality DEFAULT = ANY;
@@ -385,7 +385,6 @@ public class SPSiteQuality extends AbstractDataObject implements PropertyProvide
         public byte getPercentage() {
             return _percentage;
         }
-
 
         public String displayValue() {
             return _displayValue;
@@ -417,7 +416,6 @@ public class SPSiteQuality extends AbstractDataObject implements PropertyProvide
         public static Option<ImageQuality> read(String s) {
             return SPSiteQuality.read(s, values());
         }
-
 
         @Override
         public double getAdjustment(final BandsList bl) {
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/itc/ConditionsPanel.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/itc/ConditionsPanel.scala
@@ -63,7 +63,8 @@ final class ConditionsPanel(owner: EdIteratorFolder) extends GridBagPanel {
     cc.selection.item,
     wv.selection.item,
     sb.selection.item,
-    am.selection.item)
+    am.selection.item, 0.0, 0.0
+  )
 
   def update() = {
     // Note: site quality node can be missing (i.e. null)
--- a/bundle/edu.gemini.itc.web/src/main/resources/index.html
+++ b/bundle/edu.gemini.itc.web/src/main/resources/index.html
@@ -100,6 +100,6 @@ Source code documentation: <a href='../itcdocs/html/index.html'>../itcdocs/html/
 <br>
 Plots of data files: <a href='../itcdocs/plots/index.html'>../itcdocs/plots/</a>
 <hr>
-<footer>Last updated 2021-Mar-09</footer>
+<footer>Last updated 2021-Aug-23</footer>
 </body>
 </html>
--- a/bundle/edu.gemini.itc.web/src/main/resources/servlets/web/ITCacqCam.html
+++ b/bundle/edu.gemini.itc.web/src/main/resources/servlets/web/ITCacqCam.html
@@ -322,65 +322,64 @@
                 <td colspan="4">Wavefront sensor for tip-tilt compensation: PWFS <i>(only)</i>
                     <input value="PWFS" name="Type" type="hidden"></td>
 
-    </tr>
-    <tr>
-      <td colspan="4">
-          <p align="right"><input value="Calculate" type="submit" style="background-color: gold; padding: 3px 8px;"></p>
-      </td>
-    </tr>
-  </table>
-  <p>
-  &nbsp;
-  </p>
+            </tr>
+            <tr>
+                <td colspan="4">
+                    <p align="right"><input value="Calculate" type="submit" style="background-color: gold; padding: 3px 8px;"></p>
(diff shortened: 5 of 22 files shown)


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/itc/ConditionsPanel.scala
```diff
@@ -63,7 +63,8 @@ final class ConditionsPanel(owner: EdIteratorFolder) extends GridBagPanel {
     cc.selection.item,
     wv.selection.item,
     sb.selection.item,
-    am.selection.item)
+    am.selection.item, 0.0, 0.0
```
cquiroz, line 66: Does this mean in the OT we won't see the option to set these fields?

andrewwstephens: We would need to modify the OT UI which we probably don't want to do?

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/obscomp/SPSiteQuality.java
```diff
@@ -366,7 +365,8 @@ public double getAdjustment(final BandsList bl) {
         PERCENT_20("20%/Best",  20,  0.5),
         PERCENT_70("70%/Good",  70,  0.0),
         PERCENT_85("85%/Poor",  85, -0.5),
-        ANY(       "Any",      100, -1.0);
+        ANY(       "Any",      100, -1.0),
+        EXACT(     "Exact",      0,  0.0);
```
cquiroz, line 369: I wonder if this will have an effect on serialization. Would importing old programs still work?

cquiroz: This entry will show on the conditions selector on the UI

swalker2m: Importing from XML should still work, and I believe just adding new enum members is even binary compatible.  I'm wondering about all the usages though.  For example here:

https://github.com/gemini-hlsw/ocs/blob/develop/bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/api/AgsAnalysis.scala#L196

`EXACT` will be considered `worseOrEqual` to `ANY`.

andrewwstephens: Ah, right.  We should probably not allow selecting "EXACT" in the OT for that reason (AGS) as well.

cquiroz: Yes, ProbesLimit has one of those usages too

bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/Parameters.scala
```diff
@@ -10,7 +10,9 @@ final case class ObservingConditions(
                       cc: CloudCover,
                       wv: WaterVapor,
                       sb: SkyBackground,
-                      airmass: Double)
+                      airmass: Double,
+                      exactiq: Double,
```
cquiroz, line 14: I'm wondering if we could leave `ImageQuality` as it is and instead make these `airmass: Option[Double]` where if it is defined we take it as  an indication to use that value instead of the one from iq.
I fear a bit changing `ImageQuality` enum as it is being iterated in quite a few places.
I think this change would allow the OT to remain as it is and still let the web itc to accept exact values

bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/Parameters.scala
```diff
@@ -1,16 +1,29 @@
...
+import edu.gemini.spModel.gemini.obscomp.SPSiteQuality.{CloudCover, ImageQuality, SkyBackground, WaterVapor}
+
+import scalaz.\/
 
 // ==== Observing conditions
 
 final case class ObservingConditions(
-                      iq: ImageQuality,
-                      cc: CloudCover,
+                      iq: Double \/ ImageQuality,
```
cquiroz, line 12: Maybe me can just use `Either`? This code is being ported to scala 2.13 without `scalaz`

swalker2m: I started with `Either` but then I saw `\/` was used elsewhere in the ITC (e.g., `ConfigExtractor`) so I thought it best to be consistent with that.

cquiroz: I guess that's ok, we'll have to port the rest anyway.

At any rate, great not to modify the enum, I started fixing bugs due to that change and realized it would be fragile to have this extra entry

bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/Parameters.scala
swalker2m: That's a good point about the OT.  When do we want these changes to go into production?  Because we use TRPC for OT <-> ITC communication, unless the update is made along with the December release we'll either need to require an OT update or else make sure we provide a constructor that allows this original call to succeed:

```scala
new ObservingConditions(
     iq.selection.item,
     cc.selection.item,
     wv.selection.item,
     sb.selection.item,
     am.selection.item)
```

along with support for maintaining binary compatibility via the correct `serialVersionUID` and a `readObject` implementation.

cquiroz: Is that because `ObservingConditions` is passed in the TRPC request?
I think this update could be published to production on the web itc without affecting the OT and then we'd deploy both OT and SPDB to production together in the next release.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/obscomp/SPSiteQuality.java
cquiroz: With the enum we get things like:
image(2)

bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/Parameters.scala
swalker2m: Oh, okay the web ITC communicates with a different server (and not the SPDB)?

cquiroz: As far as I understand it the itc web has a web backend that serves the web pages and does calculations and that's the one we deploy to the itc server. In the meanwhile the OT talks directly to the itc running in spdb over TRPC

bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/Parameters.scala
cquiroz: This change is geared towards supporting exact iq and cc values on gpp however in `lucuma-core` we have that IQ and CC are limited to a strict set of values:

https://github.com/gemini-hlsw/lucuma-core/blob/master/modules/core/shared/src/main/scala/lucuma/core/enum/ImageQuality.scala#L16
https://github.com/gemini-hlsw/lucuma-core/blob/master/modules/core/shared/src/main/scala/lucuma/core/enum/CloudExtinction.scala#L9

Should we make the `lucuma-core` version a non-enum and thus remove (in the gpp itc) the need for an `Either`?

PR #1891, opened 2021-09-03
Update README.md

swalker2m:
Took me a while to remember how to do this, and I looked here hoping I'd documented it before.


cquiroz (approved):
Cool that's useful

PR #1892, opened 2021-09-10
WIP - Horizons API changes

tpolecat:
This updates our Horizons client to use the JPL's new API (the old one is going away).


swalker2m (approved):
:+1:  Would this work if put into production now or do we need to wait for a specific date?

PR #1893, opened 2021-09-27
reference coordinates

tpolecat:
ITAC users need a way to override the ephemeris for nonsidereal targets, because sometimes the reference point at mid-semester is in an RA bucket that's too full. This adds a `referenceCoordinates` method to the Phase 1 immutable `NonSiderealTarget`, which for various reasons is where it needs to go. I'm not proud of any of this.

### DIFF
--- a/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Observation.scala
+++ b/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Observation.scala
@@ -23,7 +23,7 @@ object Observation {
 
   /**
    * @param referenceCoordinates optional coordinates that will overide the ephemeris for ITAC
-   *   bing-filling. This value is ignored unless the target is nonsidereal, is used only by
+   *   bin-filling. This value is ignored unless the target is nonsidereal, is used only by
    *   ITAC, and does not exist in Phase 2.
    */
   def apply(m:M.Observation, referenceCoordinates: Option[Coordinates]) = new Observation(m, referenceCoordinates)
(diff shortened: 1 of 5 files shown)


bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/Observation.scala
```diff
@@ -20,7 +21,12 @@ object Observation {
             band:Band,
             progTime:Option[TimeAmount]) = new Observation(blueprint, condition, target, progTime, band)
 
-  def apply(m:M.Observation) = new Observation(m)
+  /**
+   * @param referenceCoordinates optional coordinates that will overide the ephemeris for ITAC
+   *   bing-filling. This value is ignored unless the target is nonsidereal, is used only by
```
swalker2m, line 26: `bin-filling` i suppose

swalker2m (approved):
I think we rejected this already but what if it were a reference instant instead and that instant were used to find the coordinates?  In theory that could survive to the scheduling block time at Phase 2.  OTOH, nobody is asking for that and I guess it would be more difficult for the poor ITAC user.  It seems odd that they can just pick any coordinate in the sky but, anyway, :+1:

tpolecat:
Yeah I'm trying to do this without changing ITAC's interface. Right now they specify new coordinates so I'm sticking with that to avoid complicating things.

swalker2m:
> Right now they specify new coordinates so I'm sticking with that to avoid complicating things.

Makes sense: +1:

cquiroz:
Would this affect the PIT?

tpolecat:
> Would this affect the PIT?

No, the member defaults to `None` and there's no reason for the PIT to touch it. It all compiles so I think we're fine.

PR #1894, opened 2021-09-28
bump versions

tpolecat:
I'm going to publish the 2.13 versions of the p1 model and related modules for ITAC, so I need to bump the versions. @cquiroz is this sensible?


cquiroz:
Seems fine

cquiroz (approved):
LGTM

PR #1897, opened 2021-10-04
REL-3985: fix time accounting summary report

swalker2m:
It's unclear why, but the 2021B programs began to include time accounting summary report rows for all partners instead of only those which allocated time.  This update:

1. Limits the report to only partners which allocated time
2. Replaces keywords in `GsaPhase1Data`.  We don't use this information any longer, but including it (for now) allows a patch release to be created from `develop` without serialization issues.

### DIFF
--- a/bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/SpProgramFactory.scala
+++ b/bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/SpProgramFactory.scala
@@ -12,7 +12,7 @@ import edu.gemini.model.p1.mutable.TooOption
 import edu.gemini.spModel.gemini.obscomp.SPProgram
 import edu.gemini.spModel.gemini.obscomp.SPProgram.{PIInfo, ProgramMode}
 import edu.gemini.shared.util.TimeValue
-
+import edu.gemini.spModel.gemini.phase1.GsaPhase1Data.Keyword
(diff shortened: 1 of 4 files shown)


bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/SpProgramFactory.scala
```diff
@@ -285,7 +285,7 @@ object SpProgramFactory {
     val category = new Gsa.Category(~proposal.category.map(_.value()))
     val pi       = gsaPhase1DataInvestigator(proposal.investigators.pi)
     val cois     = proposal.investigators.cois.map(gsaPhase1DataInvestigator).asJava
-    new Gsa(abstrakt, category, pi, cois)
+    new Gsa(abstrakt, category, java.util.Collections.emptyList[Keyword], pi, cois)
```
cquiroz, line 288: Were we sending the keywords to the archive?

cquiroz (approved):
LGTM

bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/skeleton/factory/SpProgramFactory.scala
swalker2m: There's a comment that indicates we were sending them to the archive along with categories but I can't find any evidence of sending either now.

PR #1898, opened 2021-10-04
REL-3986: adding logging

swalker2m:
REL-3986 involves dropped start/end visit events.  The code for handling events was refactored for sending FIRE messages and so no longer corresponds to what is currently in production.  Accordingly, it's difficult to tell what might be wrong with the production version.  This PR simply adds a bit of logging so that we might gather additional information in the interim between the patch release and the December update.


cquiroz (approved):
LGTM

PR #1900, opened 2021-10-05
REL-3964: respect dataflow status selections in OT browser

swalker2m:
Makes the "Dataflow Status" filter work as expected in the OT browser.


cquiroz (approved):
LGTM

PR #1901, opened 2021-10-06
REL-1306: Visitor instrument configuration with setup and readout

swalker2m:
Adds visitor instrument setup and readout time configuration and includes them in the planned time calculation.


cquiroz (approved):
Looks great. I still wonder why are we making relatively big changes to the old code base but, nonetheless, the code looks good

PR #1902, opened 2021-10-06
REL-3940: make QPT scheduled visit Observation column bigger

swalker2m:
Observation IDs have gotten longer since the QPT was first released.  In particular all program ids have indices of at least 3 digits now.  This made it so the typical observation id is truncated in the Scheduled Visits table.

This PR makes the minimum size observation id slightly bigger to accommodate the new program id standard and makes the column resizable as requested.


swalker2m:
Screen Shot 2021-10-06 at 13 44 13

PR #1904, opened 2021-10-06
REL-3924: Switch to new catalog server for all Gemini catalogs

swalker2m:
When the GAIA @ Gemini server is put into production, it will also handle queries for all existing catalogs.  This update temporarily switches to the new test server for the UCAC, PPMXL, etc. catalogs.


cquiroz (approved):
LGTM

PR #1907, opened 2021-10-07
REL-3881: adds QH 5W option to GHOST manual flat/arc

swalker2m:
Adds a 5W QH GCAL lamp.  For now, it appears only in the GHOST manual flat/arc OT component.  Also fixes a few sequence config builder issues for GHOST manual flat/arc.


cquiroz (approved):
LGTM, I wonder why the test is failing now

PR #1908, opened 2021-10-13
REL-4005: GHOST Sequence Parameters

swalker2m:
A minor update to GHOST sequence parameters to make them behave more like other instruments.  Namely there are "instrument" exposure time parameters (for blue and red) and "observe" exposure time parameters derived from them.  The "instrument" values are mostly hidden in the UI but necessary to get the sequence to work like the other instruments.  The seqexec should use the "observe" values as it presumably does with the other instruments.

* Puts the generic (not red/blue specific) exposure time parameter back since it is used throughout the OT and seems useful for the user regardless.  It shows you the expected total exposure time for a step (which is the max of the blue and red exposure times multiplied by their respective counts).

* Adds a 4 argument config injector, which ultimately went unused but could be useful at some point.


cquiroz (approved):
LGTM

PR #1915, opened 2021-11-03
REL-3965: move notes to group

swalker2m:
As requested, moves the IGRINS-specific notes back to the target group.


cquiroz (approved):
LGTM

PR #1917, opened 2021-11-04
REL-4014: correct GHOST IFU size

swalker2m:
Corrects the GHOST standard resolution IFU size.  In theory. The requirements are difficult to parse.

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/tpe/feat/TpeGhostIFUFeature.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/tpe/feat/TpeGhostIFUFeature.scala
@@ -142,34 +142,33 @@ final class TpeGhostIfuFeature extends TpeImageFeature("GHOST", "Show the patrol
   private def drawIFU(g2d: Graphics2D, p: Point2D): Unit = {
     // Transform the hexagon at the appropriate place.
     val trans: AffineTransform = {
-      val t = new AffineTransform
-      t.translate(p.getX, p.getY)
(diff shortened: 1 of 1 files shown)


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/tpe/feat/TpeGhostIFUFeature.scala
```diff
@@ -142,34 +142,33 @@ final class TpeGhostIfuFeature extends TpeImageFeature("GHOST", "Show the patrol
...
     }
 
     g2d.setColor(TpeGhostIfuFeature.IfuColor)
-    val hex: Area = TpeGhostIfuFeature.regularHexagon().createTransformedArea(trans)
-    g2d.fill(hex)
+    g2d.fill(TpeGhostIfuFeature.standardResolutionIfuArcsec.createTransformedArea(trans))
   }
 
   /**
-   * Draw the SR sky fibers relative to the center of SRIFU1. Thery are extremely small.
```
sraaphorst, line 158: Ooops.

sraaphorst (approved):
Apart from the not sure if reading correctly, this looks fine and we may as well test.

tpolecat (commented):
Well at least doing Java2D is kind of fun. How do we verify this against the actual instrument (since the spec is hard to interpret)?

swalker2m:
> How do we verify this against the actual instrument (since the spec is hard to interpret)?

I'm hoping Steve will help with that.  Presumably they can compare vs an actual image, but I expect Steve will be able to spot some issues just by looking at the TPE.  I don't have much confidence that this is correct, but I am confident that it is closer to correct than it was before so I'll await the next iteration.

PR #1918, opened 2021-11-04
REL-4010: Update to icons, images, and splash screen.

sraaphorst:
2022A icon and splash screen text update, tested for OS X and looks correct. (Obviously cannot test for Windows, but I made the ICO files match the previous version in size variations.)


cquiroz (approved):
LGTM

PR #1920, opened 2021-11-05
REL-3987 GMOS skeleton updates

tpolecat:
It is with great sadness that I announce this PR is ready for review. It updates the templates for GMOS and in so doing required adding two new features.

- Templates can now expand into organizational folders (in addition to scheduling groups, the default) as specified by the `GroupInitializer`. In order to make this work I had to add a `groupType` member to `TemplateFolder` that is consulted by the `InsantiationFunctor` at "runtime".
- For MOS observations the default MDF mask name is now set to match the program ID (without dashes, and with `-NN` at the end). In order to do this I needed to pass the `SPProgramID` down from the top to make it available in the `GroupInitializer`, which involved (among other things) turning it into a request parameter for the template reset web service.

There are tests that the proper observations make it into the template folder, that the folder type is properly recorded, and that the mask name is correctly set.

There are _not_ tests for runtime template expansion (to verify that the group type ends up correct) or for the template reset service. These will need to be monkey-tested.

### DIFF
--- a/bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/template/factory/impl/gmos/GmosBase.scala
+++ b/bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/template/factory/impl/gmos/GmosBase.scala
@@ -3,12 +3,25 @@ package edu.gemini.phase2.template.factory.impl.gmos
 import edu.gemini.spModel.template.SpBlueprint
 import edu.gemini.phase2.template.factory.impl.GroupInitializer
 import edu.gemini.spModel.gemini.gmos.GmosCommonType
+import edu.gemini.pot.sp.ISPNode
 
 trait GmosBase[B <: SpBlueprint] extends GroupInitializer[B] {
 
   /** Select the filter with wavelength closest to specified lambda. */
   def closestFilter[A <: GmosCommonType.Filter](f: A, fs:A*)(lambda:Double) = {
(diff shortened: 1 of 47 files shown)


bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/template/factory/impl/gmos/GmosBase.scala
```diff
@@ -11,4 +12,16 @@ trait GmosBase[B <: SpBlueprint] extends GroupInitializer[B] {
...
+  // SET "Custom Mask MDF" = G(N/S)YYYYS(Q/C/DD/SV/LP/FT)XXX-NN
+  //     where:
+  //     (N/S) is the site
+  //     YYYYS is the semester, e.g. 2015A
+  //     (Q/C/DD/SV/LP/FT) is the program type
+  //     XXX is the program number, e.g. 001, or 012, or 123
+  //     NN should be the string "NN" since the mask number is unknown
+  def defaultCustomMaskName(n: ISPNode): String = {
+    // or, in programmer-speak, delete the hyphens and add -NN
+    n.getProgramID.toString.filterNot(_ == '-') + "-NN"
```
swalker2m, line 24: By convention it happens to be now that the program number is always 3 digits long, but nothing guarantees it.  Maybe it doesn't matter though.

swalker2m (approved):
This all looks good to me. :+1:

tpolecat:
Alright this is ready for review. I'm sorry.

sraaphorst (approved):

PR #1921, opened 2021-11-08
REL-4015: update Gemini test catalog URLs

swalker2m:
Updates catalog server URLs.


swalker2m:
I've tested in the OT now and it works (with VPN turned on of course).

PR #1922, opened 2021-11-10
REL-3986: more logging

swalker2m:
Something sometimes happens to start/end events.  The log says they are recorded and yet they are not there.  I have a couple of theories but I'm hoping to gather more information with logging.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/pot/sp/memImpl/MemObsExecLog.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/pot/sp/memImpl/MemObsExecLog.java
@@ -1,9 +1,20 @@
 package edu.gemini.pot.sp.memImpl;
 
 import edu.gemini.pot.sp.*;
+import edu.gemini.shared.util.immutable.ImOption;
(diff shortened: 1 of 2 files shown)


bundle/edu.gemini.pot/src/main/java/edu/gemini/pot/sp/memImpl/MemObsExecLog.java
```diff
@@ -16,11 +27,68 @@
...
+
+    }
+
+    private Option<String> getObsId() {
+        return ImOption.apply(getContextObservation())
+                       .flatMap(o -> ImOption.apply(o.getObservationID()))
+                       .map(SPObservationID::stringValue);
+    }
+
+    @Override public PropagationId putClientData(String name, Object obj) {
```
swalker2m, line 63: Whenever the "shell" containing observation execution log data is updated, we'll dump some event information out.  A theory is that start/end visits are recorded correctly but subsequently get overwritten (perhaps by sync code?).  This logging should at least answer that question.

cquiroz (approved):
LGTM

PR #1923, opened 2021-11-11
REL-4015: support for v1.4, request v1.3

swalker2m:
Adds support for VO Table v1.4 but explicitly requests v1.3 from the Gemini catalog servers (for now).  There is a python bug related to the v1.4 output that we avoid by requesting v1.3.


cquiroz (approved):
LGTM

PR #1924, opened 2021-11-14
REL-4018: Reinstate missing ITC sky background data file

andrewwstephens:
Not sure how we lost it, but I found an old copy of the file from 2014.


cquiroz (approved):
LGTM

PR #1925, opened 2021-11-17
REL-3986: update logging

swalker2m:
Cuts down on extraneous logging that the previous REL-3986 introduced, yet adds a stack trace when it does matter.  While testing unrelated changes I noticed that test output was extremely chatty and wanted to focus it more on the cases that matter (when Start/End visits are added or removed).


cquiroz (approved):
LGTM

PR #1926, opened 2021-11-17
Make the fire test more reliable

swalker2m:
FIRE messages are posted asynchronously after the event that triggered them has been processed. The test case waits up to one second for the event to show up before triggering more events, but when other test cases are running in parallel and the machine is busy that is sometimes not enough.  This PR increases the wait time for an expected FIRE message to 10 seconds.

The PR only changes test code.


cquiroz (approved):
LGTM

PR #1927, opened 2021-11-17
REL-4005-B: Correct GHOST Readout Cost

swalker2m:
There should be a readout cost per exposure, not one for the entire step.  So 4x100 Red + 1x400 Blue actually has a readout cost of 4x60 (not just 60) and exposure + readout is 400 + 240 = 640.

Screen Shot 2021-11-16 at 18 13 10

Updates the planned time category so that `detail` is no longer nullable but rather an `Option`.  Adds the exposure count and label ("blue" or "red") that dominates the readout overhead in the `detail` message.

### DIFF
--- a/bundle/edu.gemini.itc.web/src/main/java/edu/gemini/itc/web/html/OverheadTablePrinter.java
+++ b/bundle/edu.gemini.itc.web/src/main/java/edu/gemini/itc/web/html/OverheadTablePrinter.java
@@ -109,7 +109,7 @@ public class OverheadTablePrinter {
 
             if (cts != null) {
                 for (PlannedTime.CategorizedTime lct : cts) {
-                    if (lct.detail.equals("LGS Reacquisition")) {
+                    if (lct.detail.contains("LGS Reacquisition")) {
                         gsaoiLgsReacqNum++;
                     }
                 }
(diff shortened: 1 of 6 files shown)


cquiroz (approved):
LGTM

bundle/edu.gemini.itc.web/src/main/java/edu/gemini/itc/web/html/OverheadTablePrinter.java
```diff
@@ -109,7 +109,7 @@ private int getGsaoiLgsReacqNum() {
 
             if (cts != null) {
                 for (PlannedTime.CategorizedTime lct : cts) {
-                    if (lct.detail.equals("LGS Reacquisition")) {
+                    if (lct.detail.contains("LGS Reacquisition")) {
```
tpolecat, line 112: Confidence high!

tpolecat (approved):
:+1:

bundle/edu.gemini.itc.web/src/main/java/edu/gemini/itc/web/html/OverheadTablePrinter.java
swalker2m: It passes the ITC "baseline" tests, so what could possibly go wrong?

PR #1928, opened 2021-11-17
REL-3987 GMOS skeleton updates (2)

tpolecat:
This fixes an issue where the `pid` parameter to the `TemplateServlet` was not being seen when passed in the multipart body, which caused the reset request to fail. There's probably a way to do it this way but I couldn't figure it out, so I moved the parameter into the URL.

Monkey-tested locally with PIT/OT/SPDB.


cquiroz (approved):
LGTM

PR #1929, opened 2021-11-17
REL-3987 GMOS skeleton updates (3)

tpolecat:
The blueprint text updates didn't mention changing the folder type for imaging, so I left it as `TYPE_SHEDULING` and in fact wrote some code to abstract over this difference. But I had interpreted the requiremeents incorrectly and all GMOS templates should in fact expand to `TYPE_FOLDER` instead. So this changes the folder type for imaging templates but leaves the abstraction there, in case they change their mind or we need to use the pattern for other instruments in the future.


swalker2m (approved):
Makes sense, looks good. :+1:

PR #1930, opened 2021-11-17
REL-3997: No such thing as GHOST HRIFU2 or HR w/o sky

swalker2m:
We incorrectly had labeled HRIFU as HRIFU1 and invented an HRIFU2 that doesn't exist.  The only high resolution option, as I understand it, is HRIFU + Sky.  This PR removes the HRIFU1 Target mode and changes HRIFU1 Target + HRIFU2 Sky to just "HRIFU Target, SRIFU2 Sky".  That is, the sky position is at SRIFU2 though it is unused.  Instead the dedicated HR sky fibers just above it are used as shown in the TPE.

In the screenshot below, I added a "User" target where the sky coordinates are.  Ordinarily nothing would show here, but slightly north of it are the HR sky fibers.

Screen Shot 2021-11-17 at 16 26 06

The requirements are still not clear though so I expect at least one more round of this.


cquiroz (approved):
LGTM

PR #1934, opened 2021-11-22
REL-2845: update probe limits

swalker2m:
The WFS probe limits have been measured carefully and require updates.  This impacts test cases as some previously valid candidates are no longer valid and previously invalid candidates are now valid.  Others that were and continue to be valid shift in required guide speed.


tpolecat (approved):
:+1:

PR #1936, opened 2021-11-23
REL-4005: exposure time again

swalker2m:
Fixes a bug with GHOST planned time calculation.  I was treating the exposure and readout times independently so a case such as 

```
Red.: 1 x 500s
Blue: 3 x 100s
```

would pick up 500s exposure from the red camera and 180s readout (3 x 60s) from the blue for a total of 680s (500 + 180).  Instead the red camera should "dominate" the calculation because 1 x (500 + 60) > 3 x (100 + 60) and we should have a total time of 560s instead of 680s.


cquiroz (approved):
LGTM

PR #1938, opened 2021-11-23
REL-3882: match GHOST and non-GHOST calibration sequences

swalker2m:
GHOST has its own manual GCAL sequence node to accommodate the two independent cameras.  Unfortunately the GHOST copy was adding `String` sequence values to the sequence instead of actual objects.  To make this part of the sequence match non-GHOST sequences, this PR switches to objects.


cquiroz (approved):
LGTM

PR #1946, opened 2022-01-10
REL-4043: PIT support for AEON part I

cquiroz:
The PIT needs to add support for multi facility proposals. This PR is the first part containing the model changes and the basic UI.
The UI needs refinements but I'm awaiting for clarification on the requirements

### DIFF
--- a/bundle/edu.gemini.model.p1/src/main/xsd/ProposalClass.xsd
+++ b/bundle/edu.gemini.model.p1/src/main/xsd/ProposalClass.xsd
@@ -52,6 +52,19 @@
         </xsd:complexContent>
     </xsd:complexType>
 
+    <xsd:complexType name="Facility">
+        <xsd:sequence>
+            <xsd:element name="name" type="xsd:string" minOccurs="0" maxOccurs="1"/>
+            <xsd:element name="aeonCompatible" type="xsd:boolean" minOccurs="0" maxOccurs="1"/>
+        </xsd:sequence>
+    </xsd:complexType>
+
+    <!-- Indicates what other facilities are included in this proposal -->
+    <xsd:complexType name="MultiFacility">
+        <xsd:sequence>
+            <xsd:element name="facility" type="Facility" minOccurs="1"/>
+        </xsd:sequence>
+    </xsd:complexType>
 
     <!--
         A Gemini queue proposal is a normal proposal (see above) with an optional band 3 request and a TOO option.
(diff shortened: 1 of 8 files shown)


bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/view/partner/PartnerView.scala
```diff
@@ -301,6 +322,94 @@ class PartnerView extends BorderPanel with BoundView[Proposal] {view =>
...
+              case MultiFacilityWrapper.Yes => Some(QueueProposalClass.multiFacility.set(q, Some(new MultiFacility())))
+              case MultiFacilityWrapper.No => Some(QueueProposalClass.multiFacility.set(q, None))
+              case _ => Some(q)
+            }
+            case Some(l: LargeProgramClass)           => model = selection.item match {
+              case MultiFacilityWrapper.Yes => Some(LargeProgramClass.multiFacility.set(l, Some(new MultiFacility())))
+              case MultiFacilityWrapper.No => Some(LargeProgramClass.multiFacility.set(l, None))
+              case _ => Some(l)
+            }
+            case _                                    => // shouldn't happen
```
swalker2m, line 385: Maybe blow up with an exception?

bundle/edu.gemini.model.p1/src/main/xsd/ProposalClass.xsd
```diff
@@ -52,6 +52,19 @@
         </xsd:complexContent>
     </xsd:complexType>
 
+    <xsd:complexType name="Facility">
+        <xsd:sequence>
+            <xsd:element name="name" type="xsd:string" minOccurs="0" maxOccurs="1"/>
+            <xsd:element name="aeonCompatible" type="xsd:boolean" minOccurs="0" maxOccurs="1"/>
+        </xsd:sequence>
+    </xsd:complexType>
```
swalker2m, line 60: I'm confused by what feeds the `aeonCompatible` element.  I mean, how does that get set in a proposal document.

swalker2m (approved):
Looks good.  With fear and trepidation about the downstream impact, :+1:

bundle/edu.gemini.model.p1/src/main/xsd/ProposalClass.xsd
cquiroz: There is a UI (still pending) that will let you select from a number of facilities and potentially the time requested at each

cquiroz:
This probably has relatively low impact downstream. The only thing that would change is the xml content and we wouldn't need to display this on the OT or take care when doing ITAC

tpolecat:
> The only thing that would change is the xml content and we wouldn't need to display this on the OT or take care when doing ITAC

What's the point of doing it at all then? It's just cosmetic?

cquiroz:
It will be used by TACs, I guess ITAC too but AFAICT is only relevant for humans checking the pdf summary rather than any actions done by code

PR #1947, opened 2022-01-10
REL-3934: Added more data to the ODB extractor.

sraaphorst:
Modification of the time accounting data sent from the ODB to the Scheduler via the program extractor.

### DIFF
--- a/bundle/edu.gemini.programexport.servlet/src/main/scala/edu/gemini/programexport/servlet/ProgramExportServlet.scala
+++ b/bundle/edu.gemini.programexport.servlet/src/main/scala/edu/gemini/programexport/servlet/ProgramExportServlet.scala
@@ -9,7 +9,7 @@ import edu.gemini.spModel.gemini.obscomp.SPSiteQuality.TimingWindow
 import edu.gemini.spModel.gemini.obscomp.{SPProgram, SPSiteQuality}
 import edu.gemini.spModel.gemini.phase1.GsaPhase1Data
 import edu.gemini.spModel.guide._
-import edu.gemini.spModel.obs.{ObsClassService, ObsQaStateService, ObservationStatus, SPObservation}
+import edu.gemini.spModel.obs.{ObsClassService, ObsQaStateService, ObsTimesService, ObservationStatus, SPObservation}
 import edu.gemini.spModel.obs.plannedtime.{PlannedTimeCalculator, SetupTime}
 import edu.gemini.spModel.obscomp.{SPGroup, SPInstObsComp, SPNote}
 import edu.gemini.spModel.obslog._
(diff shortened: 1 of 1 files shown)


bundle/edu.gemini.programexport.servlet/src/main/scala/edu/gemini/programexport/servlet/ProgramExportServlet.scala
```diff
@@ -133,19 +134,40 @@ final case class ProgramExportServlet(odb: IDBDatabaseService, user: Set[Princip
         jEmptyObject
     }
 
-  val timeAccountAwardEncodeJson: EncodeJson[(TimeAcctCategory, TimeAcctAward)] =
-    EncodeJson { case (c, a) =>
-      ("partnerTime" := TimeUnit.SECONDS.toMillis(a.getPartnerAward.getSeconds)) ->:
-        ("programTime" := TimeUnit.SECONDS.toMillis(a.getProgramAward.getSeconds)) ->:
+  val timeAccountAwardEncodeJson: EncodeJson[(TimeAcctCategory, TimeAcctAward, Long, Long)] =
+    EncodeJson { case (c, a, prog_used, part_used) =>
```
swalker2m, line 138: Sometimes I've done a non-implicit type class definition when there are multiple alternative implementations (e.g., multiple possible sort orders) and we're going to choose one explicitly and pass it in to something expecting an implicit implementation.  That's not the case here so I'm wondering why do the type class?

swalker2m (approved):
One question, but looks good!  :+1:

bundle/edu.gemini.programexport.servlet/src/main/scala/edu/gemini/programexport/servlet/ProgramExportServlet.scala
sraaphorst: I'm not sure I'm quite grokking the question... Are you asking why bother separating this out into a type class at all?

I think that is what you're asking, and honestly, I have no idea why I went that route, to be honest. I guess because it was the easiest way to modify the code? There's no reason, though, to make this a separate encoder and not include it in the only encoder that calls it, so I'll go ahead and do that for cleanliness.

swalker2m: Yes, I wasn’t very clear.  The encoder for that particular tuple didn’t make sense, I think, as an implicit so that was previously removed.  :+1:  Now since you aren’t using the `EncodeJson` implementation for anything other than to directly call its `encode` method (via `apply` I think in line 374) it just seemed that maybe it would be better to skip `EncodeJson` altogether and just extract and use the function `(ISPProgram, TimeAcctAllocation) => Json` it embodies.

I don't think we need to hold up progress or make any changes though, since this code isn't going to be maintained anyway!  Also, I may be misunderstanding something so, take with a grain of salt and all.

PR #1948, opened 2022-01-11
REL-4041: DARP support part I

cquiroz:
This contains DARP support on the p1 model and the PIT. It still needs some work on p1-monitor, skeletons and the backend servers

### DIFF
--- a/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/view/proposal/ProposalView.scala
+++ b/bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/view/proposal/ProposalView.scala
@@ -30,7 +30,7 @@ class ProposalView(advisor:ShellAdvisor) extends BorderPanel with BoundView[Prop
   implicit val boolMonoid = Monoid.instance[Boolean](_ || _,  false)
 
   // Bound
-  override def children = List(title, abstrakt, /* scheduling, */ category, attachment, investigators)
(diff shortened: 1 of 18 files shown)


bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/view/proposal/ProposalView.scala
```diff
@@ -114,54 +129,55 @@ class ProposalView(advisor:ShellAdvisor) extends BorderPanel with BoundView[Prop
     add(remove, BorderPanel.Position.East)
 
     // Panel for the buttons
-    object selectPanel extends BorderPanel {panel =>
+    lazy val selectPanel = new BorderPanel {panel =>
```
swalker2m, line 132: were there initialization issues or something?

swalker2m (approved):
I don't know what DARP is, and I think if I _did_ know I'd be morally opposed, but the code looks good to me.  :+1:   There will be no downstream impact of any of this correct?  Right? Yes?

cquiroz:
There certainly will be downstream impacts in ITAC and the ODB

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/view/proposal/ProposalView.scala
cquiroz: Yes, they are called on the setup of the parent ui control, when they become a `val` (as we have more than one `selectPanel`) the initialization fails unless we declare it `lazy`

PR #1949, opened 2022-01-12
Simplify the DARP model

cquiroz:
I wasn't very happy about the DARP attachment model and I think it could be improved a bit with this PR

### DIFF
--- a/bundle/edu.gemini.model.p1.submit/src/main/scala/edu/gemini/model/p1/submit/Post.scala
+++ b/bundle/edu.gemini.model.p1.submit/src/main/scala/edu/gemini/model/p1/submit/Post.scala
@@ -14,8 +14,8 @@ private[submit] object Post {
     con.setRequestProperty("Content-Type", s"""multipart/form-data; boundary="$BOUNDARY"""")
 
     for {
-      pdf1 <- p.meta.firstAttachment.toRight(ClientError("The proposal is missing some PDF attachment.")).right
-      pdf2 <- p.meta.secondAttachment.toRight(ClientError("The proposal is missing some PDF attachment.")).right
+      pdf1 <- p.meta.firstAttachment.toRight(ClientError("The proposal is missing PDF attachment 1.")).right
+      pdf2 <- p.meta.secondAttachment.toRight(ClientError("The proposal is missing PDF attachment 2.")).right
       _   <- doPost(con, p, pdf1, pdf2).right
      } yield Unit
   }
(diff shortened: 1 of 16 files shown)


bundle/edu.gemini.model.p1.submit/src/main/scala/edu/gemini/model/p1/submit/Post.scala
```diff
@@ -14,8 +14,8 @@ private[submit] object Post {
     con.setRequestProperty("Content-Type", s"""multipart/form-data; boundary="$BOUNDARY"""")
 
     for {
-      pdf1 <- p.meta.attachments.sortBy(_.index).map(_.name).headOption.flatten.toRight(ClientError("The proposal is missing some PDF attachment.")).right
-      pdf2 <- p.meta.attachments.sortBy(_.index).map(_.name).lastOption.flatten.toRight(ClientError("The proposal is missing some PDF attachment.")).right
+      pdf1 <- p.meta.firstAttachment.toRight(ClientError("The proposal is missing some PDF attachment.")).right
+      pdf2 <- p.meta.secondAttachment.toRight(ClientError("The proposal is missing some PDF attachment.")).right
```
swalker2m, line 18: Can/should we be more specific about which PDF attachment is missing?

bundle/edu.gemini.model.p1/src/test/resources/edu/gemini/model/p1/immutable/proposal_itac_no_ngoauthority.xml
```diff
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
 <proposal tacCategory="Extragalactic" schemaVersion="1.0.0">
     <meta>
-        <attachment name="ar_2012B_007.pdf" index="1"/>
-        <attachment name="ar_2012B_007.pdf" index="2"/>
```
swalker2m, line 5: they can be the same file in the end?

swalker2m (approved):
Seems like a great improvement.  :+1:

bundle/edu.gemini.model.p1.submit/src/main/scala/edu/gemini/model/p1/submit/Post.scala
cquiroz: Yes, good idea

bundle/edu.gemini.model.p1/src/test/resources/edu/gemini/model/p1/immutable/proposal_itac_no_ngoauthority.xml
cquiroz: So this is just a test, and currently there is no restriction. But it is a good question whether we should verify it

PR #1954, opened 2022-01-21
REL-4044: Update partner calibration overheads in PIT

cquiroz:


bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/overheads/Overheads.scala
```diff
@@ -13,66 +13,33 @@ import scala.annotation.tailrec
...
-  private case class SimpleOverheads(partnerOverheadFraction: Double) extends Overheads
+  private case class SimpleOverheads(partnerOverheadFraction: Double) extends Overheads {
+    def calculate(progTime: TimeAmount): ObservationTimes = {
+      val progTimeHrs = progTime.toHours.value
+      val partTimeHrs = progTimeHrs * partnerOverheadFraction
+      ObservationTimes(TimeAmount(progTimeHrs, TimeUnit.HR), TimeAmount(partTimeHrs, TimeUnit.HR))
+    }
+  }
+  private case class TelluricOverheads(telluricTime: TimeAmount) extends Overheads {
+    val VisitLength = 1.5
```
swalker2m, line 35: We're saying, I think, that we need a telluric for every 1.5 hours (or fraction thereof) of science?  If so, maybe `ScienceHoursPerTelluric` would be a better name, idk.  I was initially confused by the name.

swalker2m (approved):
Looks good, fewer changes than I expected for this.  :+1:

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/overheads/Overheads.scala
cquiroz: Sounds good. I'll also indicate the units

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/overheads/Overheads.scala
```diff
@@ -31,11 +31,11 @@ object Overheads extends (BlueprintBase => Option[Overheads]) {
       ObservationTimes(TimeAmount(progTimeHrs, TimeUnit.HR), TimeAmount(partTimeHrs, TimeUnit.HR))
     }
   }
-  private case class TelluricOverheads(telluricTime: TimeAmount) extends Overheads {
-    val VisitLength = 1.5
+  private case class ScienceHoursPerTelluric(telluricTime: TimeAmount) extends Overheads {
+    val VisitLength = TimeAmount(1.5, TimeUnit.HR)
```
swalker2m, line 35: Oh, it was actually the `VisitLength` name itself I was confused about.  Maybe I'm just generally confused though.

cquiroz: 😄 Te name `VisitLengh` comes directly from the original ticket

swalker2m: Ok, I don't think they care what we call our internal variables but maybe it's best to match what they put in the task.  I think I'd put the class name back to `TelluricOverheads` though.  Sorry for the hassle!  It looks good. :+1:

swalker2m: `NIRSpectroscopyOverheads` seems good.

PR #1955, opened 2022-01-24
Fixes to AEON/Multi facility feature

cquiroz:
It adds a column to the Gemini Time Required box and a few other fixes

image


swalker2m (approved):
The code looks good to me, so :+1:

I'm not sure about the requirements.  I had to read the dialog message many times to form an idea of what it might mean.  It seems like we're making it harder to make a proposal instead of easier.  Also, this seems like relevant information for a TAC.  It's unclear why it would be hidden from them unless I guess it could be used to figure out who is doing the request.

cquiroz:
I think this for internal Gemini operations. In fact this information is not to be included in the PDFs which the TACs really review

PR #1957, opened 2022-01-25
REL-3989 Select the middle semester date for non-sidereal targets

cquiroz:
Signed-off-by: Carlos Quiroz <carlos.quiroz@noirlab.edu>


swalker2m (approved):
Can't say I know all the `xsl` syntax but it looks credible to me.  :+1:

PR #1959, opened 2022-01-28
REL-4042: Accept DARP submissions on the backend

cquiroz:
Plus a set of small bug fixes


bundle/edu.gemini.model.p1.submit/src/main/scala/edu/gemini/model/p1/submit/Post.scala
```diff
@@ -15,15 +15,16 @@ private[submit] object Post {
 
     for {
       pdf1 <- p.meta.firstAttachment.toRight(ClientError("The proposal is missing PDF attachment 1.")).right
-      pdf2 <- p.meta.secondAttachment.toRight(ClientError("The proposal is missing PDF attachment 2.")).right
-      _   <- doPost(con, p, pdf1, pdf2).right
+      pdf2 <- Right(p.meta.secondAttachment).right
```
swalker2m, line 18: Do we still need the `.right` projection?  I guess we're using an old version of Scala.

swalker2m (approved):
Looks good. :+1:

bundle/edu.gemini.model.p1.submit/src/main/scala/edu/gemini/model/p1/submit/Post.scala
cquiroz: We do, this is scala-2.11

PR #1967, opened 2022-02-04
REL-4055: update ITC with new F2 filters 

andrewwstephens:
Straight-forward swap of data files.
This time with an updated unit test that checked the length of one of the data files!


sraaphorst (approved):
👍 
Looks good!

PR #1970, opened 2022-02-06
REL-4023: When using ITC exact conditions display the equivalent percentile bin

andrewwstephens:
This PR updates the ITC HTML output to display the equivalent percentile bins when the user inputs exact IQ or CC values.

The most significant changes are in HtmlPrinter.java, so please have a look and let me know if there is a better way to do things.


cquiroz (approved):
LGTM I guess we should also publish this changes for the new ITC

PR #1974, opened 2022-02-09
Setup nix

cquiroz:
Mostly this makes it easier to switch to jdk 8 for the ocs


toddburnside (approved):
Looks good, as far as I know.

PR #1983, opened 2022-03-16
REL-4080: allow PIs to change visitor instrument

swalker2m:
Currently only staff members can change the selected visitor instrument configuration.  This PR removes that restriction so that PIs can do so as well.


cquiroz (approved):
LGTM

PR #1984, opened 2022-03-16
REL-4050: mark hartmann lyot wheel options obsolete

swalker2m:
Hartmann settings remain for any existing observations, but the option disappears in the OT.


cquiroz (approved):
LGTM

PR #1986, opened 2022-03-17
REL-4059: Support multiple attachment parameters

swalker2m:
Updates the skeleton creation servlet (and the `skeleton` bash script that calls it) to allow multiple attachment arguments.  The existing single `attachment` parameter is still supported, but alternatively any number of `attachment`# parameters (`attachment1`, `attachment2`, ...) can be sent instead.  If old-style `attachment` is used, any indexed attachment parameters are ignored.


cquiroz (approved):
Thanks for taking the bullet here. LGTM

PR #1987, opened 2022-03-17
REL-4020: fixed position angle for MaroonX

swalker2m:
Fixes CR (and position angle) for MaroonX.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/visitor/VisitorInstrument.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/visitor/VisitorInstrument.java
@@ -117,6 +117,10 @@ public class VisitorInstrument extends SPInstObsComp
         }
     }
 
+    public boolean isFixed0() {
+        return visitorConfig.positionAngleMode() == VisitorPosAngleMode.Fixed0$.MODULE$;
+    }
+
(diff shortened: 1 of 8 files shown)


cquiroz (approved):
LGTM, all instruments should have a fixed PA

swalker2m:
As it stands, this bit gets added to the `tcc_tcs_config_file` XML that gets sent to the TCC:
 
```
   <param name="posAngle" value="IPA0"/>
    <paramset name="IPA0" type="rotator">
      <param name="ipa" value="0"/>
      <param name="cosys" value="Fixed"/>
    </paramset>
```

Andy reports that the SOSs "select 'fixed' in the TCC menu", not `IPA0` as shown here.  It's unclear to me exactly what needs to be sent in order to satisfy the requirement from the task:

> The MAROON-X team confirms that all MAROON-X observations should have CR=Fixed, and it would be fine to hard-code that without modifying the Visitor Instrument interface.

swalker2m:
Rotator config menu options:

CR_Config

"fixed" is selected for MaroonX apparently

swalker2m:
Updated to

```
    <param name="posAngle" value="fixed"/>
    <paramset name="fixed" type="rotator">
      <param name="ipa" value="0"/>
      <param name="cosys" value="Fixed"/>
    </paramset>
```

per discussion with @jluhrs

PR #1989, opened 2022-03-18
REL-1632: copy on-hold obs bug

swalker2m:
When copying an On Hold ToO observation to a different program we're supposed to reset the status to Phase 2.  This was working for individual observations but not for observations within a group.  The issue was the calculation of whether the group is being copied into the same or a different program.  We were making a group copy for the target program and then comparing that new group's program key to the target program key.  Instead we needed to compare the existing group's program key to the target program's key.


sraaphorst (approved):
Looks good to me.

PR #1990, opened 2022-03-21
REL-3978: Improve QPT sky brightness calculation

andrewwstephens:
This makes several improvements to the sky brightness calculation used by the QPT:
http://jira.gemini.edu:8080/browse/REL-3978


cquiroz:
@andrewwstephens I think we need to port these changes to explore, right?

swalker2m (approved):
:+1:  Thanks for taking care of this.

PR #1992, opened 2022-03-22
REL-4013: Missing Start/EndVisit logging

swalker2m:
More logging to illuminate, hopefully, what is happening in REL-4013.  The issue appears to be that execution events like start and end visit, which are kept in the `ObsExecLog`, are getting overwritten somehow in the sync/merge process.


sraaphorst (approved):
👍 This looks good to me, although it gave me some Java flashbacks.

PR #1993, opened 2022-03-25
REL-4095: StackOverflow in Merge

swalker2m:
A stack-safety issue causes synchronization of large programs (e.g., GS-2022A-LP-105 with 2000+ observations) to fail.

This PR addresses the problem in two ways:
* Changes to the implementation of VCS such that less stack is required
* An increase in maximum stack size from 1M (default) to 1536K


cquiroz (approved):
LGTM, admire your courage on taking on scalaz bugs

PR #1994, opened 2022-03-30
FIX Gaia id field parsing

cquiroz:
The 2.14 release of GAIA brought some surprises which broke ags in production for GEMS.
The main change is that when requesting the field "designation" it brings additionally an item `ID="DESIGNATION`

The usage of the ID field varies across votable implementation thus it is not just as simple as to lower case it.

Instead I added a `postProcessField` item to `CatalogAdapter` such that it is fixed only for GAIA

### DIFF
--- a/bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTable.scala
+++ b/bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTable.scala
@@ -23,11 +23,10 @@ object Ucd {
 
 /** Describes a field */
 case class FieldId(id: String, ucd: Ucd)
-case class FieldDescriptor(id: FieldId, name: String)
 
-case class TableRowItem(field: FieldDescriptor, data: String)
+case class TableRowItem(field: FieldId, data: String)
 case class TableRow(items: List[TableRowItem]) {
-  def itemsMap:Map[FieldId, String] = items.map(i => i.field.id -> i.data)(collection.breakOut)
+  def itemsMap: Map[FieldId, String] = items.map(i => i.field -> i.data)(collection.breakOut)
 }
 
 /** ParsedTable and ParsedResources contains a list of problems */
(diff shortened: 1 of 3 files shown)


bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTable.scala
```diff
@@ -23,11 +23,10 @@ object Ucd {
 
 /** Describes a field */
 case class FieldId(id: String, ucd: Ucd)
-case class FieldDescriptor(id: FieldId, name: String)
```
swalker2m, line 26: We never use `name`?

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableParser.scala
```diff
@@ -392,6 +393,9 @@ object CatalogAdapter {
 
     }
 
+    override def postProcessFieldDescriptor(fieldId: FieldId): FieldId =
+      if (fieldId.id == "DESIGNATION") fieldId.copy(id = fieldId.id.toLowerCase) else fieldId
```
swalker2m, line 397: I guess this is the crux of the matter but I don't understand.  What happens if you just always `toLowerCase` "DESIGNATION"?

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTable.scala
cquiroz: Apparently not.

bundle/edu.gemini.catalog/src/main/scala/edu/gemini/catalog/votable/VoTableParser.scala
cquiroz: you mean if you lower case all field.ids? That would be fine with Gaia as it only produces one ID. However other catalogs behave differently and a blanket change would break them (Simbad, ppxml)

In general if we wed'd read `ID` and if not found `name`. 
Previously we'd get
`<FIELD arraysize="*" datatype="char" name="designation" ucd="meta.id;meta.main">`

and now
`<FIELD ucd="meta.id;meta.main" name="designation" datatype="char" arraysize="*" ID="DESIGNATION">`

With the previous logic we'd parse this as a `FieldId("DESIGNATION")` and it wouldn't match with how we do the query which is with the field in lowercase.

An obvious solution would be to send the query with the field in uppercase but then it reverts to the previous answer and we'd fail to parse

swalker2m (approved):

PR #1995, opened 2022-03-31
Nix for arm in darwin

cquiroz:
Small update to enable nix on arm


swalker2m (approved):
Don't know what this does, but :+1:

PR #1997, opened 2022-04-11
[REL-3999] change base name for dual target

tpolecat:
This changes the base name from `Sky` to `Base` for Ghost target environments in dual-target mode.


cquiroz (approved):
LGTM

PR #1999, opened 2022-04-12
REL-4003: Add target to GHOST observation

swalker2m:
Fixes a bug in which a new target added to an existing GHOST observation is created without a GHOST asterism, leaving the observation broken.

There is still a problem in that pasting a non-GHOST target component onto a GHOST observation (with or without an existing target component) creates this situation.  This was going to be difficult to fix and at least now you can correct this situation by deleting the pasted target component and adding a new one.

### DIFF
--- a/bundle/jsky.app.ot/src/main/java/jsky/app/ot/nsp/SPTreeEditUtil.java
+++ b/bundle/jsky.app.ot/src/main/java/jsky/app/ot/nsp/SPTreeEditUtil.java
@@ -39,7 +39,7 @@ public class SPTreeEditUtil {
     private static final Logger LOG = Logger.getLogger(SPTreeEditUtil.class.getName());
 
     // Used to mark the initialized state of the dialog asking the user to confirm an operation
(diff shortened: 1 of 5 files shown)


bundle/jsky.app.ot/src/main/java/jsky/app/ot/nsp/SPTreeEditUtil.java
```diff
@@ -39,7 +39,7 @@ public class SPTreeEditUtil {
     private static final Logger LOG = Logger.getLogger(SPTreeEditUtil.class.getName());
 
     // Used to mark the initialized state of the dialog asking the user to confirm an operation
-     private static final int USER_CONFIRMATION_INITAL_STATE = -37;
+     private static final int USER_CONFIRMATION_INITIAL_STATE = -37;
```
swalker2m, line 42: The changes in `SPTreeEditUtil` are trivial and unrelated.  Sorry.

cquiroz (approved):
LGTM

PR #2000, opened 2022-04-13
REL-4001: Ghost Exposure Time Warning

swalker2m:
Warn if the red or blue exposure time in the static component or any sequence step is longer than 1800 seconds.


cquiroz (approved):
LGTM

PR #2001, opened 2022-04-13
REL-3995 Enable GMOS Fov by default

cquiroz:
This is a bit tricky as if we have started the OT before with GHOST this will be disabled and the override won't work

It is probably not worth trying to reset the value on `jsky.properties` or what do you think @swalker2m ?


cquiroz:
image

swalker2m (approved):
It's now enabled by default as requested, which is all that we needed to do.  It's true that if you've used it before your preference is recorded.  There's no way to distinguish having the FOV turned off explicitly by preference and never having turned it on before.

Looks good. :+1:

PR #2002, opened 2022-04-13
REL-3977 - Add Gaia bands to the ITC & OT

andrewwstephens:
This adds the three Gaia bands (G_BP, G, G_RP) to the OT and ITC.

Note that the tests are currently failing, so this will need a bit of help.

### DIFF
--- a/bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/MagnitudeBand.scala
+++ b/bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/MagnitudeBand.scala
@@ -2,20 +2,16 @@ package edu.gemini.spModel.core
 
 import edu.gemini.spModel.core.WavelengthConversions._
 
-sealed abstract class MagnitudeBand private (val name: String, val center: Wavelength, val width: Wavelength, val description: Option[String], val defaultSystem: MagnitudeSystem) extends Product with Serializable {
+sealed abstract class MagnitudeBand private (val name: String, val start: Wavelength, val center: Wavelength, val end: Wavelength, val description: Option[String], val defaultSystem: MagnitudeSystem) extends Product with Serializable {
 
-  private def this(name: String, center: Wavelength, width: Wavelength) =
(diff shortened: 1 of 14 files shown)


bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/MagnitudeBand.scala
```diff
@@ -2,20 +2,16 @@ package edu.gemini.spModel.core
 
 import edu.gemini.spModel.core.WavelengthConversions._
 
-sealed abstract class MagnitudeBand private (val name: String, val center: Wavelength, val width: Wavelength, val description: Option[String], val defaultSystem: MagnitudeSystem) extends Product with Serializable {
+sealed abstract class MagnitudeBand private (val name: String, val start: Wavelength, val center: Wavelength, val end: Wavelength, val description: Option[String], val defaultSystem: MagnitudeSystem) extends Product with Serializable {
```
cquiroz, line 5: Why this change? Do we need to rely on a triplet `start`, `center`, `end` rather than just `center` and `width`?

I guess we'd need to replicate this on gpp

andrewwstephens:
After poking around a bit, it looks like `center` is used by the ITC as an effective wavelength for things like calculating surface brightness and converting between systems, so it isn't necessarily the midpoint between `start` and `end` (which are used for checking that there is appropriate overlap).  I don't think this change is critical, but it does seem more accurate, at least for these Gaia filters.  I would like to review all the old filters and update their parameters, but it seemed outside the scope of this PR.

cquiroz:
Should we remove the conversion of Gaia bands to "Gemini" bands?

swalker2m:
> Should we remove the conversion of Gaia bands to "Gemini" bands?

If we do, we need to define the GeMS AGS in terms of these bands instead of the converted bands.

cquiroz:
Would we also need to port existing programs?

cquiroz (approved):
LGTM, thanks

PR #2006, opened 2022-04-18
REL-4113: AGS / GHOST / Explicit base bug

swalker2m:
The `AgsHash` did not take into consideration explicit GHOST base positions that are unlinked from targets.  This prevented AGS from knowing that an update was needed when the base position changes.

Unrelatedly, I tweaking the drawing of the bullseye marking the base position in these cases so that appears centered.


cquiroz (approved):
LGTM

PR #2007, opened 2022-04-18
REL-4024: adjust HR Sky location

swalker2m:
Places the HR Sky fibers on the exact coordinates entered by the user, adjusting the position of the SRIFU2 accordingly 3.28 arcseconds south.

### DIFF
--- a/bundle/edu.gemini.p2checker/src/main/scala/edu/gemini/p2checker/rules/ghost/GhostRule.scala
+++ b/bundle/edu.gemini.p2checker/src/main/scala/edu/gemini/p2checker/rules/ghost/GhostRule.scala
@@ -61,7 +61,7 @@ object GhostRule extends IRule {
           // The sky position is taken from the user configuration, but the
           // actual SRIFU2 is positioned just south of that.  For the range
           // check we need to use the actual SRIFU2 location.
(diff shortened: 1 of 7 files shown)


cquiroz (approved):
LGTM

swalker2m:
Here's an awkward picture from the docs.  Note that the sky position, which is determined by the coordinates of the SRIFU2, is 2 mm above.  The 2 mm has to be translated to arcsec (3.28 in this case).

Screen Shot 2022-04-18 at 14 24 26

So if the user wants to put the HR sky fibers at (0:00:00, 0:00:00) then they enter those coordinates and we ask to put SRIFU2 at (0:00:00, -0:00:03.28).

SteveMargheim:
Cristian can comment further put I believe we send the exact coordinates for the sky fiber position and tell GHOST we are sending HR Sky fiber position; no need to calculate and send an offset.

swalker2m:
> Cristian can comment further put I believe we send the exact coordinates for the sky fiber position and tell GHOST we are sending HR Sky fiber position; no need to calculate and send an offset.

Okay, switched to send the sky coordinates instead of the SRIFU2 position.  I still need to calculate the actual SRIFU2 position though for the range checking.  Unfortunately this will create sky positions that seem to be in range but actually aren't and vice versa when near the limits.

PR #2008, opened 2022-04-19
REL-4085: add resolution and asterism type to GHOST browser parameters

swalker2m:
Hacks to make it possible to query on GHOST resolution and asterism type parameters.


swalker2m:
Ain't much to look at, I'm afraid.

Screen Shot 2022-04-19 at 14 17 15

Also, the resolution mode and asterism type, while in reality are related, are here completely independent.  It's probably best to use one or the other, but if both are specified and inconsistent then no results are returned.

cquiroz:
Thanks a lot for taking this

cquiroz (approved):
LGTM

tpolecat:
Wow.

PR #2010, opened 2022-04-23
REL-4116:  Add NIRI ITC port selector and default to up-looking port

andrewwstephens:
A small bug-fix to the NIRI ITC web form.


cquiroz (approved):
LGTM

PR #2012, opened 2022-05-06
[REL-3980] add obs 24 to FPU/decker config

tpolecat:
Minor change here; Andy updated the blueprint to add the acquisition to the handler for FPU and decker.


cquiroz (approved):
LGTM

PR #2013, opened 2022-05-06
REL-3919: relax GAIA constraints

swalker2m:
From Andy:

This is working well, however, we are missing some stars that could be useful.  Could the query be relaxed to:
1. Remove the parallax constraint
2. Remove the bright magnitude limit (change phot_g_mean_mag+BETWEEN+9+AND+19 to phot_g_mean_mag<19)
3. Increase the proper motion constraint from <100 to <1000 mas/yr


cquiroz (approved):
LGTM, I suppose the pm limit is somewhat arbitrary?

swalker2m:
An arcsec per year seems like a lot of movement but I don't know how much should be disqualifying. They can even guide on nonsidereal targets if I understand correctly.

PR #2014, opened 2022-05-09
REL-4024: switch to magenta for GHOST targets

swalker2m:
By request, switching to magenta for GHOST targets to better contrast with light backgrounds.
Screen Shot 2022-05-09 at 09 39 44
.


cquiroz (approved):
LGTM

PR #2015, opened 2022-05-09
REL-4122: switch the PDF format

swalker2m:
It appears we select the wrong PDF format when the skeleton is created.


tpolecat (approved):

cquiroz (approved):
LGTM

PR #2016, opened 2022-05-12
REL-4126: synchronize XML processing

swalker2m:
I never got to the bottom of this, but catalog XML processing needs to be single-threaded.  For large result sets (e.g., 30K), the XML parsing from `String` to `Elem` to a list of targets takes a second or two which will now be paid for sequentially.


tpolecat (approved):

cquiroz (approved):
LGTM

PR #2018, opened 2022-05-14
REL-3919: tweak GAIA search limits

swalker2m:
Addresses two issues raised in REL-3919:

1. Hacks in an `isLgs` parameter in order to use different faint limits for LGS guide star searches vs. non-LGS.  Ideally we'd use the actually magnitude limits in the search parameters but for that we need to translate all the AGS limits to `phot_g_mean_mag`.
2. Switches the catalog default to Gemini GAIA instead of ESA GAIA for GeMS manual searches.

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/edu/gemini/catalog/ui/package.scala
+++ b/bundle/jsky.app.ot/src/main/scala/edu/gemini/catalog/ui/package.scala
@@ -70,8 +70,9 @@ case class ObservationInfo(ctx: Option[ObsContext],
                            offsets: Set[Offset],
                            catalog: CatalogName,
                            mt: MagnitudeTable) {
-  def catalogQuery:List[CatalogQuery] = validStrategies.collect {
-      case SupportedStrategy(s, query, _) if s == strategy => query
+  def catalogQuery: List[CatalogQuery] =
+    validStrategies.collect {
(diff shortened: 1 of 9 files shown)


bundle/jsky.app.ot/src/main/scala/edu/gemini/catalog/ui/package.scala
```diff
@@ -70,8 +70,9 @@ case class ObservationInfo(ctx: Option[ObsContext],
                            offsets: Set[Offset],
                            catalog: CatalogName,
                            mt: MagnitudeTable) {
-  def catalogQuery:List[CatalogQuery] = validStrategies.collect {
-      case SupportedStrategy(s, query, _) if s == strategy => query
+  def catalogQuery: List[CatalogQuery] =
+    validStrategies.collect {
+      case SupportedStrategy(s, query, _) if strategy.map(_.strategy).contains(s) => query
```
swalker2m, line 75: An added bonus here.  The `s == strategy` comparison we used before is an apples/oranges comparison of two different things that can never be equal.

cquiroz: Nice

cquiroz (approved):
LGTM, thanks

PR #2020, opened 2022-05-16
[REL-3980] update XML and p-code

tpolecat:
This fixes some issues in the XML and makes one code change, setting the FPU and decker for library obs 35.


swalker2m (approved):
Looks good.

tpolecat:
Test failure is a timeout in `VOTableClient` which is unrelated, so I'm merging this.

PR #2021, opened 2022-05-16
feature/REL-3977: REMOVE Gaia bands from the ITC & OT

andrewwstephens:
This UNDOES the changes in REL-3977:
https://github.com/gemini-hlsw/ocs/pull/2002

I'm not very happy with the ITC results when the input SED is normalized by one of the Gaia filters.  I suspect that this is due to the asymmetry of these filters, and to resolve the issue we may need to convolve the input SED with the normalizing filter profile.


tpolecat (approved):
:+1: pending ci

PR #2023, opened 2022-06-03
REL-4020: set a 'rotator' param instead of 'posAngle' for MaroonX

swalker2m:
Because `fixed` is not the same as `Fixed`, we were adding

```
    <param name="posAngle" value="fixed"/>
```

instead of 

```
    <param name="rotator" value="fixed"/>
```

Hilarity ensued.

### DIFF
--- a/bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/TccFieldConfig.java
+++ b/bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/TccFieldConfig.java
@@ -53,21 +53,16 @@ public class TccFieldConfig extends ParamSet {
         putParameter(TccNames.GUIDE_GROUP, getPrimaryGuideGroupName());
 
         // Add a rotator parameter and then create the config
-        // Note: If the config name contains "Fixed" in some way, it's assumed to be a Fixed rotator config
-        RotatorConfig rc = new RotatorConfig(_oe);
+        final RotatorConfig rc = new RotatorConfig(_oe);
         if (rc.build()) {
-            String rotConfigName = rc.getConfigName();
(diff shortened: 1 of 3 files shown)


bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/TccFieldConfig.java
```diff
@@ -53,21 +53,16 @@ public boolean build() throws WdbaGlueException {
...
-            if (rotConfigName.contains(TccNames.FIXED)) {
-                putParameter(TccNames.ROTATOR, rotConfigName);
-                add(rc);
-            } else {
-                putParameter(TccNames.POSANGLE, rotConfigName);
-                add(rc);
-            }
+            ImOption.apply(rc.attributeValue(TYPE)).foreach(t -> {
+                final String name = rc.getConfigName();
+                putParameter(t, name);
```
swalker2m, line 61: We always should add a parameter whose name is the same as the rotator config type (which is `posAngle` or `rotator`), so instead of checking the rotator config name and inferring the type from that I decided to so something radical and just use the type as the name directly.  This was then monkey tested and unit tested.

bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/TccFieldConfig.java
```diff
@@ -53,21 +53,16 @@ public boolean build() throws WdbaGlueException {
...
+                if (!TccNames.ALTAIR_FIXED.equals(name)) {
```
swalker2m, line 62: I don't know the reasoning behind skipping the rotator config param set addition for `AltairFixed` but I won't question it.  I assume it must refer to a file that gets loaded in the TCC or something.

cquiroz (approved):
LGTM

PR #2025, opened 2022-07-13
REL-4013: actually hold the lock throughout VCS operations

swalker2m:
Addresses a long-standing fault report in which execution events from the seqexec are being lost.  Having added additional logging in previous releases I noticed that these episodes always seem to happen when a program sync is being done at the same time that the event arrives.  These actions are supposed to be locked such that they are atomic but I found that the VCS locking implementation is incorrect.  In particular, actions are surrounded by program lock and unlock calls but instead of releasing the program lock after the wrapped (scalaz) `Task` is actually run, it was releasing the lock after construction of the `Task` instead.

This is a pretty serious flaw and I'm surprised the consequences seemingly haven't been even more serious.

### DIFF
--- a/bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs2/VcsServer.scala
+++ b/bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs2/VcsServer.scala
@@ -1,24 +1,25 @@
 package edu.gemini.sp.vcs2
 
-import edu.gemini.pot.sp.{ISPFactory, SPNodeKeyLocks, ISPProgram, SPNodeKey}
+import edu.gemini.pot.sp.{ISPFactory, ISPProgram, SPNodeKey, SPNodeKeyLocks}
 import edu.gemini.pot.sp.version.VersionMap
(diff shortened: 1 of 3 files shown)


tpolecat (approved):
This looks good. Also surprised it didn't cause bigger problems.

bundle/edu.gemini.sp.vcs/src/main/scala/edu/gemini/sp/vcs2/VcsServer.scala
```diff
@@ -179,16 +180,40 @@ class VcsServer(odb: IDBDatabaseService) { vs =>
...
+      }
+      unlock(k)
+    }
+
+    VcsAction(recordAndLock()) >>
+      // catch exceptions in the calculation of `body` itself
+      (try { body } catch { case e: Throwable => VcsFailure.vcsException(e).liftVcs[A] })
+        .run
+        .onFinish(_ => Task.delay(checkAndUnlock()))
+        .liftVcs
```
swalker2m, line 215: This little block holds the money change, insuring that the lock is released after the action is actually executed.

PR #2026, opened 2022-07-14
PIT 2023A

cquiroz:
Prepare the PIT for 2023A and update the icon


bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/transform/UpConverter.scala
```diff
@@ -33,26 +33,27 @@ object UpConverter {
...
+  val from2012B:List[SemesterConverter] = List(SemesterConverterToCurrent, SemesterConverter2022BTo2023A, SemesterConverter2022ATo2022B, SemesterConverter2021BTo2022A, SemesterConverter2021ATo2021B, SemesterConverter2020BTo2021A, SemesterConverter2020ATo2020B, SemesterConverter2019BTo2020A, SemesterConverter2019ATo2019B, SemesterConverter2018BTo2019A, SemesterConverter2018ATo2018B, SemesterConverter2017BTo2018A, SemesterConverter2016BTo2017A, SemesterConverter2016ATo2016B, SemesterConverter2015ATo2015B, SemesterConverter2014BTo2015A, SemesterConverter2014BTo2014BSV, SemesterConverter2012BTo2013A, SemesterConverter2013ATo2013B, SemesterConverter2013BTo2014A, SemesterConverter2014ATo2014B, SemesterConverter2017ATo2017B, LastStepConverter(Semester(2012, SemesterOption.B)))
```
sraaphorst, line 56: This just gets more and more depressing as we go on. 😄

sraaphorst (approved):
It's been a long time since I climbed out of the PIT, but I think this looks right. 👍

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/transform/UpConverter.scala
cquiroz: At least you have a guarantee that this will never get shorter 😀

PR #2029, opened 2022-07-15
REL-4149 Hide Gnirs HR_IFU in all cases and LR_IFU in some cases

cquiroz:


bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/dtree/inst/Gnirs.scala
```diff
@@ -87,7 +87,7 @@ object Gnirs {
     extends SingleSelectNode[(Altair, GnirsPixelScale, GnirsDisperser, GnirsCrossDisperser), GnirsFpu, (Altair, GnirsPixelScale, GnirsDisperser, GnirsCrossDisperser, GnirsFpu)](s) {
     val title = "Focal Plane Unit"
     val description = "Please select a focal plane unit."
-    val choices = GnirsFpu.values.toList
+    val choices = GnirsFpu.values.filterNot(f => (!(s._2 == GnirsPixelScale.PS_015 && s._4 == GnirsCrossDisperser.No) && f == GnirsFpu.LR_IFU) || f == GnirsFpu.HR_IFU).toList
```
tpolecat, line 90: I can't evaluate this in my head ... can you add a comment explaining the logic?

tpolecat (approved):
Looks find but a comment would help.

PR #2031, opened 2022-07-18
PIT GHOST support

cquiroz:
Base GHOST support but only with a single target


swalker2m (approved):
Looks good. :+1:

tpolecat:
Looks good. Will need GhostBlueprint support in ITAC. I'll open an issue.

PR #2035, opened 2022-07-27
Set GHOST overheads

cquiroz:
This made it impossible to set requested times for GHOST


swalker2m (approved):
Not sure what this is, but it looks good to me.

swalker2m (approved):
Pending more feedback from science this seems fine.  It is a bit inconsistent in that the dual target option doesn't specify which target is assigned to which IFU but the implication of Target + Sky and Sky + Target is (IFU1 + IFU2).

cquiroz:
I think that is the correct assumption

PR #2037, opened 2022-08-10
Add ITC charts servlet

cquiroz:
This is an extra servlet for ITC meant to serve chart data. it will be used in GPP

### DIFF
--- a/bundle/edu.gemini.itc/src/main/scala/edu/gemini/itc/service/ItcServiceImpl.scala
+++ b/bundle/edu.gemini.itc/src/main/scala/edu/gemini/itc/service/ItcServiceImpl.scala
@@ -60,6 +60,18 @@ class ItcServiceImpl extends ItcService {
     case e: Throwable => ItcResult.forException(e)
   }
 
+  def calculateCharts(p: ItcParameters): Result = try {
+    // execute ITC service call with updated parameters
+    p.observation.calculationMethod match {
+      case _: Imaging       => ItcResult.forMessage ("Imaging not implemented.")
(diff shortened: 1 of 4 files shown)


bundle/edu.gemini.itc/src/main/scala/edu/gemini/itc/service/ItcServiceImpl.scala
```diff
@@ -115,6 +134,10 @@ class ItcServiceImpl extends ItcService {
     ItcResult.forResult(s)
   }
 
-
+  private def spectroscopyResultCharts(recipe: SpectroscopyArrayRecipe): Result = {
+    val r = recipe.calculateSpectroscopy()
+    val s = recipe.serviceResult(r, false)
+    ItcResult.forResult(s)
```
swalker2m, line 140: What do `r` and `s` consist of?  This must be something that already existed?

swalker2m (approved):
Looks good. :+1:

bundle/edu.gemini.itc/src/main/scala/edu/gemini/itc/service/ItcServiceImpl.scala
cquiroz: These methods calculate the graph data via an ITC recipe. This is a copy of the existing `spoctroscopyResults` which in hindsight I could probably just reuse setting the headless value to false.
Originally I expected I needed to do some changes to the service result but in practice itc will always return the full data including graphs and calculated SN

PR #2039, opened 2022-08-15
REL-4167: ITC support for GMOS ROI CCD2

andrewwstephens:
This is a fairly simplistic modification to the ITC output when the GMOS ROI is set to CCD2.


bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/operation/DetectorsTransmissionVisitor.java
```diff
@@ -77,26 +81,38 @@ private void initialize(final double[][] data) {
...
-                // Apply the last transmission pixel value and go on without incrementing
+            } else {                   // else use the last transmission value and go on without incrementing
                 pixelData[i] = data[1][j - 1];
             }
         }
         detectorCcdIndexes.add(finalPixelValue - 1);
+        Log.fine("detectorCcdIndexes = " + detectorCcdIndexes);
+
+        // If the user has selected ROI = CCD2 then set the transmission of pixels in CCD1 and CCD3 to zero:
+        if (ROI == GmosCommonType.BuiltinROI.CCD2) {
```
cquiroz, line 106: There are three choices for ROI but it seems only CCD2 has a special treatment, Is that right?

andrewwstephens: Thanks for pointing this out.  Both CCD2 and CENTRAL_STAMP affect the output, and I have now included the effect of both.

cquiroz (approved):
LGTM

PR #2040, opened 2022-08-15
REL-3981:  ITC support for the GNIRS IFUs

andrewwstephens:
This is a first pass at adding ITC support for the new GNIRS IFUs.  This does *not* work in the OT, and I'm sure further tweaks will be required, but this version should allow the GNIRS team to do some initial testing.  There was a lot of copying and pasting from the NIFS ITC, so suggestions for improving things are welcome.


bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gnirs/IFUComponent.java
```diff
@@ -0,0 +1,184 @@
...
+        super(ITCConstants.LIB + "/" + Gnirs.INSTR_DIR + "/" + Gnirs.INSTR_PREFIX + slitwidth.toString() + Instrument.DATA_SUFFIX);
+        Log.fine("Constructing a radial set of IFU apertures...");
+
+        if (slitwidth.equals(SlitWidth.LR_IFU)) {
+           ifuElementSize = IFU_LR_SIZE;
+        } else if (slitwidth.equals(SlitWidth.HR_IFU)) {
+           ifuElementSize = IFU_HR_SIZE;
+        } else {
+           throw new RuntimeException("Unknown IFU");  // This should never happen
+        }
```
andrewwstephens, line 46: It seems redundant to have this same block of code in every constructor...

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gnirs/GnirsRecipe.java
```diff
@@ -62,16 +67,34 @@ public ItcImagingResult serviceResult(final ImagingResult r) {
     }
 
     public ItcSpectroscopyResult serviceResult(final SpectroscopyResult r, final boolean headless) {
-        final List<SpcChartData> dataSets = new ArrayList<SpcChartData>() {{
-            if (instrument.XDisp_IsUsed()) {
-                if (!headless) add(createGnirsSignalChart(r));
-                add(createGnirsS2NChart(r));
-            } else {
-                if (!headless) add(Recipe$.MODULE$.createSignalChart(r, 0));
-                add(Recipe$.MODULE$.createS2NChart(r));
```
andrewwstephens, line 71: I just realized that I dropped this check of the `headless` parameter.  I should probably put that back in, but I need to figure out why it only controls one plot and not both of them...

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gnirs/IFUComponent.java
```diff
@@ -0,0 +1,184 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.logging.Logger;
+
+/**
+ * This defines GNIRS IFU apertures.
+ */
+
+// TODO: Read in the correct IFU transmission data file
```
andrewwstephens, line 14: Oops, I did this already.

cquiroz (approved):
I can't really comment on the internals but we can certainly test it

PR #2041, opened 2022-08-17
REL-4013: hold the lock during sync updates

swalker2m:
This is a slightly modified version of (#2025) for the release branch.  It was changed to remove insignificant but possibly serialization-incompatible changes.


swalker2m:
I'll merge this on the theory that it was approved before (#2025).

PR #2042, opened 2022-08-19
Add optional version token for the itc json servlets

cquiroz:
To enable better caching of itc results we need to know if the data or algorithms from the ocs have changed.

This is added only at the edge on the Json servlets for itc rather than in the ITC itself to avoid breaking the TRPC protocol used by the OT


swalker2m (approved):
How does this get set?  :+1:

cquiroz:
I set it by default to empty. This is not used in the OT but will be used on the wrapping ITC server for GPP `lucuma-itc`
I will just set it to the build time. The actual value has no relevance, what I want is a value that will change if we do a new build with e.g. new data files

That change will invalidate all previous calculations that may be stored in the cache used on `lucuma-itc`

PR #2047, opened 2022-09-12
Keep obslog from blowing up

swalker2m:
Currently the obslog will quit with an exception if it encounters a GHOST observation.  I've copied the visitor instrument log setup and lightly edited it for GHOST so that it will at least continue to work.  The PR does not attempt to produce a great or even useful GHOST obslog.

Screen Shot 2022-09-12 at 11 51 14

### DIFF
--- a/bundle/edu.gemini.obslog/src/main/resources/ObsLogConfig.xml
+++ b/bundle/edu.gemini.obslog/src/main/resources/ObsLogConfig.xml
@@ -976,6 +976,18 @@
         <entry key="datasetcomments"/>
     </logEntry>
 
+    <logEntry key="GHOST">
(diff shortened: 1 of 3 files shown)


bundle/edu.gemini.obslog/src/main/resources/ObsLogConfig.xml
```diff
@@ -976,6 +976,18 @@
         <entry key="datasetcomments"/>
     </logEntry>
 
+    <logEntry key="GHOST">
```
cquiroz, line 979: I rather not ask what is this for 😃

cquiroz (approved):
LGTM, thanks a lot

PR #2051, opened 2022-10-06
uncomment negative lst handler

tpolecat:
This has been commented out forever but it's actually a case that happens and it was messing up the RA bucket calculation in ITAC. We'll see if it breaks any testcases.

Ok well a HORIZONS testcase is failing because a sample query we're using just started returning more answers. Not important, just added those to the testcase.

### DIFF
--- a/bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/util/ImprovedSkyCalcMethods.java
+++ b/bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/util/ImprovedSkyCalcMethods.java
@@ -922,9 +922,9 @@ public class ImprovedSkyCalcMethods {
             sid_g = sid_g + 1.0027379093 * ut - longit / 24.;
             sid_int = (long) sid_g;
             sid_g = (sid_g - sid_int) * 24.;
-    //        if (sid_g < 0.) {
-    //            sid_g = sid_g + 24.;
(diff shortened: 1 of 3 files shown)


bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/util/ImprovedSkyCalcMethods.java
```diff
@@ -922,9 +922,9 @@ protected static double lst(double jd, double longit) {
             sid_g = sid_g + 1.0027379093 * ut - longit / 24.;
             sid_int = (long) sid_g;
             sid_g = (sid_g - sid_int) * 24.;
-    //        if (sid_g < 0.) {
-    //            sid_g = sid_g + 24.;
-    //        }
+           if (sid_g < 0.) {
```
cquiroz, line 925: I wonder if this affects `lucuma-core`, we may have ported some of the code there

swalker2m (approved):

bundle/edu.gemini.qpt.client/src/main/java/edu/gemini/qpt/core/util/ImprovedSkyCalcMethods.java
tpolecat: It's worth a look.

PR #2053, opened 2022-10-07
Return ghost observing wavelength

cquiroz:
My understanding is that we return this in microns, It was specified as 655 nanometers


swalker2m:
> My understanding is that we return this in microns, It was specified as 655 nanometers

I think that's correct.  The comment in `ITccInstrumentSupport` says 
```
     * The wavelength should be in microns!
```

PR #2056, opened 2022-10-07
fix header parsing

tpolecat:
For HORIZONS records I had lazily assumed (falsely) that all records begin with three lines of metadata, followed by a delimiter, followed by the record data. In fact the number of metadata lines is arbitrary. So this changes the parser to handle any amount and just key on the delimiter line, as it always should have.


swalker2m (approved):
Seems good to me. :+1:

PR #2058, opened 2022-10-11
REL-4002: copy/paste incompatible asterism + instrument

swalker2m:
We developed a type tree / cardinality validity checker for science programs when we were working on program synchronization.  It works reasonably well but now we're confronted with an issue it cannot handle.  In particular, asterisms are stored in the target component but nevertheless tied to particular instruments.  The GHOST asterisms only apply to GHOST observations, and the non-GHOST single target asterism cannot be used with GHOST.  The validity checker doesn't prevent incpompatibilies because it doesn't drill into the component content but rather bases its decisions on component type alone.

This PR introduces an OT hack to prevent the worst issues that can arise.  After an observation component is pasted or added, an asterism/instrument compatibility check is performed and the asterism is adjusted if necessary.  Fixing or preventing incompatibilities at a lower level would be ideal, but risks program synchronization problems.


cquiroz (approved):
LGTM

PR #2062, opened 2022-10-11
REL-4161: format 2 decimal places in time accounting editor

swalker2m:
Shows time accounting information with 2 decimal places of precision.  (The data itself is stored in a `Duration`.)


cquiroz (approved):
LGTM

PR #2063, opened 2022-10-12
REL-4028: calibration + conditions rule

swalker2m:
Adds a new P2 Checker rule for GMOS calibrations with non-Any/Any/Any/Any conditions.


cquiroz (approved):
LGTM

PR #2064, opened 2022-10-12
REL-4202 Show Ghost targets on the TPE

cquiroz:
I'm not totally sure if these are correct. Some samples:

* Standard Single
Single normal zoom

* Standard Dual
Dual High zoom

* Standard + Sky
SkySR

* HR Single
HRIFU

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/tpe/feat/TpeGhostIFUFeature.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/tpe/feat/TpeGhostIFUFeature.scala
@@ -168,7 +168,7 @@ final class TpeGhostIfuFeature extends TpeImageFeature("GHOST", "Show the patrol
   private case object AsSky    extends IfuDisposition
 
   private def drawHighResolutionIfu(g2d: Graphics2D, p: Point2D): Unit =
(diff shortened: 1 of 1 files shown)


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/tpe/feat/TpeGhostIFUFeature.scala
```diff
@@ -494,8 +483,8 @@ object TpeGhostIfuFeature {
 
 
   // Color for IFU limts.
-  private val IfuFovColor: Color = Color.RED
-  private val PatrolRangeColor: Color = OtColor.SALMON
+  val IfuFovColor: Color = Color.RED
```
swalker2m, line 486: I think it was requested this way but the red labels make me think an error is being indicated.

swalker2m (approved):
Looks good to me.

bryanmiller:
This looks pretty good. I'm concerned that people will associate the red color with an error. How does it look to use the same pink color as for the cross-circle markers in two-target mode?

cquiroz:
Maybe it is better to match the user target with yellow
image

swalker2m:
That looks good to me

PR #2065, opened 2022-10-12
REL-4174 Added short red cam GNIRS p2 check

cquiroz:


bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/gnirs/GnirsRule.java
```diff
@@ -43,6 +43,33 @@ public IConfigMatcher getMatcher() {
...
+
+        @Override
+        public Problem check(final Config config, final int step, final ObservationElements elems, final Object state) {
+            final Camera camera = (Camera) SequenceRule.getInstrumentItem(config, InstGNIRS.CAMERA_PROP);
+            if (camera == null) return null;
+            switch (camera) {
+                case SHORT_BLUE:
+                case LONG_BLUE:
+                    return null;
+            }
```
swalker2m, line 57: There's also a `LONG_RED`, wouldn't that go in the list above?

bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/gnirs/GnirsRule.java
```diff
@@ -43,6 +43,33 @@ public IConfigMatcher getMatcher() {
...
+
+            final PixelScale ps = (PixelScale) SequenceRule.getInstrumentItem(config, InstGNIRS.PIXEL_SCALE_PROP);
+            if (ps != PixelScale.PS_015) return null;
+            final Wavelength l = (Wavelength) SequenceRule.getInstrumentItem(config, InstGNIRS.CENTRAL_WAVELENGTH_PROP);
+            if ((l == null) || l.doubleValue() <= 2.5) return null;
```
swalker2m, line 62: I'm missing something.  It seems like the `SHORT_RED` camera is available or it isn't.   If the pixel scale is `PS_015` or the central wavelength less than 2.5 does it become available?

swalker2m (commented):
The code is :+1: but I'm not sure I understand the task.

bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/gnirs/GnirsRule.java
cquiroz: My understanding is that there is no SHORt_RED component aa such but it is implicit when the pixel scale and central wavelength combine in a set of values

bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/gnirs/GnirsRule.java
cquiroz: I think you're right though we probably can remove this whole check altogether

bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/gnirs/GnirsRule.java
swalker2m: That may be so but in the OT at least, you can explicitly choose `SHORT_RED` as an option in the sequence component regardless of the pixel scale and wavelength.

cquiroz:
Thanks for the input. Now the check is considering the case short red is explicitly set on the sequence

bundle/edu.gemini.p2checker/src/main/java/edu/gemini/p2checker/rules/gnirs/GnirsRule.java
```diff
@@ -43,6 +43,32 @@ public IConfigMatcher getMatcher() {
         }
     };
 
+    private static final IConfigRule SHORT_RED_CAMERA_RULE = new IConfigRule() {
+        private static final String MESSAGE = "The short red camera is currently unavailable.";
+
+        @Override
+        public Problem check(final Config config, final int step, final ObservationElements elems, final Object state) {
+            final Camera camera = (Camera) SequenceRule.getInstrumentItem(config, InstGNIRS.CAMERA_PROP);
+            if (camera == null) return null;
```
swalker2m, line 52: My guess would have been that if the camera is `null`, _then_ we apply the pixel scale and wavelength check.  If not `null`, then the camera setting itself would prevail.  I don't know how it works though.

As is even if the camera is set explicitly to, say, short blue, we're doing the other property checks to make a determination.  Is this what they want?

cquiroz: Good question. Can you se e.g. short blue and the pixel scale and wavelength so that short red is implied?
I guess in that case the explicit value should win?
Can the camera ever be null?

swalker2m: I believe the camera setting is completely independent of the pixel scale and wavelength, in the `ocs` codebase at least.  I don't know how the hardware works.  It is usually `null` but can be explicitly set in engineering mode.  I would assume that when the camera is explicitly set, we use that and when not we use the properties but we should clarify.

swalker2m (approved):
Looks right to me, at any rate.  :+1:

PR #2067, opened 2022-10-16
REL-4153 - Fix GMOS ITC imaging overheads when xbin != ybin

andrewwstephens:
Override the xbinning for imaging mode to ensure that the overheads are calculated correctly.  This required updating baseline.txt since there is a test with GMOS imaging where xbin != ybin and therefore the old results were wrong.


cquiroz (approved):
LGTM

PR #2070, opened 2022-10-21
REL-4217: asterism-dependent PWFS1 limit

swalker2m:
GHOST PWFS1 inner search range limit for AGS is now asterism-dependent.  Single target mode uses a 4.7' limit while all others use 5.75'.  This required adding a parameter to the PWFS range limit method in order to obtain the chosen asterism.


cquiroz (approved):
LGTM

PR #2071, opened 2022-10-24
Change the offsets applied to GHOST targets for p2 calculations

cquiroz:
According ot the latest requirements we need to apply different offsets to GHOST targets only for purposes of p2 checks:

SRIFU1: p = -1mm = -1.64 arcsec, q = +1mm = 1.64 arcsec

SRIFU2: p = 0 arcsec, q = +1mm = 1.64 arcsec

HRIFU: p = 0 arcsec, q = -1mm = -1.64 arcsec

HRSKY: p = 0 arcsec, q = -1mm = -1.64 arcsec

### DIFF
--- a/app/spdb/build.sbt
+++ b/app/spdb/build.sbt
@@ -276,8 +276,8 @@ def cquiroz(version: Version) = AppConfig(
     "-Dedu.gemini.site=south"
   ),
   props = Map(
-    "edu.gemini.spdb.dir"                  -> "/Users/cquiroz/.spdb/",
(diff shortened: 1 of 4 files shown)


swalker2m (approved):
It says for example `SRIFU1SeparationOffset` but this offset will impact the field in range check in general.  If that's intended, then it looks good to me :+1:

cquiroz:
It is a bit unclear, I took it as it should be used for all p2 checks but maybe it is only for this one. I'll ask for clarification

cquiroz:
Feedback from users indicate this logic is correct. I'll merge it for testing

PR #2072, opened 2022-10-31
REL-3966 Draw HRIFU Sky position

cquiroz:
image

### DIFF
--- a/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/tpe/feat/TpeGhostIFUFeature.scala
+++ b/bundle/jsky.app.ot/src/main/scala/jsky/app/ot/tpe/feat/TpeGhostIFUFeature.scala
@@ -137,10 +137,9 @@ final class TpeGhostIfuFeature extends TpeImageFeature("GHOST", "Show the patrol
 
               // IFU2
               val p2 = pm.getLocationFromTag(s)
-              // We don't draw the SRIFU2, apparently, even though its position
-              // determines the position of the HR sky fibers
-              // drawStandardResolutionIfu(g2d, p2, Ifu2, AsSky)
-              drawHighResolutionSky(g2d, p2)
+              // We draw the SRIFU2 as Sky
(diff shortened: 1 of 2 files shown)


bundle/jsky.app.ot/src/main/scala/jsky/app/ot/tpe/feat/TpeGhostIFUFeature.scala
```diff
@@ -137,10 +137,9 @@ final class TpeGhostIfuFeature extends TpeImageFeature("GHOST", "Show the patrol
 
               // IFU2
               val p2 = pm.getLocationFromTag(s)
-              // We don't draw the SRIFU2, apparently, even though its position
-              // determines the position of the HR sky fibers
-              // drawStandardResolutionIfu(g2d, p2, Ifu2, AsSky)
-              drawHighResolutionSky(g2d, p2)
+              // We draw the SRIFU2 as Sky
+              drawStandardResolutionIfu(g2d, p2, Ifu2, AsSky)
+              //drawHighResolutionSky(g2d, p2)
```
cquiroz, line 142: I was a bit unsure whether to leave this on but it seems odd to draw two markers at the same position

sraaphorst (approved):
Looks fine to me. I guess this means we need a new test release of the OT built?

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/tpe/feat/TpeGhostIFUFeature.scala
swalker2m: Do the appear at the same place?  I thought the HR sky fibers were slightly offset.  At any rate, we can ask for feedback with this version.

cquiroz: Yes they show in the same place:
image

PR #2073, opened 2022-11-03
Apply offsets to sky positions

cquiroz:


bundle/edu.gemini.p2checker/src/main/scala/edu/gemini/p2checker/rules/ghost/GhostRule.scala
```diff
@@ -63,18 +63,38 @@ object GhostRule extends IRule {
...
+            base,
+            t1.coordinates(when).map(_.offset(SRIFU1SeparationOffset.p.toAngle, SRIFU1SeparationOffset.q.toAngle)),
+            t2.coordinates(when).map(_.offset(SRIFU2SeparationOffset.p.toAngle, SRIFU2SeparationOffset.q.toAngle)))
         case GhostAsterism.TargetPlusSky(t, s, _)               =>
-          checkBoth(ctx, node, base, t.coordinates(when).map(_.offset(SRIFU1SeparationOffset.p.toAngle, SRIFU1SeparationOffset.q.toAngle)), Some(s.coordinates))
+          checkBoth(ctx,
+            node,
+            base,
+            t.coordinates(when).map(_.offset(SRIFU1SeparationOffset.p.toAngle, SRIFU1SeparationOffset.q.toAngle)),
+            Some(s.coordinates.offset(SRIFU2SeparationOffset.p.toAngle, SRIFU2SeparationOffset.q.toAngle)))
```
swalker2m, line 82: Code looks good but remind me what the separation offset is again?  If you apply it to both IFUs for the check is that not the same as not applying it to either for the check?

swalker2m: Wait I see, they are different offsets for IFU1 and IFU2.

swalker2m (approved):
:+1:

bundle/edu.gemini.p2checker/src/main/scala/edu/gemini/p2checker/rules/ghost/GhostRule.scala
cquiroz: Yes, this PR is about applying the separations to the sky positions

PR #2080, opened 2022-11-23
REL-4229: adds PRV mode to GHOST HR asterism

swalker2m:
@cquiroz discovered that PRV mode was not being sent to the seqexec.  Upon investigation I learned that we aren't even storing it when selected.  This PR adds code to track it and puts `resolutionMode` (with value of type `ResolutionMode`) in the GHOST sequence.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/AsterismType.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/AsterismType.java
@@ -13,42 +13,72 @@ import java.util.TreeSet;
 import java.util.function.Supplier;
 
 public enum AsterismType implements DisplayableSpType {
-    Single("single", "Single Target", None.instance(), Asterism$.MODULE$::createSingleAsterism),
 
-    GhostSingleTarget("ghostSingleTarget", "Single Target",
-            new Some<>(AsterismConverters.GhostSingleTargetConverter$.MODULE$),
-            GhostAsterism$.MODULE$::createEmptySingleTargetAsterism),
+    Single(
+        "single",
+        "Single Target",
+        ResolutionMode.Standard,
+        None.instance()
+    ),
 
-    GhostDualTarget("ghostDualTarget", "Dual Target",
-            new Some<>(AsterismConverters.GhostDualTargetConverter$.MODULE$),
-            GhostAsterism$.MODULE$::createEmptyDualTargetAsterism),
+    GhostSingleTarget(
+        "ghostSingleTarget",
+         "Single Target",
+         ResolutionMode.GhostStandard,
+         new Some<>(AsterismConverters.GhostSingleTargetConverter$.MODULE$)
+    ),
 
-    GhostTargetPlusSky("ghostTargetPlusSky", "SRIFU1 Target, SRIFU2 Sky",
-            new Some<>(AsterismConverters.GhostTargetPlusSkyConverter$.MODULE$),
-            GhostAsterism$.MODULE$::createEmptyTargetPlusSkyAsterism),
+    GhostDualTarget(
+        "ghostDualTarget",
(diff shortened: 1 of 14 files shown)


bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/AsterismType.java
```diff
@@ -13,42 +13,72 @@
...
+    GhostHighResolutionTargetPlusSky(
+        "ghostHRTargetPlusSky",
+        "HRIFU Target, Sky",
+        ResolutionMode.GhostHigh,
+        new Some<>(AsterismConverters.GhostHRTargetPlusSkyConverter$.MODULE$)
+    ),
+
+    GhostHighResolutionTargetPlusSkyPrv(
+        "ghostHRTargetPlusSkyPrv",
+        "HRIFU Target, Sky (PRV)",
```
swalker2m, line 61: The "(PRV)" suffix here might not be appreciated.  I added it so that I could verify that the asterism type was correct when the "Precision Radial Velocity" resolution mode was selected.  I think it'd be fine to remove "(PRV)" though and have both of the HR asterism types be just "HRIU Target, Sky".

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/AsterismType.java
```diff
@@ -13,42 +13,72 @@
...
-                 final Supplier<Asterism> emptyAsterismCreator) {
-        this.tag                  = tag;
-        this.displayName          = displayName;
-        this.converter            = converter;
-        this.emptyAsterismCreator = emptyAsterismCreator;
+    AsterismType(
+        String tag,
+        String displayName,
+        ResolutionMode resolutionMode,
+        Option<AsterismConverters.AsterismConverter> converter
```
cquiroz, line 76: No creator needed anymore?

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/ResolutionMode.java
```diff
@@ -3,38 +3,25 @@
...
-    ),
-    GhostHigh("ghostHigh", "High Resolution",
-            Collections.singletonList(
-                    AsterismType.GhostHighResolutionTargetPlusSky
-            )),
-    GhostPRV("ghostPRV", "Precision Radial Velocity",
-            Collections.singletonList(
-                    AsterismType.GhostHighResolutionTargetPlusSky
-            ));
+    Standard("standard", "Standard Resolution"),
```
cquiroz, line 11: What's the difference between `Standard` and `GhostStandard`? Or is `Standard` legacy now?

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/AsterismType.java
swalker2m: I searched and it was not being used so I removed it.

cquiroz (approved):
LGTM
Do we need to do something to import existing programs?

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/ResolutionMode.java
swalker2m: I don't know, that's a good question.  I assume it just distinguishes GHOST from the others.

swalker2m:
> Do we need to do something to import existing programs?

I'm glad you asked!  We need a migration to add the `prv` parameter to old GHOST HR observations.  I've added that now.  Note there was no way to store PRV mode before so any existing GHOST HR observations are not-PRV.

astroquackers:
Thanks for figuring this out!

PR #2081, opened 2022-11-23
REL-4230: add g and v mags to ghost sequence

swalker2m:
Adds target magnitudes in g and v bands to GHOST sequence.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostCB.scala
+++ b/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostCB.scala
@@ -3,13 +3,18 @@ package edu.gemini.spModel.gemini.ghost
 import java.time.Instant
 import java.util.{Map => JMap}
 import edu.gemini.pot.sp.ISPObsComponent
+import edu.gemini.pot.sp.ISPObservation
 import edu.gemini.spModel.config.AbstractObsComponentCB
+import edu.gemini.spModel.core.Magnitude
+import edu.gemini.spModel.core.MagnitudeBand
 import edu.gemini.spModel.data.config._
 import edu.gemini.spModel.gemini.ghost.GhostAsterism.GhostTarget
 import edu.gemini.spModel.obscomp.InstConstants
 import edu.gemini.spModel.seqcomp.SeqConfigNames
+import edu.gemini.spModel.target.env.Asterism
 import edu.gemini.spModel.target.obsComp.TargetObsComp
 import edu.gemini.spModel.target.{SPCoordinates, SPSkyObject, SPTarget}
+import edu.gemini.spModel.util.SPTreeUtil
 
 import scala.collection.JavaConverters._
 import scalaz.Scalaz._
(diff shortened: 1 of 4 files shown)


bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostCB.scala
```diff
@@ -161,6 +166,42 @@ final class GhostCB(obsComp: ISPObsComponent) extends AbstractObsComponentCB(obs
...
     }
   }
+}
+
+object GhostCB {
+
+  val MagProperties: Map[MagnitudeBand, String] = Map(
+    MagnitudeBand._g -> Ghost.MAG_G_PROP,
+    MagnitudeBand.V  -> Ghost.MAG_V_PROP
+  )
```
swalker2m, line 185: @cquiroz  Are these the right bands?

cquiroz: Is that SloanG, I think we wanted Gaia G but we don't have that on the OT yet

cquiroz: Will these be stored in the map under a single key?

bundle/edu.gemini.pot/src/test/scala/edu/gemini/spModel/gemini/ghost/MagTest.scala
```diff
@@ -0,0 +1,72 @@
...
+    GhostTarget(sp, GuideFiberState.Enabled)
+  }
+
+  private def verifyMagProp(
+    magG: Option[Double],
+    magV: Option[Double]
+  ): Unit = {
+
+    val cs   = ConfigBridge.extractSequence(getObs, null, ConfigValMapInstances.IDENTITY_MAP)
+    val gKey = new ItemKey(new ItemKey(INSTRUMENT_CONFIG_NAME), Ghost.MAG_G_PROP)
```
cquiroz, line 41: Ah here's the answer to my question before

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostCB.scala
swalker2m: Yes, Sloan G.  Is this task still useful if we don't have Gaia G?  These are science targets not guide stars in any case.

cquiroz: My understanding is that we'll add Gaia G in the near future, I'll confirm if Sloan G is useful

cquiroz: From user feedback::
`The SloanG is a passable alternative to Gaia G, but V would be preferred.`

I think we should do V or else Sloan G and revesit when we have Gaia G

swalker2m: Sounds good, both will be added if both are known.

swalker2m: Should we go ahead with this then?  If we later add a new band it will be easy to swap in here.

cquiroz: Yes, let's go ahead

cquiroz (approved):
LGTM

PR #2082, opened 2022-11-24
REL-4228: fix p2 check

swalker2m:
Switches the GMOS calibration / conditions rule to work at the observation level instead of step-by-step.


cquiroz (approved):
LGTM

PR #2090, opened 2022-11-29
Feature/REL-3981:  ITC support for the GNIRS IFUs

andrewwstephens:
This fixes the issue reported with the GNIRS IFU calculations, namely taking into account the grating order and throughput.

While comparing with long-slit calculations I realized that the interpolation of the slit throughput was introducing significant errors when using a narrow slit compared to the seeing, so I added some lines to the lookup table to improve that.  It is this change that required a new baseline.txt.


cquiroz (approved):
LGTM

PR #2094, opened 2022-12-12
Dumb interface to call itc with plain strings

cquiroz:
Sorry about this


swalker2m (approved):
:+1: Nothing beats a `String`.

cquiroz:
They are the best. why did we stop using them?

PR #2096, opened 2022-12-28
Make PM calculations match lucuma-core

cquiroz:


cquiroz:
The current calculations of PM on the ocs are incorrect. This PR fixes several mistakes on conversion and a missing `cos(dec)` factor. It now matches the code on lucuma-core and includes some test cases

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/ProperMotion.scala
```diff
@@ -78,13 +75,17 @@ object ProperMotion {
...
                              parallax: Double,
                              elapsedYears: Double): Coordinates = {
     // We want the baseCoordinates in radians.
-    val (ra,  dec)  = (baseCoordinates.ra.toAngle.toSignedDegrees.toRadians, baseCoordinates.dec.toAngle.toSignedDegrees.toRadians)
-    val (dRa, dDec) = (properVelocity.p.toAngle.toSignedDegrees.toRadians,   properVelocity.q.toAngle.toSignedDegrees.toRadians)
+    val (ra,  dec)  = (baseCoordinates.ra.toAngle.toRadians, baseCoordinates.dec.toAngle.toRadians)
+    // at dec 90 this gets undetermined, while we never observe at dec 90 it wouldn't be nice to have
+    // a division by zero error. At dec 90 we don't correct by cos(dec)
+    val (dRa, dDec) = 
+      if (cos(dec) == 0.0) (properVelocity._1.toRadians, properVelocity._2.toRadians)
```
swalker2m, line 87: Looking below, `pos` is `(0,0,1)` at 90º.  The `dRa` is used to calculate the `dPos2` vector which in this case multiplies `dRa` by 0.  I don't know what the right answer is at zero but effectively it seems we could put anything here as `dRa` in this case.

swalker2m (approved):
Thanks for digging into this!  Looks great. :+1:

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/ProperMotion.scala
cquiroz: Right, we could revisit to try matching what astropy gives you

PR #2098, opened 2022-12-29
Fix "duplicate Cc" bounce

swalker2m:
Adding recipients in separate calls produces a `Cc` header per call apparently, which Google is now rejecting:

```
Bounced
550-5.7.1 [209.85.220.97] This message is not RFC 5322 compliant, the issue is: 550-5.7.1 duplicate CC headers. To reduce the amount of spam sent to Gmail, 550-5.7.1 this message has been blocked. Please review 550 5.7.1 RFC 5322 specifications for more information. s144-20020a1f2c96000000b003be18067bcasor4079229vks.5 - gsmtp
```

This is an attempt to fix the problem but adding all recipients in a single call.

### DIFF
--- a/bundle/edu.gemini.oodb.too.window/src/main/java/edu/gemini/too/email/TooEmail.java
+++ b/bundle/edu.gemini.oodb.too.window/src/main/java/edu/gemini/too/email/TooEmail.java
@@ -320,13 +320,13 @@ public class TooEmail {
 
         try {
             if (toAddrs.length > 0) {
-                mess.addRecipients(Message.RecipientType.TO, toAddrs);
+                mess.setRecipients(Message.RecipientType.TO, toAddrs);
             }
             if (ccAddrs.length > 0) {
-                mess.addRecipients(Message.RecipientType.CC, ccAddrs);
+                mess.setRecipients(Message.RecipientType.CC, ccAddrs);
             }
             if (bccAddrs.length > 0) {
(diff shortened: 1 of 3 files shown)


cquiroz:
Are we able to test this on a test instance?

bundle/edu.gemini.oodb.too.window/src/main/java/edu/gemini/too/email/TooEmail.java
```diff
@@ -320,13 +320,13 @@ public Message createMessage(ISPObservation obs)  {
...
             if (toAddrs.length > 0) {
-                mess.addRecipients(Message.RecipientType.TO, toAddrs);
+                mess.setRecipients(Message.RecipientType.TO, toAddrs);
             }
             if (ccAddrs.length > 0) {
-                mess.addRecipients(Message.RecipientType.CC, ccAddrs);
+                mess.setRecipients(Message.RecipientType.CC, ccAddrs);
             }
             if (bccAddrs.length > 0) {
-                mess.addRecipients(Message.RecipientType.BCC, bccAddrs);
```
cquiroz, line 329: Seems obvious 😝

cquiroz (approved):
LGTM

swalker2m:
> Are we able to test this on a test instance?

I'm planning to put it on `g[ns]odbtest` and see what happens.

PR #2099, opened 2022-12-29
REL-4170: GHOST Template

swalker2m:
Adds support for GHOST templates, according to the rules in `GHOST_BP.txt`.  The challenge with this template is that the resolution mode and target mode (taken together, the asterism type) is set at Phase 1 and we want to preserve this through instantiation.  Unfortunately this information is kept in the selected asterism in the target node, but there is no target node in the template group.  To compensate, a `preferredAsterismType` is added to the instrument component and overridden by GHOST in particular.  When there is no target node the instrument node's preferred asterism type is consulted instead.

### DIFF
--- a/bundle/edu.gemini.auxfile.workflow/src/main/java/edu/gemini/auxfile/copier/osgi/OsgiCopyConfig.java
+++ b/bundle/edu.gemini.auxfile.workflow/src/main/java/edu/gemini/auxfile/copier/osgi/OsgiCopyConfig.java
@@ -2,6 +2,7 @@ package edu.gemini.auxfile.copier.osgi;
 
 import edu.gemini.auxfile.copier.AuxFileType;
 import edu.gemini.auxfile.copier.CopyConfig;
+import edu.gemini.shared.util.immutable.ImOption;
 import edu.gemini.spModel.core.SPProgramID;
 import edu.gemini.util.ssh.SshConfig$;
 import org.osgi.framework.BundleContext;
@@ -19,9 +20,11 @@ final class OsgiCopyConfig extends CopyConfig {
     private static final String USER_KEY      = "user";
     private static final String PASSWORD_KEY  = "password";
     private static final String DEST_KEY      = "dest";
+    private static final String ENABLED_KEY   = "enabled";
 
-    private AuxFileType _fileType;
-    private String _destTmpl;
+    private final AuxFileType _fileType;
+    private final String _destTmpl;
+
 
     public static OsgiCopyConfig create(AuxFileType type, BundleContext ctx){
         try {
@@ -38,7 +41,9 @@ final class OsgiCopyConfig extends CopyConfig {
 	    super(getProperty(type, ctx, HOST_KEY),
 	          getProperty(type, ctx, USER_KEY),
 	          getProperty(type, ctx, PASSWORD_KEY),
-              SshConfig$.MODULE$.DEFAULT_TIMEOUT());
+              SshConfig$.MODULE$.DEFAULT_TIMEOUT(),
+              ImOption.apply(ctx.getProperty(buildPropertyName(type, ENABLED_KEY))).forall("true"::equals)
+              );
         _fileType = type;
         _destTmpl = getProperty(type, ctx, DEST_KEY);
     }
@@ -50,7 +55,7 @@ final class OsgiCopyConfig extends CopyConfig {
         final String cfg = ctx.getProperty(cfgProp);
         final String def = ctx.getProperty(defProp);
 
-        if        (cfg != null) {
+        if (cfg != null) {
             // this aux file type has a specific config, use it
             LOG.info("config for " + cfgProp + ": " + cfg);
             return cfg;
--- a/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/dtree/Root.scala
+++ b/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/dtree/Root.scala
@@ -44,6 +44,6 @@ object Root {
 
   // REL-3769: GPI not available in 2020B.
   // GHOST not yet available
-  val choices: List[Instrument] = List(Alopeke, Flamingos2, /*Ghost,*/ GmosNorth, GmosSouth, Gnirs, /*Graces,*/ Gsaoi, Igrins, MaroonX, Nifs, /*Niri,*/ Visitor, Zorro)
+  val choices: List[Instrument] = List(Alopeke, Flamingos2, Ghost, GmosNorth, GmosSouth, Gnirs, /*Graces,*/ Gsaoi, Igrins, MaroonX, Nifs, /*Niri,*/ Visitor, Zorro)
 
 }
--- a/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/dtree/Root.scala
+++ b/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/dtree/Root.scala
@@ -44,6 +44,6 @@ object Root {
 
   // REL-3769: GPI not available in 2020B.
   // GHOST not yet available
-  val choices: List[Instrument] = List(Alopeke, Flamingos2, Ghost, GmosNorth, GmosSouth, Gnirs, /*Graces,*/ Gsaoi, Igrins, MaroonX, Nifs, /*Niri,*/ Visitor, Zorro)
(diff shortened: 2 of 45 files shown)


bundle/edu.gemini.auxfile.workflow/src/main/java/edu/gemini/auxfile/copier/osgi/OsgiCopyConfig.java
```diff
@@ -38,7 +41,9 @@ private OsgiCopyConfig(AuxFileType type, BundleContext ctx) {
 	    super(getProperty(type, ctx, HOST_KEY),
 	          getProperty(type, ctx, USER_KEY),
 	          getProperty(type, ctx, PASSWORD_KEY),
-              SshConfig$.MODULE$.DEFAULT_TIMEOUT());
+              SshConfig$.MODULE$.DEFAULT_TIMEOUT(),
+              ImOption.apply(ctx.getProperty(buildPropertyName(type, ENABLED_KEY))).forall("true"::equals)
```
swalker2m, line 45: I added an `enabled` bundle to property to make it possible to skip file sftps. This greatly simplifies testing. By default copying is enabled so it should not impact production.

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/dtree/Root.scala
```diff
@@ -44,6 +44,6 @@ object Root {
 
   // REL-3769: GPI not available in 2020B.
   // GHOST not yet available
-  val choices: List[Instrument] = List(Alopeke, Flamingos2, /*Ghost,*/ GmosNorth, GmosSouth, Gnirs, Graces, Gsaoi, Igrins, MaroonX, Nifs, Niri, Visitor, Zorro)
+  val choices: List[Instrument] = List(Alopeke, Flamingos2, Ghost, GmosNorth, GmosSouth, Gnirs, Graces, Gsaoi, Igrins, MaroonX, Nifs, Niri, Visitor, Zorro)
```
swalker2m, line 47: Maybe we need to turn this back off for now?

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/target/env/AsterismType.java
```diff
@@ -91,24 +93,51 @@ public String displayValue() {
...
+
+    public static AsterismType forObservation(ISPObservation obs) {
+        return ImOption
+            .apply(SPTreeUtil.findTargetEnvNode(obs))
+            .map(n -> forTargetEnvironment(((TargetObsComp) n.getDataObject()).getTargetEnvironment()))
+            .orElse(ImOption
+                .apply(SPTreeUtil.findInstrument(obs))
+                .filter(n -> n.getDataObject() instanceof SPInstObsComp)
+                .map(n -> ((SPInstObsComp) n.getDataObject()).getPreferredAsterismType())
+            ).getOrElse(AsterismType.Single);
```
swalker2m, line 141: This is where we figure out the asterism type, checking first the target environment and then falling back to the "preferred" type recorded in the instrument.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/template/TemplateGroup.java
```diff
@@ -103,18 +63,6 @@ public void setGroupType(GroupType groupType) {
...
-    }
-
-    public void setStatus(Status status) {
-        if (this.status != status) {
-            final Status prev = this.status;
-            this.status = status;
-            firePropertyChange(PROP_STATUS, prev, status);
-        }
-    }
-
```
swalker2m, line 117: `Status` was unused so I removed it.  In theory this should be fine.  😬

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/env/Asterism.scala
```diff
@@ -190,4 +194,71 @@ object Asterism {
   def createSingleAsterism: Single = {
     Single(new SPTarget())
   }
+
+  def fromTypeAndTemplateTarget(
+    asterismType: AsterismType,
+    target:       SPTarget
+  ): Asterism = {
```
swalker2m, line 201: We have to create a proper GHOST asterism but have only one target from Phase 1.  This method compensates by adding and subtracting from the P1 target.  Presumably the PI will fill in the proper values during the phase 2 process after instantiation.

bundle/jsky.app.ot/src/main/scala/jsky/app/ot/gemini/ghost/GhostEditor.scala
```diff
@@ -628,6 +629,10 @@ final class GhostEditor extends ComponentEditor[ISPObsComponent, Ghost] {
       case SelectionChanged(`asterismComboBox`) => Swing.onEDT {
         val converter = asterismComboBox.selection.item.converter.asScalaOpt
         converter.foreach(convertAsterism)
+
+        val dataObj = getDataObject
+        dataObj.setPreferredAsterismType(asterismComboBox.selection.item)
+        getNode.setDataObject(dataObj)
```
swalker2m, line 635: The editor now records the preferred asterism type as well.

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/dtree/Root.scala
cquiroz: It was uncommented already for PIT 2023B

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/env/Asterism.scala
cquiroz: Would it make sense to generate a second target for dual-target proposal to a given offset to the one in the proposal?

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/obscomp/SPInstObsComp.java
```diff
@@ -335,34 +336,6 @@ public void setParamSet(ParamSet paramSet) {
         }
     }
 
-    /**
-     * Are two SPInstObsComp objects are equal.
-     */
-    /*  equals / hashCode for mutable objects is not well defined and this
-        is problematic anyway because it is not defined in the subclasses
-    public boolean equals(Object obj) {
```
cquiroz, line 343: Why was equals deleted?

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/env/Asterism.scala
cquiroz: Oh, I see you're already doing that

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/target/env/Asterism.scala
```diff
@@ -190,4 +194,71 @@ object Asterism {
...
+    asterismType match {
+      case AsterismType.Single                              =>
+        Single(target)
+
+      case AsterismType.GhostSingleTarget                   =>
+        GhostAsterism.SingleTarget(ghostTarget(target), None)
+
+      case AsterismType.GhostDualTarget                     =>
+        GhostAsterism.DualTarget(
+          ghostTarget(targetPlus(1)),
```
cquiroz, line 228: I guess this means we generate two targets and the one from the proposal becomes the base at the center?

cquiroz (approved):
LGTM as far as the bits I can actually understand

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/dtree/Root.scala
swalker2m: @cquiroz We now want to leave GHOST commented out correct?

cquiroz: Yes, it is commented out

bryanmiller: We are going to want to release the GHOST P2 templates in the June release, supporting GHOST in the _2023A_ PIT. There will also probably be a new 2023B PIT release later in the year with GHOST available. I will be updating the GHOST template rules soon.

PR #2100, opened 2022-12-29
Make PM calculations match lucuma-core

swalker2m:
Carlos' fixes from #2096 for the release branch.  We decided that since

a) we don't use PM other than in GHOST, and
b) it was wrong anyway

we might as well fix this as well if we are making a patch release.


cquiroz:
Thanks

cquiroz (approved):
LGTM

swalker2m:
Merged with a failing test case, which is unrelated: a timeout in catalog tests.

PR #2109, opened 2023-01-18
REL-4212 Allow open editable copies in most cases

cquiroz:


bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/view/submit/SubmitView.scala
```diff
@@ -175,11 +175,12 @@ class SubmitView(ph: ProblemRobot, newShellHandler: (Model,Option[File]) => Unit
       override def refresh(m:Option[SubmitStatus]): Unit = {
         val wasRolled = panel.model map rolledLens.get
         enabled = ~m.map {
-          case _ if tac => false
-          case Partial  => true
-          case Success  => true
+          case _ if tac   => false
+          case Partial    => true
+          case Success    => true
+          case Incomplete => true
```
swalker2m, line 181: What does an `Incomplete` submission status mean?

cquiroz: It means we are missing some data. I'm not really sure why we want to forbid opening a copy in that case but it seems to be a problem in some cases. In particular when you upconvert and e.g. are missing the attachments

swalker2m (approved):
Seems good to me :+1:

PR #2123, opened 2023-03-01
REL-3981: Update GNIRS LR-IFU throughput

andrewwstephens:
This is a trivial PR to update the GNIRS low-resolution IFU throughput with measured values.


cquiroz:
horizons has been flaky lately. I think we could merge this even with that test failing

cquiroz (approved):
LGTM

PR #2125, opened 2023-03-01
Ghost ITC

framos-gemini:
This is the first ITC Ghost draft version. All the ITC tests were passed well.


.Trash-3906/files/Ghost.scala
```diff
@@ -0,0 +1,40 @@
+package edu.gemini.model.p1.dtree.inst
```
swalker2m, line 1: Did you mean to include `.Trash-3906`?

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/calunit/smartgcal/keys/ConfigKeyGhost.java
```diff
@@ -0,0 +1,96 @@
...
+import edu.gemini.shared.util.immutable.ImList;
+import edu.gemini.spModel.gemini.calunit.smartgcal.ConfigurationKey;
+import edu.gemini.spModel.gemini.ghost.GhostType;
+import edu.gemini.spModel.gemini.gmos.InstGmosSouth;
+
+/**
+ * Base class for representation of GMOS instrument configurations for calibration lookups.
+ * Instances of this class will be used as lookup keys in the corresponding calibration map.
+ */
+public class ConfigKeyGhost implements ConfigurationKey {
```
swalker2m, line 13: Does ITC need Smart GCal?

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/ghost/GhostReadoutTime.java
```diff
@@ -0,0 +1,189 @@
...
+
+import edu.gemini.spModel.config2.Config;
+import edu.gemini.spModel.gemini.ghost.GhostType.*;
+import edu.gemini.spModel.gemini.ghost.GhostType.DetectorManufacturer;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/** Maps Ghost instrument parameters to readout times */
+public class GhostReadoutTime {
```
swalker2m, line 15: As we discussed, the readout time calculation is available in `edu.gemini.spModel.gemini.ghost.GhostCameras`.  Point me to where this class is used and I will see if I can help switch over to that?  It would be unfortunate to have two parallel readout time calculations to keep in sync so we want to avoid that.  On the other hand, if this is meant to be ITC-specific and not for general planned time calculations we can move it over to the ITC module.

bundle/project/build.properties
```diff
@@ -0,0 +1 @@
+sbt.version=1.6.2
```
cquiroz, line 1: We probably don't need this file checked in

jre.sbt
```diff
@@ -0,0 +1 @@
+ocsJreDir in ThisBuild := file("/Users/framos/jres8")
```
cquiroz, line 1: Same here this is a user specific file. We may want to add this to `.gitiggnore`

bundle/project/build.properties
swalker2m, line 1: This would seem to clash with the definition in `project/build.properties`.  Is it used by anything?  I'm skeptical that the build will work with a newer version -- it's kind of complicated by OSGi.

swalker2m (changes requested):
Thanks for putting this PR together!   I skipped-past the ITC-specific stuff since I'm not familiar with the ITC.  The main issue from my perspective is the duplicate Ghost model.  We will want to remove that in favor of the existing Ghost code.  I think the way forward might be for Carlos or I to get this branch, remove the `GhostReadoutTime`, `GhostType`, and `InstGhost` and then see what needs to be done to make the code compile with the existing model.

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/dtree/Root.scala
```diff
@@ -44,6 +44,6 @@ object Root {
 
   // REL-3769: GPI not available in 2020B.
   // GHOST not yet available
-  val choices: List[Instrument] = List(Alopeke, Flamingos2, /*Ghost,*/ GmosNorth, GmosSouth, Gnirs, /*Graces,*/ Gsaoi, Igrins, MaroonX, Nifs, /*Niri,*/ Visitor, Zorro)
```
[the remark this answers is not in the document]
cquiroz: Maybe this needs to be rebased to get the latest changes

cquiroz:
This bundle/edu.gemini.itc/src/main/resources/gnirs/.gnirs_gratings.dat.swo should probably be deleted

cquiroz:
Thanks for the PR. It was mentioned this was producing the results as expected. Do we have tests for them?
It seems this is a good start but I'd like to request a few steps before we go:
* It seems the PR was based on an older commit of `develop` it would be better to rebase this PR against the latest `develop` commit.
* Remove the duplicated Ghost model and use the one actually being used by the OT (Ghost scala model)
* It would be great to include a test suite. I'm not sure how realistic is that given the time frame

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/dtree/Root.scala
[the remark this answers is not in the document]
framos-gemini: As we have been talking about that change was not made by me although it may seem like it was. This was due to the process by which I made the merge. The solution is the one discussed with Carlos and would be to start from the HEAD of the development branch and from there include the ITC changes. I am going to make a new pull request following the steps mentioned above. 

e3da10496d (Carlos Quiroz 2023-02-21 16:00:25 -0300 47)   val choices: List[Instrument] = List(Alopeke, Flamingos2, /*Ghost,*/ GmosNorth, GmosSouth, Gnirs, /*Graces,*/ Gsaoi, Igrins, MaroonX, Nifs, /*Niri,*/ Visitor, Zorro)

cquiroz:
Superseded by #2126

PR #2126, opened 2023-03-10
First Ghost version with some java and scala code duplicate

framos-gemini:
All the tests passed well. 
This is the first Ghost ITC version.
As discussed, this version has some duplication of code between Java and Scala. We will work on uploading the final version. The goal of this version is to have a release for Ghost for science


bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/ghost/Ghost.java
```diff
@@ -0,0 +1,358 @@
...
+public final class Ghost extends Instrument implements BinningProvider, SpectroscopyInstrument {
+
+    private static final Logger Log = Logger.getLogger(Ghost.class.getName());
+
+    public static final String INSTR_DIR = "ghost";
+
+    public static final double PLATE_SCALE = 1.64;  // arcsec/mm
+
+    private static final double WellDepth = 350000;  // VENU has to confirm the correct value.
+    //private final TransmissionElement _blazeThrougthput;
```
cquiroz, line 30: Commented code

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/ghost/Ghost.java
```diff
@@ -0,0 +1,358 @@
...
+ */
+public final class Ghost extends Instrument implements BinningProvider, SpectroscopyInstrument {
+
+    private static final Logger Log = Logger.getLogger(Ghost.class.getName());
+
+    public static final String INSTR_DIR = "ghost";
+
+    public static final double PLATE_SCALE = 1.64;  // arcsec/mm
+
+    private static final double WellDepth = 350000;  // VENU has to confirm the correct value.
```
cquiroz, line 29: Does this have units?

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/ghost/Ghost.java
```diff
@@ -0,0 +1,358 @@
...
+    public Ghost(final GhostParameters gp, final ObservationDetails odp, final GhostType.DetectorManufacturer ccdColor) {
+        super(Site.GS, Bands.VISIBLE, INSTR_DIR, FILENAME);
+
+        Log.fine("Resolution: " + gp.resolution());
+        Log.fine("ccd_color: " + ccdColor.displayValue());
+        this.odp    = odp;
+        this.gp     = gp;
+
+        _ccdColor = ccdColor;
+        _detector = new Detector(getDirectory() + "/", getPrefix(),  _ccdColor.getManufacter(), _ccdColor.getManufacter());
```
cquiroz, line 88: Not sure we need the `"/"` at the end

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/ghost/Ghost.java
```diff
@@ -0,0 +1,358 @@
...
+
+        _detector.setDetectorPixels(_ccdColor.getXsize());
+        addComponent(_detector);
+        _cableThrougthput = new TransmissionElement(getDirectory()+"/"+ Ghost.INSTR_PREFIX + "cable" +getSuffix());
```
cquiroz, line 92: The file path could be constructed by the system, the slash there is system dependent

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/ghost/Ghost.java
```diff
@@ -0,0 +1,358 @@
...
+    /**
+     * Returns the subdirectory where this instrument's data files are.
+     */
+    public String getDirectory() {
+        return ITCConstants.LIB + "/" + INSTR_DIR;
+    }
+
+    @Override
+    public double wellDepth() {
+        Log.info("TODO. VENU has to give this information. Not implemented yet Ghost.java wellDepth ");
```
cquiroz, line 170: Above it was indicated that the value was confirmed, is it?

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/ghost/Ghost.java
```diff
@@ -0,0 +1,358 @@
...
+                    return 0.58;
+                return 0.53;
+            default:
+                Log.warning("Bad definition of the readMode");
+                return 0;
+        }
+    }
+
+    public double getPixelSize() {
+        //return (double) super.getPixelSize() * gp.spectralBinning().getValue();
```
cquiroz, line 198: Commented code

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/ghost/GhostGratingOptics.java
```diff
@@ -0,0 +1,30 @@
...
+    public double getStart() {
+        return get_trans().getStart();
+    }
+
+    public double getEnd() {
+        return get_trans().getEnd();
+    }
+
+    public double getPixelWidth() {
+        //return dispersion(-1) * _spectralBinning;
```
cquiroz, line 27: Commented code

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/ghost/IFUComponent.java
```diff
@@ -0,0 +1,108 @@
...
+
+/**
+ * This is the specific setup for the IFU of GMOS.
+ * It uses Hexagonal Apertures that can be in a line radialy
+ * from an arbitrary point in a gaussian.
+ */
+
+public final class IFUComponent extends TransmissionElement {
+
+    //public static final double IFU_DIAMETER_HR = 0.144;   // The apothem value is 0.072 mm.
```
cquiroz, line 20: Commented code

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/ghost/IFUComponent.java
```diff
@@ -0,0 +1,108 @@
...
+
+    private final double RADIUS_HR=0.3409428;
+
+    private final double RADIUS_SR=0.3409428;
+
+    private GhostType.Resolution _res;
+
+    public IFUComponent(GhostType.Resolution res) {
+
+        super(ITCConstants.LIB + "/" + Ghost.INSTR_DIR + "/" +Ghost.INSTR_PREFIX + "ifu" + Instrument.DATA_SUFFIX);
```
cquiroz, line 41: As before this could be constructed using `Path` or `File`

bundle/edu.gemini.itc/src/main/scala/edu/gemini/itc/base/DatFile.scala
```diff
@@ -101,8 +102,23 @@ object DatFile {
...
       val resolvingPower = s.nextInt()
       val resolution     = s.nextDouble()
-      val dispersion     = s.nextDouble()
-      l.+=(Grating(name, resolvingPower, blaze, dispersion, resolution))
+      //val dispersion     = s.nextDouble()
+      val tmp     = s.next()
+      var dispersion: Double = 0
+      try {
+        dispersion = tmp.toDouble
+        l.+=(Grating(name, resolvingPower, blaze, dispersion, null, resolution))
```
cquiroz, line 110: This could be `l += Grating(...`

bundle/edu.gemini.itc/src/main/scala/edu/gemini/itc/base/DatFile.scala
```diff
@@ -101,8 +102,23 @@ object DatFile {
...
+      } catch {
+        case e : NumberFormatException => {
+          val dDis = arrays.apply(tmp)
+          l.+=(Grating(name, resolvingPower, blaze, -999999, dDis, resolution))
+        }
+        case e : Exception => {
+          //l.+=(Grating(name, resolvingPower, blaze, -999999, arrays.apply(tmp), resolution))
+          println("NOT ANALYZED THE EXCEPTION, PLEASE CONTACT WITH A PROGRAMMER. LINE 125 OF THE DatFile.scala")
```
cquiroz, line 118: Do we want to put this message in std out? Is the message useful?

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/ghost/GhostReadoutTime.java
```diff
@@ -0,0 +1,49 @@
...
+     * @param config the current configuration
+     */
+
+    public static double getReadoutOverhead(Config config) {
+        final GhostType.ReadMode readMode                = (GhostType.ReadMode) config.getItemValue(GhostType.ReadMode.KEY);
+        final GhostType.DetectorManufacturer detMan      = (GhostType.DetectorManufacturer) config.getItemValue(GhostType.DetectorManufacturer.KEY);
+        final GhostType.Binning              xBin        = (GhostType.Binning) config.getItemValue(InstGhost.X_BIN_KEY);
+        final GhostType.Binning              yBin        = (GhostType.Binning) config.getItemValue(InstGhost.Y_BIN_KEY);
+        return calcReadOut(xBin.getValue(),yBin.getValue(), readMode.getReadRate())/1000000.0;
+        /*System.out.println("readmode displayValue" + readMode.displayValue());
```
cquiroz, line 40: Commented code

bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/ConfigExtractor.scala
```diff
@@ -142,6 +142,30 @@ object ConfigExtractor {
...
+
+    for {
+      readMode      <- extract[ReadMode]      (c, ReadModeKey)
+      specBin       <- extract[Binning]       (c, CcdXBinKey)
+      spatBin       <- extract[Binning]       (c, CcdYBinKey)
+      resolution    <- extract[Resolution]    (c, InsResolution)
+      nSkyMicrolens <- extract[Int]           (c, InsNskyMicrolens)
+      wavelen       <- extractObservingWavelength(c)
+    } yield {
+      var gain = extractGain(readMode);
```
cquiroz, line 164: This could be a `val`

bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/ConfigExtractor.scala
```diff
@@ -142,6 +142,30 @@ object ConfigExtractor {
...
+      GhostParameters(wavelen, nSkyMicrolens, resolution, gain, readMode, spatBin, specBin);
```
cquiroz, line 165: The semicolon isn't needed

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/ghost/GhostReadoutTime.java
cquiroz: About this calculation.
On the seqexec we use a lookup table to calculate readout overheads, and in fact we produce two values, one per detector.
Which one should be used?

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/ghost/IFU_Trans.java
```diff
@@ -0,0 +1,88 @@
...
+                if (_fwhm < 1.55)
+                    return polyVal(POLY_SR, _fwhm) * SR_IFU_tx;
+
+                if (_fwhm < 3)
+                    return polyVal(POLY_SR_bad, _fwhm) * SR_IFU_tx;
+
+                double alpha = 0.701 * _fwhm;     // for really bad seeing use integration in equiv circle
+                return (1 - Math.pow(1 + Math.pow((0.333 / alpha), 2), -3)) * SR_IFU_tx;
+            }
+            case HIGH:
```
cquiroz, line 60: What about PRV

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/ghost/Ghost.java
```diff
@@ -0,0 +1,358 @@
...
+    public GhostSaturLimitRule getGhostSaturLimitWarning() {
+        return _ghostSaturLimitWarning;
+    }
+
+    @Override
+    public double getSlitWidth() {
+        switch (gp.resolution()) {
+            case STANDARD:
+                return  0.32;
+            case HIGH:
```
cquiroz, line 261: What about PRV?

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/ghost/Ghost.java
```diff
@@ -0,0 +1,358 @@
...
+
+    public double getSlitLength() {
+        double slitLength=0.0;
+
+        switch (gp.resolution()) {
+            case STANDARD:
+                slitLength = GhostType.SIZE_ONE_FIBER_SR_PIXELS * 7;  // SIZE_ONE_FIBER_SR_PIXELS = 2.7
+                break;
+            case HIGH:
+                slitLength = GhostType.SIZE_ONE_FIBER_HR_PIXELS * 19;  // SIZE_ONE_FIBER_HR_PIXELS = 1.62
```
cquiroz, line 277: What about PRV?

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/obscomp/SPInstObsComp.java
```diff
@@ -435,24 +435,25 @@ public Instrument getInstrument() {
     // REL-1678: 7 seconds DHS write overhead
     // REL-1934: 10 seconds DHS write overhead for F2
 
-    public static final Map<Instrument, PlannedTime.CategorizedTime> DHS_WRITE_TIMES;
+    public static final Map<Instrument, PlannedTime.CategorizedTime> WRITE_TIMES;
```
cquiroz, line 438: Will this affect serial version? @swalker2m

swalker2m: Yeah I think this breaks serialization.  Do we need to deploy it before the June release in the production ODB?  If so we have to think about this and probably make some changes.  Otherwise, it's fine since we break serialization with every major release anyway and rely on XML migration.

swalker2m: Hmm, well I guess this is `static` so maybe not an issue.  We'll need to test carefully though in general if this is to go into production before June.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostCamera.scala
```diff
@@ -22,7 +22,7 @@ sealed trait GhostCamera extends Product with Serializable {
     oneExposure.multipliedBy(count.toLong)
 
   /** Read gain. */
-  def gain: GhostReadNoiseGain
```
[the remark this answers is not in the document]
cquiroz: Right, do you think we revert the name change or write a converter?

bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/ConfigExtractor.scala
```diff
@@ -142,6 +142,30 @@ object ConfigExtractor {
     } yield GnirsParameters(pixelScale, filter, grating, readMode, xDisp, wavelen, slitWidth, camera, wellDepth, altair)
   }
 
+  // TODO-GHOSTITC
```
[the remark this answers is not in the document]
framos-gemini: I didn't write this comment, but I have removed it

bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/ItcService.scala
```diff
@@ -92,16 +92,20 @@ final case class ItcImagingResult(ccds: List[ItcCcd]) extends ItcResult
 sealed trait SpcChartType
 case object SignalChart       extends SpcChartType { val instance: SpcChartType = this } // signal and background over wavelength [nm]
 case object S2NChart          extends SpcChartType { val instance: SpcChartType = this } // single and final S2N over wavelength [nm]
+case object S2NChartPerRes    extends SpcChartType { val instance: SpcChartType = this } // single and final S2N over wavelength [nm]
```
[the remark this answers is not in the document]
framos-gemini: this comment does not correspond to the last commit. This was fixed.

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/obscomp/SPInstObsComp.java
```diff
@@ -461,18 +462,22 @@ public Instrument getInstrument() {
...
-        DHS_WRITE_TIME = getDhsWriteTime();
+        WRITE_TIME = getWriteTime();
     }
-
-    public final PlannedTime.CategorizedTime getDhsWriteTime() {
-        return DHS_WRITE_TIMES.get(getInstrument());
+    */
+    public final PlannedTime.CategorizedTime getWriteTime() {
+        return WRITE_TIMES.get(getInstrument());
     }
```
swalker2m, line 480: This one too -- as much as I prefer (and suggested we generalize the write time since it is not all DHS now), we'd need to leave it at `getDhsWriteTime` to avoid serialization issues.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostCamera.scala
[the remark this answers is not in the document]
cquiroz: I guess we need to load test sequences already executed.
I'd vote to revert the changes to avoid the extra work of a converter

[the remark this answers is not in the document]
swalker2m: Yeah I agree, but what is the plan again?  Do we need to deploy this as a patch release or will it be a separate instance?  Is it just the web page that needs to work or does it have to work with the existing OT?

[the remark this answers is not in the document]
swalker2m: Once deployed to production you mean?  At the moment if we tried to deploy this as a minor release I think many of the programs wouldn't load into the database.  We'd need to do an import from XML and require PIs to get a new OT.

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/flamingos2/GrismOptics.java
```diff
@@ -75,10 +75,10 @@ public double getPixelWidth() {
     }
 
     public double resolutionHalfArcsecSlit() {
-        return 0.5 / plateScale * dispersion();
+        return 0.5 / plateScale * dispersion(-1);
```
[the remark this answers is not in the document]
framos-gemini: I paste the comment of the implementation of that method: 
     * The grating dispersion is defined in the file <instrument>_grating.dat
     * After the ghost implementation this value may be a scalar or an array as a function of wavelength.
     * When the dispersion is an array, this returns the dispersion nearest to the specified wavelength.
     * When wavelength = -1, this returns the average dispersion.

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/ghost/Ghost.java
framos-gemini: No, this value never was confirmed it.

bundle/edu.gemini.itc/src/main/scala/edu/gemini/itc/base/DatFile.scala
```diff
@@ -107,15 +107,15 @@ object DatFile {
...
         case e : NumberFormatException => {
           val dDis = arrays.apply(tmp)
-          l.+=(Grating(name, resolvingPower, blaze, -999999, dDis, resolution))
+          l +=(Grating(name, resolvingPower, blaze, -999999, dDis, resolution))
         }
         case e : Exception => {
-          //l.+=(Grating(name, resolvingPower, blaze, -999999, arrays.apply(tmp), resolution))
-          println("NOT ANALYZED THE EXCEPTION, PLEASE CONTACT WITH A PROGRAMMER. LINE 125 OF THE DatFile.scala")
+
+          Log.severe("NOT ANALYZED THE EXCEPTION, PLEASE CONTACT WITH A PROGRAMMER. LINE 125 OF THE DatFile.scala")
```
swalker2m, line 118: This is written as though we expect the user to see it?  Should we add the exception itself to the log message?  If we do we don't have to mention the line number (which is now incorrect anyway). Can we realistically recover from whatever we expect to happen though?  I'd be inclined to let it filter up.

bundle/edu.gemini.itc/src/main/scala/edu/gemini/itc/base/DatFile.scala
```diff
@@ -114,8 +112,8 @@ object DatFile {
           l +=(Grating(name, resolvingPower, blaze, -999999, dDis, resolution))
         }
         case e : Exception => {
-
-          Log.severe("NOT ANALYZED THE EXCEPTION, PLEASE CONTACT WITH A PROGRAMMER. LINE 125 OF THE DatFile.scala")
+          Log.log(Level.SEVERE, "The grating configuration is not correct. ", e)
+          throw e
```
swalker2m, line 116: This change is fine, :+1:, but just want to note that I don't think even an incorrect configuration will ever trigger this. If `tmp` isn't a parseable as a `Double` then the `NumberFormatException` is caught above and if there is a problem handling it (say in `arrays.apply(tmp)`, it won't be caught here.  We'll see this exception in the case of memory issues and problems that can happen anywhere else.

swalker2m: Perhaps the `Grating` constructor can also throw exceptions though in which case this log would also be relevant.

bundle/edu.gemini.itc.web/src/main/java/edu/gemini/itc/web/html/OverheadTablePrinter.java
```diff
@@ -261,7 +261,7 @@ private String printOverheadTable() {
             if (cts == null) continue;
             PlannedTime.CategorizedTime ct = cts.max(Comparator.naturalOrder());
             int coadds = p.observation().calculationMethod().coaddsOrElse(1);
-            String category = ct.detail.getOrElse(ct.category.display);
+            String category = ct.category.display;
```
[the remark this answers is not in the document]
framos-gemini: I did some verification in other instruments (gmos, gnirs, F2) before doing the commit. I saw that the new change doesn't affect label of the Overheads label.

cquiroz (approved):
Thanks a lot @framos-gemini for this PR. We can merge after the build/test are complete and make a dev release.
Future changes will need another PR but GHOST ITC will be now part of the regular releases

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/obscomp/SPInstObsComp.java
```diff
@@ -443,9 +443,9 @@ public Instrument getInstrument() {
 
 
         m.put(Instrument.Flamingos2, Duration.ofSeconds(10));
+        m.put(Instrument.Ghost,      Duration.ofSeconds(10));
```
cquiroz, line 446: This is odd, why does order matter?

framos-gemini: Totally agree, it makes no sense

swalker2m: Hmm, this fixes some test failure?  What was it?

cquiroz: It is odd but the previous last commit made the itc tests fail

swalker2m: I bet it is unrelated.  It's good to have it sorted like this, but I would have to think if you put it back the tests would still run?

swalker2m: I bet it is unrelated.  It's good to have it sorted like this, but I would have to think if you put it back the tests would still run?

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/obs/plannedtime/PlannedTime.java
```diff
@@ -57,7 +57,7 @@ public enum Category {
         CONFIG_CHANGE("Config Change"),
         EXPOSURE("Exposure"),
         READOUT("Readout"),
-        DHS_WRITE("DHS Write"),
+        WRITE("Write"),
```
andrewwstephens, line 60: I think this may break the tests which expect the output web page to say "DHS Write"

andrewwstephens: @framos-gemini - if you want to keep the new `WRITE` I think it would be sufficient to use:
WRITE("DHS Write")

framos-gemini: Hi @andrewwstephens, when I made the new commit I hadn't seen your comment. Thank you very much for that. 

I made the change so that the Observation Overheads table would show Write instead of "DHS Write" because Ghost uses GDS and not DHS. This change produced doubts in Shane and Carlos because it affects to the serialization (they could give more details). 

Also since the deadline for the Ghost ITC is next week and I have other urgent tasks. I prefer to go back, and when I have more time to make the change with the "Ghost Test" implementation. 

Screen Shot 2023-03-31 at 20 50 17

andrewwstephens: :+1:

cquiroz:
Thanks @framos-gemini and @andrewwstephens for the fix.
I tested the latest changes with the seed that triggered the test failing and it is now working fine.

Thanks a lot and this is ready to merge

PR #2128, opened 2023-03-16
Improvements to the ITC json api

cquiroz:
These are some changes to make the latest calculations amenable to be used with explore
This should be merged after #2126


swalker2m (approved):
Don't know much about ITC but looks good to me. :+1:

PR #2130, opened 2023-03-24
REL-4265: Static component for Igrins2

cquiroz:
This is the first part of the OCS model for Igrins2. It contains the requested properties as  well as the basic UI editor

More work is needed to complete the task but it would be good to get initial feedback

image


bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/igrins2/SeqConfigIgrins2.java
```diff
@@ -0,0 +1,40 @@
...
+import edu.gemini.spModel.data.property.PropertySupport;
+import edu.gemini.spModel.gemini.init.ComponentNodeInitializer;
+import edu.gemini.spModel.seqcomp.SeqConfigNames;
+import edu.gemini.spModel.seqcomp.SeqConfigObsBase;
+
+import java.beans.PropertyDescriptor;
+import java.util.Collections;
+import java.util.Map;
+
+final public class SeqConfigIgrins2 extends SeqConfigObsBase implements PropertyProvider {
```
cquiroz, line 17: I'm not totally sure this is needed, it was essentially copied from the Ghost model

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/igrins2/SeqConfigIgrins2CB.java
```diff
@@ -0,0 +1,30 @@
+package edu.gemini.spModel.gemini.igrins2;
+
+import edu.gemini.pot.sp.ISPSeqComponent;
+import edu.gemini.spModel.config.HelperSeqCompCB;
+import edu.gemini.spModel.data.config.IConfig;
+
+import java.util.Map;
+
+final public class SeqConfigIgrins2CB extends HelperSeqCompCB {
```
cquiroz, line 9: I'm not totally sure this is needed, it was essentially copied from the Ghost model

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/igrins2/Igrins2.scala
```diff
@@ -0,0 +1,227 @@
...
+import java.beans.PropertyDescriptor
+import scala.collection.immutable.TreeMap
+import scala.collection.JavaConverters._
+import scala.math.sqrt
+
+/*
+ ** The Igrins2 instrument SP model.
+ * Note that we do not override clone since private variables are immutable.
+ */
+final class Igrins2 extends ParallacticAngleSupportInst(Igrins2.SP_TYPE) with PropertyProvider with PosAngleConstraintAware with Igrins2Mixin with IssPortProvider {
```
cquiroz, line 30: Should we specify a serial version?

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/igrins2/SeqConfigIgrins2.java
swalker2m: If it has an iterator, even a simple one where you can only change the exposure time, then I think we need one of these.

bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/igrins2/Igrins2.scala
sraaphorst: Do we ever use that? I've never changed or upgraded one before.

swalker2m: Probably not.  Mostly we only change the binary format between major releases, when we do a migration anyway.  If we add one and then make a breaking change in a supposedly minor release it would only cover it up.

swalker2m (approved):
Seems good to me.  :+1:

PR #2133, opened 2023-03-29
REL-4280: Setup Igrins2 obslog segment

cquiroz:


swalker2m (approved):
I can't remember how this works, but the code looks good. :+1:

PR #2134, opened 2023-03-31
REL-4265: Set Igrins2 overheads

cquiroz:


bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/igrins2/Igrins2.scala
```diff
@@ -198,6 +228,11 @@ object Igrins2 {
     AllowedFowlerSamples.minBy(cur => math.abs (cur - nFowler))
   }
 
+  def readoutTime(expTime: Time): Time = {
+    val nFowler = fowlerSamples(expTime)
+    1.45479.seconds / nFowler
```
andrewwstephens, line 233: Should this be 1.45 * nFowler?

PR #2139, opened 2023-04-03
REL-4203: Hide guide fiber options

swalker2m:
Hides the "Enable guide fibers" options in the OT and removes them from the sequence configuration.  I'm not committing to removing them altogether yet since it changes the asterism definition drastically (`GhostTarget` would go away altogether) and it seems like it could easily come back.


cquiroz (approved):
LGTM

PR #2140, opened 2023-04-03
REL-4295: Ghost readout + write times

swalker2m:
Ditches the formula for calculating readout time in favor of a simple lookup.  Changes the write time.


bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostCamera.scala
```diff
@@ -139,6 +118,46 @@ object GhostCamera {
         configToBinning(c, Ghost.RED_BINNING_KEY)
       )
 
+    val ReadoutTime: Map[(GhostBinning, GhostReadNoiseGain), Duration] = {
```
cquiroz, line 121: Cool, we can use this in the seqexec

cquiroz (approved):
LGTM

PR #2143, opened 2023-04-05
REL-4265: Igrins2 fixes

cquiroz:
This PR includes 3 changes for Igrins2:
* Remove the SlitViewingCamera porperty
* Allow the creation of an Igrins2-iterator
* Fix a UI bug when changing the exposure time


bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/igrins2/SeqConfigIgrins2CB.java
```diff
@@ -21,6 +36,12 @@ public Object clone() {
     @Override
     protected void thisApplyNext(final IConfig config, final IConfig prevFull) {
         super.thisApplyNext(config, prevFull);
+
+        copyToObserve(config, Igrins2.EXPOSURE_TIME_PROP());
+
+        config.putParameter(SeqConfigNames.INSTRUMENT_CONFIG_NAME,
+                StringParameter.getInstance(InstConstants.INSTRUMENT_NAME_PROP,
+                        Igrins2.INSTRUMENT_NAME_PROP()));
```
swalker2m, line 44: I can't remember the details, but somehow this happens for `Flamingos2` without having to explicitly put the parameter in the config builder.

swalker2m (approved):
Seems good to me.  :+1:

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/igrins2/SeqConfigIgrins2CB.java
cquiroz: This was taken from Ghost, and the same was done for GPI but maybe I should remove it?

swalker2m: It's not worth spending any time on it.  If it works then :+1:

PR #2149, opened 2023-04-11
REL-4289: Setup Igrins2 Tcc Support

cquiroz:
There are a few unknowns about the parameters but let's open this PR so we don't forget


swalker2m (approved):
Looks good.  :+1: If you want to merge it now and update it later if necessary that seems fine to me.

PR #2154, opened 2023-04-28
REL-4321- Fix ITC sampling of redshifted SED

andrewwstephens:
The ITC used a fixed wavelength sampling size for the input SED, even when the SED was redshifted. This resulted in undersampling of the SED which distorted the line shape and amplitude and yielded incorrect S/N estimates.

This PR scales the wavelength sample size with redshfit so that after redshifting the sample size is correct.

Some of the baseline tests had a non-zero redshift and those broke, so this PR also includes a new baseline.txt file.


andrewwstephens:
For future reference, here's an emission line at various redshifts with the old and new code:
itc_old
itc_new

PR #2155, opened 2023-04-28
Support PIT 2023.2.[1|2] versions

cquiroz:
We have two versions of the PIT 2023 in the field and at the moment spdb con only handle one.
This PR fixes this allowing us to accept skeletons from either version


swalker2m (approved):
Not sure what this does, but it looks good to me.  :-)

PR #2158, opened 2023-05-05
Feature/REL-3981 - GNIRS IFU ITC updates

andrewwstephens:
This pr includes 3 changes:
1. Cleanup the GNIRS IFU web form based on suggestions from the GNIRS team.
2. Fix a bug where the grating transmission was being applied inside the IFU element loop.
3. Fix a bug where the SED was being incorrectly smoothed for every IFU element loop.


cquiroz (approved):
LGTM

PR #2159, opened 2023-05-05
Fix offset calculation

swalker2m:
When we instantiate GHOST templates we add a fake target (for Dual Target mode) or fake sky positions (for the + sky modes).  These are created such that they are separated by 2 arcmins from the known target position.  Unfortunately this was being done as a simple offset in RA when in reality it should be angular separation.

### DIFF
--- a/bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/Offset.scala
+++ b/bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/Offset.scala
@@ -12,6 +12,7 @@ import Scalaz._
  * base position and a second position.
  */
 case class Offset(p: OffsetP, q: OffsetQ) {
+
   /**
    * Calculates the absolute distance between this Offset position and the
    * given position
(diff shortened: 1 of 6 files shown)


cquiroz:
Will the sequence contain the correct target coordinates or are they unrolled when get the sequence
This may affect the seqexec

bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/Offset.scala
```diff
@@ -29,6 +30,15 @@ case class Offset(p: OffsetP, q: OffsetQ) {
     Angle.fromDegrees(d)
   }
 
+  /**
+   * Calculates the bearing from 0,0 as an Angle clockwise from north 0º.
+   */
+  def bearing: Angle = {
+    val x = -p.toAngle.toSignedDegrees
+    val y =  q.toAngle.toSignedDegrees
+    Angle.fromRadians(atan2(x, y))
```
cquiroz, line 39: wow, real math 😄

cquiroz (approved):
LGTM

swalker2m:
> Will the sequence contain the correct target coordinates or are they unrolled when get the sequence. This may affect the seqexec

This will only impact the initial sky position (or IFU2 target position for dual target) when we instantiate GHOST observations from the templates.  How we pass the information to the seqexec is unchanged.   In all likelihood the values will get adjusted during the phase 2 refinement regardless.

PR #2163, opened 2023-05-09
Logging Cleanup

swalker2m:
This PR would change the log level to `FINE` for some rather verbose log messages that were added specifically to track down a problem with synchronization.  That problem has been solved and the log messages are distracting from actual issues that might appear so it seems appropriate to tone them down a bit before the 2023B release.


cquiroz (approved):
LGTM

PR #2170, opened 2023-05-17
REL-4313 Indicate if GHOST is using P2

cquiroz:


bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/GhostSupport.java
```diff
@@ -36,34 +38,44 @@ public String getPositionAngle() {
...
             case GhostHighResolutionTargetPlusSkyPrv:
-                result = "GHOST_HIGH_RESOLUTION";
+                result = "GHOST_HIGH_RESOLUTION" + usesP2;
                 break;
 
             default:
-                result = "GHOST";
+                result = "GHOST" + usesP2;
         }
         return result;
```
swalker2m, line 80: I suppose you could factor the `+ usesP2` down to the `return`:
```
return result + usesP2;
```

swalker2m (approved):
Looks good.  :+1:

bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/GhostSupport.java
cquiroz: Oh right, good point

PR #2173, opened 2023-06-16
REL-4338: Set the GMOS readnoise according to the read mode

andrewwstephens:
This changes the previously static read noise value to one that is determined from the selected "Amplifier Gain" and "Amplifier Read Mode".


bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gmos/Gmos.java
```diff
@@ -226,6 +226,71 @@ public double getPixelSize() {
         }
     }
 
+    // Derive the read noise from the Instrument, Amp Gain, and Amp Read Mode, overriding the default method in Instrument.java
```
sraaphorst, line 229: This method looked really messy in GitHub, and I was going to suggest splitting it into multiple functions if possible, but seeing it in the raw view, it looks fine and very clear. Good stuff.

sraaphorst (approved):
👍

cquiroz (approved):
LGTM

PR #2184, opened 2023-07-27
REL-4345 Fix some bugs on the upconversion of old GHOST proposals

cquiroz:


bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/transform/UpConverter.scala
```diff
@@ -174,7 +174,25 @@ case class LastStepConverter(semester: Semester) extends SemesterConverter {
...
  */
 case object SemesterConverter2023BTo2024A extends SemesterConverter {
-  override val transformers: List[TransformFunction] = Nil
+  val renameSRPlusSky: TransformFunction = {
+    case p @ <ghost>{ns @ _*}</ghost> =>
+      object srTransformer extends BasicTransformer {
+        override def transform(n: xml.Node): xml.NodeSeq = n match {
+          case <name>{name}</name>                                   =>
+            // There is an interim model with some typos we need to consider
+            <name>{name.text.replaceAll("Ghost", "GHOST").replaceAll("SRIFU1 Target \\+ SRIFU2 Sky", "SRIFU + Sky").replaceAll("SRIFU1 Sky \\+ SRIUF2 Target", "SRIFU + Sky").replaceAll("SRIFU1 Sky \\+ SRIFU2 Target", "SRIFU + Sky")}</name>
```
swalker2m, line 183: Since it isn't a regex, should it be `replace`?

Remind me why all the options, which are actually distinct cases, get mapped to `SRIFU + Sky`?

swalker2m (approved):
Looks good, assuming we should be combining all those cases.  :+1:

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/transform/UpConverter.scala
cquiroz: Right, it could be `replace`.

We have
* SRIFU1 Target + SRIFU2 Sky
* SRIFU1 Sky + SRIFU2 Target

And I found a sample with a typo, instead of SRIFU2 it was SRIUF2 so I'm considering that case too
All those cases map to `SRIFU + Sky`

swalker2m: The distinction between the cases gets lost?  I'm sure there's a reason I just can't remember.

swalker2m: I guess because it is Phase 1 and we don't know how it will ultimately shake out at Phase 2?

PR #2185, opened 2023-07-28
Use a regex to search for the old target modes

cquiroz:
For some reason this didn't get folded into the previous PR


swalker2m (approved):
Looks good. :+1:

PR #2188, opened 2023-08-03
REL-4345 Make the model for the 2023B release match 2024A

cquiroz:
The model for 2023B test and 2024A have diverged and need to be sync'd to match
This PR incorporates those changes and also fixes the conversion code to make sure previous proposals can be imported


tpolecat (approved):
😭 👍🏻

PR #2192, opened 2023-08-28
Update signing process for mac

cquiroz:
This PR contains updates to the build such that the signed mac installer can also be notarized:.
This requires three chaanges:

* Sign each binary in the jdk separately
* Mark the binaries as "hardened"
* Add a set of entitlements to let the JDK run properly

The notarization process is not automated just yet, it has to be done manually

More details at
https://github.com/gemini-hlsw/ocs/wiki/Signing-Applications-for-OSX#osx-notarization

### DIFF
--- a/app/pit/dist/MacOS/entitlements.plist
+++ b/app/pit/dist/MacOS/entitlements.plist
@@ -1,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+    <key>com.apple.security.cs.allow-jit</key>
+    <true/>
+    <key>com.apple.security.cs.allow-unsigned-executable-memory</key>
+    <true/>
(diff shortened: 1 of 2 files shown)


swalker2m:
This is excellent, thanks for working on this!  

To make the notarization part work, I assume we need to switch the a newer 1.8 JRE?  But after applying these changes the  OT release continues to work as it does today without changing the JRE correct?  How do you update the JRE for the PIT release w/o impacting the ODB?  Do you have to manually switch and switch back?

Have we tried notarizing the OT?  I'm wondering if an OT with the latest JRE can still talk to the ODB with the old 1.8 version.

cquiroz:
I haven't tried the OT and yes we'd need to switch across jdk versions.
The JRE I used for the PIT is 1.8 but a much newer build (382) whose binary is amenable to notarization

swalker2m (approved):
It'd be good to confirm that the Mac OT installer works at least as well as it does today with these changes.

For December we should try with a properly notarized OT with newer 1.8 JRE talking to an ODB with the old 1.8 version it is stuck with.

cquiroz:
I tested the OT and it gets notarized correctly and starts just fine.
We may want to test against spdb built with the older jdk on linux to verify trpc has no problems with two different jres

PR #2194, opened 2023-08-31
Last minute PIT changes

cquiroz:
Changes the logic of converting certain NIFS proposals and adjusts the location of the AGS server

### DIFF
--- a/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/transform/UpConverter.scala
+++ b/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/transform/UpConverter.scala
@@ -204,7 +204,7 @@ case object SemesterConverter2023BTo2024A extends SemesterConverter {
 
   def transformNifsName(name: String) = name match {
     case nifsNameRegexAo(a, b, _) => s"""GNIRS Spectroscopy $a$b 0.15"/pix 111 l/mm grating None LR-IFU &lt; 2.5um"""
-    case nifsNameRegexNonAo(_)    => s"""GNIRS Spectroscopy LGS/PWFS1 0.15"/pix 111 l/mm grating None LR-IFU &lt; 2.5um"""
+    case nifsNameRegexNonAo(_)    => s"""GNIRS Spectroscopy  0.15"/pix 111 l/mm grating None LR-IFU &lt; 2.5um"""
     case _                        => name
   }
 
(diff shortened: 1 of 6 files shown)


bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/immutable/transform/UpConverter.scala
```diff
@@ -204,7 +204,7 @@ case object SemesterConverter2023BTo2024A extends SemesterConverter {
 
   def transformNifsName(name: String) = name match {
     case nifsNameRegexAo(a, b, _) => s"""GNIRS Spectroscopy $a$b 0.15"/pix 111 l/mm grating None LR-IFU &lt; 2.5um"""
-    case nifsNameRegexNonAo(_)    => s"""GNIRS Spectroscopy  0.15"/pix 111 l/mm grating None LR-IFU &lt; 2.5um"""
+    case nifsNameRegexNonAo(_)    => s"""GNIRS Spectroscopy LGS/PWFS1 0.15"/pix 111 l/mm grating None LR-IFU &lt; 2.5um"""
```
swalker2m, line 207: It's `NonAo` but uses LGS?

cquiroz: This has been clarified, we'll change this logic

bundle/edu.gemini.pit/src/main/scala/edu/gemini/pit/ui/robot/ProblemRobot.scala
```diff
@@ -237,6 +237,14 @@ class ProblemRobot(s: ShellAdvisor) extends Robot {
         case _                 => false
       }
 
+    def bpIsGnirsLRIFU(b: BlueprintBase): Boolean =
+      b match {
+        case GnirsBlueprintSpectroscopy(AltairNone, _,  _, _, GnirsFpu.LR_IFU, _)           => false
+        case GnirsBlueprintSpectroscopy(AltairLGS(true, _, _), _, _, _, GnirsFpu.LR_IFU, _) => false
```
swalker2m, line 243: On the surface of it, since it is `GnirsFpu.LR_IFU`, I would expect this to be `true` for these two cases as well.  Maybe a comment about what is going on here?

cquiroz: I'll add a comment but basically LR_IFU is only compatible with altair None or Altair LGS/PWFS1

cquiroz:
After thinking a bit more about this I'm instead of having a p2 check, we will allow selecting LR-IFU only on the permissible altair modes

swalker2m (approved):
Looks great. :+1:

PR #2203, opened 2023-11-10
REL-4366: publish to pachon first for GS

swalker2m:
Since publishing to GN is currently unavailable, this update changes the publishing order at GS to `PACHON` and then `INTERNAL` (GN).  It will still time out when attempting to publish at GN, but the site-local publish will in theory have happened by that point.

One caveat: this is untested since it involves network connections that aren't available to me.


cquiroz (approved):
LGTM

PR #2204, opened 2023-11-13
REL-4345: Apply Proper Motion Corrections in `GhostRule` Coordinate Checks

swalker2m:
Adds `pmCorrectedCoordinates` to `GhostTarget` and `pmCorrectedBasePosition` to `Asterism`.  These are unused except in the `GhostRule` as requested.  This is unfortunately confusing for multiple reasons:

* Asking for a target's coordinates takes a time parameter (`when`) which implies that it will be PM corrected.  It is not.  Only sidereal targets use the `when` parameter.
* Fixing the problem above might open Pandora's box, so now there are `pmCorrected` options that also take a time `when` but do in fact use it for sidereal targets.
* Applying proper motion to `GhostRule` coordinate checks will mean that it no longer matches what we plot in the position editor, generating angst.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostAsterism.scala
+++ b/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostAsterism.scala
@@ -59,8 +59,18 @@ object GhostAsterism {
   final case class GhostTarget(spTarget: SPTarget,
                                guideFiberState: GuideFiberState) {
 
+    /** Target coordinates, not PM corrected. */
     def coordinates(when: Option[Instant]): Option[Coordinates] =
       spTarget.getCoordinates(when.map(_.toEpochMilli))
+
+    /** Target coordinates, PM corrected. */
+    def pmCorrectedCoordinates(when: Option[Instant]): Option[Coordinates] = {
+      val corrected = for {
+        pm <- spTarget.getProperMotion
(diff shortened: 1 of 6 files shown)


bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/ghost/GhostAsterism.scala
```diff
@@ -59,9 +59,19 @@ object GhostAsterism {
...
     def coordinates(when: Option[Instant]): Option[Coordinates] =
       spTarget.getCoordinates(when.map(_.toEpochMilli))
 
+    /** Target coordinates, PM corrected. */
+    def pmCorrectedCoordinates(when: Option[Instant]): Option[Coordinates] = {
+      val corrected = for {
+        pm <- spTarget.getProperMotion
+        w  <- when
+      } yield pm.calculateAt(spTarget.getTarget, w)
+      corrected orElse coordinates(when)
```
cquiroz, line 72: Isn't `pm.calculateAt` doing the same as `getCoordinates(when)`?

swalker2m: You might think so, but no we don't apply proper motion despite asking for the `when` parameter.   It is ignored except for non-sidereal targets.

cquiroz (approved):
LGTM

PR #2205, opened 2023-11-13
[REL-4364] Add IGRINS-2 to inst enum

tpolecat:
This adds IGRINS-2 to the `Inst` enumeration in QPT as an available instrument at GN. It shows up in the tree ok, and IGRINS observations appear correctly in the candidate view.


swalker2m (approved):
Excellent. :+1:

PR #2211, opened 2023-11-14
REL-3922: non-sidereal target lookup UI event loop issues

swalker2m:
Andy reports problems with non-sidereal name resolution hanging occasionally.  He's even had to kill the OT to resolve it.  It works for me, but makes me suspicious that we're mixing in UI updates from non-event loop threads.  We seem to be pretty meticulous about keeping UI updates in the event loop, but I did find one spot where we weren't doing it.  Perhaps this will fix the issue?  Is there a better way?


tpolecat:
Ugh, sorry. That was my mistake. Hopefully this improves things. Can't hurt.

I don't know of any systematic way to deal with this other than just looking through the code, which we have done a number of times. It's all terrible.

tpolecat (approved):

swalker2m:
I found a couple more instances of this, the most egregious of which was definitely my fault and is most likely the source of the mishap.

PR #2215, opened 2023-12-21
REL-4280 Fixes for IGRINS-2 support on obslog

cquiroz:


bundle/edu.gemini.obslog/src/main/java/edu/gemini/obslog/instruments/Igrins2LogSegment.java
```diff
@@ -17,7 +17,7 @@
 public class Igrins2LogSegment extends InstrumentLogSegment {
     public static final Logger LOG = Logger.getLogger(Igrins2LogSegment.class.getName());
 
-    private static final String NARROW_TYPE = "Igrins2";
+    private static final String NARROW_TYPE = "IGRINS-2";
```
swalker2m, line 20: Does this have to match the `SpComponentType`?  If so, maybe we should just take it from there:
```java
private static final String NARROW_TYPE = SpComponentType.INSTRUMENT_IGRINS2.narrowType;
```

swalker2m (approved):
Looks good to me. :+1:

bundle/edu.gemini.obslog/src/main/java/edu/gemini/obslog/instruments/Igrins2LogSegment.java
cquiroz: Right, let me try that

swalker2m (approved):
:+1:

PR #2230, opened 2024-01-31
REL-4209 Use Gaia and UCAC4 for PIT AGS estimate

cquiroz:
This is a low-effort solution for the particular case of phase I proposals that use altair. There is a small number of cases where gaia doesn't provide a star but UCAC4 could.

In this solution we only override `estimate` to collect stars from both catalogues and use them to find a star. This showed to work on the provided examples.
This is `estimate` specific as it isn't clear whether we want to do the same for the OT which should remain unchanged.

Also we are not providing a UI just the ags backend server will be updated to automatically do this in the PIT


bundle/edu.gemini.ags/src/main/scala/edu/gemini/ags/impl/SingleProbeStrategyParams.scala
```diff
@@ -40,7 +38,9 @@ sealed trait SingleProbeStrategyParams {
         case SingleProbeStrategyParams.AltairAowfsParams => true
         case _                                           => false
       }
-      CatalogQuery.coneSearch(base.toNewModel, rc, ml, catalogName, isLgs)
+      // for Altair two catalogs can be searched
+      val cats = if (isLgs) NonEmptyList(GaiaGemini, UCAC4) else catalogs
+      cats.map(c => CatalogQuery.coneSearch(base.toNewModel, rc, ml, c, isLgs))
```
swalker2m, line 43: There is an `AltairAowfsParams` entry.  Perhaps we could override `catalogs` there to return `GaiaGemini` and `UCAC4`?  I suppose it doesn't make much difference though since it seems we are computing `isLgs` for some other reason.

swalker2m (approved):
Seems reasonable given the circumstances.  :+1:

PR #2238, opened 2024-03-02
REL-4410 Fix F2 ITC NaNs

andrewwstephens:
This includes minor tweaks to two F2 data files to avoid getting NaNs when calculating the peak signal.


cquiroz (approved):
LGTM

PR #2241, opened 2024-04-12
REL-4411: 24B default ObsClass for arcs

swalker2m:
An embarrassing hack to change the default `OcsClass` for arcs in 2024B and beyond.  It's important to maintain the old default for old programs since the value impacts time accounting, which isn't recorded anywhere but instead recalculated on demand.  So, this is a _wee bit_ sketchy.  To avoid checking when you're asking, we instead check the program id and strip the semester from it.  If the semester is defined and pre-2024B we keep the old default.  Otherwise we use the new default.

### DIFF
--- a/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/seqcomp/DefaultArcObsClass.scala
+++ b/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/seqcomp/DefaultArcObsClass.scala
@@ -1,0 +1,25 @@
+package edu.gemini.spModel.gemini.seqcomp
+
+import edu.gemini.pot.sp.ISPNode
+import edu.gemini.spModel.core.Semester
+import edu.gemini.spModel.obsclass.ObsClass
+import edu.gemini.spModel.rich.core._
+import edu.gemini.spModel.rich.pot.sp._
+
+// REL-4411: The default ObsClass for arcs changed in 24B.  Before it would
+// default to PARTNER_CAL but now should default to PROG_CAL.
+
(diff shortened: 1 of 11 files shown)


bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/seqcomp/DefaultArcObsClass.scala
```diff
@@ -0,0 +1,25 @@
...
+  private val Semester24B: Semester =
+    Semester.parseOptional("2024B").get
+
+  private def uses24BDefault(c: ISPNode): Boolean =
+    Option(c)
+      .flatMap(_.semesterOption)
+      .forall(_ >= Semester24B)
+
+  def forNode(c: ISPNode): ObsClass =
+    if (uses24BDefault(c)) ObsClass.PARTNER_CAL else ObsClass.PROG_CAL;
```
cquiroz, line 23: cool

swalker2m:
> I wonder if this has some downstream effects on the seqexec 

The seqexec has to be prepared for any of the possible `observe:obsClass` values since they are (and always have Benn) user editable.  I'm guessing it just changes what it writes into the FITS header?  In the ODB, we use it in time accounting calculations.

cquiroz:
I checked usages of `obsClass` iin the seqexec and gpi and ghost use it for logic
in the case of GPI we have a special class for `acq` where it triggers a call to `aliignAndCalib`. Since GPI is not active this is not an issue
For GHOST we have a special behavior for `dayCal`
So I think this does not  affect the seqexec

cquiroz (approved):
LGTM

PR #2242, opened 2024-04-12
REL-4413: Preserve order in InstantiationFunctor

swalker2m:
Updates the `InstantiationFunctor` so that the order of parameters is preserved on instantiation.  It does this employing, rather awkwardly, Scala's `ListMap` and `ListSet` ~in the Java `InstantiationFunctor` instead of the Java `HashMap` and `HashSet`~.

**Edit:** It turned out to be too challenging to get this to compile as an update to the Java functor in both Scala 2.11 and Scala 2.13 so I converted the `InstantiationFunctor` itself to Scala.


bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/template/InstantiationFunctor.java
```diff
@@ -34,29 +42,29 @@ public class InstantiationFunctor extends DBAbstractFunctor {
...
 
     public void execute(IDBDatabaseService db, ISPNode node, Set<Principal> principals) {
         try {
             List<ISPGroup> newGroups = new ArrayList<>();
             final ISPProgram prog = db.lookupProgram(node.getProgramKey());
-            for (Map.Entry<ISPTemplateGroup, Set<ISPTemplateParameters>> e : selection.entrySet()) {
-                for (ISPTemplateParameters ps : e.getValue()) {
-                    final ISPTemplateGroup templateGroup = e.getKey();
+            for (Tuple2<ISPTemplateGroup, ListSet<ISPTemplateParameters>> e : WrapAsJava$.MODULE$.asJavaIterable(selection.toList())) {
+                for (ISPTemplateParameters ps : WrapAsJava$.MODULE$.asJavaIterable(e._2.reversed())) {
```
swalker2m, line 66: Oddly, I have to reverse the `Set` of parameters.  New parameters are added using the `+` method (translated to `.$plus` in Java).  I confirmed they `add` method is called in the expected order so `+` must be prepending instead of appending?

cquiroz (approved):
LGTM

PR #2244, opened 2024-04-15
REL-4420: IGRINS2 Template

swalker2m:
Welcomes IGRINS2 to the world of templates.

### DIFF
--- a/bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/template/factory/impl/igrins2/Igrins2.scala
+++ b/bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/template/factory/impl/igrins2/Igrins2.scala
@@ -1,0 +1,64 @@
+package edu.gemini.phase2.template.factory.impl
+package igrins2
+
+import edu.gemini.pot.sp.{ISPGroup, SPComponentType}
+import edu.gemini.spModel.core.SPProgramID
+import edu.gemini.spModel.gemini.igrins2.NoddingOption
+import edu.gemini.spModel.gemini.igrins2.{Igrins2 => InstIgrins2}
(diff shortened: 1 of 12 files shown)


bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/template/factory/impl/igrins2/Igrins2.scala
```diff
@@ -0,0 +1,64 @@
...
+  //
+  // INCLUDE {4} IN target-specific Scheduling Group
+  //
+  // INCLUDE the note "Observer Instructions" IN the target-specific Scheduling Group
+
+  val n = blueprint.getNoddingOption match {
+    case NoddingOption.KEEP_TARGET_IN_SLIT => 2
+    case NoddingOption.NOD_TO_SKY          => 3
+  }
+  include(1, 5, n, 4) in TargetGroup
```
swalker2m, line 61: Everything here works, except that the order of inclusion above is not respected.  It always keeps the order of library id, which means it ends up: `1, n, 4, 5`.  :-/  Any idea why?

cquiroz (approved):
LGTM

bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/template/factory/impl/igrins2/Igrins2.scala
swalker2m: https://github.com/gemini-hlsw/ocs/blob/develop/bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/template/factory/impl/TemplateFactoryImpl.scala#L141-L142

PR #2245, opened 2024-04-15
REL-4390: back off daily sync

swalker2m:
Reduces the insistency of the daily full database / archive sync, reducing the thread count from 10 to 2 and backing the thread priority down by one.  This was in response to a request from Paul to avoid hitting the FITS servers with too many simultaneous requests.

Because all of the polling was done by shared worker pools, I had to segregate the jobs and put the daily full sync in its own pool.  Because the work is done by a priority queue, with the program sync being the lowest, I then needed to lower the thread priority of the separate daily program sync.


cquiroz (approved):
LGTM

PR #2249, opened 2024-04-16
REL-4414: IGRINS SmartGcal

swalker2m:
Support for SmartGcal with IGRINS.  There is no instrument configuration to match.  It appears to be these constant values.
Screen Shot 2024-04-16 at 4 39 07 PM


cquiroz (approved):
LGTM

PR #2251, opened 2024-04-22
Strip ephemeris data on export

swalker2m:
When we export a non sidereal observation to the seqexec, we include binary compressed ephemeris data.  Unfortunately the seqexec uses a newer Scala library and cannot read this information.  Fortunately, it also doesn't use this information and no other clients use this servlet.  I've modified the export to strip the ephemeris data and replace it with an empty map (since the codec requires something valid to be written there).

### DIFF
--- a/bundle/edu.gemini.spModel.io/src/main/scala/edu/gemini/spModel/io/ocs3/Ocs3ExportFunctor.scala
+++ b/bundle/edu.gemini.spModel.io/src/main/scala/edu/gemini/spModel/io/ocs3/Ocs3ExportFunctor.scala
@@ -1,12 +1,14 @@
 package edu.gemini.spModel.io.ocs3
 
 import edu.gemini.pot.sp._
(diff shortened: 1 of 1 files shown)


tpolecat (approved):
Ah shit, yeah that would do it. This must have been hard to figure out. So the seqexec can decode `Nil` but not the list elements themselves. Bah. 👍🏻

swalker2m:
>  So the seqexec can decode Nil but not the list elements themselves. 

I'm not sure yet, Carlos is going to test it.  I may need to add a "catchingAll" `ParamSet` decoder or something :-/

tpolecat:
They have taken some care to ensure that things like `Nil` and `None` always deserialize to their singletons, so it seems likely to work.

cquiroz:
I was a bit surprised to notice that things like `Tuple3` don't have a `serialVersionID` thus I don't trust serialization promises 😄

cquiroz:
🤯 🔫

swalker2m:
Still doesn't work.  Serialization issues.  Closed in favor of #2252.

PR #2252, opened 2024-04-22
Don't calculate non-sidereal coordinates for GHOST

cquiroz:


swalker2m (approved):
Looks good, under the circumstances.  :+1:

cquiroz:
It will be worth it...

PR #2254, opened 2024-04-30
REL-4392: GNIRS template updates

swalker2m:
Based on a diff of the pseudocode.


bundle/edu.gemini.phase2.skeleton.servlet/src/main/scala/edu/gemini/phase2/template/factory/impl/gnirs/GnirsImaging.scala
```diff
@@ -40,7 +40,7 @@ case class GnirsImaging(blueprint:SpGnirsBlueprintImaging) extends GnirsBase[SpG
     case _              => oldFilter
   }
 
-  include(16, 17, 18, 19, 20, 21) in TargetGroup
+  include(16, 17, 18, 19, 20, 21, 31) in TargetGroup
```
cquiroz, line 43: Why not 39?

swalker2m: D'oh, you're right.  Thanks for pointing this out!

cquiroz (approved):
LGTM

PR #2255, opened 2024-05-01
REL-4446 Update TPE visualization Igrins2ScienceAreaGeometry.scala

arturog8m:
Small request from science that looks [trivial](https://noirlab.atlassian.net/browse/REL-4446)

but please review and see if I'm missing something as I haven't touched this in years.


cquiroz (approved):
LGTM

swalker2m:
Are there any p2 check rules associated with this?

arturog8m:
I couldn't find any, but this is the OT, and the OT always prevails, and it will prove me wrong.

PR #2257, opened 2024-05-02
REL-4428: WDBA / TCC instrument config updates

swalker2m:
Adds P1/P2-specific configs for select instruments.


tpolecat (approved):
This looks good. 👍🏻

PR #2261, opened 2024-05-13
REL-4429 Renaming of IG2 nodding options

cquiroz:
It requires a p1 model bump

### DIFF
--- a/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/igrins2/blueprint/SpIgrins2Blueprint.java
+++ b/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/igrins2/blueprint/SpIgrins2Blueprint.java
@@ -25,7 +25,7 @@ public final class SpIgrins2Blueprint extends SpBlueprint {
     }
 
     public SpIgrins2Blueprint(ParamSet paramSet) {
-        noddingOption = Pio.getEnumValue(paramSet, PARAM_NODDING, NoddingOption.KEEP_TARGET_IN_SLIT);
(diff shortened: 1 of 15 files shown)


swalker2m (approved):
Looks great. :+1:

cquiroz:
@swalker2m This was updated after some changes on the task description and also it includes now REL-4420

bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/igrins2/blueprint/SpIgrins2Blueprint.java
```diff
@@ -25,15 +25,15 @@ public NoddingOption getNoddingOption() {
     }
 
     public SpIgrins2Blueprint(ParamSet paramSet) {
-        noddingOption = Pio.getEnumValue(paramSet, PARAM_NODDING, NoddingOption.KEEP_TARGET_IN_SLIT);
+        noddingOption = Pio.getEnumValue(paramSet, PARAM_NODDING, NoddingOption.NOD_ALONG_SLIT);
```
swalker2m, line 28: The blueprints are stored in the science program, so I think we may need to migrate the nodding option unfortunately assuming this was in the 2024A model?

cquiroz: That wasn't very clear to me. We haven't imported any IG2 p1 proposal so far

swalker2m: Right, I guess this is new for this semester and we don't use `NoddingOption` in the Phase 2 model.  It should be fine. hehe

PR #2264, opened 2024-05-21
Include target coordinates for sidereals

cquiroz:
A side effect of the changes for GHOST non-sidereals was to remove the coordinates for the sidereal targets if they have a null proper motion.
This breaks all GHOST modes in the seqexec unless pm is set

This was tested against a local instance of the seqeexc


swalker2m (approved):
Looks great.

PR #2265, opened 2024-05-22
REL-4421: Fix error in displayed final S/N when specifying the S/N and using coadding

andrewwstephens:
This fixes the overhead table after my last "fix" to the S/N.
The baseline had to be updated for all the cases that used the `ImagingInt` calculation method.


cquiroz (approved):
LGTM

PR #2267, opened 2024-05-22
REL-4390: update summit FITS storage service host

swalker2m:
`cpofits-lv1.cl.gemini.edu` is apparently the old summit storage host.  This PR switches `gsodbtest` to use the same alias used on `gsodb`.


cquiroz (approved):
LGTM

PR #2270, opened 2024-06-05
REL-4481: allow viewing targets in observations without obs conditions

swalker2m:
There was a FR (https://noirlab.atlassian.net/browse/GSFR-43564) which complains of not being able to see the FITS background image in an observation.  The problem turned out to be that the observation was missing observing conditions, which is really irrelevant for the purpose of showing a FITS image.


cquiroz:
Interesting, how is that this was never noted earlier

cquiroz (approved):
LGTM

swalker2m:
> Interesting, how is that this was never noted earlier

I think it's just that every observation comes with observing conditions by default and they are only ever removed by accident.

PR #2274, opened 2024-07-12
Preparation for PIT 2025A

cquiroz:


swalker2m (approved):
Looks good.  Wish we could do something about the semester converters.  :+1:

cquiroz:
> Looks good. Wish we could do something about the semester converters. 👍

In GPP they will be a migratioin

PR #2285, opened 2024-10-02
Update QPT and OT icons

cquiroz:


swalker2m (approved):
:+1: on the completion of the task, :-1: on changing icons every semester

PR #2288, opened 2024-10-06
 REL-4572: Add stellar SEDs to the ITC

andrewwstephens:
This augments some of the more poorly-represented stellar SEDs in the ITC, namely subgiants, subdwarfs, and white dwarfs.

This does not replace any of the existing SEDs.

The long list of SEDs was moved out of the individual instrument HTML forms into two files (seds_stellar.html and seds_nonstellar.html) that are used by all the forms.

I removed the .toLowerCase() from the stellar SED reader so that the file names may now use upper case (as was always true for non-stellar SEDs) which makes the names a bit more natural (and makes it easy to distinguish the old SEDs which are in lower case from the new SEDs which use upper case).


cquiroz:
I'd say we just merge it

PR #2295, opened 2024-10-28
REL-4573: Extend the F2 R3K grism K-band (m=3) throughput to 2.5 microns

andrewwstephens:
One-line update of the Flamingos-2 R3000 grism throughput in order m=3.


cquiroz (approved):
LGTM

PR #2297, opened 2024-10-28
REL-4448 - Include correct acquisition overheads (Slit or MOS) in GMOS ITC

andrewwstephens:
This adds separate options in the GMOS ITC web form for either Long Slit or MOS so that the correct acquisition overhead is given.  The S/N calculations should be unchanged.

### DIFF
--- a/bundle/edu.gemini.itc.web/src/main/scala/edu/gemini/itc/web/ITCRequest.scala
+++ b/bundle/edu.gemini.itc.web/src/main/scala/edu/gemini/itc/web/ITCRequest.scala
@@ -337,8 +337,8 @@ object ITCRequest {
     fpMask match {
       case fpMask if fpMask.startsWith("CUSTOM_WIDTH") =>
         site match {
-          case Site.GN => FPUnitNorth.CUSTOM_MASK
-          case Site.GS => FPUnitSouth.CUSTOM_MASK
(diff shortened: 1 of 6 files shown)


bundle/edu.gemini.itc.web/src/main/scala/edu/gemini/itc/web/ITCRequest.scala
```diff
@@ -330,6 +331,34 @@ object ITCRequest {
...
   }
 
+  def fpMaskParameters(r: ITCRequest): GmosCommonType.FPUnit = {
+    val site = r.enumParameter(classOf[Site])
+    val fpMask = r.parameter("instrumentFPMask")
+    fpMask match {
+      case fpMask if fpMask.startsWith("CUSTOM_WIDTH") =>
+        site match {
+          case Site.GN => FPUnitNorth.CUSTOM_MASK.asInstanceOf[GmosCommonType.FPUnit]
+          case Site.GS => FPUnitSouth.CUSTOM_MASK.asInstanceOf[GmosCommonType.FPUnit]
```
andrewwstephens, line 341: Casting to GmosCommonType.FPUnit is to avoid an error flagged by IntelliJ.

cquiroz (approved):
LGTM

PR #2299, opened 2024-11-10
REL-4532 Fix GMOS ITC saturation level for imaging when spectral binning set

andrewwstephens:
The problem was that the saturation limit calculation was using the spectral binning in imaging mode.  The solution implemented here is to simply set the spectral binning equal to the spatial binning in imaging mode.  The output web page was also updated to always print both the x- and y-binning to make it clear what is happening, i.e. that the spectral binning is ignored in imaging mode.


cquiroz (approved):
LGTM

PR #2300, opened 2024-11-15
REL-4602: Do not add recentering overheads in the IGRINS-2 ITC

andrewwstephens:
This is a trivial change to avoid adding re-centering overheads in the IGRINS-2 ITC.  Note that even though it is no longer used, I reset the IGRINS-2 re-acquisition time back to 5 minutes since that had been accidentally set to the acquisition time of 7 minutes.


bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/igrins2/Igrins2.scala
```diff
@@ -179,7 +179,7 @@ final class Igrins2 extends ParallacticAngleSupportInst(Igrins2.SP_TYPE)
     Duration.ofSeconds(Igrins2.SetupTime.toSeconds.toLong)
 
   override def getReacquisitionTime(config: Config): Duration =
-    Duration.ofSeconds(Igrins2.SetupTime.toSeconds.toLong)
+    Duration.ofMinutes(5)
```
cquiroz, line 182: Are you sure about this. This was recently changed as per:
https://noirlab.atlassian.net/browse/REL-4531

cquiroz (approved):
LGTM

PR #2303, opened 2024-11-18
Clean scala compiler

cquiroz:


project/src/main/scala/edu/gemini/osgi/tools/app/MacDistHandler.scala
```diff
@@ -138,20 +132,22 @@ case class MacDistHandler(jre: Option[String], jreName: String)
...
           override def visitFile(
-              file: Path,
-              attrs: BasicFileAttributes
+            file:  Path,
+            attrs: BasicFileAttributes
           ): FileVisitResult = {
+            if (file.toFile.getName.startsWith("scala-compiler")) {
+              println(s"Remove native libs from scala-compiler jar")
+              val cmd = Seq("zip", "-d", file.toFile.getAbsolutePath, """META-INF/native/**""" )
+              cmd.lines.foreach(println)
```
swalker2m, line 141: Where does it actually do the removal?

cquiroz: When you call `cmd.lines` it  executes and produces an output

swalker2m: Seems to be a method on `ProcessBuilder`, I wonder how `Seq` becomes a `ProcessBuilder`.  But I don't wonder enough to investigate. :+1:

swalker2m (approved):

project/src/main/scala/edu/gemini/osgi/tools/app/MacDistHandler.scala
cquiroz: It is syntax added on the scala lib
https://www.scala-lang.org/api/2.12.8/scala/sys/process/index.html

PR #2305, opened 2024-11-19
REL-4572: Add ITC SEDs (this time with unique names)

andrewwstephens:
This pr renames the new SEDs so that they are unique when case is ignored.


cquiroz:
Great, we need this for osx work

cquiroz (approved):
LGTM

PR #2308, opened 2024-11-22
REL-4535 Enable both versions of the R400 grating

cquiroz:


bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gmos/GmosNorthType.java
```diff
@@ -79,9 +79,7 @@ public enum DisperserNorth implements GmosCommonType.Disperser, ObsoletableSpTyp
         B600_G5307(  "B600_G5307",  "B600",  600, Ictd.track("B600")),
         R600_G5304(  "R600_G5304",  "R600",  600, Ictd.track("R600")),
         B480_G5309(  "B480_G5309",  "B480",  480, Ictd.track("B480")),
-        R400_G5305(  "R400_G5305",  "R400",  400, Ictd.unavailable()) {
-            @Override public boolean isObsolete() { return true; }
-        },
+        R400_G5305(  "R400_G5305",  "R400",  400, Ictd.track("R400")),
```
cquiroz, line 82: @swalker2m It is unclear to me the effect of having two gratings tracking the same value in ICTD

swalker2m: This needs to match the corresponding key in the ICTD.  If they both have the same key, I think the availability will need to be the same for both of them.

swalker2m: This will show up in the QPT and QV, where both these gratings will be available or unavailable according to the information in the ICTD under "R400".

cquiroz:
I changed the id for `R400` and will  do a release  butt still needs some investigation about how will it interact with ICTD

PR #2313, opened 2024-12-17
Support multiple asterisms type on the program exporter

cquiroz:
This fixes the exception when trying to use program exporter with Ghost Asterism.
The output is not documented or well defined enough to account for the particular characteristics of Ghost Asterisms. Thus this is a simple solution that just takes the main target (in case of duals, the first target) and returns it.

Thus ghost programs with an explicit base will return the selected main target instead


swalker2m (approved):
Seems like a reasonable workaround for now. :+1:

PR #2321, opened 2025-01-23
REL-4624 Update model and all dependencies

cquiroz:
I misunderstood the requirements. This is a rework of REL-4624

### DIFF
--- a/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/overheads/Overheads.scala
+++ b/bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/overheads/Overheads.scala
@@ -1,12 +1,8 @@
 package edu.gemini.model.p1.overheads
 
 import edu.gemini.model.p1.immutable._
-
 import scalaz._
 import Scalaz._
-import squants.time._
-
-import scala.annotation.tailrec
 
 // This is not the ideal package for this to live in, but to avoid circular bundle references, we put it here.
 // Much of this is rendered obsolete due to the change from REL-2985 to REL-2926, but I am leaving it here in case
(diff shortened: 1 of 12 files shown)


bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/overheads/Overheads.scala
```diff
@@ -41,9 +37,15 @@ object Overheads extends (BlueprintBase => Option[Overheads]) {
     }
   }
 
-  private case class IGRINS2Overheads(telluricTime: TimeAmount, nTelluric: Int) extends Overheads {
+  private case class IGRINS2Overheads(telluricTime: TimeAmount, telluricStars: Igrins2TelluricStars) extends Overheads {
     override def calculate(progTime: TimeAmount): ObservationTimes = {
       val progTimeHrs: Double = progTime.toHours.value
+      val nTelluric = telluricStars match {
+        case Igrins2TelluricStars.Default  => (progTimeHrs / 1.5).toInt
```
swalker2m, line 44: What happens when `progTimeHrs` is >= 4.5 hours?  That is, is it allowed to be > 2?  Seems like it could be rounded as well?

swalker2m (approved):
I like the enum idea.  :+1:

bundle/edu.gemini.model.p1/src/main/scala/edu/gemini/model/p1/overheads/Overheads.scala
cquiroz: There is no upper bound, if you ask for 6h you'd get 4 stars.
It seems most observations are short though

PR #2323, opened 2025-01-25
REL-4638: F2 ITC calculation of exposure time & number of exposures to achieve S/N

andrewwstephens:
This updates the Flamingos2 ITC to support specifying only the desired S/N and return the exposure time, number of exposures, and the read mode.

I see that the read mode is not being automatically set for imaging mode, so this will still need a bit of work.


bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/flamingos2/Flamingos2Recipe.java
```diff
@@ -67,6 +75,162 @@ public ItcSpectroscopyResult serviceResult(final SpectroscopyResult r, final boo
...
+            Log.fine(String.format("desiredSNR = %.2f", desiredSNR));
+
+            double wavelength = ((SpectroscopyInt) _obsDetailParameters.calculationMethod()).wavelength();
+            Log.fine(String.format("Wavelength = %.2f nm", wavelength));
+
+            double maxFlux = instrument.maxFlux();  // maximum useful flux
+            Log.fine(String.format("maxFlux = %.0f e-", maxFlux));
+
+            // Perform an initial run of the ITC to get the signal and background for this target at this wavelength
+            readMode = ReadMode.MEDIUM_OBJECT_SPEC;  // arbitrary choice
```
cquiroz, line 107: Will this be a parmeter in the future?

cquiroz (approved):
LGTM

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/flamingos2/Flamingos2Recipe.java
andrewwstephens: We may be able to avoid it by setting the read mode based on the returned exposure time.

PR #2326, opened 2025-01-31
REL-4637 Fix IGRINS-2 ITC reporting of Fowler Samples when requesting time to achive a S/N

andrewwstephens:
This PR fixes a bug  where the number of Fowler samples and read noise was derived from the exposure time set in the web form even though the user had asked for the integration time to achieve a S/N.  This also optimizes the calculations so that only 2 iterations (per arm) are required.

### DIFF
--- a/bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/igrins2/Igrins2Recipe.java
+++ b/bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/igrins2/Igrins2Recipe.java
@@ -71,20 +71,168 @@ public final class Igrins2Recipe {
     }
 
     public SpectroscopyResult[] calculateSpectroscopy() {
-        Log.fine("calcMethod = " + _calcMethod);
-        if (_calcMethod instanceof SpectroscopyS2N) {
+
+        Igrins2Arm arm;
+        int instIndex;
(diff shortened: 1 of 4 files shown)


bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/igrins2/Igrins2Recipe.java
```diff
@@ -261,11 +269,45 @@ private SpectroscopyResult calculateSpectroscopy(
         calcSource.sed.accept(specS2N);
 
         final SpecS2N[] specS2Narr = new SpecS2N[]{specS2N};
+
+        VisitableSampledSpectrum finalS2NSpectrum = specS2N.getFinalS2NSpectrum();
+        double snr = finalS2NSpectrum.getY(SnrWavelength);
+        Log.fine(String.format("S/N @ %.1f nm = %.3f", SnrWavelength, snr));
+
         return new SpectroscopyResult(p, instrument, IQcalc, specS2Narr, slit, throughput.throughput(), altair,
                 Option.apply(new ExposureCalculation(exposureTime, numberExposures, snr)));
```
cquiroz, line 278: As discussed offline we may need to enhance this to support a set of calculations in case we have multiple detectors as for IGRINS-2

cquiroz (approved):
LGTM

PR #2327, opened 2025-02-02
REL-4503: Update Flamingos-2 minimum exposure times and only allow integer values

andrewwstephens:
This PR makes **some** of the requested changes:
- updates F2 minimum exposure time for each read mode
- updates the F2 ITC to return the exposure time rounded up to an integer
- updates the F2 ITC web form to require integer input
- fixes an ITC bug where the minimum exposure time was not honored in imaging mode

This PR does **NOT** update the OT so that it only accepts integer exposure times.


bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/flamingos2/Flamingos2Recipe.java
```diff
@@ -327,8 +329,8 @@ public ImagingResult calculateImaging() {
         return ImagingResult.apply(p, instrument, IQcalc, SFcalc, peak_pixel_count, IS2Ncalc);
     }
 
-    public double getExposureTime() {
-        return exposureTime;
+    public int getExposureTime() {
+        return (int) exposureTime;
```
swalker2m, line 333: > updates the F2 ITC to return the exposure time rounded up to an integer

Maybe it is unimportant but wouldn't this round down?  7.9 seconds would become 7 seconds.

PR #2330, opened 2025-02-24
Ignore keystore and skip validating the cert chain

cquiroz:
Once again we are having test problems because we have our own SSL certificate store but certificates on the wild change, e.g. horizons just got a new intermediate cert. This has happened several times in the middle of the semester requiring an emergency release.

This PR makes it so that we don't validate the certificate chain thus making it robust against changes in remote serves
This lowers the security checks however I claim it is not as big a problem.

In most cases, we are doing a client connection to a remote server and we'll still be encrypting the connection, though we could be susceptible to spoofing but most queries are for reading info.
The one case where we use the custom store doing trpc calls to the odb. In that case, we could be susceptible if an attacker could replace the certificate and take control of the DNS entry for the odb. I think the risk is relatively small


swalker2m (approved):
Let's go with it. :+1:

PR #2331, opened 2025-02-24
REL-4650: Add ToO option to Subaru proposals

cquiroz:


swalker2m (approved):
Looks good.  I think I would have needed a month to figure that out.

swalker2m:
Oops, mean to say "Looks good. I think I *would* have needed a month to figure that out."

PR #2340, opened 2025-03-13
Fix some edge cases for f2

cquiroz:


swalker2m (approved):
:+1:  What happens when an exception is thrown?   I mean, how is it handled and reported?

cquiroz:
Not sure if you want to know 😄 
It is put on an `Either` then written via a json serializer and passed to the gpp ITC where it is parrsed back into a `Left(exception)`

PR #2341, opened 2025-03-18
Better hierarchy

cquiroz:
I know we are not supposed to change the ocs much but the hierarchy of calculation methods in ITC is wrong and it is confusing for the work being done on gpp itc

### DIFF
--- a/bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/Parameters.scala
+++ b/bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/Parameters.scala
@@ -144,7 +144,7 @@ sealed trait IntMethod extends CalculationMethod {
 }
 
 // Return the exposure time and number of exposures given the desired S/N
-final case class ImagingExp(
+final case class ImagingInt(
                              sigma: Double,
                              coadds: Option[Int],
                              sourceFraction: Double,
@@ -168,8 +168,8 @@ final case class SpectroscopyInt(
 }
 
 
-// Return the number of exposures (Integration Time) given the exposure time, desired S/N, etc.
-final case class ImagingInt(
+// Return the number of exposures (count) given the exposure time, desired S/N, etc.
+final case class ImagingExpCount(
                     sigma: Double,
                     exposureTime: Double,
                     coadds: Option[Int],
--- a/bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/Parameters.scala
+++ b/bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/Parameters.scala
@@ -13,7 +13,7 @@ final case class ExactIq private (toArcsec: Double)
 
 object ExactIq {
 
-  def fromArcsec(d: Double): Option[ExactIq] =
+  private def fromArcsec(d: Double): Option[ExactIq] =
     if ((0.05 <= d) && (d <= 5.0)) Some(ExactIq(d)) else None
 
   def unsafeFromArcsec(d: Double): ExactIq =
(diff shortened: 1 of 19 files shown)


bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/Parameters.scala
```diff
@@ -108,84 +108,79 @@ sealed trait CalculationMethod {
...
+                                  offset: Double) extends Spectroscopy with IntMethod {
   val exposureTime: Double = 1200
   val exposures: Int = 1
   def isSpectroscopy: Boolean = true
   def isImaging: Boolean = false
 }
 
 
+// Return the number of exposures (count) given the exposure time, desired S/N, etc.
+final case class ImagingExpCount(
```
cquiroz, line 172: This is another source of confusion as this is neither `IntMethod` nor `S2NMethod`

bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/Parameters.scala
```diff
@@ -108,84 +108,79 @@ sealed trait CalculationMethod {
   require(isSpectroscopy != isImaging)
```
swalker2m, line 108: Random observation unrelated to changes: maybe we could just define `final def isImaging: Boolean = !isSpectroscopy` and not `require` anything.

bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/Parameters.scala
```diff
@@ -108,84 +108,79 @@ sealed trait CalculationMethod {
...
 }
 
+// Calculations that return signal to noise out of an amount of exposures
```
swalker2m, line 111: Do you not also need the exposure time to do this?

bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/Parameters.scala
```diff
@@ -108,84 +108,79 @@ sealed trait CalculationMethod {
...
-}
-
 // Return the spectroscopic integration time (exposure time & number of exposures) given the desired S/N
 final case class SpectroscopyInt(
                                   sigma: Double,
                                   wavelengthAt: Double,
                                   coadds: Option[Int],
                                   sourceFraction: Double,
-                                  offset: Double) extends Spectroscopy with SpecIntMethod with S2NMethod {
+                                  offset: Double) extends Spectroscopy with IntMethod {
```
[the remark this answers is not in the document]
swalker2m: I wouldn't be opposed to changing the `Int` in `IntMethod` to something more descriptive.  Is that for `Integration`?

swalker2m (approved):
This seems like a good improvement.  I don't really know ITC so my comments are probably dumber than usual.  :+1:

bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/Parameters.scala
cquiroz: right, good idea

bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/Parameters.scala
cquiroz: yes

bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/Parameters.scala
```diff
@@ -108,84 +108,79 @@ sealed trait CalculationMethod {
...
-                    exposureTime: Double,
-                    coadds: Option[Int],
-                    sourceFraction: Double,
-                    offset: Double) extends Imaging with IntMethod {
-
-  def isSpectroscopy: Boolean = false
-  def isImaging: Boolean = true
+// Calculations that return exposure time to achivee a given SN
+sealed trait IntMethod extends CalculationMethod {
+  def sigma: Double
```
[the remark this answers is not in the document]
cquiroz: that's right

cquiroz:
Right we could change it to `Integration` I have to say this is still in gerenal a bit confusing. does it mean it produces integration time, or requests integration time (time x count)

PR #2343, opened 2025-03-24
Correctly return SignalToNoiseAt for the S2N method on F2

cquiroz:


bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/flamingos2/Flamingos2Recipe.java
```diff
@@ -287,6 +287,9 @@ private SpectroscopyResult calculateSpectroscopy(
         final double finalSnr = specS2N.getFinalS2NSpectrum().getY(wavelength);
         Log.fine(String.format("single S/N @ %.1f nm = %.3f", wavelength, singleSnr));
         Log.fine(String.format("final S/N @ %.1f nm = %.3f", wavelength, finalSnr));
+        // 0 is used by the web forms, ignore it as a degenerate case
+        if (wavelength > 0 && (wavelength < specS2N.getExpS2NSpectrum().getStart() || wavelength > specS2N.getExpS2NSpectrum().getEnd()))
+            throw new RuntimeException(String.format("Wavelength %.1f out of range", wavelength));
```
swalker2m, line 292: If the wavelength is 0, we don't throw an exception and just use?

swalker2m (approved):
:+1:

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/flamingos2/Flamingos2Recipe.java
cquiroz: I was hoping this was going to go unnoticed under the radar
gpp will never send a 0 as we don't allow 0 wavelengths but the web form hardcode a 0.
This is a cheap way to tell them apart. Ideally this should be an `Option` but I don't want to mess with the web forms to set them to None

swalker2m: Makes sense.  What happens if it _is_ a 0 from the webform?

cquiroz: on one side we ignore the exception. The web form doesn't print the Signal Noise At value, which is expected to be 0. Users of the web form use the graph to calculate the signal-to-noise value at any wavelength

PR #2347, opened 2025-04-25
REL-4509 Add P2 check for GMOS-N observations with E2V

cquiroz:
you can still copy paste an old program using gmos e2v

A phase 2 check will indicate that this is not allowed, and you should change it to Hammamatsu. 

The UI now only lets you ever change from E2V to Hammamatsu


swalker2m (approved):
:+1:

PR #2348, opened 2025-04-26
REL-4706: limit the calculated number of exposures to 1000

andrewwstephens:
When calculating the number of exposures required to reach a desired S/N, limit the number of exposures to 1000 to prevent bad things from happening.


cquiroz (approved):
LGTM

PR #2349, opened 2025-04-27
REL-2501 "obsoleting" Y filter that was never purchased.

arturog8m:
Making the F2 Y filter obsolete. REL-2501 requested this `Y` filter and `F1056`  and `F1063` to be hidden - but the last two were taken care in REL-3561


swalker2m (approved):
👍

PR #2351, opened 2025-04-29
REL-4701: Improve OT window placement on multi-screen computers

andrewwstephens:
This pr attempts to put new OT windows on the same screen as the mouse.  There may be a better way to do this...

### DIFF
--- a/bundle/jsky.util/src/main/java/jsky/util/Preferences.java
+++ b/bundle/jsky.util/src/main/java/jsky/util/Preferences.java
@@ -327,8 +327,12 @@ public class Preferences {
     private static void _centerComponent(Component component) {
         Dimension dim = component.getPreferredSize();
         if (dim.width > 1 && dim.height > 1) {
-            Dimension screen = Toolkit.getDefaultToolkit().getScreenSize();
-            component.setLocation(screen.width / 2 - dim.width / 2, screen.height / 2 - dim.height / 2);
+            PointerInfo pointerInfo = MouseInfo.getPointerInfo();
+            if (pointerInfo != null) {
+                LOG.info("Placing " + component.getName() + " on " + pointerInfo.getDevice().getIDstring());
+                Rectangle screen = pointerInfo.getDevice().getDefaultConfiguration().getBounds();
+                component.setLocation(screen.x + (screen.width - dim.width) / 2, screen.y + (screen.height - dim.height) / 2);
+            }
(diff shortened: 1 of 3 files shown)


bundle/jsky.util/src/main/java/jsky/util/Preferences.java
```diff
@@ -327,8 +327,12 @@ public void componentMoved(ComponentEvent e) {
     private static void _centerComponent(Component component) {
         Dimension dim = component.getPreferredSize();
         if (dim.width > 1 && dim.height > 1) {
-            Dimension screen = Toolkit.getDefaultToolkit().getScreenSize();
-            component.setLocation(screen.width / 2 - dim.width / 2, screen.height / 2 - dim.height / 2);
+            PointerInfo pointerInfo = MouseInfo.getPointerInfo();
+            if (pointerInfo != null) {
+                LOG.info("Placing " + component.getName() + " on " + pointerInfo.getDevice().getIDstring());
+                Rectangle screen = pointerInfo.getDevice().getDefaultConfiguration().getBounds();
+                component.setLocation(screen.x + (screen.width - dim.width) / 2, screen.y + (screen.height - dim.height) / 2);
```
swalker2m, line 334: I can see where it centers the component in the coordinates where it would be in the center if it were placed on the screen where the mouse is.  I don't see how it selects a particular screen though.  How does that work?  What did it do before this change in a multi-screen environment?

swalker2m (approved):

bundle/jsky.util/src/main/java/jsky/util/Preferences.java
andrewwstephens: The motivation for this change is that the new EL9 telops workstations treat the 6-8 monitors as a single monolithic desktop.  If I remember correctly, the splash screen appears exactly in the center, while the program chooser always appears on in the top-left monitor.  When a program opens it is usually where the last program was because I think we save that as a user preference.

`MouseInfo.getPointerInfo().getDevice()` returns the `GraphicsDevice` where the mouse pointer is currently located.  On my dual-monitor computer, `IDstring` is `:0.0` on the left screen and `:0.1` on the right screen.  Using a test program that recenters a popup on the screen with the mouse, this seems to work well.  However, testing the OT seems to be less reliable, perhaps because of the aforementioned saving of user preferences.

PR #2354, opened 2025-05-20
REL-4724: Use preferred browser to view QPT plan under Linux

andrewwstephens:
It seems that "firefox -remote openURL" no longer works (it opens firefox, but not the requested HTML file).  This pr suggests switching to "xdg-open" under Linux to use the user's preferred browser.  I left the "else use firefox" in case there is something that is not Windows, Mac, or Linux, but maybe that isn't really necessary...


cquiroz:
I wonder if we should use the java way
https://www.geeksforgeeks.org/java-program-to-open-input-url-in-system-default-browser-in-windows/

In theory, that will work for any OS but it would need testing

andrewwstephens:
Whatever you think is best.  :shrug:  It might be nice to deploy something soon though since the current mechanism does not work under el9.

PR #2397, opened 2026-04-03
Make GHOST central wavelength a constant

toddburnside:
It turns out the central wavelength for GHOST was hardcoded as a hidden parameter in the html file and treated as a variable in GhostParameters. Since the central wavelength for GHOST is a fixed value, I made it a constant in GhostParameters and removed it from the web page and the JSON interface.


cquiroz (approved):
LGTM

PR #2398, opened 2026-04-10
Fix ghost charts

toddburnside:
GHOST was returning some chart enums we didn't have encoders for. Since we don't need those in GPP I just don't generate those charts for GPP. This also goes one step of the way toward decreasing the size of the data we throw over the wall to GPP.

NOTE: I think ignoring whitespace changes will probably greatly reduce the size of the diff...

### DIFF
--- a/bundle/edu.gemini.itc/src/main/scala/edu/gemini/itc/service/ItcServiceImpl.scala
+++ b/bundle/edu.gemini.itc/src/main/scala/edu/gemini/itc/service/ItcServiceImpl.scala
@@ -66,7 +66,7 @@ class ItcServiceImpl extends ItcService {
     // execute ITC service call with updated parameters
     p.observation.calculationMethod match {
       case _: Imaging       => ItcResult.forMessage ("Imaging not implemented.")
-      case _: Spectroscopy  => calculateSpectroscopyCharts(p)
+      case _: Spectroscopy  => calculateSpectroscopy(p, excludeCharts = false)
     }
 
   } catch {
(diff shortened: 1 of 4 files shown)


bundle/edu.gemini.json/src/main/scala/edu/gemini/json/CoproductCodec.scala
```diff
@@ -26,7 +26,7 @@ trait CoproductCodec {
   class CoproductCodec[A] private { prev =>
 
     def encode(a: A): Json =
-      sys.error("Can't encode. No case was provided for $a.")
+      sys.error(s"Can't encode. No case was provided for $a.")
```
toddburnside, line 29: @swalker2m This is where you've been getting that stray `s` from.

bundle/edu.gemini.itc/src/main/scala/edu/gemini/itc/service/ItcServiceImpl.scala
```diff
@@ -104,45 +104,38 @@ class ItcServiceImpl extends ItcService {
...
+  private def spectroscopyResult(recipe: SpectroscopyArrayRecipe, excludeCharts: Boolean): Result = {
     val r = recipe.calculateSpectroscopy()
-    val s = recipe.serviceResult(r, includeCharts)
+    val s = recipe.serviceResult(r, excludeCharts)
     ItcResult.forResult(s)
   }
 
+  private def ghostSpectroscopyResult(recipe: GHostRecipe, excludeCharts: Boolean): Result = {
+    val r = recipe.calculateSpectroscopy()
+    // we use a special method to not produce S2NChartPerRes charts or the combined charts.
```
cquiroz, line 134: Per res?

bundle/edu.gemini.itc/src/main/scala/edu/gemini/itc/service/ItcServiceImpl.scala
```diff
@@ -104,45 +104,38 @@ class ItcServiceImpl extends ItcService {
...
+    val s = recipe.serviceResult(r, excludeCharts, true)
```
cquiroz, line 135: are me always calling for gpp on the json servlet?

bundle/edu.gemini.itc/src/main/scala/edu/gemini/itc/service/ItcServiceImpl.scala
toddburnside: Yeah. I have no idea what the PerRes charts are, but we don't show them in GPP.

bundle/edu.gemini.itc/src/main/scala/edu/gemini/itc/service/ItcServiceImpl.scala
toddburnside: Yes, we are. It is my understanding that only GPP is using the json servlet for GHOST.

cquiroz (approved):
LGTM

PR #2399, opened 2026-04-14
Round the ITC graph data returned from the JSONServlet to 6 sigfigs

toddburnside:
There are timeout issues with the ghost charts because there is so much more data than for the other modes. This PR will greatly reduce the size of the serialized data and hopefully will reduce the time to serialize/deserialize a bit. How much is uncertain since the `number` of data points to serialize/deserialize remains the same. I'm hoping we can get away with this for now and can optimize further when we have that mythical more time.

I discussed with Andy and he said 6 significant figures is plenty for any instrument in both GPP and the OT.

### DIFF
--- a/bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/ItcService.scala
+++ b/bundle/edu.gemini.itc.shared/src/main/scala/edu/gemini/itc/shared/ItcService.scala
@@ -122,6 +122,11 @@ final case class SpcSeriesData(dataType: SpcDataType, title: String, data: Array
     ssdCopy
   }
 
+  def withSignificantFigures(n: Int): SpcSeriesData = {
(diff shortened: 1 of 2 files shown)


cquiroz:
We do this on lucuma-itc, why move it here?

toddburnside:
> We do this on lucuma-itc, why move it here?

See description. I'm hoping it can reduce serialization/deserialization time somewhat. I will test this in explore to see if it is true before merging this, but I'm waiting on a explore/odb combo that will work for GHOST.

cquiroz:
At least there will be less memory pressure, which is a good thing
I've thought about passing these as binary using egg CBOR but that seems a much larger task

toddburnside:
> At least there will be less memory pressure, which is a good thing I've thought about passing these as binary using egg CBOR but that seems a much larger task

Right. I'm trying to do the simplest thing possible...

PR #2401, opened 2026-04-28
Change the selection of signalToNoiseAt for GHOST

toddburnside:


bundle/edu.gemini.itc/src/main/scala/edu/gemini/itc/base/Recipe.scala
```diff
@@ -171,17 +171,46 @@ object Recipe {
...
 
+
+  // Based on serviceGroupedResult above, but specialized for GHOST
+  def serviceGhostResult(rs: Array[SpectroscopyResult], charts: JList[JList[SpcChartData]], excludeCharts: Boolean): ItcSpectroscopyResult = {
+    // We know that GHOST has 2 results
+    val rBlue = rs(0);
+    val rRed = rs(1);
+
+    // We return a single signalToNoiseAt response. From discussions with Andy, if we have choices we will
+    // return the one with the greatest final S/N. We'll also use that as the "selected" IntegrationTime,
```
swalker2m, line 184: I thought `signalToNoiseAt` was an input, but I'm not familiar with the ITC.  What is going on here?

swalker2m (approved):
Maybe I'm not the indicated person to review this, but it seems fine to me.

bundle/edu.gemini.itc/src/main/scala/edu/gemini/itc/base/Recipe.scala
toddburnside: Yeah, like so many terms in this project, signalToNoiseAt is an overloaded term without clear indication of the meaning. In this case, the input is a `requested` s/n or time and count at a specified wavelength. The output is the `calculated` single exposure s/n and final s/n at that requested wavelength.

bundle/edu.gemini.itc.web/src/main/scala/edu/gemini/itc/web/servlets/JsonServlet.scala
```diff
@@ -35,7 +35,7 @@ object ItcCalculation extends ItcParametersCodec with ItcResultCodec {
 
     (for {
       itcReq <- Parse.decodeEither[ItcParameters](json)
-      itcRes <- itc.calculate(itcReq, true).toEither.leftMap(_.msg)
+      itcRes <- itc.calculate(itcReq, headless = true).toEither.leftMap(_.msg)
```
cquiroz, line 38: Unrelated but it would be nice if somedy we could get rid of this headless business. This is related to the charting lib which uses AWT and we want to avoid having an awt thread live

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/ghost/GHostRecipe.java
```diff
@@ -193,14 +181,17 @@ private SpectroscopyResult calculateSpectroscopy(Ghost instrument) {
         VisitableSampledSpectrum singleS2NSpectrum = specS2N.getExpS2NSpectrum();
         VisitableSampledSpectrum finalS2NSpectrum = specS2N.getFinalS2NSpectrum();
 
+        final SpectroscopyS2N calcMethod = getSpectroscopyS2N(instrument.getCamera());
+        final AllIntegrationTimes times = AllIntegrationTimes.single(new IntegrationTime(calcMethod.exposureTime(), calcMethod.exposures()));
```
cquiroz, line 185: Why is it single? shouldn't be there one per detector?

bundle/edu.gemini.itc/src/main/scala/edu/gemini/itc/base/Recipe.scala
cquiroz: as an extra they dont't match. the requested and the result s/n are different presummably due to numeric approximations, though they should be similar

bundle/edu.gemini.itc.web/src/main/scala/edu/gemini/itc/web/servlets/JsonServlet.scala
toddburnside: Not sure I understand. Do you want the parameter name changed? Or something about the functionality changed? The renaming seems to be happening slowly to `excludeCharts`.

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/ghost/GHostRecipe.java
toddburnside: This is the calculation for a single detector. They are combined in `Recipe.serviceGhostResult`.

bundle/edu.gemini.itc.web/src/main/scala/edu/gemini/itc/web/servlets/JsonServlet.scala
cquiroz: I guess I meant that it would be ideal not to depend on awt here. We do that only for historical reasons. As you found out we could perhaps transfer spectra data directly rather than chart series

cquiroz (approved):
LGTM

bundle/edu.gemini.itc.web/src/main/scala/edu/gemini/itc/web/servlets/JsonServlet.scala
toddburnside: Ah, I see. Yeah, I hope to change it to not create the charts for GPP in an upcoming PR. But, I don't think awt is actually being used here, yet. The "charts" that are being created at this point is just another data structure with data being copied from the spectra. No awt involved, just a bunch of new arrays. Well, to be honest, it's a bit hard to tell if they are new arrays or just references to existing ones... I'm trying to figure that out right now.

PR #2402, opened 2026-04-29
[itc] Return s2nAt and times in Gnirs calculations

rpiaggio:
Adds `times` and `signalToNoiseAt` to returned Gnirs ITC calculations.

- Since Gnirs only supports T+C modes, `times` just echoes back the passed `exposureTime` but `numberExposures` = original `numberExposures` * `coadds`.
- I skipped this for IFU mode for the moment, I'm not sure how to do this since IFU mode computes multiple charts.

### DIFF
--- a/bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gnirs/GnirsRecipe.java
+++ b/bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gnirs/GnirsRecipe.java
@@ -590,9 +590,12 @@ public final class GnirsRecipe implements ImagingRecipe, SpectroscopyRecipe {
         int numberExposures = 0;
         if (calcMethod instanceof ImagingS2N) {
             ImagingS2N s2nMethod = (ImagingS2N) calcMethod;
-            expTime = s2nMethod.exposures();
+            expTime = s2nMethod.exposureTime();
             numberExposures = s2nMethod.exposures() * s2nMethod.coaddsOrElse(1);
         } else if (calcMethod instanceof ImagingIntegrationTime) {
+            expTime = IS2Ncalc.getExposureTime();
+            numberExposures = IS2Ncalc.numberSourceExposures(); // Already has coadds factored.
+        } else if (calcMethod instanceof ImagingExposureCount) {
             expTime = IS2Ncalc.getExposureTime();
             numberExposures = IS2Ncalc.numberSourceExposures(); // Already has coadds factored.
         } else {
--- a/bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gnirs/GnirsRecipe.java
+++ b/bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gnirs/GnirsRecipe.java
@@ -603,7 +603,7 @@ public final class GnirsRecipe implements ImagingRecipe, SpectroscopyRecipe {
         }
         Log.fine("numberExposures = " + numberExposures);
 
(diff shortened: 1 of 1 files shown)


bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gnirs/GnirsRecipe.java
```diff
@@ -566,7 +586,25 @@ public ImagingResult calculateImaging() {
         }
         IS2Ncalc.calculate();
 
-        return new ImagingResult(p, instrument, IQcalc, SFcalc, peak_pixel_count, IS2Ncalc, altair, Option.empty());
+        double expTime = 0;
```
toddburnside, line 589: I see expTime being set but not used, am I missing something?

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gnirs/GnirsRecipe.java
```diff
@@ -566,7 +586,25 @@ public ImagingResult calculateImaging() {
...
+        int numberExposures = 0;
+        if (calcMethod instanceof ImagingS2N) {
+            ImagingS2N s2nMethod = (ImagingS2N) calcMethod;
+            expTime = s2nMethod.exposureTime();
+            numberExposures = s2nMethod.exposures() * s2nMethod.coaddsOrElse(1);
+        } else if (calcMethod instanceof ImagingIntegrationTime) {
+            expTime = IS2Ncalc.getExposureTime();
```
toddburnside, line 596: It seems this is hardcoded to `60`. Does it make sense to use it? I mean, is exposure time really always 60 for this mode?

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gnirs/GnirsRecipe.java
```diff
@@ -493,6 +510,9 @@ public ImagingResult calculateImaging() {
         // calculates: redshifted SED
         // output: redshifteed SED
 
+        final CalculationMethod calcMethod = _obsDetailParameters.calculationMethod();
+        Log.fine("calcMethod = " + calcMethod);
```
toddburnside, line 514: Do we need to keep all of this new logging in this file?

toddburnside (approved):
LGTM with a few comments.

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gnirs/GnirsRecipe.java
rpiaggio: Good catch, I'm fixing in new commit

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gnirs/GnirsRecipe.java
rpiaggio: I'd say to leave it, as we have done in other places in the ITC. If we need to collect information when this is working, we can adjust the log level.

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gnirs/GnirsRecipe.java
rpiaggio: I'm not sure about this. In any case, this case is not used ATM. Gnirs ITC only supports the first case. I had to add the other two cases to make the baseline tests happy.

PR #2403, opened 2026-05-05
[itc] only return y-data and axis information for charts to GPP

toddburnside:
For the series data sent to GPP, only return the y data and enough information to reconstruct the x data (which is equally spaced wavelengths). 

Claude Sonnet wrote some of the code that altered implicit resolution so I could provide an alternate `CodecJson[SpcSeriesData]` for the GPP endpoints. I did have to pretty much tell it how to do it, but it did figure out how make it all work and deal with all the types for me.

Note: I have no idea why `jline-2.10.7` is suddenly failing to resolve in CI. I deleted the ivy cache for it locally and it resolved just fine. Maybe it will work tomorrow.

### DIFF
--- a/bundle/edu.gemini.itc.web/src/main/scala/edu/gemini/itc/web/servlets/JsonServlet.scala
+++ b/bundle/edu.gemini.itc.web/src/main/scala/edu/gemini/itc/web/servlets/JsonServlet.scala
@@ -1,13 +1,16 @@
 package edu.gemini.itc.web.servlets
 
-import argonaut._, Argonaut._
-import edu.gemini.itc.shared.{ ItcParameters, ItcResult, ItcService }
+import argonaut._
+import Argonaut._
+import edu.gemini.itc.shared.{ItcParameters, ItcResult, ItcService, SpcSeriesData}
 import edu.gemini.itc.service.ItcServiceImpl
-import edu.gemini.itc.web.json.{ ItcParametersCodec, ItcResultCodec }
-import javax.servlet.http.{ HttpServlet, HttpServletRequest, HttpServletResponse }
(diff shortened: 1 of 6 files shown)


cquiroz:
Should me put these into a separate branch?

toddburnside:
> Should me put these into a separate branch?

Yeah. I was going to do it later, but I just created a `gpp-ocs` branch at `main` and made it the base branch for this PR. When merged, it will merge into `gpp-ocs`.

bundle/edu.gemini.itc.web/src/main/scala/edu/gemini/itc/web/servlets/JsonServlet.scala
```diff
@@ -19,10 +22,46 @@ import scalaz._, Scalaz._
...
+    // sends x-axis data and only the y data values to save encoding time and bandwidth.
+    implicit val SpcSeriesDataCodec: CodecJson[SpcSeriesData] =
+      CodecJson(
+        (s: SpcSeriesData) =>
+          ("dataType" :=  s.dataType) ->:
+          ("title"    :=  s.title)    ->:
+          ("dataY"    :=  s.data(1).map(roundToSix))  ->:
+          ("xAxis"    :=  ItcXAxis.calcAxis(s.data(0)))  ->:
+          jEmptyObject,
+        (c: HCursor) => ??? // the decoder is never used
```
cquiroz, line 62: Should we just define the encoder?

toddburnside: I could try that, but I don't really know much about how argonaut does things and whether I can implicitly override just the encoder part of a codec. I figured since we never send graph data from gpp to ocs, this was safe enough since this codec can only be used here.

PR #2404, opened 2026-05-15
Only calculate GHOST spectrum data in the range of the detector

toddburnside:
Vibe coded with Claude Sonnet. 

Andy tested it and it fixes a bug in the web itc as well. The `peak pixel signal + background` for the blue detector was wrong because the data it looked in had the full wavelength range and it found the peak somewhere above 542 nm.

### DIFF
--- a/bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/ghost/GHostRecipe.java
+++ b/bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/ghost/GHostRecipe.java
@@ -23,7 +23,8 @@ public final class GHostRecipe implements SpectroscopyArrayRecipe {
     private final ObservingConditions _obsConditionParameters;
     private final TelescopeDetails _telescope;
 
-    private final short WAVELENGTH_BLUE_END = 542; // nm . Range 347 - 542 nm
+    private static final double WAVELENGTH_BLUE_END = 542;  // nm. Blue arm range: ~347 – 542 nm
+    private static final double WAVELENGTH_RED_START = 520; // nm. Red arm range:  520 – redEnd nm
 
     /**
      * Constructs a GhostRecipe given the parameters. Useful for testing.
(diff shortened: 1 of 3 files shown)


bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/ghost/GHostRecipe.java
```diff
@@ -23,7 +23,8 @@ public final class GHostRecipe implements SpectroscopyArrayRecipe {
     private final ObservingConditions _obsConditionParameters;
     private final TelescopeDetails _telescope;
 
-    private final short WAVELENGTH_BLUE_END = 542; // nm . Range 347 - 542 nm
+    private static final double WAVELENGTH_BLUE_END = 542;  // nm. Blue arm range: ~347 – 542 nm
+    private static final double WAVELENGTH_RED_START = 520; // nm. Red arm range:  520 – redEnd nm
```
cquiroz, line 27: are there blue_start and red_end constants? Interesting that there is an overlap

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/ghost/GHostRecipe.java
```diff
@@ -69,23 +70,31 @@ public ItcSpectroscopyResult serviceResult(final SpectroscopyResult[] r, final b
...
-                backGroundSpectComb.setY(i, r[0].specS2N()[0].getBackgroundSpectrum().getY(i));
-                expS2NComb.setY(i, r[0].specS2N()[0].getExpS2NSpectrum().getY(i));
-                finalS2NSpecComb.setY(i, r[0].specS2N()[0].getFinalS2NSpectrum().getY(i));
-                singleS2NCombPerRes.setY(i, listPerResElement.get(0).getY(i));
-                finalS2NSpecCombPerRes.setY(i, listPerResElement.get(1).getY(i));
-            }
+            // Build combined full-range spectra by stitching blue and red arm spectra.
+            // Blue arm data is used up to its computed end (≈ WAVELENGTH_BLUE_END);
+            // red arm data covers the remainder. The combined grid uses blue's pixel
+            // width and spans from blueStart to redEnd.
```
cquiroz, line 76: What about the overlap?

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/ghost/GHostRecipe.java
```diff
@@ -189,6 +209,30 @@ private SpectroscopyResult calculateSpectroscopy(Ghost instrument) {
...
 
+    /**
+     * Builds a combined full-range spectrum by stitching blue and red arm spectra.
+     * Blue arm data is used for wavelengths up to its computed end (≈ WAVELENGTH_BLUE_END);
+     * red arm data is used beyond that point.
+     * The result has uniform sampling equal to the blue arm's pixel width and spans
+     * from blueSpec.getStart() to redSpec.getEnd(). Linear interpolation (getY(double))
+     * maps each arm onto the combined pixel grid, returning 0 outside a spectrum's range.
+     */
+    private VisitableSampledSpectrum buildCombinedSpectrum(
```
cquiroz, line 220: I think for igrins2 we are returning a spectrum per ccd rather than a combined one, should we unify the criteria?

toddburnside: The combined spectra are not created for GPP. They are used by the web ITC. In GPP we get the spectra per detector and they will display similar to igrins2 - except that there is overlap in this case.

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/ghost/GHostRecipe.java
toddburnside: There is overlap because...well, just because. That's how the light gets split, I guess. There don't appear to be any constants anywhere but here.

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/ghost/GHostRecipe.java
toddburnside: Blue is used in the overlap range. This is how the existing code worked. Except, of course, it generated way more data than needed.

toddburnside: Personally, I suspect the combined charts were made for the web itc because the individual charts were not worth anything because they contained data for wavelengths outside of the detector. We might be able to eliminate those now, but I don't want to mess with the web itc. I just made it so that GPP doesn't pay the cost of charts it doesn't need.

PR #2406, opened 2026-05-17
REL-4887: Update GHOST gain and read noise

andrewwstephens:
Attempt number 2.  This time with support for parameters that are inter-camera dependent.  The HashMap was suggested by claude, so please double-check.


cquiroz (approved):
LGTM

PR #2407, opened 2026-05-17
REL-4931: Add support for GNIRS S/N ETM

andrewwstephens:
This has only been lightly tested, but nominally adds S/N Exposure Time Mode support in the GNIRS ITC.

Note that all of the baseline tests passed _before_ printing the one-line ReadMode in the output web page.  Since the spectroscopic read mode is calculated in S/N mode, I added it to the ouput and updated baseline.txt accordingly.


bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gnirs/GnirsRecipe.java
```diff
@@ -98,27 +113,192 @@ public ItcSpectroscopyResult serviceResult(final SpectroscopyResult r, final boo
...
+        Log.fine("Running ITC with final input parameters:");
         Log.fine("numberExposures = " + numberExposures);
-        Log.fine(String.format("WavelengthAt = %.2f nm", wavelengthAt));
+        Log.fine("exposureTime = " + exposureTime);
+        Log.fine("coadds = " + coadds);
+        Log.fine("readMode = " + readMode + " -> readNoise = " + readMode.getReadNoise() + " e-");
+        Log.fine(String.format("wavelengthAt = %.2f nm", wavelengthAt));
+
+        // Run the ITC to generate the output graphs
+        return calculateSpectroscopy(instrument, readMode, exposureTime, numberExposures, coadds, wavelengthAt).withTimes(
```
cquiroz, line 283: If I understand correctly the read mode is an output as we iterate over them to find the optimal one?

cquiroz (approved):
LGTM

PR #2413, opened 2026-06-08
REL-4931: Auto-select the GNIRS and F2 read mode for imaging S/N ETM

andrewwstephens:
This updates the ITC to calculate the best read mode for GNIRS and F2 imaging when the user requests S/N.

Note that this does *not* use the exposure time guidelines (which are really designed for spectroscopy) but rather calculates the total time for each read mode and chooses the shortest one.  This seems to yield good results although it does take a bit longer to calculate.

Note that every case I tested with Flamingos2 resulted in "Bright Object Mode".  We could probably just hard-code the read mode and save a bit of calculation time, but it is not a huge overhead since the calculation breaks off as soon as the next mode takes longer than the previous.

This also includes a fix to the GMOS S/N imaging mode where the number of exposures was not being propagated through to the overheads table.

### DIFF
--- a/bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/operation/ExposureTimeCalculator.java
+++ b/bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/operation/ExposureTimeCalculator.java
@@ -1,0 +1,40 @@
+package edu.gemini.itc.operation;
+
+import java.util.logging.Logger;
+
+public final class ExposureTimeCalculator {
+
+    private static final Logger Log = Logger.getLogger(ExposureTimeCalculator.class.getName());
+
+    private ExposureTimeCalculator() {}  // prevent instantiation
+
+    /**
+     * Calculate the exposure time required to achieve a desired Signal-to-Noise on a single frame.
+     * @param signal     The signal per second summed over the pixels in the aperture.
+     * @param background The background per second summed over the pixels in the aperture.
+     * @param darkNoise  The dark current per second summed over the pixels in the aperture.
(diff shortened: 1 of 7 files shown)


bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/operation/ExposureTimeCalculator.java
```diff
@@ -0,0 +1,40 @@
...
+            double SNR
+    ) {
+        final double f = 1. + (1. / skyAper);  // noise factor
+        // SNR = S / sqrt(S + f(B + D + R)) = st / sqrt(st + f(bt + dt + R)) = x
+        // This can be expressed as a quadratic equation:  (ss/xx)t^2 - (s + fb + fd)t - fR = 0
+        // and solved using the quadratic formula: t = (-b + sqrt(b^2 - 4ac)) / 2a
+        double a = signal * signal / (SNR * SNR);
+        double b = -(signal + f * background + f * darkNoise);
+        double c = -f * readNoise;
+        double exposureTime = (-b + Math.sqrt(b*b - 4.*a*c)) / (2.*a);
```
cquiroz, line 37: Should we add some guards here? We have had cases where signal is negative (probably due to rounding errors) and make the sqrt param negative, or the whole exptime negative too

andrewwstephens: Good point.  If the signal is <= 0 then there is no way to proceed, so it now throws an error.

cquiroz (approved):
LGTM

PR #2416, opened 2026-06-25
Don't discard single exposure S/N in Gnirs XD

rpiaggio:


bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/gnirs/GnirsRecipe.java
```diff
@@ -588,11 +595,14 @@ private static SpcChartData createGnirsSignalChart(final SpectroscopyResult resu
     }
 
     private static SpcChartData createGnirsS2NChart(final SpectroscopyResult result) {
-        final String title = "Final S/N";
+        final String title = "Intermediate Single Exp and Final S/N";
```
cquiroz, line 598: What does it mean to be `Intermediate`

rpiaggio: That's it's a single exposure's S/N (intermediate step). This is the same text that's used in other modes.

cquiroz (approved):
LGTM

PR #2425, opened 2026-08-27
Improve performance of rounding itc series

cquiroz:
`SpcSeriesData.roundToSignificantFigures` took a `BigDecimal`, but its only caller (`ItcCalculation.roundToSix`) passes a `Double`.

The implicit widening made the first line `if (num == 0)` route through `java.math.BigDecimal.longValueExact`, which throws an `ArithmeticException` for every non-integral value.

It's caught and discarded inside`isValidLong`, but `fillInStackTrace` has already run — once per graph data point, inside the argonaut encoder.

Changing the parameter to `Double` (what every call site already had) removes it.

## Effect

`ItcCalculation.calculateCharts` on a GMOS-N longslit fixture:

| | before | after |
|---|---|---|
| wall time | 113.4 ms | 50.8 ms (-55%) |
| bytes allocated | 142.8 MB | 62.9 MB (-56%) |


toddburnside (approved):
LGTM

PR #2426, opened 2026-08-31
Speed up ArraySpectrum lower index lookup

cquiroz:
`DefaultArraySpectrum.getLowerIndex` ran a full binary search on every call. Visitors such as `TransmissionElement.visit` call it once per data point with x (wavelength) in increasing order.

This change keeps a cursor with the index returned by the previous lookup. Each call first checks whether x falls in the cursor's bin or the next one and answers in O(1) if so; otherwise it falls back to the binary search.

This is useful performance wise because `visit` calls getLowerIndex sequentially with an increasing X, and that was benchmarked as the hot path.

```
        for (int i = 0; i < sed.getLength(); i++) {
            final double startval = sed.getX(i);
            final double multiplier = _trans.getY(startval);
            sed.setY(i, sed.getY(i) * multiplier);
        }
```

Testing

Measured on lucuma-itc: time-only call 43.7 → 15.9 ms, graphs 51.1 → 22.7 ms.

The change adds an internal mutable field but it is safe for serialization as the class is never serialized and works fine ina multi-thread call environment.

There are some alternatives to achieve the same result without a mutable field but they all require API changes I'd rather avoid

### DIFF
--- a/bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/base/DefaultArraySpectrum.java
+++ b/bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/base/DefaultArraySpectrum.java
@@ -12,9 +12,19 @@ import java.util.logging.Logger;
 public final class DefaultArraySpectrum implements ArraySpectrum {
     private static final Logger Log = Logger.getLogger(DefaultArraySpectrum.class.getName());
     // The spectral data.  _data = new double[2][num_data_points]
-    // data[0][i] = x values
+    // data[0][i] = x values (wavelength in nm)
     // data[1][i] = y values
+    // The x axis is assumed non-decreasing: the .dat resource files list wavelengths in
+    // ascending order by convention (nothing validates it), and getStart/getEnd, the
+    // interpolation in getY and the search in getLowerIndex all rely on that ordering.
     private final double[][] _data;
+
+    // The index getLowerIndex returned last time. Callers usually walk x in increasing order,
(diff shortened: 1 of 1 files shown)


bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/base/DefaultArraySpectrum.java
```diff
@@ -12,10 +12,20 @@
...
+    // ascending order by convention (nothing validates it), and getStart/getEnd, the
+    // interpolation in getY and the search in getLowerIndex all rely on that ordering.
     private final double[][] _data;
 
+    // The index getLowerIndex returned last time. Callers usually walk x in increasing order,
+    // so the next lookup tends to land in this bin or the one after.
+    // The cursor is only a hint and needs no synchronisation: getLowerIndex checks it against
+    // the data before using it, so at worst a bad value causes a fallback to the binary
+    // search, never a wrong result.
+    private int _lowerIndexCursor = 0;
```
toddburnside, line 27: 🤮 But, there are worse dragons about. 🐉

toddburnside (approved):
Seems reasonable to me.

bundle/edu.gemini.itc/src/main/java/edu/gemini/itc/base/DefaultArraySpectrum.java
cquiroz: i had to read this a few times to be sure it is kosher. This looks scary but works fine

PR #2427, opened 2026-09-09
Reduce ITC Allocations

cquiroz:
Two follow-ups to the ITC allocation profiling (#2425, #2426). I was always suspect of the number of times arrays were copied in ITC and this confirm it. We Tremove roughly a third of the bytes allocated per `calculateExposureTime` call. 

**Avoid double array copy in `reset` internals.** The sampling constructors, `clone`, `trim`, `rescaleX` and `rescaleXwithFixedSampleSize` all build a fresh local array and hand it to `reset`, which copied it again. Those call sites now adopt the array directly.

The public constructor and public `reset` still copy, so no external contract changes.

**Tidy `smoothY` and `trim`.** `smoothY` allocated its scratch array before the early return, and both methods used `new Double(x).intValue()` as an int cast. `Double.intValue` is the same narrowing conversion as `(int)`, so results are identical.

This was verified on lucuma-itc comparing the results before and after.


toddburnside (approved):
LGTM. Good catch.

